PR #1598, opened 2017-04-12
Clean up the tests

dave-tucker:
This PR cleans up the tests a little.

It moves the packages in to a `tests/pkg` directory.
The YAML files are moved to `tests/cases`.

This follows the convention in the root of the repo and allows for better modularity.

The `test/test.yml` is now `tests/cases/test-kernel-config.yml` which better reflects its purpose.
I also removed `dhcpd` from the `test-kernel-config.yml` and used the new `poweroff` pkg to shutdown on completion.


dave-tucker:
Closing due to rename

PR #1620, opened 2017-04-13
[ci skip] ARM64 project -- NOT FOR MERGE

mor1:
Kernel build should work.
Still working towards an `initrd`.


mor1:
Closing following the great renaming, having recreated as #1654

PR #1623, opened 2017-04-13
Ebpf jit verifying

tych0:
Here's a patch that verifies the EBPF jit buffer output, see patch notes for details.

One question I have: I've changed the version from 4.9.x to 4.9.22-1. Should I be force pushing to the 4.9.x tag instead or something?


justincormack:
Needs re cherry picking over the current repo, so closing. Also need to resolve #1628

l0kod:
You may want to talk with @thestinger about integrating this patch in https://github.com/thestinger/linux-hardened/

thestinger:
I don't disagree with spender's analysis of this though. It's better than nothing but if the attacker has that primitive they can interfere with the verification just as they interfered with the initial generation of the code. I'll take the same approach of hard-wiring off the JIT compiler. The interpreter itself is still a problem even then. This hasn't been designed with security in mind.

tych0:
I agree r.e. the interpreter has a similar problem, and I suppose a similar trick might (?) be able to work for that, although I haven't looked closely.

Can you elaborate on what you mean by an attacker can interfere with the verification? I'm assuming here that the attacker can't modify kernel text, and that the hash function is Good Enough.

thestinger:
It's an improvement but hash calculation depends on a bunch of state exposed to tampering. It's not comparable to a way of verifying it by simply scanning over the generated code and verifying that it's sane. An attacker can interfere with the initial code generation and then the hash calculation. It's a valid mitigation but it doesn't provide the same boundary against arbitrary code injection as there was in the past.

I'm not sure there's much you can do about this beyond partially mitigating it like this. eBPF is a lot closer to arbitrary kernel code execution by design than the previous infrastructure so it's harder to verify it, ensure correctness of the implementation and mitigate attacks via other bugs against the state it uses.

There are lots of other missing pieces in a mainline kernel like lack of protection for page tables, no CFI, etc. so it's probably true that there's a lot of lower hanging fruit there but ebpf and then the JIT on top of it are chipping away at the potential of mitigations.

thestinger:
> chipping away at the potential of mitigations

At least those targeting exploit techniques (CFI, more pervasive use of RO memory, etc.) rather than ones catching memory corruption bugs directly. The security aspects would have needed to be considered as part of designing it vs. trying to mitigate the impact after the fact with stopgap measures. It's ironic that part of the push for it is having even more options for userspace access control. There's already so much fine-grained userspace access control with so many different choices and overlap between all the features. On the other hand, it's all meaningless if an attacker can just exploit the kernel which is already the main way to escape sandboxes, etc. Going to have contained applications using this to implement internal sandboxes and then they need access to it directly as attack surface not just as a target for other bugs.

tych0:
> It's an improvement but hash calculation depends on a bunch of state exposed to tampering.

Yes, although if you choose a good hash, doesn't this mean that it's hard to do this tampering and end up with a collision?

> It's a valid mitigation but it doesn't provide the same boundary against arbitrary code injection as there was in the past.

Point well taken, thanks for your input!

thestinger:
> Yes, although if you choose a good hash, doesn't this mean that it's hard to do this tampering and end up with a collision?

Tampering of the state it uses to generate code and then the same state used to verify code, so the strength of the hash doesn't really come into it.

tych0:
Right, I guess that leaves us with your first point about the interpreter having the same problem that needs to be addressed too.

Thanks!

l0kod:
> At least those targeting exploit techniques (CFI, more pervasive use of RO memory, etc.) rather than ones catching memory corruption bugs directly. The security aspects would have needed to be considered as part of designing it vs. trying to mitigate the impact after the fact with stopgap measures.

Sure

> It's ironic that part of the push for it is having even more options for userspace access control. There's already so much fine-grained userspace access control with so many different choices and overlap between all the features. On the other hand, it's all meaningless if an attacker can just exploit the kernel which is already the main way to escape sandboxes, etc. Going to have contained applications using this to implement internal sandboxes and then they need access to it directly as attack surface not just as a target for other bugs.

The goal of Landlock is not to mimic SELinux and other access-control systems, which would be pointless and rejected by upstream anyway. Landlock is more close to seccomp-bpf which was the starting point. Without going in details, it enables to create standalone sandboxes, which BTW may be interesting for Android applications. ;)  There is no equivalent of XNU Sandbox, Capsicum or Pledge for Linux.

I agree that when you add a new userland interface to the kernel it increases its attack surface, but the SUID/SGID interface (used by sandboxes) may increase the *system* attack surface too. However, the issue here is not the eBPF interface but the way the kernel translate and use eBPF code. This is an issue whoever uses it, be it kprobes, an XDP program, a socket filter or a seccomp filter.

PR #1662, opened 2017-04-19
Fix for prefix for linuxkit

thebsdbox:

**- What I did**
I've added in the capability to look for the `vmdk` from the use of the prefix

**- How I did it**
Duplicated the same functionality from `qemu` and `hypervisor`

**- How to verify it**
`./moby run vmware vmware`
 
Modified the default behaviour of the `qemu` plugin to match  expected <prefix> behaviour.

### DIFF
--- a/src/cmd/moby/run_vmware.go
+++ b/src/cmd/moby/run_vmware.go
@@ -151,6 +151,13 @@ func buildVMX(cpus int, mem int, diskPath string, prefix string) string {
 
 	if diskPath != "" {
 		returnString += fmt.Sprintf(vmxDisk, diskPath)
+	} else {
+		vmdkPath := prefix + ".vmdk"
+		if _, err := os.Stat(vmdkPath); os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


rneugeba:
Could you add a short explanation to the commit message? Thanks

src/cmd/moby/run_vmware.go
```diff
@@ -151,6 +151,13 @@ func buildVMX(cpus int, mem int, diskPath string, prefix string) string {
 
 	if diskPath != "" {
 		returnString += fmt.Sprintf(vmxDisk, diskPath)
+	} else {
+		vmdkPath := prefix + ".vmdk"
+		if _, err := os.Stat(vmdkPath); os.IsNotExist(err) {
```
riyazdf: should we also bail if we have an error that is not `os.IsNotExist`?

thebsdbox: The error will be reported from `vmrun` if it can't access the vmdk file (permissions etc.), we could also pre-test for that condition?

DieterReuter:
Works for me, thanks!

Issue #1671, opened 2017-04-20
Fallback to using containers if qemu is not installed locally


dave-tucker:
`moby run qemu` should fallback to using the qemu container from `tools/qemu` if QEMU is not installed locally.

This is needed for `moby run` to be used in CI which doesn't presently have QEMU installed in the image.

PR #1672, opened 2017-04-20
Let moby run qemu fallback to using a container

dave-tucker:
This PR deprecates scripts/qemu.sh and uses `moby run` exclusively
In order to do this, a fallback to using the qemu container from `tools/qemu` was added

If `qemu-system-$ARCH` isn't found on the `$PATH`, container fallback is enabled.

Fixes: #1671


src/cmd/moby/run_qemu.go
```diff
@@ -73,18 +91,38 @@ func runQemu(args []string) {
 	} else {
 		qemuArgs = append(qemuArgs, "-enable-kvm")
 		qemuArgs = append(qemuArgs, "-machine", "q35,accel=kvm:tcg")
+		if container {
+			dockerArgs = append(dockerArgs, "", "")
+			copy(dockerArgs[5:], dockerArgs[3:])
```
rneugeba: this looks odd. You set up dockerArgs above and here you are munging it and it hasn't been touched in between. I don't understand why this is necessary.

src/cmd/moby/run_qemu.go
```diff
@@ -73,18 +91,38 @@ func runQemu(args []string) {
...
+		if err != nil {
+			log.Fatal("Unable to find docker in the $PATH")
+		}
+		binArgs := append(dockerArgs, qemuBinPath)
+		imgArgs := append(dockerArgs, qemuImgPath)
+		qemuCmd = exec.Command(dockerPath, binArgs...)
+		qemuImgCmd = exec.Command(dockerPath, imgArgs...)
+	} else {
+		qemuCmd = exec.Command(fullQemuPath)
+		qemuImgCmd = exec.Command(fullQemuImgPath)
```
rneugeba: this also looks odd, you have a conditional above for container, and then another just after. How about just one conditional where you set up the args either for local or container and then execute it without conditional?

src/cmd/moby/run_qemu.go
dave-tucker: To add `--device /dev/kvm` if you are running on a system that supports KVM.
The `dockerArgs` set up above ends with image name and thus can't be appended to as `--device` has to come before it.

I could just do a plain old `append` here and then have this afterwards.
```
if container {
    dockerArgs = append(dockerArgs, QemuImg)
}
```

src/cmd/moby/run_qemu.go
dave-tucker: the conditional above is for `container && /dev/kvm is present` and adds the necessary `--device /dev/kvm` runes to the docker run command. this conditional sets up the `*exec.Command` for both qemu and qemu image so the same set of args (built from the flags supplied) can be used.

there are still cases where we need to check whether we're in a container or not:
e.g: lines 143 and 176
as some options aren't compatible with containers (e.g GUI) or the paths that are being tested for are inside the container itself (e.g the OVMF firmware)

src/cmd/moby/run_qemu.go
rneugeba: your patch patch sets up dockerArgs so why don't you set it up so that these gymnastics are not necessary? Your code looks a bot like:
```
a = 5
...
a = a + 3
a = a - 1
```
when it could look like:
```
a = 7
```

src/cmd/moby/run_qemu.go
rneugeba: sorry, pre coffee. let me take another look

src/cmd/moby/run_qemu.go
rneugeba: yeah, that would be better. Probably best in the conditional on container further below

dave-tucker:
@rneugeba having considered your comments i think it's true that the code in its current form doesn't read too well. I've now refactored this and split the code in to:

- Building the qemu commandline
- Run qemu in a container
- Run qemu locally

I think this makes it much easier to grok what's going on and to find the right place to add validation checks etc..

src/cmd/moby/run_qemu.go
```diff
@@ -53,78 +75,185 @@ func runQemu(args []string) {
...
+				return fmt.Errorf("Error creating disk [%s]:  %s", config.DiskPath, err.Error())
+			}
+		} else {
+			log.Infof("Using existing disk [%s]", config.DiskPath)
+		}
+	}
+
+	// Check for OVMF firmware before running
+	if config.UEFI {
+		_, err := os.Stat(config.FWPath)
```
riyazdf: nit: can collapse err-checking into one line

```
if _, err := os.Stat(config.FWPath); err != nil {
...
```

src/cmd/moby/run_qemu.go
```diff
@@ -53,78 +75,185 @@ func runQemu(args []string) {
...
+	}
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+}
+
+func runQemuLocal(config QemuConfig, args []string) error {
+	if config.DiskPath != "" {
+		// If disk doesn't exist then create one
+		if _, err := os.Stat(config.DiskPath); os.IsNotExist(err) {
```
riyazdf: I noticed this pattern on @thebsdbox's PR too - I'm inclined to bail on this error if it is not `nil` and is also `os.IsNotExist(err) == false` for extra safety and clearer debugging.

Also it would make it consistent with how you handle the OVMF firmware below: https://github.com/linuxkit/linuxkit/pull/1672/files#diff-7ea76eb7a56000c30ad47347644f6c66R123

src/cmd/moby/run_qemu.go
```diff
@@ -53,78 +75,185 @@ func runQemu(args []string) {
...
+			}
+			return err
+		}
+	}
+
+	qemuCmd := exec.Command(config.QemuBinPath, args...)
+	// If verbosity is enabled print out the full path/arguments
+	log.Debugf("%v\n", qemuCmd.Args)
+
+	// If we're not using a seperate window then link the execution to stdin/out
```
riyazdf: typo: `separate`

src/cmd/moby/run_qemu.go
```diff
@@ -53,78 +75,185 @@ func runQemu(args []string) {
...
+	}
+
+	dockerPath, err := exec.LookPath("docker")
+	if err != nil {
+		return fmt.Errorf("Unable to find docker in the $PATH")
+	}
+
+	if config.DiskPath != "" {
+		// If disk doesn't exist then create one
+		if _, err = os.Stat(config.DiskPath); os.IsNotExist(err) {
```
riyazdf: same comment as above re: error handling

src/cmd/moby/run_qemu.go
```diff
@@ -53,78 +75,185 @@ func runQemu(args []string) {
...
+		}
+	}
+
+	qemuArgs := append(dockerArgs, QemuImg, "qemu-system-"+config.Arch)
+	qemuArgs = append(qemuArgs, args...)
+	qemuCmd := exec.Command(dockerPath, qemuArgs...)
+	// If verbosity is enabled print out the full path/arguments
+	log.Debugf("%v\n", qemuCmd.Args)
+
+	// GUI mode not currentlty supported in a container. Although it could be in future.
```
riyazdf: nit typo: `currently`

src/cmd/moby/run_qemu.go
riyazdf: Also: could we share these common blocks of code between `runQemuContainer` and `runQemuLocal`?

src/cmd/moby/run_qemu.go
```diff
@@ -160,4 +275,5 @@ func buildPath(prefix string, postfix string) string {
 		log.Fatalf("File [%s] does not exist in current directory", path)
 	}
 	return path
+
```
riyazdf: nit: extra line

src/cmd/moby/run_qemu.go
justincormack: verbose error checking is fine...

justincormack:
@dave-tucker any chance of fixing the issue before I totally mess up any ability to merge this?

dave-tucker:
@justincormack give me 30 mins and i'll push an update with all comments addressed

src/cmd/moby/run_qemu.go
dave-tucker: I thought about it and tbh it makes the code harder to read IMHO

dave-tucker:
@justincormack done...

talex5:
Build on master failed, with:
```
Status: Downloaded newer image for linuxkit/qemu:17f052263d63c8a2b641ad91c589edcbb8a18c82
qemu: could not load kernel 'test-bzImage': No such file or directory
```
Maybe `make ci` needs updating?

justincormack:
Hmm, @dave-tucker any idea? How could this pass on the PR CI and not on merge?

justincormack:
only difference is having `test-ltp`

dave-tucker:
No idea! @talex5 if you retrigger the master build does it reliably fail? Or is it intermittent?

justincormack:
No its failing always on master now. Also this broke running CI locally with a tty see https://github.com/linuxkit/linuxkit/pull/1696 for attempt to revert (but which breaks CI).

justincormack:
This is one reason I am really against having tests that are only run on merge.

PR #1686, opened 2017-04-22
use docker build to build linuxkit

friism:
Signed-off-by: Michael Friis <friism@gmail.com>

**- What I did**

Changed Makefile to use official golang image and `docker build`

**- How to verify it**

```
make
make bin/infrakit-instance-hyperkit
```

I'm not sure if this covers all the logic hidden inside the old builder image - let me know if something is missing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM golang:1.8
+
+RUN go get -u github.com/golang/lint/golint && \
+	go get -u github.com/gordonklaus/ineffassign && \
+	go get -u github.com/LK4D4/vndr
+
+ADD . /go/src/github.com/linuxkit/linuxkit
+
+ARG target=moby
+ARG ldflags
+ARG GOOS
+ARG GOARCH
+
+WORKDIR /go/src/github.com/linuxkit/linuxkit/src/cmd/$target
+
+RUN files=$(find . -type f -name '*.go' -not -path "./vendor/*" -not -name '*.pb.*') && \
+	echo "gofmt..." && test -z $(gofmt -s -l $files | tee /dev/stderr) && \
+	echo "go vet..." && test -z $(GOOS=linux go tool vet -printf=false . 2>&1 | grep -v vendor/ | tee /dev/stderr) && \
+	echo "golintr..." && test -z $(golint $files | tee /dev/stderr) && \
+	echo "ineffassign..." && test -z $(for file in $files ; do (ineffassign $file | tee /dev/stderr) ; done)
+
+RUN go build -ldflags "${ldflags}" -buildmode pie -o /out/$target .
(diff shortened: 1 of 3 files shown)


rneugeba:
The existing process has the advantage that tools like golint etc have a fixed, known version, while with this approach we get whatever the latest is. Also, by using `ADD .` it will transfer the entire directory structure, including any images to the daemon, which can be quite large.

Having said that, the `tar cf - ...| docker ... | tar xf -` pipeline is quite brittle.

friism:
> tools like golint etc have a fixed, known version

We can use the same approach as in the moby/moby Dockerfiles: https://github.com/moby/moby/blob/master/Dockerfile#L151

Happy to update.

> it will transfer the entire directory structure, including any images to the daemon
No, I added a `.dockerignore` file: https://github.com/linuxkit/linuxkit/pull/1686/files#diff-f7c5b4068637e2def526f9bbc7200c4e

justincormack:
Why?  I know it is slightly non standard, but to generate a binary `docker run` is actually more appropriate, what I want to do is run the Go compiler, not build a Docker image. It is as fast as the cached build case, and much faster than the uncached case...

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ARG ldflags
+ARG GOOS
+ARG GOARCH
+
+WORKDIR /go/src/github.com/linuxkit/linuxkit/src/cmd/$target
+
+RUN files=$(find . -type f -name '*.go' -not -path "./vendor/*" -not -name '*.pb.*') && \
+	echo "gofmt..." && test -z $(gofmt -s -l $files | tee /dev/stderr) && \
+	echo "go vet..." && test -z $(GOOS=linux go tool vet -printf=false . 2>&1 | grep -v vendor/ | tee /dev/stderr) && \
+	echo "golintr..." && test -z $(golint $files | tee /dev/stderr) && \
```
AkihiroSuda: golintr →golint

Makefile
```diff
@@ -22,17 +20,17 @@ PREFIX?=/usr/local/
...
 
 INFRAKIT_DEPS=$(wildcard src/cmd/infrakit-instance-hyperkit/*.go) Makefile vendor.conf
 bin/infrakit-instance-hyperkit: $(INFRAKIT_DEPS) | bin
-	tar cf - vendor -C src/cmd/infrakit-instance-hyperkit . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit -o $@ > tmp_infrakit_instance_hyperkit_bin.tar
-	tar xf tmp_infrakit_instance_hyperkit_bin.tar > $@
-	rm tmp_infrakit_instance_hyperkit_bin.tar
-	touch $@
+	docker build --build-arg target=infrakit-instance-hyperkit -t $(GIT_COMMIT)-builder .
+	$(eval cid = $(shell docker create $(GIT_COMMIT)-builder))
+	docker cp $(cid):/out/moby $@
```
AkihiroSuda: target is not moby here

justincormack:
No longer applies as `moby` tool not in this repo.

PR #1687, opened 2017-04-23
Convert some images to multi-stage builds

rneugeba:
This converts `tini`, `media-toybox`, `rngd`, and `test-virtsock` to use the new multi-stage build. It also moves `tini` and `media-toybox` to `./tools` which seems a more appropriate place.

`c-compile` has also been extended to allow it to be used to compile the above C code


avsm:
This doesn't work yet on the Hub autobuilds does it?

justincormack:
@avsm no not yet, but we are not using those

pkg/rngd/Dockerfile
```diff
@@ -1,3 +1,47 @@
...
+  linux-headers \
+  make \
+  musl-dev \
+  patch \
+  && true
+
+COPY . /
+
+ENV pkgname=rng-tools pkgver=5
+RUN curl -O -sSL http://downloads.sourceforge.net/project/gkernel/$pkgname/$pkgver/$pkgname-$pkgver.tar.gz
```
justincormack: sourceforge...

pkg/rngd/Dockerfile
```diff
@@ -1,3 +1,47 @@
...
+  strip /usr/sbin/rngd
+RUN mkdir -p /tmp/dev /tmp/proc /tmp/sys
+RUN cat ${HASHFILES} | sha1sum | sed 's/ .*//' > hash
+
+
+FROM linuxkit/tini:a847727dd242b65975e6e5e007bcfe31fbd30fef@sha256:587b2af6eb9a0657c51fccd496a910b06fd71bb14e932a9bd24f21e0f63315a1
+COPY --from=rng-build usr/sbin/rngd usr/sbin/rngd
+COPY --from=rng-build hash hash
+# Can't create directories, so copy empty dirs from previous context
+COPY --from=rng-build tmp/dev dev
```
justincormack: why cant you just `RUN mkdir ...`?

justincormack:
My plan was to get rid of the whole hashing thing by having a base Alpine image with all the packages available in a filesystem package repo (not installed, as its still good to see which ones are needed to be installed for each build).

pkg/rngd/Dockerfile
rneugeba: no `mkdir` in the `tini` image

justincormack: ah

rneugeba:
I'm not sure how a base Alpine image will help. It get's rid of some of the hashes, but not all. We still have images which are created based on input files and we need to detect if/when the input changed...

Also, let's do things one step at a time. I suggest we move to multi-stage builds first.

What's your view on unified (ie containing all necessary packages) c-compile image? Will make the dependent images simpler

justincormack:
Yes, it will get rid of the non source changes.

What doesnt c-compile include?

rneugeba:
Let me find out when I update :)

justincormack:
But anyway this looks fine as a stage 1

rneugeba:
Pushed an update with an updated `c-compile`. Simplifies the builds a lot.

Makefile
```diff
@@ -71,16 +71,26 @@ test: test-initrd.img test-bzImage test-cmdline
 	tar cf - $^ | ./scripts/qemu.sh 2>&1 | tee test.log
 	$(call check_test_log, test.log)
 
+test-ltp.img.tar.gz: $(MOBY) test/ltp/test-ltp.yml
+	$(MOBY) build test/ltp/test-ltp.yml
```
justincormack: these ltp parts are unrelated and already upstream...

pkg/rngd/Dockerfile
justincormack: we check the sha256sum instead

pkg/rngd/fix-textrels-on-PIC-x86.patch
```diff
@@ -1,17 +1,17 @@
 --- rng-tools/rdrand_asm.S
 +++ rng-tools/rdrand_asm.S
 @@ -49,6 +49,7 @@
- 	ret
+	ret
```
justincormack: why did this change?

src/cmd/moby/run_gcp.go
```diff
@@ -45,6 +45,7 @@ func runGcp(args []string) {
 	familyFlag := gcpCmd.String("family", "", "GCP Image Family. A group of images where the family name points to the most recent image. *Optional* when 'prefix' is a filename")
 	nameFlag := gcpCmd.String("img-name", "", "Overrides the Name used to identify the file in Google Storage, Image and Instance. Defaults to [name]")
 	diskSizeFlag := gcpCmd.Int("disk-size", 0, "Size of system disk in GB")
+	skipCleanup := gcpCmd.Bool("skip-cleanup", false, "Don't remove images or VMs")
```
justincormack: this shouldnt be in here either?

Makefile
riyazdf: yep - should just need a rebase for this and the gcp flags

pkg/rngd/fix-textrels-on-PIC-x86.patch
rneugeba: my git changed the whitespace, should be benign

rneugeba: Hmm, more of this file changed. Let me check

rneugeba: ah, looks like GH's web interface doesn't grok patch files very well. The changes are just whitespace changes, so should be fine

PR #1693, opened 2017-04-24
Split build and push, and remove push code from run

### LINKED ISSUE #1618: how should `moby build` work for cloud platforms?
## Summary of current behaviour

The `linuxkit.yml` contains a number of fields like `Bucket` and `Project` that are used to allow the `gcp-img` artifact to be uploaded to Google Storage and for an Google Compute Engine image to be created from this file.

## Problems with the current approach

- `linuxkit.yml` files need to be edited by each user to contain the correct variables and for an image upload to proceed. As a user, this means that I don't bother using the `format: gcp` and prefer to use `format: gcp-img` and let `moby run` handle the upload and image creation. It also means that the `examples/gcp.yml` can't be used without being edited

- The `format` section of `linuxkit.yml` will grow exponentially as more cloud providers are added, each needed a dozen or so additional fields or format types

## Proposed solutions

These were discussed in #1607

### Configuration using CLI flags and environment variables

This was the approach taken by #1607
It simplifies the `format` section and allows the `examples/gcp.yml` to be used without editing.
The drawback of this approach is that we end up with dozens of flags available to `moby build`.

### Introduce an intermediate command `moby upload` or `moby cloud`

In this scenario, `moby build` would only output local files. Akin to the `gcp-img` format that is available today. The result of `moby build` could then be passed to `moby upload` or `moby cloud` to create an image on the targeted cloud platform. This command could be structured such that there is a subcommand for each cloud provider and would re-use code (and possibly flags) from `moby run`.

This would simplify the `format` section a lot, as it only needs to know about local formats.
It would also mean that the examples would be portable too.

The drawback is that the promise of `moby build` providing you something that you can run is broken if you were hoping to run on a cloud platform.

For example:

```
moby build examples/gcp.yml
moby cloud gcp -bucket foo -project bar -replace true gcp
moby run gcp gcp
```

### Add a supplemental configuration file to `moby build`

This could be, in a default location for example, `~/.moby/provider.yml` or provided using a command line flag e.g `moby build --config /secrets/provider.yml` and would contain:
```
gcp:
  bucket: foo
  project: bar
  keys: /path/to/keyfile.json
aws:
  ...
packet:
  ...
```

`moby build` would then test for the presence of this file before performing a build that required access to a cloud provider.

This is possibly the most elegant solution but it comes at the cost of additional onboarding for a user. For example, a `moby build examples/gcp` may fail with `[ERROR]: Google Cloud Credentials not present. Please ensure that you have configured ~/.moby/providers.yml`

justincormack:
This currently only changes the `gcp` target, but is the new
model - the `build` command will only do things locally, then
you need to `push` to an image store such as GCP or other ones
in order to `run` for platforms that cannot boot directly from
a local image.

Fix #1618

cc @dave-tucker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,8 @@ test-hyperkit: $(MOBY) test-initrd.img test-bzImage test-cmdline
 
 .PHONY: test-gcp
 test-gcp: $(MOBY) test.img.tar.gz
-	$(MOBY) run gcp test.img.tar.gz | tee test-gcp.log
+	$(MOBY) push gcp test.img.tar.gz
+	$(MOBY) run gcp test | tee test-gcp.log
 	$(call check_test_log, test-gcp.log)
 
 .PHONY: test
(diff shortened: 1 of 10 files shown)


justincormack:
(LGTM fro @dave-tucker)

PR #1695, opened 2017-04-24
Split cli into moby and linuxkit

justincormack:
`moby` just does the simple `build` cases, while `linuxkit` does
`push` and `run`.

This allows splitting out just `moby`.


justincormack:
Amended the `README` too to refer to the new tools.

README.md
```diff
@@ -15,17 +15,18 @@ LinuxKit, a toolkit for building custom minimal, immutable Linux distributions.
 
 ## Getting Started
 
-### Build the `moby` tool
+### Build the `moby` and `linuxkit` tools
```
riyazdf: Should we add a brief explanation of the difference between the linuxkit and moby tools?  I realize we'll continue to shuffle things around but might be helpful in the intermediate stage

justincormack:
Added a very short explanation, will expand when split out (soon).

alexellis:
Updated tutorial @ http://blog.alexellis.io/boot-linuxkit-in-10-mins/

README.md
deitch: I still am trying to wrap my head around it. Is the short form: `moby` is a tool for building images, and `linuxkit` is a tool for pushing and running images? So `moby == docker build` and `linuxkit == docker run / docker push`?

rneugeba: yes. And the `moby` tool will be moved and extended to build assemblies other than linux images. This was discussed at length at DockerCon and various meetings after. I'm not sure there is a good writeup yet

PR #1655, opened 2017-04-18
Clean up the tests

dave-tucker:
This PR cleans up the tests a little.

It moves the packages in to a `tests/pkg` directory.
The YAML files are moved to `tests/cases`.

This follows the convention in the root of the repo and allows for better modularity.

The `test/test.yml` is now `tests/cases/test-kernel-config.yml` which better reflects its purpose.
I also removed `dhcpd` from the `test-kernel-config.yml` and used the new `poweroff` pkg to shutdown on completion.

Rework of #1598

PR #1696, opened 2017-04-24
Restore the tty workaround for the test runner

justincormack:
Deleted in #1672 and stops the test script returning right result with
a tty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test-bzImage: test-initrd.img
 # interactive versions need to use volume mounts
 .PHONY: test-qemu-efi
 test-qemu-efi: test-efi.iso
-	$(MOBY) run $^ | tee test-efi.log
+	script -q /dev/null $(MOBY) run $^ | tee test-efi.log
 	$(call check_test_log, test-efi.log)
 
 bin:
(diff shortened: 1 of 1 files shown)


justincormack:
Breaks CI locally.

justincormack:
Hmm, this then breaks CI cc @dave-tucker @talex5 - we really need this to run locally and in CI...

dave-tucker:
I don't think this is needed.
The error in `make test` is caused due to derpy stdout/stderr from HyperKit.
A normal `./bin/linuxkit run test | tee test.log` doesn't contain the console output from the VM.
If you redirect stderr to stdout then it does...

dave-tucker:
Note: as `make test` doesn't specify a backend, it will use the default backend for each platform (hyperkit on OSX)

justincormack:
I just want `make test` to not error locally... eg so I can use for bisect or whatever...

Can we remove test-hyperkit target then.

dave-tucker:
that should be fixed in #1655
I can raise a PR to fix it sooner if you'd rather

justincormack:
ah yes, no thats ok. But how do we get `make test` locally to not error?

dave-tucker:
@justincormack my current strategy is to explicitly use the `qemu` backend. e.g `/bin/moby run qemu test`

justincormack:
@dave-tucker can you change `make test` to do this then?

PR #1700, opened 2017-04-25
update gcp docs

AkihiroSuda:
* removed incomplete command from README.md; it does not work without specifying GCP project and bucket
* `moby <push|run>` -> `linuxkit ...`
* `myfile` -> `myprefix`, because there is no file named `myfile`

Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ See `linuxkit run --help`.
 
 `make test` or `make test-hyperkit` will run the test suite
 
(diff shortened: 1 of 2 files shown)


rneugeba:
Thanks for the documentation update. The GCP backend also supports passing this information via environment variables (https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_gcp.go#L16). Would you mind updating the PR to reference them as an alternative?

AkihiroSuda:
done. PTAL.

Issue #1661, opened 2017-04-18
With containerd, sometimes service containers don't get started


rneugeba:
#1636 introduced the ability to use containerd, instead of runc, to run containers. This seems to work fine with HyperKit, but when booted on some other platform, like Hyper-V and packet.net, I've seen that sometimes the service containers don't get started. This looks like a race condition to me. 

I'll try to get a more reliable repo, but thought I add an issue to track

justincormack:
This may be fixed, unclear.

rneugeba:
I haven't seen this issue with since we updated containerd, so closing this for now

Issue #1660, opened 2017-04-18
Logs sometimes don't work when using containerd


rneugeba:
https://github.com/linuxkit/linuxkit/pull/1636 introduced the ability to use `containerd`, instead of `runc`, to run containers. For service containers, sometimes the log files are empty.

justincormack:
@MagnusS was looking at this - we should file an upstream bug...

MagnusS:
I can reproduce this locally with containerd master. Upstream issue in https://github.com/containerd/containerd/issues/770

PR #1701, opened 2017-04-25
*.yml: move dhcpcd from onboot to services

AkihiroSuda:
**- What I did**
move dhcpcd from onboot to services, so that the instance can invoke DHCP request on network reconnection and on lease timeout

**- How I did it**
Edit *.yml

**- How to verify it**
```
host$ moby build minimal.yml && linuxkit run minimal
guest$ runc list
ID          PID         STATUS      BUNDLE                        CREATED                          OWNER           
dhcpcd      301         running     /containers/services/dhcpcd   2017-04-25T05:13:03.762592514Z   root
```

penguin

Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

### DIFF
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -2,12 +2,12 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551
+  - linuxkit/init:63eed9ca7a09d2ce4c0c5e7238ac005fa44f564b
   - linuxkit/runc:b0fb122e10dbb7e4e45115177a61a3f8d68c19a9
-  - linuxkit/containerd:60e2486a74c665ba4df57e561729aec20758daed
-onboot:
+  - linuxkit/containerd:18eaf72f3f4f9a9f29ca1951f66df701f873060b
+services:
   - name: dhcpcd
-    image: "linuxkit/dhcpcd:48e249ebef6a521eed886b3bce032db69fbb4afa"
+    image: "linuxkit/dhcpcd:57a8ef29d3a910645b2b24c124f9ce9ef53ce703"
     binds:
      - /var:/var
      - /tmp/etc:/etc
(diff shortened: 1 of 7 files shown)


rneugeba:
The reason for invoking `dhcp` in onboot is so that we can guarantee that we have an IP address for everything starting afterwards. If we just start dhcp in the service section other services, like for example an nginx contianer, need to check/wait till the network is up. We had discussed to use dhcpd in oneshot mode in `onboot` *and* as a service

rneugeba:
This commit also seems to have other unrelated changes, like changing the SHA for init and containerd. Please don't mix them in one commit and provide them in a separate commit with an appropriate commit message

examples/minimal.yml
```diff
@@ -2,12 +2,12 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551
+  - linuxkit/init:63eed9ca7a09d2ce4c0c5e7238ac005fa44f564b
   - linuxkit/runc:b0fb122e10dbb7e4e45115177a61a3f8d68c19a9
-  - linuxkit/containerd:60e2486a74c665ba4df57e561729aec20758daed
-onboot:
+  - linuxkit/containerd:18eaf72f3f4f9a9f29ca1951f66df701f873060b
```
AkihiroSuda: Updating containerd (and init to the corresponding version) here is intentional. The old version does not start up for some reason:
```
/ # cat /var/log/dhcpcd.log
ctr: rpc error: code = 14 desc = grpc: the connection is unavailable
```

AkihiroSuda:
@rneugeba 

Thank you for the information.
Before duplicating dhcpd on-boots in all the yaml files and keeping them up-to-date, I think we should better implement https://github.com/moby/tool/issues/5 (inheritance) or https://github.com/moby/tool/issues/4 (include) in the builder.

So I'm temporarily closing this PR.

As mentioned above, other modifications are intentional, but I'll investigate and fix up in another issue/PR.

rneugeba:
Thanks. the `init` version needs to be matched with the `runc` and `containerd` version. There is at least one version of `init` which starts a hacked version of `containerd` instead of using `runc`. The hacked version of `containerd` has a modification to `ctr` to use a `runc` style rootfs but has some issues with logging and seems to be racey during start (#1660, #1661). We added it before DockerCon, so we could do `ctr list` for the demo. The plan is to create a `containerd` style imagestore with `moby` to properly (and fix the `containerd` issues.

PR #1703, opened 2017-04-25
General cleanup

rneugeba:
- Remove hyperkit/infrakit plugin as it is now in https://github.com/docker/infrakit
- Removed vendored packages only needed by the plugin
- Add macOS specific documentation
- consolidate files used for the DockerCon demo
  - move redis-os to ./examples
  - rename demo to etcd


justincormack:
There is also a line in the README "You can use go get -u github.com/linuxkit/linuxkit/src/cmd/infrakit-instance-hyperkit to get the hyperkit infrakit tool." - maybe move/remove/update this?

rneugeba:
A missed that, removed in the pushed update

justincormack:
yay less vendored stuff!

PR #1704, opened 2017-04-25
Switch fallback hostname to linuxkit-* from moby-*

ijc25:
This was missed when things were renamed.

The intention with this code was (apparently) to provide a (pseudo)unique
hostname in the case where something more specific was not provided (e.g. by
DHCP). Make this a little clearer by using '(none)' rather than 'linuxkit' as
the default, in the normal case this will be overwritten by something more
specific and if it isn't we will change it to something somewhat unique derived
from the MAC address (as before). nb: '(none)' is already used by Debian so I
think it is a safe choice as the sentinel value.

The use of both CONFIG_DEFAULT_HOSTNAME and the explicit /etc/hostname from
mkimage.sh is likely to be redundant in some cases, but neither seems to
completely cover all cases so keep both.

**- What I did**

Fixed the fallback to a pseudo-unique hostname in the absence of a more specific hostname having been provided.

**- How I did it**

Fixed the initscript and default hostname.

**- How to verify it**

Boot in a configuration where no more specific hostname is provided.

Fix fallback to pseudo-unique hostname.

Ahab, The Call of the Wretched Sea


ijc25:
I updated the YAML files with:
```
git grep -l linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551 | xargs sed -i -e 's,linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551,linuxkit/init:1c8cd75ec89313f4058b069449e9bac966cd96b1,g'
```

where `linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551` was currently in use by `linuxkit.yml` and `linuxkit/init:1c8cd75ec89313f4058b069449e9bac966cd96b1` was the output of `make -C pkg/init/ tag` IOW I updated everything which was already using the same thing as `linuxkit.yml` and left the others alone.

PR #1707, opened 2017-04-25
Add additional kernel options

### LINKED ISSUE #1689: support devicemapper for encrypted swap or other block or general usage
devicemapper support, of course, requires the `dm` kernel module. If you want to do anything like, e.g. encrypted swap (or any other encrypted or loopback), you need devicemapper, which needs `dm`. How to get it done? Is there a sane way? I don't relish the idea of building my own kernel to replace `linuxkit/kernel` (actually, that would be rather silly of me to do). 

Is there a way to get the module in? Or is there an alternate path? @justincormack has suggested libguestfs to use for mounting when making build images, but in this case I need mounted encrypted devices not as a on-off but ongoing to run the system.

justincormack:
- XFS filesystem support, as we plan to support this
- Quota support, as XFS has good support
- NVMe PCI support
- Per file encryption
- Device Mapper support, with main options
- BTRFS support (as a module, as it slows boot otherwise by several seconds).

I haven't built these yet... need to set up my signing keys.

fix #1689


kernel/Makefile
```diff
@@ -20,18 +20,18 @@ all:	bzImage tag
...
 ifeq ($(KERNEL),v4.10)
 KERNEL_VERSION=4.10.10
-IMAGE_VERSION=$(KERNEL_VERSION)-0
+IMAGE_VERSION=$(KERNEL_VERSION)-1
 IMAGE_MAJOR_VERSION=4.10.x
 DEPS=Dockerfile.4.10 Makefile kernel_config kernel_config.debug patches-4.10
 else
 KERNEL_VERSION=4.9.22
-IMAGE_VERSION=$(KERNEL_VERSION)-0
+IMAGE_VERSION=$(KERNEL_VERSION)-1
```
AkihiroSuda: seems deduplicatable across kernel versions

rneugeba: not sure I understand? This will go away very soon anyway...

AkihiroSuda: You dont need to duplicate this line 3 times.
This line could be moved to outside of if...endif block

rneugeba: as I said, this will go away soon. the commit is about something else, so it should be done in a separate commit...which is not necessary given that we change it soon

AkihiroSuda: ok

kernel/kernel_config
```diff
@@ -2892,14 +2930,20 @@ CONFIG_EXPORTFS=y
 # CONFIG_EXPORTFS_BLOCK_OPS is not set
 CONFIG_FILE_LOCKING=y
 CONFIG_MANDATORY_FILE_LOCKING=y
-# CONFIG_FS_ENCRYPTION is not set
+CONFIG_FS_ENCRYPTION=y
```
riyazdf: to confirm my understanding: this option enables the possibility for fs encryption, but doesn't actually perform it right?  The docs page is a bit terse but seems to read that way

kernel/kernel_config
```diff
@@ -1684,7 +1686,32 @@ CONFIG_PATA_SIS=y
...
+# CONFIG_DM_CACHE is not set
+# CONFIG_DM_ERA is not set
+# CONFIG_DM_MIRROR is not set
+# CONFIG_DM_RAID is not set
+# CONFIG_DM_ZERO is not set
+# CONFIG_DM_MULTIPATH is not set
+# CONFIG_DM_DELAY is not set
+# CONFIG_DM_UEVENT is not set
+# CONFIG_DM_FLAKEY is not set
+# CONFIG_DM_VERITY is not set
```
riyazdf: I'll probably circle back and enable dm-verity on a followup PR :)

rneugeba:
needs a rebase/change in kernel suffix due to https://github.com/linuxkit/linuxkit/pull/1704

justincormack:
As discussed, removed the version bumps as the kernel build process is about to change, images will be updated then.

Issue #1698, opened 2017-04-24
Network Options for the qemu runner


fntlnz:
Right now the qemu runner has a very limited set of options available, but the thing I most need is the ability to specify network options to the VM.

The  hyperkit runner for example uses a  VPNKit key generated from the IP address to allow you to bridge a network interface from the host to the VM.

In qemu things would be easier because it already has all the needed options to do bridging/port forwarding, for example, I achieved a simple 8080:80 port forwarding in my fork like this:

```go
qemuArgs = append(qemuArgs, "-net", "user,hostfwd=tcp::8080-:80")
qemuArgs = append(qemuArgs, "-net", "nic")
```

Also, I think that there should be some shortcut like hyperkit's `-ip` to allow some sort of consistency between different runners when the runner is not specified and the default is used.

Wdyt?

fntlnz:
BTW, I would be glad to submit a pr to achieve this and discuss on the code if there's interest

justincormack:
cc @thebsdbox @dave-tucker

thebsdbox:
Sure sounds like a good idea, I was a little concerned about wrapping too many of the options for `qemu` and i'm not sure if it would be easier just having the capability to pass through a string containing any additional/advanced params/arguments. Would simplify the use of multiple host forwarded ports. 

Happy to discuss the code.

alexellis:
For developer experience a unified CLI would be helpful - so that `-ip` works the same for either back-end where applicable. Docker-machine had a "plugin" model - does this make sense for `moby`?

rneugeba:
Networking is quite specific to the VMM used and the host platform so I doubt we'll get consistent options across different runner.

The PR to expose ports with qemu has been merged.

Can this be closed, or is there something in the discussion I missed?

fntlnz:
@rneugeba completely agree :+1: but keeping an eye on having the most consistent possible set of options

Closing this

PR #1713, opened 2017-04-26
Port forwarding in qemu runner

fntlnz:
I added the `-publish` option to the `qemu` runner for the linuxkit cmd

**Usage**

```bash
$ linuxkit run qemu -publish 8080:80 -publish 8081:80 example
```

or even

```bash
linuxkit run qemu -publish 1111:1111/udp -publish 8081:80/tcp example
```

**Help output**
```
Please specify the prefix to the image to boot
USAGE: linuxkit run qemu [options] prefix

'prefix' specifies the path to the VM image.

Options:
  -arch string
        Type of architecture to use, e.g. x86_64, aarch64 (default "x86_64")
  -cpus string
        Number of CPUs (default "1")
  -diskpath string
        Path to disk image to use
  -disksize string
        Size of disk to create, only created if it doesn't exist
  -fwpath string
        Path to OVMF firmware for UEFI boot (default "/usr/share/ovmf/bios.bin")
  -gui
        Set qemu to use video output instead of stdio
  -iso
        Set Legacy BIOS boot from 'prefix'.iso
  -kernel
        Set boot using 'prefix'-bzImage/-initrd/-cmdline (default true)
  -mem string
        Amount of memory in MB (default "1024")
  -publish value
        Publish a vm's port(s) to the host (default [])
  -uefi
        Set UEFI boot from 'prefix'-efi.iso
```

## Related work
This work is related to #1698 - this PR addresses only half the issue.
This is just port forwarding other part would be to do linux briding in order to have a veth pair within the guest vm and being able to access it directly.

## Moar

This is just some initial implementation of my idea, for example I put all the custom flags and build functions in the same `run_qemu` file because didn't know if there's a better place, any suggestion is warmly welcomed :angel: 

## Also

I don't know **how much** I'm breaking, there's any integration test suite I'm not aware of?

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 )
@@ -16,21 +18,22 @@ const QemuImg = "linuxkit/qemu:17f052263d63c8a2b641ad91c589edcbb8a18c82"
 
 // QemuConfig contains the config for Qemu
 type QemuConfig struct {
-	Prefix        string
-	ISO           bool
-	UEFI          bool
-	Kernel        bool
-	GUI           bool
-	DiskPath      string
-	DiskSize      string
-	FWPath        string
-	Arch          string
-	CPUs          string
-	Memory        string
-	KVM           bool
-	Containerized bool
-	QemuBinPath   string
-	QemuImgPath   string
+	Prefix         string
+	ISO            bool
+	UEFI           bool
+	Kernel         bool
+	GUI            bool
+	DiskPath       string
+	DiskSize       string
+	FWPath         string
+	Arch           string
+	CPUs           string
+	Memory         string
+	KVM            bool
+	Containerized  bool
+	QemuBinPath    string
+	QemuImgPath    string
+	PublishedPorts []string
 }
 
 func runQemu(args []string) {
@@ -60,6 +63,9 @@ func runQemu(args []string) {
 	qemuCPUs := qemuFlags.String("cpus", "1", "Number of CPUs")
 	qemuMem := qemuFlags.String("mem", "1024", "Amount of memory in MB")
 
+	publishFlags := multipleFlag{}
+	qemuFlags.Var(&publishFlags, "publish", "Publish a vm's port(s) to the host (default [])")
+
 	if err := qemuFlags.Parse(args); err != nil {
 		log.Fatal("Unable to parse args")
 	}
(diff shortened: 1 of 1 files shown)


fntlnz:
ping  @thebsdbox @dave-tucker @justincormack

src/cmd/linuxkit/run_qemu.go
```diff
@@ -15,21 +17,22 @@ const QemuImg = "linuxkit/qemu:17f052263d63c8a2b641ad91c589edcbb8a18c82"
...
-	DiskSize      string
-	FWPath        string
-	Arch          string
-	CPUs          string
-	Memory        string
-	KVM           bool
-	Containerized bool
-	QemuBinPath   string
-	QemuImgPath   string
+	Prefix         string
```
rneugeba: what changed here?

src/cmd/linuxkit/run_qemu.go
```diff
@@ -283,3 +295,77 @@ func buildPath(prefix string, postfix string) string {
 	}
 	return path
 }
+
+type multipleFlag []string
```
dave-tucker: I'd be tempted to move this to `util.go`

src/cmd/linuxkit/run_qemu.go
```diff
@@ -283,3 +295,77 @@ func buildPath(prefix string, postfix string) string {
...
+
+type publishedPorts struct {
+	guest    int
+	host     int
+	protocol string
+}
+
+func (f *multipleFlag) String() string {
```
dave-tucker: And these functions too...

rneugeba:
I don't think we will standardise on one set of arguments across all backends. They are just too different. It's good to have a "common" set, though and if more than one backend supports a particular feature than obviously they should use the same option name...

It would be good if the commit message contained the example or even the help text. The PR text is not easily available via the git command line...

src/cmd/linuxkit/run_qemu.go
fntlnz: Now it works in a container :tada: , haven't noticed that there was a way to run it in a container, could be worth to make it configurable via a flag instead than just checking if qemu is installed?

Example use case: for this pr not having that flag required me to hardcode `true` for containerized

fntlnz:
@rneugeba agree for the backends, opened  an issue regarding this #1732 - also because things are diverging very quickly

I added the message to the pr too :)

fntlnz:
There's a logic issue with container forwardings, please don't merge this yet

justincormack:
You will also need to rebase to fix CI I think

fntlnz:
@justincormack fixed, rebased and pushed - I'm wondering about writing some integration tests, wdyt?

justincormack:
Tests are always good!

cc @dave-tucker

dave-tucker:
yes we should probably have some integration tests but i think we should wait until we have a proper framework for running them. Overall this LGTM!

dave-tucker:
Note: This will probably need a rebase to pick up the latest iteration of CI changes

justincormack:
yes, rebase should hopefully fix CI...

fntlnz:
@dave-tucker rebased :rabbit:  CI Fixed too!

@justincormack I see that something's happening around testing like here: https://github.com/linuxkit/linuxkit/tree/master/test/cases

dave-tucker:
@fntlnz yes indeed there is!
I've got a PR brewing with a proper test suite.
See: https://github.com/dave-tucker/linuxkit/tree/rtf-cases/test/cases for a sneak preview

Hopefully we can add a nice integration test in`test/cases/010_platforms/000_qemu/` soon, but I don't think this should prevent merging this PR.

PR #1714, opened 2017-04-26
Removed moby from the linuxkit's push help

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/src/cmd/linuxkit/push.go
+++ b/src/cmd/linuxkit/push.go
@@ -3,19 +3,21 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
 )
 
 func pushUsage() {
-	fmt.Printf("USAGE: %s push [backend] [options] [prefix]\n\n", os.Args[0])
+	invoked := filepath.Base(os.Args[0])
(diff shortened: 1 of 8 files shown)


rneugeba:
actually, could you change this to what the `moby` tool does here: https://github.com/moby/tool/blob/master/cmd/moby/main.go#L48

fntlnz:
@rneugeba at first instance I implemented it in that exact way but after I looked at [the other commands](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run.go#L24) and for consistency with them I choose to "hardcode" it.

If you prefer we can change those too

wdyt?

rneugeba:
@fntlnz ah, I didn't see that this is also hardcoded elsewhere. Must have happened during the split. Would you mind changing the other hardcoded instances with `filepath.Base(os.Args[0])`? I think that is better/cleaner. But do hope we won't rename this again ;)

fntlnz:
@rneugeba did it, it's quite a simple command line. I actually prefer this approach  instead of using some more complete but even heavy and complex command line library *at this stage*

rneugeba:
@fntlnz thanks for the extra work beyond the initial fix

Issue #1694, opened 2017-04-24
Proposal: example for swap created and configured by onboot


deitch:
I have had a need for an instance with swapfile, configured at build time. The sample at https://github.com/deitch/linuxkit-swap includes that, as well as adding conditions (e.g. "only make swapfile if `/var/external` is its own filesystem and >20GB"). Encrypted swap is in partially, but largely depends on devicemapper.

justincormack:
We do want a generic swap service in the repo. We currently use swap on the desktop (but not cloud) Docker Editions so will use for those anyway. I think it selects if there is enough disk space and not lots of memory, so would be great to do a PR for this.

justincormack:
(it used to be in the format code in d4m/d4w but I removed it to make things more modular so we dont have anything right now)

deitch:
Sure, I will PR it. Wait for resolving encrypted swap first? Or as is?

justincormack:
I think we can add encryption later.

deitch:
Closing in favour of #1715

PR #1715, opened 2017-04-26
First cut of swap image and example

deitch:
**- What I did**

Add support for creating swap files in linuxkit image

**- How I did it**

add onboot image `linuxkit/swap`, which contains script to read params and set up swapfile

**- How to verify it**

Run with various options:

* no `--path` provided
* invalid swapfile
* no mounted disk (cannot use in `/var` because it is tmpfs
* disk too small
* check conditions

Add image with support for swapfiles to be used in onboot

eclectus

** Notes **
1. I used the Makefile from [pkg/mount](https://github.com/linuxkit/linuxkit/tree/master/pkg/mount) as a template. I was unsure the level of adherence to the shahash versioning scheme, but figured something that already is in and is similar is a good start.
2. Needs encryption support, which will come when devmapper works. See https://github.com/linuxkit/linuxkit/issues/1689
3. Closes https://github.com/linuxkit/linuxkit/issues/1694 per @justincormack

### DIFF
--- a/examples/swap.yml
+++ b/examples/swap.yml
@@ -1,0 +1,83 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 console=tty0 page_poison=1"
+init:
+  - linuxkit/init:42fe8cb1508b3afed39eb89821906e3cc7a70551
+  - linuxkit/runc:b0fb122e10dbb7e4e45115177a61a3f8d68c19a9
+  - linuxkit/containerd:60e2486a74c665ba4df57e561729aec20758daed
+  - linuxkit/ca-certificates:eabc5a6e59f05aa91529d80e9a595b85b046f935
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:2cf2f9d5b4d314ba1bfc22b2fe931924af666d8c"
+    net: host
+    pid: host
+    ipc: host
+    capabilities:
+     - CAP_SYS_ADMIN
+    readonly: true
+  - name: binfmt
+    image: "linuxkit/binfmt:8881283ac627be1542811bd25c85e7782aebc692"
+    binds:
+     - /proc/sys/fs/binfmt_misc:/binfmt_misc
+    readonly: true
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:48e249ebef6a521eed886b3bce032db69fbb4afa"
+    binds:
+     - /var:/var
+     - /tmp/etc:/etc
+    capabilities:
+     - CAP_NET_ADMIN
+     - CAP_NET_BIND_SERVICE
+     - CAP_NET_RAW
+    net: host
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+  - name: format
+    image: "linuxkit/format:53748000acf515549d398e6ae68545c26c0f3a2e"
+    binds:
+     - /dev:/dev
+    capabilities:
+     - CAP_SYS_ADMIN
+     - CAP_MKNOD
+  - name: mount
+    image: "linuxkit/mount:d2669e7c8ddda99fa0618a414d44261eba6e299a"
+    binds:
+     - /dev:/dev
+     - /var:/var:rshared,rbind
+    capabilities:
+     - CAP_SYS_ADMIN
+    rootfsPropagation: shared
+    command: ["/mount.sh", "/var/external"]
+  - name: swap
+    image: "linuxkit/swap:68deee354ec4bbed405c5efedb25d60bb799b07d"
+    net: host
+    pid: host
+    capabilities:
+     - CAP_SYS_ADMIN
+    readonly: true
+    binds:
+     - /var:/var
+     - /dev:/dev
+    command: ["swap.sh", "--path", "/var/external/swap", "--size","1G"]
+services:
+  - name: rngd
+    image: "linuxkit/rngd:3dad6dd43270fa632ac031e99d1947f20b22eec9"
+    capabilities:
+     - CAP_SYS_ADMIN
+    oomScoreAdj: -800
+    readonly: true
+  - name: nginx
+    image: "nginx:alpine"
+    capabilities:
+     - CAP_NET_BIND_SERVICE
+     - CAP_CHOWN
+     - CAP_SETUID
+     - CAP_SETGID
+     - CAP_DAC_OVERRIDE
+    net: host
+trust:
+  image:
+    - linuxkit/kernel
+outputs:
+  - format: kernel+initrd
+  - format: iso-bios
+  - format: iso-efi
--- a/pkg/swap/Dockerfile
+++ b/pkg/swap/Dockerfile
@@ -1,0 +1,13 @@
+FROM alpine:3.5
+
+# swap command - only minimal Alpine install
+
+# use util-linux to get swapfile utils
+#   at some point, hopefully use our own mkswap and swapon
+RUN apk add --update util-linux
+
+# add the entrypoint and make it executable
+COPY . ./
+
+# ENTRYPOINT
+ENTRYPOINT swap.sh
--- a/pkg/swap/swap.sh
+++ b/pkg/swap/swap.sh
@@ -1,0 +1,135 @@
+#!/bin/sh
+
+set -e
+
+while [ $# -gt 1 ]; do
+	key="$1"
+
+	case $key in
+		--debug)
+			set -x
+			;;
+		--path)
+			path="$2"
+			shift # past argument
+			;;
+		--size)
+			size="$2"
+			shift # past argument
+			;;
+		--encrypt)
+			ENCRYPT=true
+			;;
+		--condition)
+			CONDITIONS="$CONDITIONS $2"
+			shift
+			;;
+		*)
+			echo "Unknown option passed to swapmaker: $key"      # unknown option
+			exit 1
+			;;
+	esac
+	shift # past argument or value
+done
+
+function disksize_to_count {
+	local blocksize=$1
+	local origsize=$2
+	local ret
+	case $origsize in
+		*G)
+			ret=$(( ${origsize%%G} * 1024 * 1024 * 1024 ))
+			;;
+		*M)
+			ret=$(( ${origsize%%M} * 1024 * 1024 ))
+			;;
+		*K)
+			ret=$(( ${origsize%%K} * 1024 ))
+			;;
+		*)
+			ret=$origsize
+			;;
+	esac
+	ret=$(( $ret / $blocksize ))
+	echo $ret
+}
+
+
+## make sure path is valid
+if [ -z "${path}" ]; then
+	echo "swap: --file <path> must be defined"
+	exit 1
+fi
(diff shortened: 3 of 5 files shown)


deitch:
> Please sign your commits

Oh oops. Doing now.

examples/swap.yml
```diff
@@ -0,0 +1,86 @@
...
+    image: "linuxkit/swap:68deee354ec4bbed405c5efedb25d60bb799b07d"
+    net: host
+    pid: host
+    capabilities:
+     - CAP_SYS_ADMIN
+    readonly: true
+    binds:
+     - /var:/var
+     - /dev:/dev
+    command: ["swap","--path","/var/external/swap","--size","1G"]
```
justincormack: very nitpicky, but I think we usually have spaces after the commas....

examples/swap.yml
```diff
@@ -0,0 +1,86 @@
...
+  - name: nginx
+    image: "nginx:alpine"
+    capabilities:
+     - CAP_NET_BIND_SERVICE
+     - CAP_CHOWN
+     - CAP_SETUID
+     - CAP_SETGID
+     - CAP_DAC_OVERRIDE
+    net: host
+files:
```
justincormack: can you delete the files section from the example (it is only in the top level one to show what you can do but its somewhat confusing...)

pkg/swap/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:3.5
+
+# swap command - only minimal Alpine install
+MAINTAINER Avi Deitcher <https://github.com/deitch>
```
justincormack: `MAINTAINER` is deprecated, and its not helpful in this context can you remove...

deitch:
Nitpick away, I updated the spaces after commas. :-)

pkg/swap/Dockerfile
deitch: Oh, yeah. That is what you get from copying over. :-) Fixing.

pkg/swap/swap.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
```
justincormack: please use `sh` not `bash` - if a script can't be written in Posix shell it should be written in a real programming language...

examples/swap.yml
deitch: Yup. Again, mostly copying/pasting.

pkg/swap/swap.sh
```diff
@@ -0,0 +1,122 @@
...
+			;;
+	esac
+done
+## if a condition failed:
+### Required? exit 1
+### Else? exit 0
+
+
+## Allocate the file
+fallocate -l $size $path
```
justincormack: I usually use `dd` as from memory performance of sparse swap files is not that great.

pkg/swap/swap.sh
deitch: Out of curiosity, why? I don't care all that much (other than having to sh-ify the script), but lighter and easier to install bash than, say, python or ruby or node or whatever.

pkg/swap/swap.sh
deitch: `dd` is quite slow to initialize, which is why I used `fallocate`. But you know far more about this stuff than I do, so happy to accommodate. It does slow startup time as it initializes the blocks. Worth the trade-off?

pkg/swap/swap.sh
justincormack: absolutely not python or ruby or node, no, but we do plan to rewrite most of the shell in linuxkit in Rust/Go actually. We already have too much shell code. Also the fewer things we need to install the more we will be able to share between images once we have the `containerd` image store. Also installing `bash` in Alpine pulls in a lot of dependencies (ncurses etc), and bloats Alpine up to 12MB...

deitch:
Hey, no one comments on my choice of an Eclectus for the animal? :-)

Whose idea to ask people to put up an animal, by the way?

justincormack:
It is from the `moby/moby` template, I don't know when it was added to `docker`though maybe @thaJeztah does. The Docker ones have their own twitter feed now https://twitter.com/DockerCuties

pkg/swap/swap.sh
deitch: > bloats Alpine up to 12MB

Is it that much? Ouch. As it is, swap has to install `util-linux` because it needs `mkswap` and `swapon`. and eventually will need `cryptsetup`. I viewed bash as small potatoes comparatively. I wish I could get *just* `mkswap` and `swapon` without all of `util-linux`; if you know how, happy to take a pointer.

I guess I will try to rewrite to `sh`.

deitch:
We gotta get LinuxKit animals into that feed. If I had known, I would have put in a picture of the Eclectus we used to own (and the cockatiel we do now).

pkg/swap/swap.sh
justincormack: I think the advantage is you get consecutive allocations if you do it up front with `dd` (which is why it is slow). On spinning rust that was important. On SSD it probably helps a bit too as real write units are 128k or so too. So probably still worth paying the cost upfront.

On Docker for mac we only create the file if it doesnt already exist so you only pay the cost on first boot.

pkg/swap/swap.sh
justincormack: Its probably worth writing `mkswap` in Go/Rust at some point, I suspect it is pretty trivial. `swapon` is a single system call, so thats easy. I don't know what the crypto setup involves.

deitch: Ok, I think I got it sh-ified. Certainly it will be good enough for now. When we git `mkswap` and `swapon` as golang calls, I will look at rewriting this as well, and then it is a single binary.

deitch: sh done and pushed.

pkg/swap/swap.sh
deitch: OK, replaced `fallocate` with `dd`.

deitch:
I think I got all of the comments in. Have another look-see please?

pkg/swap/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+# swap command - only minimal Alpine install
+
+# use util-linux to get swapfile utils
+#   at some point, hopefully use our own mkswap and swapon
+RUN apk add --update util-linux
+
+# add the entrypoint and make it executable
+COPY . ./
+RUN chmod +x swap.sh
```
justincormack: can you chmod +x the `swap.sh` in the git repo then you can remove this...

pkg/swap/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN ln -s swap.sh swap
```
justincormack: also just include it as the filename you want...

pkg/swap/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+# ENTRYPOINT
+ENTRYPOINT swap
```
justincormack: we dont currently support `ENTRYPOINT` in the config file, ie if you specify a `command` it replaces the whole line, so this is probably not much use as you always need to specify arguments. We should probably fix this though...

pkg/swap/swap.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/sh
+
+set -e
+
+while [ $# -gt 1 ]; do
+	key="$1"
+
+	case $key in
+		--debug)
```
justincormack: indentation a bit inconsistent here

pkg/swap/swap.sh
```diff
@@ -0,0 +1,138 @@
...
+	    shift # past argument
+	    ;;
+    --size)
+	    size="$2"
+	    shift # past argument
+	    ;;
+    --encrypt)
+	    ENCRYPT=true
+	    ;;
+		--condition)
```
justincormack: and here

pkg/swap/Dockerfile
deitch: I did, this is more of a safety thing. Too many years seeing too many mistakes leaves me happy with a no-cost additional safety net. You really care?

pkg/swap/Dockerfile
deitch: Yeah, but the hashing algorithm you used depends on having everything in the container. I like `swap.sh` because of vim and Atom and other editors, but `swap` as a command is cleaner. :-)

pkg/swap/Dockerfile
deitch: Probably should. Again, it doesn't hurt.

pkg/swap/swap.sh
deitch: Ouch. Atom. Fixing.

deitch: Atom seems to be doing indentation inconsistently. Mostly tabs, sometimes spaces.

pkg/swap/Dockerfile
justincormack: I just like the uniformity that all Dockerfiles should be 
```
FROM x
COPY . ./
CMD ...
```
and rarely anything else where possible - the less processing of the inputs the better...

pkg/swap/Dockerfile
justincormack: we are going to change the hashing to be based on the git tree hash I think once some preconditions are sorted out.

justincormack:
Can you squash the commits and make Gordon happy again?

pkg/swap/Dockerfile
deitch: OK, I cleaned that up

deitch:
> Can you squash the commits and make Gordon happy again?

And I specifically kept them separate to make it easier to see. OK, fine.

deitch:
All squashed.

deitch:
Did I miss anything else?

justincormack:
The commit message is a bit messy and Gordon still doesnt think its signed properly....

pkg/swap/swap.sh
```diff
@@ -0,0 +1,135 @@
...
+			;;
+	esac
+	ret=$(( $ret / $blocksize ))
+	echo $ret
+}
+
+
+## make sure path is valid
+if [ -z "${path}" ]; then
+    echo "swap: --file <path> must be defined"
```
justincormack: these indents are inconsistent too, bad Atom...

pkg/swap/swap.sh
```diff
@@ -0,0 +1,135 @@
...
+
+
+## Allocate the file
+dd if=/dev/zero of=$path bs=1024 count=$(disksize_to_count 1024 $size)
+chmod 0600 $path
+
+## was it encrypted? use cryptsetup and get the mapped device
+if [ "$ENCRYPT" == "true" ]; then
+	# might need
+	#loop=$(losetup -f)
```
justincormack: and here

justincormack:
Apart from indent and signing issues LGTM

justincormack:
Thanks!

deitch:
Excellent! Next step is getting encrypted swap working.

thaJeztah:
The "cute animal" picture was a tradition started by @calavera. Decided to continue the tradition here as it makes pull requests a lot more fun to review; you can view them all on Twitter here; https://twitter.com/DockerCuties

deitch:
I like the idea @thaJeztah . Add a little humanity ("animality"? "Bachmanity"? ha! ) to the process.

Issue #1480, opened 2017-04-04
Convert package builds to multi-stage builds


rn:
17.05 will have a new feature call [multi-stage build](http://blog.alexellis.io/mutli-stage-docker-builds/). This seems ideal for building packages and we should convert the package build to it once available

rn:
Most/All packages are now on multi-stage builds. Closing

Issue #1074, opened 2017-01-20
sign built images in CI and push to hub


justincormack:
Once https://github.com/docker/moby/issues/957 is done let us sign them in CI.

cc @riyazdf

dave-tucker:
We had a little chat about this when I went to Cambridge and we decided on the following...

Changes to the `tools` directory will be handled manually by those with access to the org in hub.

Changes to `pkg` and `kernel` should be handled by CI

For `kernel`, a bigger machine should be spawned to make the build faster!

To make this easy for CI, there will be a top-level `make check` target which will delegate to `make check` in each package.

This target will determine if the image is available on hub or not.
If it doesn't exist, it will be built locally with `make tag` on a PR build
If this is a master or tag build, `make push` will be used

Assuming the push is successful and notary has been configured correctly, then CI should be able to issue a `make sign` afterwards.

rneugeba:
thanks for the summary.

PR #1720, opened 2017-04-26
Major rewrite of the kernel build

### LINKED ISSUE #1430: Improve kernel build process
The current build process for the kernel has a number of issues
- We should check before building the kernel (via the sha1) if it has already been build. Don't build if it has
- The above implies to also push an image with a sha1 tag (on top of the specific kernel version and the more generic version), ie each kernel should be on hub as, e.g.: `:4.9.21-0`, `:4.9.x`, `:<sha1>`.
- Building for different kernels requires a `make clean` because otherwise the build will just pickup potentially old and wrong vmlinux and inittrds. The build artefacts should go into different subdirs prefixed by `KERNEL` or `KERNEL_VERSION`

rneugeba:
- Move kernel-build container to ./tools (and remove `./base`)
- Change the kernel build system
    - Use multi-stage builds
    - Use a single Dockerfile for all kernels
      Kernel version and series are passed in as arguments
    - Use a separate kernel config per kernel version
      These have been copied from kernel_config and ran
      through oldconfig to tidy them up
    - Rename patch directories
    - Refactor the Makefile to use a template
      - Allows building of all kernels without arguments to make
      - Use git tree hash as the image tag
      - Don't build the image if the tag already exists
- Use `-linuxkit` as the kernel version suffix
- Update docs

resolves #1430
wip #1074
wip #1480

image


rneugeba:
Note, I haven't pushed the images, but can kick it off before merging, once I get a LGTM of sort.

kernel/Dockerfile
```diff
@@ -18,7 +19,7 @@ RUN if [ $DEBUG -ne "0" ]; then \
     fi
 
 # Apply local patches
-COPY patches-4.9 /patches
+COPY patches-${KERNEL_SERIES} /patches
```
justincormack: There was an issue in the past if there were no patches, as the directory would be empty, so not in git.

kernel/Makefile
```diff
@@ -1,92 +1,66 @@
...
-clean:
-	rm -rf x86_64 lib usr sbin kernel.tag Dockerfile.media bzImage kernel.tar
+#
+# Build Targets
+#
+$(eval $(call kernel,4.10.10,4.10.x))
+$(eval $(call kernel,4.9.22,4.9.x))
+$(eval $(call kernel,4.4.61,4.4.x))
 
 .DELETE_ON_ERROR:
```
justincormack: dont really need this is we dont output any files I guess

justincormack:
Two minor comments but LGTM generally

docs/kernels.md
```diff
@@ -12,10 +30,13 @@ Patches are located in `kernel/patches-<kernel version>` and should follow these
   which they are applied.
 - If patches are cherry-picked, they *must* be cherry-picked with `-x`
   to contain the original commit ID.
+- If patches are from a different git tree (other than the stable
+  tree), or from a mailing list posting they should container a
```
riyazdf: nit typo: `contain an`

kernel/Makefile
```diff
@@ -1,92 +1,66 @@
...
-# We push the image to hub twice, once with the full kernel version of
-# "mobylinux/kernel:<kernel version>.<major version>.<minor version>-<n>",
-# where "<n>" is a monotonically increasing config number, and as
-# "mobylinux/kernel:<kernel version>.<major version>.x". This version
-# number is stored in IMAGE_VERSION.
-#
-# We expect most users to us the "<kernel version>.<major version>.x"
-# variant as this simply is the latest version of a given major kernel
-# version. This version number is stored in IMAGE_MAJOR_VERSION.
+# This build the support LinuxKit kernels. Kernels are wrapped up in a
```
riyazdf: nit typo: `This builds the supported...`

kernel/Dockerfile
rneugeba: there doesn't seem to be an easy solution to this. There were proposals (e.g. https://github.com/moby/moby/issues/13045) and patches (e.g. https://github.com/moby/moby/pull/13813) to ignore errors but they weren't accepted.

This is also not a regression to what we had before, and I would suggest to ignore it for now and address, if we hit it. A possible workaround would be always create the directory and then later check that `*.patches` is not empty...

rneugeba:
Address the nits (thanks for spotting) except the patches directory one. I've also veryfied on a different computer that the git tree hash returns the same results.

Kicked off the builds and will merge tomorrow morning unless @justincormack wants the empty/non-extistant patch-dir issue addressed before

kernel/Dockerfile
justincormack: Well I think it is a regression, that was one reason why I dah these files split, as I ran into this before testing a new version... You can do the mkdir in the Makefile I guess.

justincormack:
That can wait until its an issue again...

rneugeba:
had to force push a minor tweak (local build image wasn't prefixed with `linuxkit/`) but all kernels build and pushed and verified with the 4.9 kernel and `moby build -pull` that it's ok.

merging

PR #1721, opened 2017-04-26
Signing targets for binfmt, rngd, compilers, toybox, tini

riyazdf:
Set up notary delegations and Makefile `sign` targets for:
```
docker.io/linuxkit/binfmt
docker.io/linuxkit/rngd
docker.io/linuxkit/go-compile
docker.io/linuxkit/c-compile
docker.io/linuxkit/toybox-media
docker.io/linuxkit/tini
```

Also updated yml files and Makefiles to verify signatures where appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,15 +18,13 @@ endif
 PREFIX?=/usr/local/
 
 bin/moby: | bin
-	DOCKER_CONTENT_TRUST=1 docker pull $(GO_COMPILE)
-	DOCKER_CONTENT_TRUST=1 docker run --rm --log-driver=none $(CROSS) $(GO_COMPILE) --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_moby_bin.tar
+	docker run --rm --log-driver=none $(CROSS) $(GO_COMPILE) --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_moby_bin.tar
 	tar xf tmp_moby_bin.tar > $@
 	rm tmp_moby_bin.tar
 	touch $@
 
 LINUXKIT_DEPS=$(wildcard src/cmd/linuxkit/*.go) Makefile vendor.conf
 bin/linuxkit: $(LINUXKIT_DEPS) | bin
-	DOCKER_CONTENT_TRUST=1 docker pull $(GO_COMPILE)
 	tar cf - vendor -C src/cmd/linuxkit . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_linuxkit_bin.tar
 	tar xf tmp_linuxkit_bin.tar > $@
 	rm tmp_linuxkit_bin.tar
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ all: default
 VERSION="0.0" # dummy for now
 GIT_COMMIT=$(shell git rev-list -1 HEAD)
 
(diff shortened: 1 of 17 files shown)


pkg/binfmt/Makefile
```diff
@@ -23,6 +23,12 @@ push: tag
 	rm -f hash
 	docker rmi $(IMAGE):build || true
```
justincormack: this Makefile doesn't `pull` the base image, so is missing the trust check on that

justincormack:
Not sure why CI is failing

pkg/binfmt/Makefile
```diff
@@ -10,6 +11,7 @@ DEPS=Dockerfile Makefile main.go $(wildcard etc/binmft.d/*)
 # the hashing stuff soon anyway, so there is little point in faffing
 # about with computing a hash during the build and get it out.
 hash: $(DEPS)
+	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
```
justincormack: put this under `tag` where the build is done not here, it makes more sense.

pkg/rngd/Makefile
```diff
@@ -2,10 +2,12 @@
 default: push
 
 IMAGE=rngd
+BASE=linuxkit/c-compile:f52f485825c890d581e82a62af6906c1d33d8e5d
 SHA_IMAGE=alpine:3.5@sha256:dfbd4a3a8ebca874ebd2474f044a0b33600d4523d03b0df76e5c5986cb02d7e8
 DEPS=Dockerfile Makefile
 
 hash: $(DEPS)
+	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
```
justincormack: and here

Makefile
```diff
@@ -18,13 +18,15 @@ endif
 PREFIX?=/usr/local/
 
 bin/moby: | bin
-	docker run --rm --log-driver=none $(CROSS) $(GO_COMPILE) --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_moby_bin.tar
+	DOCKER_CONTENT_TRUST=1 docker pull $(GO_COMPILE)
+	DOCKER_CONTENT_TRUST=1 docker run --rm --log-driver=none $(CROSS) $(GO_COMPILE) --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_moby_bin.tar
```
justincormack: not sure we should set DCT for the tool builds, as it breaks ability to compile offline, and this is not part of the product. Users can do this themselves, but long term we will provide other ways to get the binaries that have signing, eg package managers etc.

Makefile
```diff
@@ -18,13 +18,15 @@ endif
...
 	tar xf tmp_moby_bin.tar > $@
 	rm tmp_moby_bin.tar
 	touch $@
 
 LINUXKIT_DEPS=$(wildcard src/cmd/linuxkit/*.go) Makefile vendor.conf
 bin/linuxkit: $(LINUXKIT_DEPS) | bin
+	DOCKER_CONTENT_TRUST=1 docker pull $(GO_COMPILE)
```
justincormack: and here

pkg/binfmt/Makefile
```diff
@@ -14,7 +15,8 @@ hash: $(DEPS)
 
 tag: hash
 	docker pull linuxkit/$(IMAGE):$(shell cat hash) || \
```
justincormack: this pull should have DCT too? Then we replace unsigned version with signed one. If its not signed its not real!

justincormack: this should set DCT at least in the case we are doing `sign` else if there is an existing unsigned tag it wont be replaced. Maybe means splitting signed and unsigend cases though.

Makefile
justincormack: actually for `moby` you cant currently build offline i guess but for linuxkit you can.

pkg/binfmt/Makefile
riyazdf: split out a `signed-tag` for all `sign` targets

pkg/binfmt/Makefile
riyazdf: fixed

PR #1727, opened 2017-04-27
scripts: Add simple Powershell script to manage  Hyper-V VMs

rneugeba:
This script can create/start/stop/remove Hyper-V VMs. The
VMs are configured as Gen2 VMs which boot from EFI ISOs.

Network and Disk configuration is optional.


justincormack:
Whats up with CI?

rneugeba:
dunno

Issue #635, opened 2016-10-26
better way to manage kernel config


justincormack:
Want to specify the config options that actually matter (versioned by kernel version perhaps) and then generate the others from dependencies.

Currently a kernel upgrade means a messy diff, and it is easy to not do oldconfig too.

justincormack:
Another option is to manage like Android does see https://source.android.com/devices/tech/config/kernel with sorted sets of options https://android.googlesource.com/kernel/common/+/android-4.9/kernel/configs/android-base.config

tych0:
Hi guys, what about something like the following:

* kernel_config.base - everything that isn't arch or version specific
  * e.g. CONFIG_NET, more detailed as we want to turn options on and off
* kernel_config.x86 - stuff that's arch (but not version) specific
  * e.g. X86_64, X86, all implied ARCH_* settings (and friends)
* kernel_config.4.9.x - stuff that's === 4.9.x specific
  * optional
* kernel_config.x86.4.9.x - stuff that's arch specific to 4.9.x
  * optional

When building a kernel, we'd cat kernel_config.base kernel_config.arch 
kernel_config.{versions <= your specified version} kernel_config.arch.{versions 
<= your specified version} all together and do old(def?)config, and use that as 
the result.

When we drop support for an old version, we can just take kernel_config.4.4.x 
and put it in kernel_config.base (similar with the .arch one).

For features (e.g. a .hardened config) or something, we'll need something 
slightly more intelligent than cat, since the features might overlap. I think 
it should be enough to make the e.g. .hardened config /replace/ all options 
that are specified in the base config, so we cat the configs together as above,
and then replace everything that is specified in .hardened.

Finally, it would be nice to be able to optionally include projects, which 
could use the same replacement strategy, I think. You'd give the tool a yaml 
file (or something) which says "I want selinux and wireguard, but not 
landlock", and it would add the patches, add the kernel config, and the 
userspace tools necessary. This would involve changing the format of the 
projects/ directly slightly, but I think it would be nice to have. In any case, 
it's not necessary for the stuff above.

justincormack:
I would generally be keen on doing something close to Debian or Android, but it may not work for us. I would suggest doing a prototype in `projects/hardened` or `projects/kernelconfig` and we can look at it as its the detail of how manageable it is eg how we handle adding and removing options for different versions and how easy it is to see what happens over time (eg diffs and PRs) that will make the difference.

tych0:
[...]

PR #1730, opened 2017-04-27
add kernel config update mechanism and remove some kernel config options

tych0:
**- What I did**

Disable a few unnecessary (and potentially scary) kernel features (see patch notes for details). 
I don't know much about CONFIG_X86_MSR, but the other four seem like they can probably go with little to no consequence.

**- How I did it**

Added some make targets to easily update kernel configs, then updated the kernel configs.

Remove some potentially scary kernel config options.


kernel/Dockerfile
```diff
@@ -54,7 +55,8 @@ RUN DVER=$(basename $(find /tmp/kernel-modules/lib/modules/ -mindepth 1 -maxdept
 RUN printf "KERNEL_SOURCE=${KERNEL_SOURCE}\n" > /kernel-source-info
 
 
-FROM linuxkit/toybox-media:eee3dd4d72cd784801e95b1781e6c4f9d8a5e5eb@sha256:7f940e687164ee2676e11c61705c79f7dd2d144ee87ad17a494848a7045f5f53
+# FROM linuxkit/toybox-media:eee3dd4d72cd784801e95b1781e6c4f9d8a5e5eb@sha256:7f940e687164ee2676e11c61705c79f7dd2d144ee87ad17a494848a7045f5f53
+FROM 9375904404b1
```
tych0: I don't have +w to toybox-media, but I'm happy to change this when something gets uploaded that has cat in it.

rneugeba: why do you need `cat` other then for convenience?

rneugeba: NM, just saw it further down

tych0: Is there another way to get the contents of a file out of a docker image that's easy? I don't need it, but I couldn't figure out anything else.

kernel/Makefile
```diff
@@ -51,9 +53,12 @@ sign_$(2): build_$(2)
 		 docker tag linuxkit/$(IMAGE):$(1)-$(HASH) linuxkit/$(IMAGE):$(2) && \
 		 DOCKER_CONTENT_TRUST=1 docker push linuxkit/$(IMAGE):$(2))
 
+update-config_$(2): build_$(2)
+	docker run linuxkit/$(IMAGE):$(1)-$(HASH) cat /kernel_config > kernel_config-$(2)
```
rneugeba: I'm not sure about this. What this does is it builds the kernel docker image and then copies the kernel config out (which is nice to have). But if we then commit it, this will change the git tree hash and thus we have to rebuild, right? So as a developer, I'd have to do a full kernel compile when a download kernel, apply patches, copy config and `make defconfig && make oldconfig` would suffice. I think, I'd prefer having a separate Dockerfile to update the kernel config for quicker turnaround

kernel/kernel_config-4.9.x
```diff
@@ -243,16 +243,13 @@ CONFIG_OPROFILE=y
 # CONFIG_OPROFILE_EVENT_MULTIPLEX is not set
 CONFIG_HAVE_OPROFILE=y
 CONFIG_OPROFILE_NMI_TIMER=y
-CONFIG_KPROBES=y
+# CONFIG_KPROBES is not set
```
rneugeba: Is KPROBES problematic? Does it stop `perf` from working? We definitely would like to enable perf debugging on kernels

kernel/kernel_config-4.9.x
```diff
@@ -3302,7 +3294,6 @@ CONFIG_BRANCH_PROFILE_NONE=y
 # CONFIG_PROFILE_ALL_BRANCHES is not set
 CONFIG_STACK_TRACER=y
 CONFIG_BLK_DEV_IO_TRACE=y
-CONFIG_KPROBE_EVENT=y
```
rneugeba: see above re debugging/performance monitoring

rneugeba:
Thanks for auditing. I made a couple of comments. I think the having a separate dockerfile to update the kernel config would be better (also then doesn't require `cat` in the toybox, so that might be nice to have anyway)

kernel/Dockerfile
rneugeba: there is, I had some runes...I think you can do a `docker cp` but need to have an existing container (can't do it from the image) https://docs.docker.com/engine/reference/commandline/cp/

kernel/Makefile
tych0: I was just forcing builds with HASH=foo, so that you can do it before commit, but it shouldn't be hard to do a separate Dockerfile either.

rneugeba: Yeah, that's why it's `HASH?=` but it still compiles the full kernel, which on my Mac takes a lot longer than just doing the `oldconfig`

kernel/Dockerfile
tych0: Ok. I'd prefer cat over some hackery, but I'll just do a separate Dockerfile instead.

kernel/Makefile
tych0: Yep, it does. I'll do a separate Dockerfile.

kernel/kernel_config-4.9.x
tych0: I think it would probably block the kprobe instrumenting parts of perf, but not the events and such (that's under a different flag). I can drop it from the series if you like.

rneugeba: no let's leave it. `kprobes` are quite invasive. Events are nice and more important to have for `perf`. Do you know if it blocks the eBPF style hooks for monitoring or is this done via a different flags as well?

tych0: That would be controlled by `CONFIG_BPF_SYSCALL`, which we currently enable, so those will work.

rneugeba: thanks for checking

justincormack: Are you sure though? I would think that attaching bpf programs to kprobes needs kprobes... Currently this is the only way to eg monitor most security related things as there are no tracepoints (hint hint).

kernel/Dockerfile
justincormack: Personally I just use `tar`.

kernel/Dockerfile.config
```diff
@@ -0,0 +1,24 @@
+FROM linuxkit/kernel-compile:1b396c221af673757703258159ddc8539843b02b@sha256:6b32d205bfc6407568324337b707d195d027328dbfec554428ea93e7b0a8299b AS kernel-build
+
+ARG KERNEL_VERSION
+ARG KERNEL_SERIES
+
```
justincormack: You don't need another `Dockerfile` to do this. Just copy the `.config` to a location like `/export/kernel_config` at the end of the normal Dockerfile and use a volume mount there to extract it.

justincormack:
Absolutely no to adding a new largely duplicated Dockerfile. And can you split this into multiple PRs as the parts are unrelated.

kernel/Dockerfile.config
rneugeba: But this would still compile the kernel. We just want to run `oldconfig` to update the config...or do you suggest the person updating the kernel just to hit ^C once that is done? I personally would prefer a single `make` target, which I run after updating kernel/patches to get the updated config file.

kernel/kernel_config-4.9.x
tych0: Sorry, I didn't ebpf style hooks used kprobes. Then yes, obviously we need kprobes :)

tych0:
That's exactly what I had before Rolf asked me to do it this way :). Please work out how you'd like me to do this, and let me know and I'll resubmit.

justincormack:
I do not want security changes or kernel config in the same PR as any other changes.

tych0:
Understood. How should I do the config updates?

tych0:
Sorry, by that i mean the make target that updates config.

justincormack:
I don't know but not by duplicating Dockerfiles. I think it would be better to rework the whole config management area - see https://github.com/linuxkit/linuxkit/issues/635 rather than making small changes, as we need to accomodate arm64 soon and we know this needs improving so I would rather see a fully worked out proposal not small changes.

PR #1733, opened 2017-04-28
Change to use kernel not bzImage everywhere

### LINKED ISSUE #1630: Support non bzImage kernels
We didnt really think when designing the kernel support initially, so we hardcoded `bzImage` in various places, and this needs to be fixed for 
(a) architectures that use other formats eg aarch64 https://github.com/docker/linuxkit/pull/1620
(b) projects/boot methods that want to use another format eg Clear Containers https://github.com/docker/linuxkit/pull/1625

There are a few options - either we collect all the potential kernels that we find in the kernel image (bzImage, zImage, vmlinux etc) and then pass them all to the output methods so they can select the one we want, or we can have a selector in the kernel image stanza saying which one to use (with a per architecture default).

The second method is probably simpler to implement...

justincormack:
Previously we hardcoded `bzImage` which is not used for all
use cases or architectures.

fix #1630

~~Requires https://github.com/moby/tool/pull/8 to be merged~~

### DIFF
--- a/tools/mkimage-gcp/make-gcp
+++ b/tools/mkimage-gcp/make-gcp
@@ -5,7 +5,7 @@ set -e
 mkdir -p /tmp/image
 cd /tmp/image
 
-# input is a tarball of vmlinuz64 and initrd.img on stdin
+# input is a tarball of kernel and initrd.img on stdin
 # output is a compressed tarball of a raw disk image on stdout
 
 mkdir -p files
@@ -17,10 +17,10 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name vmlinuz64 -or -name bzImage)"
+KERNEL="$(find . -name kernel -or -name bzImage)"
 CMDLINE="$*"
 
-[ "$KERNEL" = "./vmlinuz64" ] || mv "$KERNEL" vmlinuz64
+[ "$KERNEL" = "./kernel" ] || mv "$KERNEL" kernel
 [ "$INITRD" = "./initrd.img" ] || mv "$INITRD" initrd.img
 
 # clean up subdirectories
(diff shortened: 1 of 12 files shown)


justincormack:
cc @mor1 @avsm should make arm64 easier

justincormack:
Also now includes https://github.com/moby/tool/pull/10 to fix pull/create issue as we need to update `moby` in this PR.

tools/mkimage-gcp/make-gcp
```diff
@@ -17,10 +17,10 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name vmlinuz64 -or -name bzImage)"
+KERNEL="$(find . -name kernel -or -name bzImage)"
 CMDLINE="$*"
 
-[ "$KERNEL" = "./vmlinuz64" ] || mv "$KERNEL" vmlinuz64
+[ "$KERNEL" = "./kernel" ] || mv "$KERNEL" kernel
```
dave-tucker: Per above comment this would make this line unnecessary

tools/mkimage-gcp/make-gcp
```diff
@@ -17,10 +17,10 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name vmlinuz64 -or -name bzImage)"
+KERNEL="$(find . -name kernel -or -name bzImage)"
```
dave-tucker: Do we want to drop `bzImage` support here too?

justincormack: Its a temporary thing, we can delete in a bit

dave-tucker: nvm, i didn't read the title properly...

PR #1734, opened 2017-04-28
Pull images on CI

justincormack:
Always do `moby build --pull` on CI.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

PR #1736, opened 2017-04-28
use updated kernel and correct path in examples/swap.yml

deitch:

**- What I did**

Fix `examples/swap.yml` to:

1. Use correct path for `swap.sh`
2. Updated to a more recent kernel version

**- How I did it**

**- How to verify it**

Run.

Update kernel in swap.yml example and use correct path to script for onboot container

koala

### DIFF
--- a/examples/swap.yml
+++ b/examples/swap.yml
@@ -48,7 +48,7 @@ onboot:
     rootfsPropagation: shared
     command: ["/mount.sh", "/var/external"]
   - name: swap
-    image: "linuxkit/swap:edd74ce7ced2b97dba5075012f497570c5490299"
+    image: "linuxkit/swap:d089b31acad8a5f6a5f1c368ddd1dfe1d049b100"
     net: host
     pid: host
     capabilities:
@@ -57,7 +57,7 @@ onboot:
     binds:
      - /var:/var
      - /dev:/dev
-    command: ["swap.sh", "--path", "/var/external/swap", "--size","1G"]
+    command: ["/swap.sh", "--path", "/var/external/swap", "--size", "1G"]
 services:
   - name: rngd
     image: "linuxkit/rngd:3dad6dd43270fa632ac031e99d1947f20b22eec9"
--- a/pkg/swap/swap.sh
+++ b/pkg/swap/swap.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-while [ $# -gt 1 ]; do
+while [ $# -ge 1 ]; do
 	key="$1"
 
 	case $key in
@@ -74,7 +74,9 @@ fi
 ## check each of our conditions
 for cond in $CONDITIONS; do
 	# split the condition parts
-	IFS=: read condtype arg1 arg2 arg3 arg4 <<< "$cond"
+	IFS=: read condtype arg1 arg2 arg3 arg4 <<EOF
+$cond
+EOF
 	case $condtype in
 		part)
 			partition=$arg1


justincormack:
You need to rebase to fix CI I think

deitch:
I thought I did, but I must have just missed it. No worries, I will do it again. In any case, I found another (dumb) issue with `swap.sh` (only came to fore when working on the encrypted, but could be triggered in other cases).

deitch:
Very strange. It is complaining about:

```
time="2017-04-28T14:23:41Z" level=fatal msg="Failed to extract kernel image and tarball: Failed to docker create image linuxkit/kernel:4.9.x: Error: No such image: linuxkit/kernel:4.9.x" 
```

`kernel:4.9.x` was in the _old_ one, not the _new_ one. And there *is* a tag `4.9.x` in the docker hub for `linuxkit/kernel`?

justincormack:
Yes, seems that https://github.com/linuxkit/linuxkit/pull/1734 did fix it though. Its unclear how it got in that state...

deitch:
No idea! :-)

I fixed the other issue, and submitted it as a single commit and signed (to keep Mr. Turtle happy)

justincormack:
Doesnt look like you rebased against the latest master

deitch:
Dang, multiple windows at once. Oops.

deitch:
I think I got it this time.

justincormack:
Well it may work, but I found the bug...

deitch:
The `-gt` one?

deitch:
It occurs to me that this fix means that the hash is changed, so I need to change the example, right?

deitch:
OK, will add to this PR

deitch:
Added.

justincormack:
You don't need to add 4.10.x to get the config changes, they will be in 4.9.x so long as you re-pull (eg `moby build --pull ...`) - unless you need specific 4.10 features?

deitch:
No, I don't, but why not use the latest? Any reason not to use 4.10?

justincormack:
Well its just kind of odd if one example uses 4.10 and the rest don't people will assume it needs 4.10 features. Currently we have not done extensive testing of 4.10 although we plan to soon; it is also going to be EOL shortly as 4.11 is out this weekend...

deitch:
No worries. I will back to 4.9 over the weekend. Hold off on the merge until another push here.

justincormack:
needs a rebase too, sorry

deitch:
Ugh. Oh well.

deitch:
OK, kernel restored to `4.9.x`, need to rebase, though

deitch:
OK, rebased. Used the hash from the fixed build.

deitch:
Ready for approval?

justincormack:
Ah but you havent got push access have you, whats your docker hub id? I can give you push to linuxkit/swap for now

deitch:
Same as my GitHub, https://hub.docker.com/u/deitch/

Hold on a sec on merging this please. Just want to be paranoid and run one more check

justincormack:
Ok, you should be able to push to linuxkit/swap now so you can add that image

deitch:
I copied the Makefile from `pkg/mount`, it has a push which includes the hash. I assume that is good enough?

justincormack:
yes that should work

deitch:
OK, one more paranoid check, then I will push

deitch:
Any plans to break these packages to their own `linuxkit/*` repos?

justincormack:
Maybe, not decided yet, will see what makes sense.

deitch:
I knew there was another issue in `swap.sh`, was hoping to defer to another PR. Even though it is not in current path, it still trips it up. Well, I might as well fix it and squash into this one.

deitch:
Yeah, herestring `<<< $var` is a bash thing, need to convert to heredoc `<<EOF`. A minute or two

deitch:
Fixed, new hash updated as well. Should I wait for next approval to push the image?

deitch:
Image is in https://hub.docker.com/r/linuxkit/swap/

PR #1737, opened 2017-04-28
kernel: disable some unnecessary features

tych0:
There are four top level configs which we disable, and the rest are
cascading effects of those:

* CONFIG_CROSS_MEMORY_ATTACH: this adds two syscalls
  (process_vm_readv/writev) which allow tasks to access each others'
  memory. I don't know of any applications using these syscalls, and it
  sounds scary :)
* CONFIG_DEVMEM: allows direct access to memory. Perhaps useful for
  debugging, but we don't need it in general.
* CONFIG_IP_PNP: used to allow network configuration during diskless boot.
  We're not diskless.
* CONFIG_X86_MSR: export /dev/cpu/*/msr

Thanks to Elena Reshetova for pointing these out.


agshew:
With regard to CONFIG_CROSS_MEMORY_ATTACH, it is important for the performance of applications using Message Passing Interface libraries. So, disabling it may prevent a significant amount of scientific and engineering software from running well.

- https://www.open-mpi.org/faq/?category=sm
- http://mvapich.cse.ohio-state.edu/overview/

[Fast interprocess messaging](https://lwn.net/Articles/405346/):
> Either both processes must have the same ownership or the copying process must have the CAP_SYS_PTRACE capability; otherwise the copy will not be allowed.

tych0:
Thanks for the pointers! I think the idea here is that if they can use the syscalls in the container, they can also use them if they've managed to break out and set up the perms right, making further exploit easy. Whatever the case, if we need them we need them. I'll drop CROSS_MEMORY_ATTACH from the patch.

justincormack:
This process seems to me to be totally ad-hoc, and I think it needs some engineering.

1. Zero testing. This is not even going through the test suite we do have. Plus we deliberately introduced a test suite for kernel options as at least a first pass to make sure they remained set or unset in future see https://github.com/linuxkit/linuxkit/blob/master/test/check/check-kernel-config.sh ; testing whether the actual feature remains non available would be significantly better. Testing is still improving, but bypassing what we do have is not ok.
2. Multiple unrelated changes in one PR. These have totally different security issues. One of them it was already pointed out is used (its used elsewhere too); you can't really disable non obsolete syscalls. One of the others is I believe not needed, but the justification is incorrect (we absolutely do support a diskless setup). Two of them are fairly unclear and need further examination. So having them in one PR and one commit is not helpful.
3. Poor justification other than "I don't think people use this". They do, they definitely do, our users are incredible diverse, and so is code. The questions are around what do they use it for, what are the userspace APIs, what are the risks, and how are these options already gated by eg capabilities. What mitigations are there already in place? How exposed are we? For example the attack surface of `CONFIG_IP_PNP` is quite obscure (if it exists) while `/dev/mem` is clearer, but there is a bunch of detail involved. We can make these decisions against the users, eg `docker build` still cannot compile Emacs in the default setup, because we decided that allowing users to disable ASLR was going too far, but there are workable workarounds for those users who want to do that. Kernel config options are much more drastic, as there are no workarounds other than don't use your kernel.
4. We need a better way for managing sets of options; we know already that managing kernel options for multiple architectures is going to be a problem. It seems fairly clear that it would make sense to have a hardened config, but we do not have the tooling to support this.
5. Splitting out a hardened config with a graduation process for moving config to the standard config would help to test these changes, and would allow adding some high impact changes that some of our users can't use right now but we can push them in the right direction. For example removing 32 bit emulation would remove about, say 20,000 poorly tested entrypoints from userspace that have been shown to be high risk. Some of our customers need emulation, but many do not. Then we can add the more grey things to this setup and more gradually push them to the standard setup after users have a choice.

tych0:
On Mon, May 01, 2017 at 12:11:34PM -0700, Justin Cormack wrote:
> This process seems to me to be totally ad-hoc, and I think it needs some engineering.

Sounds good :)

> 1. Zero testing. This is not even going through the test suite we do have. Plus we deliberately introduced a test suite for kernel options as at least a first pass to make sure they remained set or unset in future see https://github.com/linuxkit/linuxkit/blob/master/test/check/check-kernel-config.sh ; testing whether the actual feature remains non available would be significantly better. Testing is still improving, but bypassing what we do have is not ok.

Ah thanks for this. I didn't know we had it.

> 2. Multiple unrelated changes in one PR. These have totally different security issues. One of them it was already pointed out is used (its used elsewhere too); you can't really disable non obsolete syscalls. One of the others is I believe not needed, but the justification is incorrect (we absolutely do support a diskless setup). Two of them are fairly unclear and need further examination. So having them in one PR and one commit is not helpful.
> 3. Poor justification other than "I don't think people use this". They do, they definitely do, our users are incredible diverse, and so is code. The questions are around what do they use it for, what are the userspace APIs, what are the risks, and how are these options already gated by eg capabilities. What mitigations are there already in place? How exposed are we? For example the attack surface of `CONFIG_IP_PNP` is quite obscure (if it exists) while `/dev/mem` is clearer, but there is a bunch of detail involved. We can make these decisions against the users, eg `docker build` still cannot compile Emacs in the default setup, because we decided that allowing users to disable ASLR was going too far, but there are workable workarounds for those users who want to do that. Kernel config options are much more drastic, as there are no workarounds other than don't use your kernel.

Yes, hopefully we can have some discussion in this PR for what we can
disable or not. I don't pretend to know what every user is doing :)

> 4. We need a better way for managing sets of options; we know already that managing kernel options for multiple architectures is going to be a problem. It seems fairly clear that it would make sense to have a hardened config, but we do not have the tooling to support this.
> 5. Splitting out a hardened config with a graduation process for moving config to the standard config would help to test these changes, and would allow adding some high impact changes that some of our users can't use right now but we can push them in the right direction. For example removing 32 bit emulation would remove about, say 20,000 poorly tested entrypoints from userspace that have been shown to be high risk. Some of our customers need emulation, but many do not. Then we can add the more grey things to this setup and more gradually push them to the standard setup after users have a choice.

Ok, so should I just focus on figuring out kernel config management
for now, and close this PR?

justincormack:
Well we don't need `CONFIG_IP_PNP` although we support diskless as we don't support `nfs` rootfs and have no intention of doing so, so we can remove it.

`/dev/mem` should be able to be removed in a hardened config, but not sure in standard (historically it was mainly used by userspace device drivers, eg `X`; it is also used in IoT apps for IO). I am less clear about `msr`, I think some code may read these via these devices but I don't know. Github code search might help here perhaps, should be able to match on this with the Google bigquery github search, or maybe even the github search.

The kernel config management is a priority, as we need it for multiarch (last time we did Arm we had this issue, and ended up dropping the support), and also if we are to support a hardened config.

tych0:
Closing in favor of doing this in the way #1787 describes.

Issue #1265, opened 2017-03-09
Add targets for unpatched stable kernels


rneugeba:
suggested by @justincormack in the pub: We should have targets and images for all unpatched stable releases to aid testing and triage of kernel issues for things like `runc` etc

Issue #1431, opened 2017-03-30
Allow building images with "foreign" kernels


rneugeba:
For testing it would be nice if we could build bootable images with kernels from other sources (ie from `.deb` or `.rpm` sources)

PR #1738, opened 2017-04-28
Add support for building mainline, Ubuntu and Debian kernel packages

rneugeba:
This PR adds support for repackaging kernel debian packages from the Ubuntu mainline PPA (which tracks kernel.org releases) as well as the main Ubuntu and Debian kernel packages.

The scripts are written in a way that they should be run daily/weekly as they scrape the relevant websites and determine the kernel version. It then compares this with images on Hub before building/signing/pushing.

I have not started the scripts yet, but will when merged.

wip: #1431
closes: #1265


justincormack:
Great, looking forward to using this!

PR #1740, opened 2017-04-28
kernels: Fix annoying error in shell scripts

rneugeba:
The previous PR had two syntax errors from testing...

image


rneugeba:
update with another minor snag

Issue #1742, opened 2017-04-29
Handle kernel module loading better


rn:
With vendor kernels being mostly modules based I noticed that we are not loading modules for device drivers on boot.

We currently can't run `modprobe` as we don't have `modules.dep`. The vendor kernels (at least centos and ubuntu) don't have them in the package. We can't run `depmod` as we have a read-only root filesystem.

Ideally we run `depmod` pre-boot, but I'm not sure we can do that...

justincormack:
Hmm, we do have `modules.dep` in our builds (and I was thinking of setting up a service to run it, as we now have modular `btrfs`). They must have them in another package? We could bind mount and make `modules.dep` but its a bit slow and clunky. You can run it at the end of the kernel build/packaging I think, from memory.

rneugeba:
Ideally we run it at end of kernel build so we don't have to mess around with it during boot

rneugeba:
we also need to consider if we want something like `udev` to load modules for devices on boot

justincormack:
We are currently using busybox `mdev` which should do that in theory, but not entirely sure how well it works.

rneugeba:
I tried it with foreign kernels and it didn't seem to work...need to poke around a bit more.

justincormack:
Yeah, it didnt autoload `btrfs` on our kernel when I tried either (which is our only major module), I thought `mkfs` would try to autoload but not sure if it will or not.

pwFoo:
I just have done some short testing with mdev yet, but it should work for hotplugged devices during the system is up and running.

**hotplug**
* echo /sbin/mdev > /proc/sys/kernel/hotplug
* echo '$MODALIAS=.*    0:0 660 @modprobe "$MODALIAS"' > /etc/mdev.conf
* /sbin/mdev -s


But you should need a function to load modules during boot.
At the moment I still use a simple (and slow) script during boot.

```
MODALIAS=`find /sys/devices -iname modalias -exec cat {} \;`
TMODALIAS="" 

while [ `echo "$MODALIAS"|wc -l` -gt `echo "$TMODALIAS"|wc -l` ]; do
    for i in $MODALIAS; do
        /sbin/modprobe $i 2>/dev/null
    done

    [ "$VERBOSE" ] && echo Loaded `echo "$MODALIAS"|wc -l` modaliases
    [ "$DRV_WAIT" -gt 0 ] && sleep $DRV_WAIT
    TMODALIAS="$MODALIAS" 
[...]

PR #1744, opened 2017-04-30
Add CentOS/Fedora kernels. Fix/Improve kernel images

### LINKED ISSUE #1431: Allow building images with "foreign" kernels

rneugeba:
- Add selected Fedora and CentOS kernels (resolves #1431)
- Improve scripts:
  - Don't pull to check if an image is on hub. For some repos there are many, many images
  - Run `depmod` on modules. Allows `modprobe` to work. (wip #1742)
  - Clean images after build
  - Add `System.map` to kernel image
- Update documentation

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -1,19 +1,36 @@
-# LinuxKit kernels
+# Linux kernels
 
-Currently, LinuxKit supports a number of kernels. These kernels are
-typically based on the latest stable releases and are updated
-frequently to include bug and security fixes.  For some kernels we do
-carry some additional patches, which are mostly back-ported fixes from
-newer kernels. The full kernel source with patches is on
-[github](https://github.com/linuxkit/linux).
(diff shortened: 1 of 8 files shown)


rneugeba:
CentOS, Fedora, Ubuntu, Debian packages all pushed. Mainline kernels still being pushed (Currently at 4.4.4).

Tested the images with HyperKit and qemu. Module loading via `modprobe` seems to work.

AkihiroSuda:
Do you plan to support CentOS / Fedora rootfs as well?

rneugeba:
Not at the moment

justincormack:
@AkihiroSuda if you use the kernel and the userspace in a container, it is a pretty good version of an OS for testing. Adding apparmor and SELinux enable and policies would be a better simulation too, depending what you are testing, but mostly the straight kernel gives you what you need for testing.

PR #1750, opened 2017-05-01
Fix typo in docs/kernels.md

solidnerd:
Hey,
i think there is a typo at the end of `docs/kernels.md`.


solidnerd:
This was the one that I found. But sure we are all not perfect ;)

PR #1753, opened 2017-05-02
Fix CI

dave-tucker:
This PR brings two commits that should fix CI for LinuxKit

One fixes a regression where `linuxkit push gcp` no longer allowed for a
custom destination to be provided via the `-img-name` flag or the
`CLOUDSDK_IMAGE_NAME` environment variable

The second ensures that when a custom name is set, the same name is issued to the `run` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,9 +58,10 @@ test-hyperkit: $(LINUXKIT) test-initrd.img test-bzImage test-cmdline
 	$(call check_test_log, test.log)
 
 .PHONY: test-gcp
+test-gcp: export CLOUDSDK_IMAGE_NAME?=test
 test-gcp: $(LINUXKIT) test.img.tar.gz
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -58,9 +58,10 @@ test-hyperkit: $(LINUXKIT) test-initrd.img test-bzImage test-cmdline
 	$(call check_test_log, test.log)
 
 .PHONY: test-gcp
+test-gcp: export CLOUDSDK_IMAGE_NAME?=test
```
rneugeba: Is there a reason this is specified as a separate target line?

dave-tucker: AFAICT it won't work any other way!
see: https://www.gnu.org/software/make/manual/make.html#Target_002dspecific for example usage...

If I try and put this on the target line, I get:
```
Makefile:62: *** commands commence before first target.  Stop.
```

PR #1756, opened 2017-05-02
projects/kubernetes: Fixup boot-*.sh

ijc25:
It seems that `-disk-size` is now `-disksize`.

Also, drop `hyperkit` from the `linuxkit run` invocation, thus causing the
linuxkit tool to pick the platform's default backend (which is qemu on my Linux
system, which works better than hyperkit in this environment).

(full disclosure: I've booted master, booting node seems to require me to read more of the README, but I propagated the same changes blindly nonetheless)

**- What I did**

Updated `projects/kubernetes/boot-*.sh`

**- How I did it**

emacs

**- How to verify it**

Run the `boot-master.sh` script.

Mastodon, Leviathan

### DIFF
--- a/projects/kubernetes/boot-master.sh
+++ b/projects/kubernetes/boot-master.sh
@@ -2,4 +2,4 @@
 disk="kube-master-disk.img"
 set -x
 rm -f "${disk}"
-../../bin/linuxkit run hyperkit -cpus 2 -mem 4096 -disk-size 4096 kube-master
+../../bin/linuxkit run -cpus 2 -mem 4096 -disk-size 4096 kube-master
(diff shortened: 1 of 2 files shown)


justincormack:
It is currently inconsistent between `disk-size` and `disksize`, so that needs fixing first.

rneugeba:
ping @errordeveloper

justincormack:
Er no not good, as this will just break hyperkit which uses `disk-size`. I prefer the hyphenated form, but yes consistentcy first.

ijc25:
I just assumed the script hadn't kept up with some other change, it never occurred to me the various backends might be inconsistent. I'll rebase and remove the unnecessary changes.

ijc25:
Reworked commit message due to changed scope. Now:

```
    projects/kubernetes: Allow linuxkit to pick default backend
    
    Drop `hyperkit` from the `linuxkit run` invocation, thus causing the linuxkit
    tool to pick the platform's default backend (which is qemu on my Linux system,
    which works better than hyperkit in this environment).
    
```

PR #1763, opened 2017-05-03
Use Alpine base image that includes packages

justincormack:
This means that we limit to one place where we use network access
 in building, and in future all other package builds can be deterministic.

Update `ca-certificates` to use this as a model.

84933-cute-hedgehog


pkg/ca-certificates/Makefile
```diff
@@ -1,19 +1,17 @@
...
 hash: Dockerfile
-	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
-	tar cf - $^ | docker build --no-cache -t $(IMAGE):build -
-	CONTAINER=$$(docker create $(IMAGE):build /dev/null); \
-		mkdir tmp && docker export $${CONTAINER} | tar xf - -C tmp && \
-		cat tmp/etc/ssl/certs/ca-certificates.crt | docker run --rm -i $(SHA_IMAGE) sha1sum - | sed 's/ .*//' > $@ && \
-		rm -rf tmp && docker rm $${CONTAINER}
+	docker pull $(BASE)
+	tar cf - $^ | docker build --network=none -t $(IMAGE):build -
+	echo $(HASH) > $@
```
rneugeba: These three steps should not be necessary anymore as we use the git tree hash as the hash. So there is no need  for the `hash` step. In the `./kernel` I just have a `build` target which does a `docker build -t $(IMAGE):$(HASH)` conditional on a `pull` of `$(IMAGE):$(HASH)` failing, but I added that to be able to build and test kernels locally without pushing them.

pkg/ca-certificates/Makefile
```diff
@@ -1,19 +1,17 @@
...
 default: push
 
 hash: Dockerfile
-	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
-	tar cf - $^ | docker build --no-cache -t $(IMAGE):build -
-	CONTAINER=$$(docker create $(IMAGE):build /dev/null); \
-		mkdir tmp && docker export $${CONTAINER} | tar xf - -C tmp && \
-		cat tmp/etc/ssl/certs/ca-certificates.crt | docker run --rm -i $(SHA_IMAGE) sha1sum - | sed 's/ .*//' > $@ && \
-		rm -rf tmp && docker rm $${CONTAINER}
+	docker pull $(BASE)
```
rneugeba: Not sure why `BASE` is pulled here

tools/alpine/versions
```diff
@@ -0,0 +1,162 @@
+alpine-baselayout-3.0.4-r0
```
rneugeba: This file seems to be autogenerated. Can we indicated this somewhere?

pkg/ca-certificates/Makefile
justincormack: hmm, no its not needed actually as if we set `DOCKER_CONTENT_TRUST` I think `docker build` will do the right thing.

pkg/ca-certificates/Makefile
justincormack: ah yes

tools/alpine/packages
```diff
@@ -0,0 +1,162 @@
+alpine-baselayout
```
rneugeba: How is this generated? It looks like it is the some of all the packages used in the repository *plus* their dependencies. What are the runes I have type if I want to add another package? Can we add a README in `./tools/alpine` to describe the process?

tools/alpine/versions
justincormack: yes, this is what we used to have for the alpine base image, it was useful to see what the security updates were, I will add a note to the top.

tools/alpine/packages
justincormack: Potentially I could generate it from a shorter list by having another stage that installs all the packages in the list you give and then uses the list that is then installed, ie with dependencies, as the list to mirror...

tools/alpine/versions
rneugeba: I agree that having a list of exact versions if good to have. A note at the top would be sufficient.

tools/alpine/packages
rneugeba: I think that would be good. The current list is pretty comprehensive, but I'm sure soon we'd want to add a package or two, or maybe even remove some, so it would be good to have the full list being autogenerated

justincormack:
Ok, generated from short list now, and has a "generated" comment on the versions file

justincormack:
And removed the `tar` in `docker build`

PR #1764, opened 2017-05-03
Makefile: gather all build artifacts in one location

dave-tucker:
This makes it easier for CI to grab artifacts from the ephemeral build
VMs and provides a stable location so they can be used in subsequent
tests. For example, running the LTP tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ Dockerfile.media
 *.qcow2
 *-kernel
 *-cmdline
+artifacts/*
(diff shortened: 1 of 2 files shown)


dave-tucker:
Note: use of `cp` here is so not to break the CI as it stands.
Once the CI has been changed to use the artifacts directory exclusively then it can be changed to `mv`...

While writing this I realise it'd be helpful for `moby` to support a `-o | -output-dir` flag to specify the location of the files it creates. I'll open an issue for that.

dave-tucker:
ping @justincormack @rneugeba I'm waiting on this before I can make progress with #1655

justincormack:
Oh, sorry. LGTM

Issue #1568, opened 2017-04-10
ship kernel headers in image?


justincormack:
this makes it easier to compile external modules, eg sysdig, nvidia. They could be in a different package in the `init` section, or in the `kernel.tar`.

rneugeba:
The source is already in the image on hub (and I'll add them for foreign kernels as well). So may we can add a option to the kernel section in the YAML to add it on demand

rneugeba:
pointed out on slack https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=849357
may be relevant

brycekahle:
A newer option is setting `CONFIG_IKHEADERS` in the kernel config for 5.2+ kernels, which exposes the headers as `/sys/kernel/kheaders.tar.xz`

PR #1765, opened 2017-05-03
Add support for building custom kernel modules

### LINKED ISSUE #1263: Create proper Linux header to compile against
We currently have minimum support for extracting Linux header for the kernel we run. However this is not sufficient to, for example compile kernel modules against.

The kernel build system does not have a standard way for packaging this up, however `scripts/package/buildeb` has the sample code for debian packages. We should plunder as much as possible from there

rneugeba:
- Fix the contents of `kernel-dev.tar`. This now contains all the necessary headers/files to compile kernel modules against the specific kernel
- Add sample/test for kernel module building and loading
- Add documentation on how to build modules

resolves #1263 
wip #1568

image

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -36,6 +36,35 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Compiling kernel modules
+
+The LinuxKit kernel packages include `kernel-dev.tar` which contains
+the headers and other files required to compile kernel modules against
+the specific version of the kernel. Currently, the headers are not
+included in the initial RAM disk, but it is possible to compile custom
+modules offline and include then include the modules in the initial
+RAM disk.
+
+There is a [example](../tests/kmod), but basically one can use a
+multi-stage build to compile the kernel modules:
+```
+FROM linuxkit/kernel:4.9.x AS ksrc
+# Extract headers and compile module
+FROM linuxkit/kernel-compile:1b396c221af673757703258159ddc8539843b02b@sha256:6b32d205bfc6407568324337b707d195d027328dbfec554428ea93e7b0a8299b AS build
+COPY --from=ksrc /kernel-dev.tar /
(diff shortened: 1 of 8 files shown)


docs/kernels.md
```diff
@@ -36,6 +36,35 @@ In summary, LinuxKit offers a choice of the following kernels:
...
+modules offline and include then include the modules in the initial
+RAM disk.
+
+There is a [example](../tests/kmod), but basically one can use a
+multi-stage build to compile the kernel modules:
+```
+FROM linuxkit/kernel:4.9.x AS ksrc
+# Extract headers and compile module
+FROM linuxkit/kernel-compile:1b396c221af673757703258159ddc8539843b02b@sha256:6b32d205bfc6407568324337b707d195d027328dbfec554428ea93e7b0a8299b AS build
+COPY --from=ksrc /kernel-dev.tar /
```
justincormack: I guess you can just do `ADD --from=ksrc /kernel-dev.tar /` which will untar it as well....

test/kmod/kmod.yml
```diff
@@ -0,0 +1,17 @@
...
+  - linuxkit/runc:b0fb122e10dbb7e4e45115177a61a3f8d68c19a9
+  - linuxkit/containerd:18eaf72f3f4f9a9f29ca1951f66df701f873060b
+onboot:
+  - name: check
+    image: "kmod-test"
+    binds:
+     - /dev:/dev
+     - /lib/modules:/lib/modules
+    capabilities:
+     - all
```
justincormack: you should only need `CAP_SYS_MODULE` - its nice to show minimal examples.

rneugeba: I tried that and I got a `EPERM`

justincormack: hmm. ok. Wonder what else it does need.

docs/kernels.md
rneugeba: apparently not:
```
Step 3/10 : ADD --from=ksrc /kernel-dev.tar /
Unknown flag: from
```

justincormack: ugh, ok. Well I guess its kind of deprecated

PR #1768, opened 2017-05-03
Eliminated the need for users to copy their public key.

mbruzek:
**- What I did**
The Makefile now copies the public key for the user.

**- How I did it**
Edited the Makefile to exit if no public key exists, otherwise uses sed to copy the key to the files.

**- How to verify it**
Run make build-vm-images and check `kube-master.yml` and `kube-node.yml`

nope

### DIFF
--- a/projects/kubernetes/Makefile
+++ b/projects/kubernetes/Makefile
@@ -20,9 +20,13 @@ push-container-images: build-container-images cache-images
 build-vm-images: kube-master-initrd.img kube-node-initrd.img
 
 kube-master-initrd.img: kube-master.yml
+	[ ! -f ~/.ssh/id_rsa.pub ] || echo "The public key was not found at ~/.ssh/id_rsa.pub" || exit 1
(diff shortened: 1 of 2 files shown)


justincormack:
The plan is to use metadata services to pass ssh keys, we are working on support even for the local hypervisors, so that the code looks just the same as on the cloud platforms. There is initial support for this on hyperkit, but we need to integrate this with ssh keys. I would rather fix it properly I think, but I guess we could do this temporarily...

cc @errordeveloper

mbruzek:
Using metadata services makes sense I am just trying to help improve the user experience so they don't have to dig up their public key and paste it in the files. Automate all the things!

justincormack:
https://github.com/linuxkit/linuxkit/pull/1868 updates `moby` to allow you to specify `source` in the files section instead of file contents, so I think this can be replaced with use of that.

mbruzek:
Closing in favor of #1868

PR #1770, opened 2017-05-04
Fix make clean to remove new -kernel files

justincormack:
Missed this in the changeover


thebsdbox:
Also `*.vmx` and `*.vmdk` I noticed a few minutes ago

justincormack:
@thebsdbox ah yes, added those too

PR #1772, opened 2017-05-04
Update packages

rneugeba:
Update packages:
- Use new alpine base package
- Use multi-stage build
- Use git sub-tree hashes

Update YAML files

These are mostly simplification, but quite extensive changes. I've tested the `examples/docker.yml` image, which seems to use most of the packages and it still seems to work


justincormack:
Definitely an improvement.

I dont think we are at the optimal thing yet - the whole squash thing is annoying, but getting better...

rneugeba:
yeah, it's incremental. I think this is a big step in the right direction. Need to move the other packages forward and then iterate

PR #1784, opened 2017-05-08
Code in the initrd generator to split a single tarball

justincormack:
In the WIP code in `moby` we now have a standard base tarball format,
that includes the kernel and cmdline as files in `/boot` so that the
entire output of the yaml file can default to a single tarball. Then
this can be split back up by LinuxKit into initrd, kernel and cmdline
as needed. This will probably become the only output of the `moby build`
stage, with a `moby package` stage dealing with output formats.

We may remove the output format specification from the yaml file as well,
and just have it in the command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,5 +65,5 @@ ci-pr:
 
 .PHONY: clean
 clean:
-	rm -rf bin *.log *-kernel *-cmdline *.img *.iso *.tar.gz *.qcow2 *.vhd *.vmx *.vmdk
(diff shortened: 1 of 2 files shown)


rneugeba:
Where is the WIP code in `moby`?

justincormack:
https://github.com/justincormack/tool/tree/tar-boot-rework

rneugeba:
I see this creates a dependency of `moby` on `linuxkit` which doesn't seem nice.

What do you want with this PR? The code itself LGTM

justincormack:
`moby` already has a dependency on `linuxkit` `initrd` code, will remove that later. This makes it easier to split, as `moby build` only needs to know how to make a tarball and the packaging step can be split between moby and linuxkit...

PR #1785, opened 2017-05-08
kernel: Add initial support for the 4.11 kernel

rneugeba:
This add a new version of the Hyper-V socket support based on
https://github.com/dcui/linux/commits/decui/hv_sock/next-20170504

Note, this changes the Linux side API to Hyper-V sockets as the
support is now based on the VMware/virtio socket implementation.
This means that the Address Family and the addressing changes.

Other patches from the 4.10 kernel are no longer needed as they
were already upstream.

The new Hyper-V socket code has not been tested, but the kernel
boots fine on HyperKit and Hyper-V.


rneugeba:
Having this in now, will make testing of the Hyper-V socket code easier. The kernel itself seems to boot fine

PR #1787, opened 2017-05-08
projects: add a project that re-works how we manage kernel config

tych0:
This is an implementation of what I proposed in #635. The bulk of the interesting-ness happens in the second patch in the series, see the notes there for discussion.


tych0:
Looks like the CI timed out or something?

rneugeba:
Thanks for taking a stab at this. Is there a reason you split up the single Dockerfile into multiple files, other than saving the download for debug vs non-debug kernels? It seems that it complicates the build and Makefile considerably...

tych0:
Mostly it's just saving the download on every build (debug or not), which is more annoying than the additional makefile stuff (which isn't that much, but obviously that's opinion :)

rneugeba:
I don't like it. It splits up the build process over three Dockerfiles and adds additional dependencies/steps to the Makefile.

Here is a half-baked idea (ie untested): How about adding an argument for the kernel URL, defaulting to what it is and then do a `ADD` instead of the `curl`. That way, a developer could point it to a local copy but CI would download? As said, haven't tried it...but if it works, it'll be cleaner/simpler

tych0:
Sure, I can play around with it. As I said in the patch notes: one of the splits is totally artificial, and isn't really needed. The only thing it's useful for is checking config without building a whole kernel. So really it would just be a split into two targets if we don't want that.

justincormack:
Why not check the config on the host not in a container?

tych0:
Because we need to merge all the configs and run them through "make oldconfig" to see what pops out. We have to do it in a build container, so I figured I'd just check it there too. Is there some reason not to?

justincormack:
You arent really building anything though - we switched to using `docker run` for non build cases and that worked ok. So you can have a script in the final kernel image that will do this checking for you...

tych0:
Well, we're not building the kernel, but `make defconfig` does compile everything in scripts/kconfig/,

```
  HOSTCC  scripts/basic/fixdep
  HOSTCC  scripts/kconfig/conf.o
  SHIPPED scripts/kconfig/zconf.tab.c
  SHIPPED scripts/kconfig/zconf.lex.c
  SHIPPED scripts/kconfig/zconf.hash.c
  HOSTCC  scripts/kconfig/zconf.tab.o
  HOSTLD  scripts/kconfig/conf
*** Default configuration is based on 'x86_64_defconfig'
#
# configuration written to .config
#
```

so we need a C-compiler for those bits. What's a good way to use docker run for that?

tych0:
Ok, I collapsed the build back into one file. Let me know what to do r.e. makeconfig.sh and where it runs or how to split it up. Thanks for looking!

tych0:
Thanks, @rneugeba! I'll see about importing the other kernel config options and if there's a nice/sane way to do that.

PR #1788, opened 2017-05-08
Update containerd

justincormack:
My patch was upstreamed so update, even though we won't need it soon.

Remove `dist` from bundle as we don't need it.

Not going to update anything that uses it as needs shifting to new Alpine package anyway (needs Go added still, will do this).


ijc25:
> Remove `dist` from bundle as we don't need it.

We should perhaps consider how this could be made available to users who want to `dist pull && ctr run` a new container at runtime (I tried to do this the other day). Perhaps add a dist container?

justincormack:
Yes, want to do same with docker container and remove CLI, then have a dev
container with user tools.

I think we have an issue around Dev container already

PR #1789, opened 2017-05-08
Multistage builds for containerd and runc

justincormack:
From new Alpine base, with Go.


rneugeba:
There are several YAMLs where conainerd is not updated. Intentional?

justincormack:
I updated all the ones that were on the same commit as the `linuxkit.yml` - I think these are all the ones using `ctr` right now. The bug still hasnt been fixed so not updated the ones just using `runc` yet.

PR #1798, opened 2017-05-10
Update virtsock test

rneugeba:
I've updated https://github.com/rneugeba/virtsock/

This PR updates the test package and test image


justincormack:
Are you going to move it to `linuxkit/`? If so should update paths here

rneugeba:
I will move, first want to update the C code as well. Can hold off with this PR till then, if you prefer. Just flushing out local changes

justincormack:
dont mind...

PR #1802, opened 2017-05-10
Metadata fix + a simpler file content option

ijc25:
**- What I did**

Fixed a bug in parsing the input JSON and added a less verbose mechanism to populate a data (i.e. not executable) file 

**- How I did it**

If a file key corresponds to a string rather than a map take that to be the desired file contents and write it with the default rw permissions (moduo `umask`). If the key corresponds to a map then the existing scheme (with separate `content` and `perm` keys) is maintained.

**- How to verify it**

Provide e.g. `{ "foo": { "bar": "some contents" } }` and observe that `/var/config/foo/bar` contains `some contents`.

Metadata JSON can now contain a simple string which will be written to the file

Faith No More, Angel Dust

### DIFF
--- a/pkg/metadata/main.go
+++ b/pkg/metadata/main.go
@@ -167,25 +167,31 @@ func processUserData(data []byte) error {
 			continue
 		}
 		for f, i := range files {
-			fi, ok := i.(map[string]interface{})
-			if !ok {
+			p := uint64(0644)
+			var c string
+
+			switch fi := i.(type) {
+			case map[string]interface{}:
+				if _, ok := fi["perm"]; !ok {
+					log.Printf("No permission provided %s:%s", f, fi)
+					continue
+				}
+				if _, ok := fi["content"]; !ok {
+					log.Printf("No content provided %s:%s", f, fi)
+					continue
+				}
+				c = fi["content"].(string)
+				if p, err = strconv.ParseUint(fi["perm"].(string), 8, 32); err != nil {
+					log.Printf("Failed to parse permission %s: %s", fi, err)
+					continue
+				}
+			case string:
+				c = fi
+			default:
 				log.Printf("Couldn't convert JSON for items: %s", i)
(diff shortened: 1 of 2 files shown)


pkg/metadata/main.go
```diff
@@ -167,25 +167,31 @@ func processUserData(data []byte) error {
...
+				if _, ok := fi["perm"]; !ok {
+					log.Printf("No permission provided %s:%s", f, fi)
+					continue
+				}
+				if _, ok := fi["content"]; !ok {
+					log.Printf("No content provided %s:%s", f, fi)
+					continue
+				}
+				c = fi["content"].(string)
+				if p, err = strconv.ParseUint(fi["perm"].(string), 8, 32); err != nil {
```
rneugeba: does this set `p` if there is an error?

justincormack: not defined what the value is if `err` set.

rneugeba: NM, misread the code slightly.

rneugeba:
Actually, could you add the new format to https://github.com/linuxkit/linuxkit/blob/master/docs/metadata.md as well?

pkg/metadata/main.go
ijc25: Yeah, if err is set it'll hit the `continue` and never look at `p` and skip the file (like it did before)

ijc25:
I added the docs.

I also changed the default perm from `0666` to `0644` having observed that the default `umask` in Alpine (and therefore the metadata container) is `0022`, so `0644` is the actual result. Since the `umask` used in the metadata container isn't exposed to the user my attempts to document the permissions as `0666-umask` were just getting silly and this makes more sense anyway I think.

ijc25:
@riyazdf did you mean something more than the docs entry with your "would be great to note the default perms for string input" comment?

PR #1803, opened 2017-05-10
Add `linuxkit metadata create` command

ijc25:
This is a thin wrapper to create an iso file in the format expected by
`pkg/metadata`.

Currently very basic and just takes the content on the command line but could
be extended to read from a file etc.

**- What I did**

Added a new `linuxkit` subcommand to create a metadata iso

**- How I did it**

Emacs

**- How to verify it**

Run `linuxkit mkmetadata meta.iso '{ "foo": { "bar": "some contents" } }'` and boot an image (with #1802 merged) using that ISO. Observe that `/var/config/foo/bar` contains `some contents`

Added new `mkmetadata` subcommand to `linuxkit`

Satyricon, Volcano

### DIFF
--- a/src/cmd/linuxkit/metadata.go
+++ b/src/cmd/linuxkit/metadata.go
@@ -9,6 +9,17 @@ import (
 	"github.com/rneugeba/iso9660wrap"
 )
 
+// WriteMetadataISO writes a metadata ISO file in a format usable by pkg/metadata
+func WriteMetadataISO(path string, content []byte) error {
+	outfh, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return err
+	}
+	defer outfh.Close()
+
+	return iso9660wrap.WriteBuffer(outfh, content, "config")
+}
+
 func metadataCreateUsage() {
 	invoked := filepath.Base(os.Args[0])
 	fmt.Printf("USAGE: %s metadata create [file.iso] [metadata]\n\n", invoked)
@@ -32,15 +43,8 @@ func metadataCreate(args []string) {
 	isoImage := args[0]
 	metadata := args[1]
 
-	outfh, err := os.OpenFile(isoImage, os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		log.Fatal("Cannot create user data ISOs", "err", err)
-	}
-	defer outfh.Close()
-
-	err = iso9660wrap.WriteBuffer(outfh, []byte(metadata), "config")
-	if err != nil {
-		log.Fatal("Cannot write user data ISO", "err", err)
+	if err := WriteMetadataISO(isoImage, []byte(metadata)); err != nil {
+		log.Fatal("Failed to write user data ISO: ", err)
 	}
 }
 
--- a/src/cmd/linuxkit/metadata.go
+++ b/src/cmd/linuxkit/metadata.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/rneugeba/iso9660wrap"
+)
+
+func metadataCreateUsage() {
+	invoked := filepath.Base(os.Args[0])
+	fmt.Printf("USAGE: %s metadata create [file.iso] [metadata]\n\n", invoked)
+
+	fmt.Printf("'file.iso' is the file to create.\n")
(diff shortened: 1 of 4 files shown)


ijc25:
Not sure about adding this to `linuxkit`, should perhaps be an independent tool, either under `pkg/metadata` or a new subdir of `src/cmd`.

rneugeba:
the hyperkit runner has a `-data` option to create this on the fly:
```
linuxkit run hyperkit --help
USAGE: linuxkit run hyperkit [options] prefix

'prefix' specifies the path to the VM image.

Options:
  -cpus int
    	Number of CPUs (default 1)
  -data string
    	Metadata to pass to VM (either a path to a file or a string)
  -disk string
    	Path to disk image to used
  -disk-size int
    	Size of Disk in MB
  -hyperkit string
    	Path to hyperkit binary (if not in default location)
  -ip string
    	IP address for the VM
  -mem int
    	Amount of memory in MB (default 1024)
``` 

PS: Not a cute animal

ijc25:
> the hyperkit runner has a -data option to create this on the fly:

Ah, useful. Sadly I was using this with libvirt so no `linuxkit run` for me, perhaps it could share some code, although there is so little of it I'm not sure it is worth it.

> PS: Not a cute animal

Opinions may differ ;-)

rneugeba:
I was going to add `-data` to the other runners. I think it's the right place as we can also add it to GCP/AWS/Azure meta data services, where it's not really a ISO

So I'm not sure that `linuxkit mkmetadata` is the right way to do it...though I see that a common/separate way to create a ISO with metadata might be useful

ijc25:
Unfortunately adding `-data` to the other runners doesn't help me since I'm not using `linuxkit run`, so I do need a separate tool somewhere.

justincormack:
Yes, we have always said that `run` is just a demo, and we should support other tools.

I would prefer it was `linuxkit metadata create` possibly, in case we need other options later.

rneugeba:
FWIW, I'm not objecting having this in `linuxkit`, it's probably the right place. Just not sure what the best option would be.

`linuxkit metadata createiso` might be better since there may be other ways. IIRC qemu/kvm has some other way of injecting metadata other than via ISOs

ijc25:
> IIRC qemu/kvm has some other way of injecting metadata other than via ISOs

I'm not sure, perhaps it has a way of tagging the VM with management type metadata (but even that I think might be a libvirt layer thing), but I don't know of anything for exposing metadata to the guest OS. Even if it did from a LinuxKit point of view I think we would probably go the ISO route anyway.

ijc25:
WRT `metadate create` vs `metadata createiso` I'd be inclined to go with the former and an (implicit, for now) `-t iso`, we can add `-t somethingelse` in the future then.

rneugeba:
ah, found the link: Qemu has a firmware configuration device of sorts: https://github.com/qemu/qemu/blob/d75aa4372f0414c9960534026a562b0302fcff29/docs/specs/fw_cfg.txt CoreOS ignition seems to use it

let's go with your inclination and stick with `linuxkit metadata create`

ijc25:
Ah yes, `fw_cfg` is how qemu exposes things to SeaBIOS/UEFI within the guest e.g. for building ACPI tables and stuff, I didn't think it was intended as a general guest metadata type thing, although you could probably use it that way.

ijc25:
Switched to `linuxkit metadata create`.

rneugeba:
Would be good if the hyperkit runner would re-use this code, do have one less place where `/config` is hard coded.

ijc25:
How about I create a `src/metadata` package with `func WriteMetadataISO(path string, content []byte) error` in it and use it from both?

ijc25:
The way `go-compile` works precludes simply adding a package (it only supports vendor and the package it is building) and I don't much feel like reworking all that. So I'll put this in a function within the linuxkit package instead.

src/cmd/linuxkit/metadata.go
```diff
@@ -0,0 +1,77 @@
...
+	case "help", "-h", "-help", "--help":
+		metadataUsage()
+		os.Exit(0)
+	}
+
+	switch args[0] {
+	case "create":
+		metadataCreate(args[1:])
+	default:
+		fmt.Printf("%q is not valid metadata command.\n\n", args[0])
```
riyazdf: typo nit "%q is not a valid metadata command"

ijc25:
Thanks @riyazdf. Fixed the typo and added some words to `docs/metadata.md`

PR #1804, opened 2017-05-11
Switch to containerd

justincormack:
Updated to containerd that has the fix for the logging issue #1660. I havent managed to replicate #1661 yet so that is probably fixed.

wabbit

### DIFF
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -2,9 +2,9 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:63eed9ca7a09d2ce4c0c5e7238ac005fa44f564b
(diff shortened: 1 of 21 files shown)


rneugeba:
did you push the init image?

justincormack:
Seems to have been a transient hub failure, that was the one I built yesterday...

examples/docker.yml
```diff
@@ -2,9 +2,9 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:63eed9ca7a09d2ce4c0c5e7238ac005fa44f564b
+  - linuxkit/init:deea956a9ab07bf262083e93a86930bdc610cc2fb
```
rneugeba: the `b` here is wrong. The tag on hub is `deea956a9ab07bf262083e93a86930bdc610cc2f`

PR #1808, opened 2017-05-12
Update Moby tool

justincormack:


justincormack:
ok, this should fix the various issues people have had...

PR #1809, opened 2017-05-12
Add image builder for compressed disk image

justincormack:
This is a fairly generic bootable disk with syslinux. Should
work if you `dd` it onto a USB stick, and should also work for AWS.

You need to uncompress it of course! Default size is 1G.

`moby` PR coming shortly.

hedgehogs


justincormack:
Added the updated `moby` tool

PR #1813, opened 2017-05-12
Added matching capability to create a disk

thebsdbox:
**- What I did**
Added the capability to re-use or create a disk from the linuxkit VMware cli

**- How I did it**
Using the `vmware-diskmanager` that comes with Fusion/Workstation

**- How to verify it**
Run `linuxkit run vmware` with the `-disk` `-disk-size` arguments

Persistent disk functionality added into the VMware backend.

### DIFF
--- a/src/cmd/linuxkit/run_vmware.go
+++ b/src/cmd/linuxkit/run_vmware.go
@@ -22,14 +22,20 @@ floppy0.present = "FALSE"
 displayName = "%s"
 numvcpus = "%d"
 memsize = "%d"
-`
-
-const vmxDisk string = `scsi0.present = "TRUE"
+scsi0.present = "TRUE"
 scsi0.sharedBus = "none"
 scsi0.virtualDev = "lsilogic"
+`
+
+const vmxDisk string = `
 scsi0:0.present = "TRUE"
 scsi0:0.fileName = "%s"
 scsi0:0.deviceType = "scsi-hardDisk"
+`
+
+const vmxDiskPersistent string = `scsi0:1.present = "TRUE"
+scsi0:1.fileName = "%s"
+scsi0:1.deviceType = "scsi-hardDisk"
 `
 
 const vmxCdrom string = `ide1:0.present = "TRUE"
@@ -68,9 +74,10 @@ func runVMware(args []string) {
 		fmt.Printf("Options:\n")
 		vmwareArgs.PrintDefaults()
 	}
-	runCPUs := vmwareArgs.Int("cpus", 1, "Number of CPUs")
-	runMem := vmwareArgs.Int("mem", 1024, "Amount of memory in MB")
-	runDisk := vmwareArgs.String("disk", "", "Path to disk image to use")
+	vmwareCPUs := vmwareArgs.Int("cpus", 1, "Number of CPUs")
+	vmwareMem := vmwareArgs.Int("mem", 1024, "Amount of memory in MB")
+	vmwareDisk := vmwareArgs.String("disk", "", "Path to disk image to use")
+	vmwareDiskSize := vmwareArgs.String("disk-size", "", "Size of the disk to create, only created if it doesn't exist")
 
 	if err := vmwareArgs.Parse(args); err != nil {
 		log.Fatal("Unable to parse args")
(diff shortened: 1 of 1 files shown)


src/cmd/linuxkit/run_vmware.go
```diff
@@ -68,9 +74,10 @@ func runVMware(args []string) {
 		fmt.Printf("Options:\n")
 		vmwareArgs.PrintDefaults()
 	}
-	runCPUs := vmwareArgs.Int("cpus", 1, "Number of CPUs")
-	runMem := vmwareArgs.Int("mem", 1024, "Amount of memory in MB")
-	runDisk := vmwareArgs.String("disk", "", "Path to disk image to use")
+	vmwareCPUs := vmwareArgs.Int("cpus", 1, "Number of CPUs")
+	vmwareMem := vmwareArgs.Int("mem", 1024, "Amount of memory in MB")
+	vmwareDisk := vmwareArgs.String("disk", "", "Path to disk image to use (.vmdk appended if disk is created)")
```
justincormack: why do we append `.vmdk` rather than having the user specify a full path?

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
-		path = fullVMrunPath
+		vmrunPath = fullVMrunPath
 		vmrunArgs = []string{"-T", "ws", "start", prefix + ".vmx"}
 	}
 
-	// Check executables exist before attempting to execute
-	if _, err := os.Stat(path); os.IsNotExist(err) {
+	// Check vmrunPath exist before attempting to execute
+	if _, err := os.Stat(vmrunPath); os.IsNotExist(err) {
 		log.Fatalf("ERROR VMware exectuables can not be found, ensure software is installed")
```
justincormack: typo here....

src/cmd/linuxkit/run_vmware.go
thebsdbox: Thought it was easier for the user, as we don't specify extensions for .iso or .vmdk. I suppose explicit paths will make it easier for a user to test something by pulling in an existing disk from elsewhere on a system.

justincormack: I don't really like not specifying extensions at all... its confusing.

justincormack: I know we do it for `linuxkit run` but I might remove that soon

thebsdbox: Understood. I'll change it to be explicit.

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
 
 	if runtime.GOOS == "linux" {
-		path = "vmrun"
-		fullVMrunPath, err := exec.LookPath(path)
+		vmrunPath = "vmrun"
+		vmDiskManagerPath = "vmware-vdiskmanager"
+		fullVMrunPath, err := exec.LookPath(vmrunPath)
 		if err != nil {
-			log.Fatalf("Unable to find %s within the $PATH", path)
+			// Kept as seperate error as people may manually change their environment vars
```
riyazdf: typo: `separate`

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
+			log.Fatalf("Unable to find %s within the $PATH", vmrunPath)
 		}
-		path = fullVMrunPath
+		vmrunPath = fullVMrunPath
 		vmrunArgs = []string{"-T", "ws", "start", prefix + ".vmx"}
 	}
 
-	// Check executables exist before attempting to execute
-	if _, err := os.Stat(path); os.IsNotExist(err) {
+	// Check vmrunPath exist before attempting to execute
```
riyazdf: nit typo: exists

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
+	if _, err := os.Stat(vmrunPath); os.IsNotExist(err) {
 		log.Fatalf("ERROR VMware exectuables can not be found, ensure software is installed")
 	}
 
+	if *vmwareDisk != "" {
+		// Check vmDiskManagerPath exist before attempting to execute
```
riyazdf: nit typo: exists

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
+		if _, err := os.Stat(vmDiskManagerPath); os.IsNotExist(err) {
+			log.Fatalf("ERROR VMware Disk Manager exectuables can not be found, ensure software is installed")
```
riyazdf: nit typo: `executables`

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
+		}
```
riyazdf: should we have an `else` case that bails if the err is not nil and `os.IsNotExist(err) != true`?

src/cmd/linuxkit/run_vmware.go
```diff
@@ -84,49 +91,74 @@ func runVMware(args []string) {
...
+
+		// If disk doesn't exist then create one
+		if _, err := os.Stat(*vmwareDisk); err != nil {
+			if os.IsNotExist(err) {
+				log.Infof("Creating new VMware disk [%s]", *vmwareDisk)
+				vmDiskCmd := exec.Command(vmDiskManagerPath, "-c", "-s", *vmwareDiskSize, "-a", "lsilogic", "-t", "0", *vmwareDisk)
+				if err := vmDiskCmd.Run(); err != nil {
+					log.Errorf("Error creating disk [%s]:  %s", *vmwareDisk, err.Error())
+				}
+			}
```
riyazdf: same here: should we have an `else` case that bails if the err is not nil and `os.IsNotExist(err) != true`?

src/cmd/linuxkit/run_vmware.go
riyazdf: I suppose the `vmx := buildVMX(*vmwareCPUs, *vmwareMem, *vmwareDisk, prefix)` step may fail, but I would prefer to error here so we can fail-fast with a more descriptive error message

src/cmd/linuxkit/run_vmware.go
```diff
@@ -151,15 +183,15 @@ func buildVMX(cpus int, mem int, diskPath string, prefix string) string {
 		returnString += fmt.Sprintf(vmxCdrom, cdromPath)
 	}
 
-	if diskPath != "" {
-		returnString += fmt.Sprintf(vmxDisk, diskPath)
+	vmdkPath := prefix + ".vmdk"
+	if _, err := os.Stat(vmdkPath); os.IsNotExist(err) {
+		log.Fatalf("File [%s] does not exist in current directory", vmdkPath)
 	} else {
```
riyazdf: same on the `else` case here: I think we should distinguish between `err != nil && !os.IsNotExist(err)` vs. `err == nil`

src/cmd/linuxkit/run_vmware.go
thebsdbox: There is a second check that will take place on line 137-138, we identify that the binaries exist here and then we will attempt to execute the found binary. If the VMware disk manager won't start/is corrupted then we print out the error on line 139. We could alternatively check permissions explicitly but I think the `exec.command` will report the same problem.

thebsdbox:
A quick note around checking for missing files or permissions, VMware Fusion/Workstation will also give a detailed issue in the event of a problem such as missing disk image/incorrect permissions or a Disk image that isn't in the correct VMDK format. On the off chance people try to use an existing disk image from QEMU etc.

justincormack:
Merging as the issues have been addressed, we can fix anything else later.

PR #1822, opened 2017-05-14
cmd: Enable virtio sockets for HyperKit

rneugeba:
HyperKit optionally exposes virtio sockets via a named pipes
in a directory. This patch passes a state directory and sets
the VSock flag to enable this feature.

Note, this will create a directory called 'prefix'-hyperkit
on every execution of 'linuxkit run hyperkit'.

While at it, also fix the .gitignore settings for kernels.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ Dockerfile.media
 *.vhdx
 *.efi
 *.qcow2
-*-kernel$
+*-kernel
 *-cmdline
+*-hyperkit
 artifacts/*
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -87,7 +87,9 @@ func runHyperKit(args []string) {
 		*disk = prefix + "-disk.img"
 	}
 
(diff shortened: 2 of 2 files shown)


rneugeba:
ping @MagnusS i think you were going to look into this as well. I need this for some virtsock testing

justincormack:
We need a better naming for state dir. With this I can't run multiple copies in the same place. And generally state dirs should be in `$TMPDIR` by default. I would use the pid of the hyperkit process or something.

rneugeba:
It also needs to be easily discoverable. We already create the disk in the same directory and the same name, so that also prevents us from running the multiple copies from the same place.

I'm not sure this should be a aim of `linuxkit run` to run multiple copies. That's what infrakit should be for

justincormack:
The disk at least gives you an option to specify it so you can put it somewhere else.

rneugeba:
How about adding a `state` option? It can default to `<prefix>-state/` and is used for the disk and metatdata ISO if not explicitly specified.

justincormack:
That sounds better, as other backenfs have state too.

rneugeba:
closing. will do a `-state` one seperately

PR #1827, opened 2017-05-15
Use a "state" directory for linuxkit run state and enable virtio sockets

rneugeba:
Some runners require state to be kept somewhere while they are executing. Examples are the virtio socket unix domain socket for HyperKit and various log and config files for the VMware runner. This patch set introduces a new `-state` argument to the HyperKit and VMware runners. These are the only ones which require such state (the qemu backend ony optionally creates a userdata disk). The state directory defaults to `<prefix>-state\` but can be overwritten.

Having a state directory allows us to enable the virtio socket interface for the HyperKit runner.

While at it, also clean up the backend code somewhat. The different runners were written by different people and the variables and other implementation details were inconsistent.

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -15,31 +15,31 @@ import (
 
 // Process the run arguments and execute run
 func runHyperKit(args []string) {
-	hyperkitCmd := flag.NewFlagSet("hyperkit", flag.ExitOnError)
+	flags := flag.NewFlagSet("hyperkit", flag.ExitOnError)
 	invoked := filepath.Base(os.Args[0])
-	hyperkitCmd.Usage = func() {
+	flags.Usage = func() {
 		fmt.Printf("USAGE: %s run hyperkit [options] prefix\n\n", invoked)
 		fmt.Printf("'prefix' specifies the path to the VM image.\n")
 		fmt.Printf("\n")
 		fmt.Printf("Options:\n")
-		hyperkitCmd.PrintDefaults()
+		flags.PrintDefaults()
 	}
-	hyperkitPath := hyperkitCmd.String("hyperkit", "", "Path to hyperkit binary (if not in default location)")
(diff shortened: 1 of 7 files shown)


src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -51,10 +59,10 @@ func runHyperKit(args []string) {
 		} else {
 			d, err = ioutil.ReadFile(*data)
 			if err != nil {
-				log.Fatalf("Cannot read user data: %v", err)
+				log.Fatalf("Cannot read user data: %s", err)
```
justincormack: `%v` is correct for `error` as it is a type not a string.

src/cmd/linuxkit/run_vmware.go
```diff
@@ -115,54 +123,56 @@ func runVMware(args []string) {
...
-
+	vmx := buildVMX(*cpus, *mem, *disk, prefix)
 	if vmx == "" {
 		log.Fatalf("VMware .vmx file could not be generated, please confirm inputs")
 	}
 
 	// Create the .vmx file
-	err := ioutil.WriteFile(prefix+".vmx", []byte(vmx), 0644)
-
+	vmxPath := filepath.Join(*state, "linuxkit.vmx")
```
thebsdbox: Using the identical `.vmx` file name will result in only being able to start one Virtual machine. VMware won't start any other LinuxKit Virtual machines with the same name.

thebsdbox: Actually this should only be a problem on VMware vSphere, on the Workstation products different directories are OK. With vSphere I'd imagine InfraKit will spawn numerous VMs and label them accordingly.

rneugeba: It does. I just tested by copying `examples/vmware.yml` to `examples/vmware2.yml` and build and run

src/cmd/linuxkit/run_hyperkit.go
rneugeba: should be fixed in the update. Also in the VMware code

PR #1830, opened 2017-05-15
Tidy up the tests

rneugeba:
Re-organise the tests
- kernel tests go under `linuxkit.kernel`
- build tests go under `linuxkit.build`
- don't pass artefacts from one test to another (like the qemu build/run tests did)
- Remove tests which weren't run/didn't work
- Move hack to run GCP tests into main Makefile (instead of hiding it in `cases`
- Remove README. It was pretty inaccurate.


rneugeba:
@dave-tucker any idea why the GCP test is failing? Something somewhere is expecting a GCP image in artefacts:
```
[2017-05-15 16:59.26] Failed: Unix.Unix_error(Unix.ENOENT, "open", "/tmp/linuxkit-79fabde/src/artifacts/test.img.tar.gz")
```

rneugeba:
I think I remember it's related to credentials not being in the build VM or something

dave-tucker:
Overall LGTM.

Yes @rneugeba the issue is that CI is expecting to be able to download a file called `artifacts/test.img.tar.gz` from the VM. This is then used to attempt to run that image on GCP.

I think what we ought to do is:

1) Run the tests on a disposable GCP VM
2) Copy back the `test/_results` folder but nothing else
3) If the tests passed, run the `linuxkit.platform` tests on the privileged CI machine

As a side effect this would also fix linuxkit/linuxkit-ci#1

/cc @talex5 WDYT?

rneugeba:
I think it isn't right that `rtf` builds some images which are then used by a *different* `rtf` run somewhere else. If we need some image for a `rtf` run, say to test if we can boot images on GCP, then we should build said images with `make` as a dependency, just like we build `moby`, `linuxkit` etc.
 
So, as a short term fix, I was going to add that to this PR, while also trying to support the local dev who has their own GCP setup.

However, this doesn't really scale that well once we also want to test LinuxKit images on AWS, Azure, packet.net etc. Once we get to that stage, I don't think we can run these on every PR, so we should maybe only do these nighlty and make sure that the "local" tests (ie Linux with qemu, macOS with HyperKit and Windows with Hyper-V) cover most of the cases

dave-tucker:
> I think it isn't right that rtf builds some images which are then used by a different rtf run somewhere else. 

Like I said in response to your comment in my PR, I 100% agree with this

> If we need some image for a rtf run, say to test if we can boot images on GCP, then we should build said images with make as a dependency, just like we build moby, linuxkit etc.

That won't work as the running of images on GCP is handled by Python scripts in LinuxKitCI and `make` is never invoked.

> However, this doesn't really scale that well once we also want to test LinuxKit images on AWS, Azure, packet.net etc. Once we get to that stage, I don't think we can run these on every PR, so we should maybe only do these nighlty and make sure that the "local" tests (ie Linux with qemu, macOS with HyperKit and Windows with Hyper-V) cover most of the case

Also agreed. That's why I think the best option is to change the CI code, re-run the tests for this PR and merge it as-is.

rneugeba:
> That won't work as the running of images on GCP is handled by Python scripts in LinuxKitCI and make is never invoked.

I don't understand this comment. Currently we run `make ci-pr` in the disposable VM which doesn't have secrets etc. This then, with the `build` label creates a GCP image and copies it to `$LINUXKIT_ARTIFACTS_DIR`. This directory is then copied back to the LinuxKit-CI controller which then somewhere runs `rtf` with the `gcp` label.

So, why can't `make ci-pr` build the gcp images and put them into `$LINUXKIT_ARTIFACTS_DIR`

dave-tucker:
You are right up to this point.

> This directory is then copied back to the LinuxKit-CI controller which then somewhere runs rtf with the gcp label.

I set up the test framework so it **could** do this but so far it hasn't been implemented in CI.
It's still using using Python. See: https://github.com/linuxkit/linuxkit-ci/blob/master/src/linuxkit_test.ml#L54

> So, why can't make ci-pr build the gcp images and put them into $LINUXKIT_ARTIFACTS_DIR

`make ci-pr` could handle copying in to the `$LINUXKIT_ARTIFACTS_DIR` and then it would be copied back to the LinuxKit-CI controller and all should be well.

I misunderstood your comment the first time around as I thought you wanted to add the file as a dependency of a non-existent `make test-gcp` target.

I'd much rather clean up the CI than propagating hacks... it's been a headache to incrementally change things + add the test framework without changing CI too much and this has led to a very confusing setup.

rneugeba:
Ah, I just found found that also. I'll add a quick hack for the GCP test because I want the structure of the tests merged so I (and others) can add more tests while the CI gets fixed

rneugeba:
@justincormack do you have an opinion on this PR? I've checked the CI and the now the test logs are captured in datakit (yeah) and everything else seems ok too

test/cases/010_platforms/100_packet/group.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
-# SUMMARY: LinuxKit configuration tests
-# LABELS:
+# SUMMARY: LinuxKit packet.net tests
+# LABELS: darwin
```
justincormack: why label `darwin`?

justincormack:
Just one question but otherwise LGTM

rneugeba:
I think I copy/pasted the packet.net `group.sh`. Pushed a fix and merge once CI is done

PR #1831, opened 2017-05-15
Add option for forwarding vsock ports in linuxrun for hyperkit

MagnusS:
Adds a new option `-vsock-ports` that can be used to make listed vsock ports in the guest available as unix domain sockets in the state-directory.

Also updates the hyperkit version in `/vendor`. 

(Running `vndr` from the root dir removed several of the existing dependencies as they were reported to not be in use, but I didn't include the deleted files in this PR.)

3675289_orig

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -32,6 +32,7 @@ func runHyperKit(args []string) {
 	data := flags.String("data", "", "Metadata to pass to VM (either a path to a file or a string)")
 	ipStr := flags.String("ip", "", "IP address for the VM")
 	state := flags.String("state", "", "Path to directory to keep VM state in")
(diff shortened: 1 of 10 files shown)


justincormack:
@MagnusS have you got an up to date copy of `vndr` - it keeps changing

justincormack:
Ah there do seem to be some missing deletions in vendoring...

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -7,12 +7,29 @@ import (
...
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/moby/hyperkit/go"
 	"github.com/satori/go.uuid"
 )
 
+func stringToIntArray(l string, sep string) ([]int, error) {
```
justincormack: put this in `util.go` as likely to be useful for other CLI options later?

MagnusS:
Rebased and moved `stringToIntArray` to `util.go`

PR #1832, opened 2017-05-15
Update vendoring and fix Windows build

justincormack:
We were missing one repo needed for Windows; but we have also not deleted
a bunch of vendored code that we are no longer using.

windowlicker


justincormack:
Yes its a mix of stuff that didnt get deleted plus I think `vndr` is now quite good at removing parts of code you dont use...

PR #1833, opened 2017-05-15
Add security SIG info and notes

riyazdf:
Following up on https://github.com/moby/mobywebsite/pull/11, adding more info and agenda/notes to the linuxkit repo under a `sigs` top level dir.  I've setup the zoom meeting and forum topic as well.  Aiming for May 24th as the first meeting, though this is flexible (intro email coming soon)

I'll update the mobywebsite repo once this is merged

cc @chanezon 

</img>

### DIFF
--- a/reports/sig-security/2017-05-24.md
+++ b/reports/sig-security/2017-05-24.md
@@ -1,0 +1,17 @@
+# 2017-05-24
+Video link: https://docker.zoom.us/j/779801882
+
+## Agenda
+- Introductions
+- Discuss goals of SIG
+- Updates on security `/projects`:
+  - clear-containers
+  - kernel-config
+  - kspp
+  - landlock
+  - miragesdk
+  - okernel
+  - wireguard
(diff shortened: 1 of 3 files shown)


justincormack:
Actually, it would be better to put meeting notes in top level `reports/`?

riyazdf:
@justincormack that works too, in a subdir such as `reports/security-sig`?

I also like the idea of keeping the agenda and meeting notes in the same place, so that each doc is self-contained

justincormack:
yeah, that makes sense to me

riyazdf:
ok - did some shuffling around so that we have both top-level `/sigs` for high level description, and `/reports/sigs/security` for notes

justincormack:
I think `reports/sig-security` is a bit more concise for now as we dont have any other sub-parts yet and its a lot of hierarchy...

riyazdf:
@justincormack: agreed, done.

PR #1842, opened 2017-05-16
Fix Github language detection

dave-tucker:
Ignore go files in `vendor`
Kernel config files are text and not Logos

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+vendor/* linguist-vendored
+kernel/*.x linguist-language=text


dave-tucker:
Adjusted language breakdown is now
```
26.99%  C
21.49%  Go
18.50%  Shell
16.68%  OCaml
13.77%  Makefile
1.49%   PowerShell
1.02%   Ruby
0.07%   Cap'n Proto
```

.gitattributes
```diff
@@ -0,0 +1,3 @@
+vendor/* linguist-vendored
+kernel/*.x linguist-language=text
+
```
justincormack: remove blank line?

dave-tucker: done!

Issue #1824, opened 2017-05-14
Set docker hub repository linuxkit as variable 


pwFoo:
To be able to use a custom repository (for example to build a custom kernel) "linuxkit" repository shouldn't be hardcoded. It should be replaced by a variable.

rneugeba:
Someone else (forgot who) suggested that as well. We could change, e.g.:
```
IMAGE:=kernel
```
to
```
IMAGE?=linuxkit/kernel
```
In the `Makefile`s

justincormack:
I think having a way to easily just specify the repo/org is most useful so
you can easily push a build to your own registry. Overriding the full name
is less useful.

mor1:
@rneugeba I certainly suggested this in conversation on Thursday, maybe others have too :) 

FWIW I have been replacing `linuxkit` with `$(ORG)` and setting `ORG=mor1` while doing the ARM64 PoC.

I don't have a very strong opinion on creating an `ORG` variable vs setting `IMAGE` to include the org. AFAICS the only places where `IMAGE` isn't used without `ORG` attached is where the temporary local build container is created. As each package is a thing unto itself, and there doesn't seem to be a need to declare `ORG` across an entire repo, I'd probably vote for `IMAGE` to include ORG per @rneugeba suggestion above.

justincormack:
The case to declare org across the repo is when you want to build your own
local mirror.

I can't see the use case for changing the actual image name eg kernel at
all, the only mentioned use case is for changing org.

pwFoo:
Hi @rneugeba,

please add to kernel doc:
> make ... ORG=foo

PR #1844, opened 2017-05-16
Update Hyper-V patches for 4.11 and add Hub org override for kernel builds

rneugeba:
While testing the new support for Hyper-V sockets in 4.11 we discovered some issues. The updated patches from Microsoft in this PR address this issue. I have run some light stress tests and they passed (after failing previously). I will run more tests over night but this is a definite improvement over the previous patches.

While at it, also make the Hub org configurable during build, ie:
```
make push_4.11.x ORG=rneugeba
```
pushes the image to my Hub account for testing (wip #1824)

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -10,7 +10,8 @@
 
 # Git tree hash of this directory. Override to force build
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-# Name on Hub
+# Name and Org on Hub
+ORG?=linuxkit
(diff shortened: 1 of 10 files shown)


rneugeba:
Ha, of course the further stress tests failed. But still should merge as this is an improvement...

@justincormack are you ok with the `ORG` make argument? Can rename it to something else and then we should roll this out to all Makefile

justincormack:
I think `ORG` is ok. Technically you can put registry and org in there (eg `docker.io/linuxkit`), but I can't think of a better name...

PR #1853, opened 2017-05-18
[RFC] Reconfigure the CI

dave-tucker:
This PR is both an RFC and a test case for reconfiguring the CI.

- Remove existing hacks in place to copy artifacts from ephemeral VM to the LinuxKit CI controller
- Control LTP test runs through `rtf` and labels and not through a make target called by CI directly
- Move the GCP test (run via Python scripts in CI) to a `rtf` test case
- Run the `linuxkit.platforms` tests on PR's and Master from *after* the first test run has been completed

This PR makes the necessary changes in LinuxKit and update the documentation.

Once CI has been update and reconfigured this PR should be green and ready for merge.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -67,19 +67,18 @@ The test logs are available via the `Details` link but also via this [website](h
 
 ## PR Testing
 
-Each PR is tested on disposable VM's spawned in Google Cloud Platform
+Each PR is tested on disposable VMs spawned in Google Cloud Platform
 This machine has no privileges or credentials to talk to GCP or other cloud platforms.
 
 TODO: Add instructions on how to build a base image for LinuxKit CI in GCP.
 
 LinuxKit CI runs `make ci-pr` in the VM.
 This target runs the tests using `rtf` and the results directory is `scp`'ed back to the controller.
-The test results will be stored in DataKit for additional access
-Additionally, the `./artifacts` folder is `scp`'ed back to the controller.
+The test results are stored in DataKit.
 
-If the tests passed the next step is to check if the kernel config test image runs on GCP.
-The `./artifacts/test.img.tar.gz` file is used to create a GCP image by the Python scripts
-that are part of LiunxKit CI. It runs this image and greps for the success message in the logs.
+If the tests passed the next step is to run additional platform tests.
+LinxKit CI runs `make ci-platforms` from the controller.
+This will use `rtf` to run tests on all local and/or cloud platforms in the test suite.
 
 ## Branch and Tag testing
 
@@ -87,8 +86,8 @@ Branches and Tags are tested on a dedicated machine that runs in GCP.
 
 LinuxKit CI runs `make ci` or `make ci-tag` in the VM
 This target runs the tests using `rtf` and the results directory is SCP'd back to the controller.
+The labels supplied to `rtf` via the `Makefile` will enable slower tests like the Linux Testing Project tests
+for more thorough test coverage.
 The test results will be stored in DataKit.
 
-If the tests pass, the GCP test is run in the same manner as described for PR tests.
-
-Finally, CI will run `make test-ltp` which will run the Linux Testing Project tests.
+If the tests pass, the platform tests are run in the same manner as described for PR's. 
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,9 +1,9 @@
 .PHONY: default pr all
 
-default: check-deps test 
-pr: check-deps test-pr
-# TODO: should have a separate target
-all: check-deps test-pr ltp
+pr: check-deps test
+all: check-deps test-all
+platforms: check-deps test-platforms
+default: pr
 
 MOBY:=$(shell command -v moby 2> /dev/null)
 LINUXKIT:=$(shell command -v linuxkit 2> /dev/null)
@@ -21,33 +21,14 @@ ifndef RTF
 	$(error "rtf is not available. please install it!")
 endif
 
+test: 
+	@rtf -x run
 
-# TODO: Remove this section once we no longer depend on this in CI
-### -------
-# Currently the linuxkit-ci runs GCP tests outside of rtf and expects some
-# files in ../artifacts. This hacky target puts them there until
-# the CI can use rtf for running GCP tests
-gcp-hack: ../artifacts/test.img.tar.gz
-../artifacts/test.img.tar.gz:
-	rm -rf ../artifacts
-	mkdir -p ../artifacts
-	$(MOBY) build -output gcp-img -pull -name ../artifacts/test hack/test.yml
+test-pr:
+	@rtf -x run
 
-define check_test_log
-	@cat $1 |grep -q 'test suite PASSED'
-endef
+test-all:
+	@rtf -l slow,gcp -x run
 
-.PHONY: ltp
-ltp: export CLOUDSDK_IMAGE_NAME?=test-ltp
-ltp: $(LINUXKIT) test-ltp.img.tar.gz
-	$(MOBY) build -output gcp-img -pull hack/test-ltp.yml
-	$(LINUXKIT) push gcp test-ltp.img.tar.gz
-	$(LINUXKIT) run gcp -skip-cleanup -machine n1-highcpu-4 $(CLOUDSDK_IMAGE_NAME) | tee test-ltp.log
-	$(call check_test_log, test-ltp.log)
-### ------
-
-test: gcp-hack
-	@rtf -l build -x run
-
-test-pr: gcp-hack
-	@rtf -l build -x run
+test-platforms:
+	@rtf -l gcp -x run linuxkit.platforms
--- a/Makefile
+++ b/Makefile
@@ -53,24 +53,29 @@ test:
 .PHONY: collect-artifacts
 collect-artifacts: artifacts/test.img.tar.gz artifacts/test-ltp.img.tar.gz
 
-.PHONY: ci ci-tag ci-pr
+.PHONY: ci ci-tag ci-pr ci-platforms
 ci:
 	$(MAKE) clean
 	$(MAKE)
 	$(MAKE) install
 	$(MAKE) -C test all
+	$(MAKE) -C test platforms
 
 ci-tag:
 	$(MAKE) clean
 	$(MAKE)
 	$(MAKE) install
 	$(MAKE) -C test all
+	$(MAKE) -C test platforms
 
 ci-pr:
 	$(MAKE) clean
 	$(MAKE)
 	$(MAKE) install
 	$(MAKE) -C test pr
+
+ci-platforms:
+	$(MAKE) -C test platforms
 
 .PHONY: clean
 clean:
--- a/test/cases/010_platforms/110_gcp/000_minimal/test.sh
+++ b/test/cases/010_platforms/110_gcp/000_minimal/test.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+# SUMMARY: Test that a minimal image boots on GCP
+# LABELS:
+# AUTHOR: Dave Tucker <dt@docker.com>
+
+set -e
+
+# Source libraries. Uncomment if needed/defined
+#. "${RT_LIB}"
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+
+IMAGE_NAME=gcp-minimal
+
+clean_up() {
+	rm -rf ${IMAGE_NAME}*
+}
+
+trap clean_up EXIT
+
+# Test code goes here
+moby build -name ${IMAGE_NAME} test.yml
+linuxkit push gcp ${IMAGE_NAME}.img.tar.gz
+RESULT=$(linuxkit run gcp "${CLOUDSDK_IMAGE_NAME}")
+echo "$RESULT"| grep -q "I Can Haz Linux?"
+exit 0
--- a/test/cases/010_platforms/110_gcp/000_minimal/test.yml
+++ b/test/cases/010_platforms/110_gcp/000_minimal/test.yml
@@ -1,8 +1,6 @@
-# FIXME: This should use the minimal example
-# We continue to use the kernel-config-test as CI is currently expecting to see a success message
 kernel:
   image: "linuxkit/kernel:4.9.x"
-  cmdline: "console=ttyS0"
+  cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037
   - linuxkit/runc:2649198589ef0020d99f613adaeda45ce0093a38
@@ -13,6 +11,9 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: check-kernel-config
     image: "linuxkit/test-kernel-config:ecff41279ccbc408079a3996a956432651c6eb9c"
+  - name: whalesay
+    image: "docker/whalesay"
+    command: ["/usr/local/bin/cowsay", "I Can Haz Linux?"]
     readonly: true
   - name: poweroff
     image: "linuxkit/poweroff:a8f1e4ad8d459f1fdaad9e4b007512cb3b504ae8"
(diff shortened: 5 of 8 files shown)


justincormack:
So whats the new security story for access to cloud resources?

dave-tucker:
PR's are tested on an ephemeral VM without an access to cloud resources.
Once tests have passed (which includes running images in QEMU etc...) we should have enough confidence to be able to test these images on GCP and other platforms. This is done from the privileged section of the CI that has access to cloud resources.

dave-tucker:
In other words, the security story is unchanged.
The only practical difference is in the source of the image that gets pushed to GCP for testing (which only occurs after a successful test run).

Today, it gets `scp`'d back from the VM.

With these changes, we'll build it ourselves.

docs/testing.md
```diff
@@ -67,28 +67,27 @@ The test logs are available via the `Details` link but also via this [website](h
...
+Each PR is tested on disposable VMs spawned in Google Cloud Platform
 This machine has no privileges or credentials to talk to GCP or other cloud platforms.
 
 TODO: Add instructions on how to build a base image for LinuxKit CI in GCP.
 
 LinuxKit CI runs `make ci-pr` in the VM.
 This target runs the tests using `rtf` and the results directory is `scp`'ed back to the controller.
-The test results will be stored in DataKit for additional access
-Additionally, the `./artifacts` folder is `scp`'ed back to the controller.
+The test results will be stored in DataKit.
```
rneugeba: will -> are

test/Makefile
```diff
@@ -1,9 +1,9 @@
 .PHONY: default pr all
 
 default: check-deps test 
-pr: check-deps test-pr
-# TODO: should have a separate target
-all: check-deps test-pr ltp
+platforms: check-deps test-platforms
+pr: default
```
rneugeba: what's the difference between `default` and `pr`. Can one of them be removed, or a more common pattern would be to have `default: pr` and then `pr: ...`

dave-tucker: no difference so yes we can remove PR completely.

rneugeba: I wouldn't remove it I would just re-arrange it because it's clearer

dave-tucker: ack

justincormack:
Needs a rebase

dave-tucker:
Rebased. Addressed Comments. Fixed Flakiness in GCP Tests.

@talex5 here is a summary of the work required in LinuxKitCI:

1) Stop copying the `artifacts` directory - we no longer need these files as we actually have tests that are run during the `make ci-pr`. We do still need the `test/_results` directory

2) **ON PRs ONLY** run `make ci-platforms` from the LinuxKitCI controller (the bit with the necessary creds for talking to GCP) *after* the `make ci-pr` completed successfully in the GCP VM.
This will also generate a `test/_results` directory which will need storing in DataKit.

3) Disable running the GCP tests - this is taken care of in `rtf` and via `make ci-platforms`

4) Disable the LTP tests on master and tag builds - these are now enabled when a `make ci` or `make ci-tag` are done and run by `rtf`

I'm hoping this will be the last set of "major" CI changes for a while.

talex5:
How can we "run `make ci-platforms` from the LinuxKitCI controller" for PRs? The `Makefile` is from the attacker, and the controller has access to the secrets.

dave-tucker:
@talex5 I think in this case we should probably whitelist some PR's (e.g those by org owners) and wait for a "ok to test" comment for others. IIRC @justincormack suggested something similar to this in Slack when I last suggested it but the scrollback doesn't go far enough for me to find a link

justincormack:
Yes that's an option.

dave-tucker:
The changes outlined above will simplify the CI setup and maximise the ability for LinuxKit contributors to make trivial CI changes. CI will call certain `make` targets and `rtf` controls which tests get run through labels etc...

The downside is that in order to test PR's against the cloud platforms we might need to run a potentially compromised `Makefile` on a machine that has access to the cloud platforms we need to test on.

I can see a few options here:

1) Don't worry about it and hope that our tests catch any mischief

2) Don't test running images on the cloud platforms on PR's

3) Have LinuxKitCI detect `Makefile` changes and either fail the build either immediately or at the `Test Cloud Platforms` step. This would then require approval from a maintainer in order to continue the tests.

There are probably other options I've missed so please feel free to make suggestions.

@riyazdf what would you recommend?

justincormack:
I think some of the Docker projects eg `notary` require approval to run CI. In this case we could run the base tests but not the platform tests. We would want to be able to whitelist contributors, not just maintainers.

2. is not an option - there is no point in CI if we don't run it.

dave-tucker:
@justincormack if we went down the approval route (which I think is best) would you:

a) Run platform tests only if contributor is on the whitelist
b) Run platform tests only if Makefile was changed && contributor is on the whitelist

justincormack:
B seems the least friction, but changing Makefile is fairly rare so either
ok

On 23 May 2017 16:39, "Dave Tucker" <notifications@github.com> wrote:

@justincormack <https://github.com/justincormack> if we went down the
approval route (which I think is best) would you:

a) Run platform tests only if contributor is on the whitelist
b) Run platform tests only if Makefile was changed && contributor is on the
whitelist

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/linuxkit/linuxkit/pull/1853#issuecomment-303438999>,
or mute
the thread

dave-tucker:
@talex5 it looks like we have a solution for the modification of the `Makefile`.
Would you be able to look at making the changes outlined here in LinuxKitCI?

ijc25:
>  b) Run platform tests only if Makefile was changed && contributor is on the whitelist

Shouldn't that be:
b) Run platform tests only if Makefile was not changed || contributor is on the whitelist
?

Also if we go the `ok for test` route (not sure if we are) then don't we need to protect against the user subsequently pushing more commits to the PR?

And in the "b) Run platform tests only if Makefile was not changed || contributor is on the whitelist" we need to be careful of races between evaluating that condition and pulling from the PR branch, i.e. we should carry the sha which we evaluated over and not repull the branch by name.

talex5:
Has @riyazdf approved this plan? It sounds pretty risky to me, but I'll implement it if he's happy.

dave-tucker:
@ijc25 yes you are right. thanks for the correction.

> then don't we need to protect against the user subsequently pushing more commits to the PR?

Yep I think that the approval would be per SHA and not for the PR as a whole.

> And in the "b) Run platform tests only if Makefile was not changed || contributor is on the whitelist" we need to be careful of races between evaluating that condition and pulling from the PR branch, i.e. we should carry the sha which we evaluated over and not repull the branch by name.

Ack. LinuxKitCI deals with SHA's anyway and not branch names so we should be ok in this regard.

dave-tucker:
@talex5 ack. waiting on a comment from @riyazdf

riyazdf:
I agree with @ijc25: `Run platform tests only if Makefile was not changed || contributor is on the whitelist`.   As @justincormack notes, we have a similar setup in Notary and it's worked well for us.

@dave-tucker: Where are planning on maintaining/storing the whitelist?
@talex5: do you have additional concerns?

Also, instead of `ok to test` can we use the Jenkins terminology `test this please` (or close to it)?  `test this please` is for running one CI run against the current state of the PR, `ok to test` is for running CI for this state and for any future commits.

dave-tucker:
@riyazdf the whitelist would be checked in to the `linuxkit/linuxkit-ci` repo.
👍 on all other points

riyazdf:
@dave-tucker: 👍  thank you, sounds good!

talex5:
@riyazdf : basically what we want to know is: is it safe to run `make` (a decades old C tool not designed for security) in a directory with a given set of files?

e.g.

- Does `make` read any files besides `Makefile`?
- Does the Makefile include other files (now or in the future)?
- Does the Makefile rule run any software from the checkout (now or in the future)?
- etc

If not, we give all our cloud keys (GCP, AWS, etc) to the attacker.

riyazdf:
@talex5: if the Makefile or our setup expands in scope to pull in other files, we'll have to also update the CI gating rules
ex: `Run platform tests only if Makefile && <other_file> was not changed || contributor is on the whitelist

As for `make` itself: I'm ok with using that program - of course we'll have to monitor it for serious security issues.  This is already the case today even if we don't opt to use it in LinuxKit, since we use it across other projects for critical CI operations

WDYT?

ijc25:
Pretty ignorant about our CI setup, but just brainstorming...

> If not, we give all our cloud keys (GCP, AWS, etc) to the attacker.

Do we necessarily need to expose those to all processes spawned by the CI or could we restrict them to just a smaller subset? e.g. a small set of "do the cloud bits" shell scripts to which we apply the restrictions we are talking about adding to modifications of the `Makefile`

Do we have the option of having the CI itself use the primary keys to generate short term, much more restricted keys, either on a per run basis or simply to turn them over more quickly (i.e. daily)?

If not could we run a proxy which accepts short term keys which we control and turns them into requests with the proper keys to the actual cloud services.

Can we use different keys (ephemeral or otherwise) with much more specific sets of abilities for different operations and apply different restrictions based on which ones are more powerful? i.e. read only keys for things which only need that privilege.

Could we put some/all of the bits which need the keys into the CI rather than the repo?

dave-tucker:
> Do we necessarily need to expose those to all processes spawned by the CI or could we restrict them to just a smaller subset? e.g. a small set of "do the cloud bits" shell scripts to which we apply the restrictions we are talking about adding to modifications of the Makefile

We're already strict about who has access to the secrets. Alas, "do the cloud bits" isn't a shell script, it's a bunch of tests in `./test/cases` orchestrated by `rtf`.  Having `make` take care of this solves dependency problems (are `rtf`, `moby` and `linuxkit` installed and of the correct version?) and allows us the flexibility to change the labels provided to `rtf` easily.

> Do we have the option of having the CI itself use the primary keys to generate short term, much more restricted keys, either on a per run basis or simply to turn them over more quickly (i.e. daily)?

That's definitely an option. I don't think it'd be much work to generate single use access keys using GCP's IAM API. I'm not sure how that would interface with `docker secrets` though but maybe that's not a concern if the keys are short lived.

> Could we put some/all of the bits which need the keys into the CI rather than the repo?

That's not really an option as it ends up duplicating effort. For example, we've duplicated `linuxkit run gcp` in a set of python scripts in LinuxKitCI instead of dogfooding. That won't scale for all the other cloud platforms we intend on testing.

ijc25:
> Alas, "do the cloud bits" isn't a shell script, it's a bunch of tests in ./test/cases orchestrated by rtf. Having make take care of this solves dependency problems (are rtf, moby and linuxkit installed and of the correct version?) and allows us the flexibility to change the labels provided to rtf easily.

So don't we need to gate changes to the tests in the same way as the `Makefile`, otherwise wouldn't I just write `tests/cases/exfiltrate-secrets.test`? This is essentially @talex5's "Does the Makefile rule run any software from the checkout (now or in the future)?" point I think.

dave-tucker:
@ijc25 good point. i'm debating whether or not to suggest that we just gate all PR's that are from non-whitelisted users. that way, we don't run the risk of forgetting to gate certain paths (i.e the tests)

justincormack:
@dave-tucker thats probably fine, we will whitelist all the regular contributors pretty quickly. Apparently some of the bits for this have a whitelist from maintainers via comments on the PR which would make it easy.

dave-tucker:
ping @talex5 - do you have a response to @riyazdf's comment?
per @ijc25 and @justincormack's comments I think it's sensible to only test PR's from whitelisted contributors unless approval is explicitly given.

where i'd like to eventually take this is per this diagram I pasted in Slack last week:
linuxkitci flow 1

the tests could be broken down in to a number of containers that can run in parallel.
each container would only contain the secrets required for it's tests (e.g **only gcp**)
ideally, the secrets in the container could be single use and generated by LinuxKitCI itself using the cloud providers IAM APIs (full credit to @ijc25 for that idea)

talex5:
@dave-tucker : As @ijc25 pointed out later, this would require running the test scripts with privileges too, so clearly the `Makefile` isn't the only file that needs to be trusted. Given this, it seems (even more) impossible to secure the system.

And even with whitelisting, you're still effectively giving the credentials to everyone on the whitelist. Who is responsible if a whitelisted contributor misuses them? And if someone submits a buggy PR, who is responsible for finding all the VMs and removing them? The CI can't track them if they're created by the tests.

It's certainly possible to proxy the requests through a restricted API provided by the CI (as @ijc25 suggests), but then you're not testing the tool against the real services.

> For example, we've duplicated linuxkit run gcp in a set of python scripts in LinuxKitCI instead of dogfooding. That won't scale for all the other cloud platforms we intend on testing.

Scaling isn't the problem. We can use `linuxkit run` to test the images on various plaforms; we just have to use a known-good version of it (not the one from the PR). The issue is whether it's worth accepting all the above security risks in order to test the *tool* (in addition to testing the images).

justincormack:
@talex5 I trust everyone who is whitelisted, we can add a nightly cron job to test if there are any stray machines created, and yes we absolutely do want to test the tool itself thats the whole point.

riyazdf:
@talex5: Maintainers are ultimately responsible for trusting those who they wish to put on the whitelist, and ensuring they will not abuse the system.  Maintainers are collectively responsible for any security response, CI related or otherwise.

I'm in agreement with @justincormack that I trust the folks on the whitelist and additional cron jobs and tests will help keep us on top of things.  We absolutely need to test the tool in an automated fashion across multiple platforms, so it's great that we're having this in-depth discussion to ensure we do so in a way that is safe.

rneugeba:
Just to chime in, let's not try to do any gymnastics trying to figure out if a PR is safe to run by an untrusted, ie not-whitelisted, user. Nothing should be run unless the PR is from a whitelisted user or a whitelisted user (or maintainer) hits a button on the PR. That way we can keep the whitelist in linuxkit-ci but move more of the logic of what is run on PRs/merge/etc in linuxkit's Makefiles

ijc25:
Perhaps we should arrange for something (CI, gordon, poule, etc) to label PRs which need checking by a maintainer and marking as safe, so we can filter on them?

dave-tucker:
> That way we can keep the whitelist in linuxkit-ci but move more of the logic of what is run on PRs/merge/etc in linuxkit's Makefiles

It did occur to me that it might be easier for LinuxKitCI to be configured using a YAML file similar to that of CircleCI. Each section (e.g `checkout`, `build`. `test`, `deploy`) would be an array of commands that would be executed in order. Defaults for each section would be documented. They could be overridden in a `linuxkitci.yml` which could be stored either in this repo or in the LinuxKitCI repo.

This would also document the relationship between LinuxKitCI and Makefiles so there was less magic involved.

justincormack:
The whole whitelisting thing is well understood and we should not try to reinvent anything...

justincormack:
eg https://github.com/jenkinsci/ghprb-plugin is the canonical example

talex5:
Given that we're not actually using any features of DataKitCI (the "pipeline" is a single step that calls `make`, which then coordinates everything), perhaps we should just use Jenkins?

dave-tucker:
@talex5 while that may be true as of now, I'd very much like for it to be more than one `make` incantation and would be happy to work with you to make better use of DatakitCI's features. I think the diagram I linked above goes a good way to explaning what I have in mind - comments welcome. I think we also all agree that having logs in git is much better that anything Jenkins provides today.

It seems we have consensus that testing PR's from only from whitelisted contributors is a reasonable way to protect our secrets while also enabling the `linuxkit` tool to be tested.

We also have some existing whitelisting implementations to borrow from.

The question is whether or not this feature can be added to DataKitCI and how soon we could make use of it.

justincormack:
Thinking about where to store the whitelist, I think having it in a file in the CI repo would be fine, happy to do PRs to add people there. Else it could be in the repo itself but it would have to be in the master branch not the checked out branch...

talex5:
@samoht: can we get the PR owner from the bridge metadata?

samoht:
@talex5 yup sure, I am adding this shortly. Let me know if you need more metadata.

dave-tucker:
Closing as I'm not actively working on this.

PR #1856, opened 2017-05-18
Update packages to use alpine base (plus other cleanups)

### LINKED ISSUE #1824: Set docker hub repository linuxkit as variable

rneugeba:
- Build packages previously using `alpine:3.5` or other to use the `linuxkit/alpine` as the base
  - Added additional packages to base
  - Typically installed required packages one stage in `/out` and then copy them into a `scratch` stage
- `alpine:edge` comes with tini, so switch packages to use that and remove our home-build tini package
- remove the toybox package. It was only used by the kernel packages and is not essential
- Make Hub org/repository build-time configurable use `make ... ORG=foo` (resolves #1824)
- Update all YAML files to use the new packages

I've test this with the `linuxkit` and `docker` images and hyperkit and qemu

image


pkg/init/Dockerfile
```diff
@@ -1,7 +1,14 @@
-# Use sha256 here to get a fixed version
-FROM alpine:edge@sha256:99588bc8883c955c157d18fc3eaa4a3c1400c223e6c7cabca5f600a3e9f8d5cd
+FROM linuxkit/alpine:6336329f15b4166514782eaa555cf0ffd35c519c@sha256:f6c2ce92910b1d6e4e5557850a554f4a3ae9f66c1e89ad86a24d6c6e550f165e AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out alpine-baselayout busybox musl
+
+# Remove apk residuals. We have a read-only rootfs, so apk is of no use.
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
```
justincormack: actually `init` is where I still sometimes use `apk add` for debug (you can remount `rw`), but I guess we should get rid of these now...

rneugeba: hopefully soon we have a dev container in which we should maybe not disabled `apk`

PR #1857, opened 2017-05-18
projects: add IMA namespacing project

tych0:
This PR adds the ima-namespacing project, which is a version of the IMA namespacing patches rebased on to 4.11.

There isn't a linuxkit/kernel-ima or linuxkit/init-ima project on the docker hub, so I haven't pushed the images there (I'm not really sure what goes in to creating one of these). But, here's the patches.

cc @riyazdf @guimagalhaes


justincormack:
I may be blind, but I can't see any difference in the `init` package here from the standard one.

@riyazdf can give you push access to linuxkit/kernel-ima etc on hub so you can push the builds.

Also Can you add an explan of what it is to the README, I am missing an overview. What does IMA stand for? The patch intro doesn't help much.

tych0:
On Thu, May 18, 2017 at 12:46 PM, Justin Cormack <notifications@github.com>
wrote:

> I may be blind, but I can't see any difference in the init package here
> from the standard one.
>
It includes the openssl and setfacl binaries, so that people can play
around with it on the host.


> @riyazdf <https://github.com/riyazdf> can give you push access to
> linuxkit/kernel-ima etc on hub so you can push the builds.
>
> Also Can you add an explan of what it is to the README, I am missing an
> overview. What does IMA stand for? The patch intro doesn't help much.
>
Yep, no problem. I'll push an update shortly.

justincormack:
If you just want to add binaries to the base system you can just add another package that just includes them into the `init` section.

tych0:
Right, thanks. I think I've sorted that out now.

justincormack:
Thanks makes sense now!

tych0:
Cool, thanks!

PR #1862, opened 2017-05-19
Label support

justincormack:
Add support for using labels to specify the image configuration yaml, as done in https://github.com/moby/tool/pull/39 - use `org.mobyproject.config` as the label with a JSON version of the yaml image config.

First example I have done `sysctl`, which I also converted to modern style build, and added a test to make sure it was still working correctly. I also added the `test-sysctl` one and `poweroff` to clean up the tests.

Once we do this elsewhere this will make the config files much simpler for the `linuxkit` images; external images will not have the correct labels.

dunnock_4185


avsm:
From a Slack convo:

> @avsm: does the old style still work? it was kind of nice to be able to grep for admin caps
> @avsm: a CI could still add them (along with hashes) for a fully elaborated yaml file
> @justincormack: yes putting config in the yaml overrides the label
> @justincormack: you can also see the generated config if you use `-v` on build

justincormack:
cc @dave-tucker I changed the default command for `poweroff` to have a timeout of `3` but left overrides for the ones that were `10`.

test/hack/test-ltp.yml
```diff
@@ -16,11 +16,7 @@ onboot:
     capabilities:
      - all
   - name: poweroff
-    image: "linuxkit/poweroff:961412b8ef5c5285de0d40ec076701d955eaa084"
-    pid: host
-    capabilities:
-      - CAP_SYS_BOOT
-    readonly: true
+    image: "linuxkit/poweroff:5673236900c6beae35bbceec90ded4b9add3f5ae"
```
dave-tucker: Could you presvere the old behaviour by adding a 30 second delay here?

justincormack: hmm, but actually I didnt change the default, I should change back the other ones that were `3`

justincormack: oh yes I did but I shouldnt have, going to revert. It makes little sense having `command` in yaml as it is already in image

justincormack:
@dave-tucker they should be unchanged now

pkg/sysctl/Dockerfile
```diff
@@ -1,3 +1,16 @@
...
+
 FROM scratch
-COPY . ./
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /go/bin/sysctl /usr/bin/sysctl
+COPY etc/ /etc/
 CMD ["/usr/bin/sysctl"]
+LABEL org.mobyproject.config='{"net": "host","pid": "host", "ipc": "host", "readonly": true, "capabilities": ["CAP_SYS_ADMIN"]}'
```
rneugeba: would it be possible to make this multi-line for better readability?

test/cases/040_packages/group.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+# SUMMARY: LinuxKit security tests
```
rneugeba: not security tests

test/pkg/sysctl/Makefile
```diff
@@ -0,0 +1,29 @@
+.PHONY: tag push
+
+BASE=alpine:3.5
+IMAGE=test-sysctl
+
+default: push
+
+hash: Dockerfile check.sh
```
rneugeba: Is there a reason this package makefile is old-style rather than the new more compact form with git tree hash etc

pkg/sysctl/Dockerfile
justincormack: The advantage of this format is it is the only way to write something you can cut and paste as JSON, as it needs backslashes if it has line break. Hence the single quotes so the double ones don't need to be escaped.

test/pkg/sysctl/Makefile
justincormack: all the test ones need redoing, was easier to leave this old style for now

test/cases/040_packages/group.sh
justincormack: fixing

pkg/sysctl/Dockerfile
rneugeba: OK, makes sense. escaping quotes makes this error prone and less readable

test/pkg/sysctl/Makefile
rneugeba: Ok, agree best to re-do all the test ones new style in one PR

rneugeba:
are you planning to convert the other packages using `linuxkit/go-compile` as well? Otherwise it's introducing more inconsistency while we should strive for more consistency...

justincormack:
@rneugeba yes, but didnt want to do anything else in the same PR its got enough in...

PR #1865, opened 2017-05-20
more Go compile fixes

justincormack:
Minor tweaks to both versions of go-compile, update sysfs to use alpine base image and same build as sysctl.


rneugeba:
since `linuxkit/alpine` changed (and some of the versions(, shouldn't all the packages using it be changed too...

justincormack:
No, we can't make everyone recompile all packages whenever anything changes when we don't have automation. Once all the packages have the same base we can do an update pass.

PR #1867, opened 2017-05-22
Allow specifying the format of a qemu drive

justincormack:
Default remains qcow2 but you can specify raw or other supported formats.

great_tit

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -26,6 +26,7 @@ type QemuConfig struct {
 	GUI            bool
 	DiskPath       string
 	DiskSize       string
+	DiskFormat     string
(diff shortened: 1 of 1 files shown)


rneugeba:
Out of curiosity, what's the use case for this option?

justincormack:
I am switching the code in `moby` that makes raw images to using linuxkit/qemu rather than libguestfs. libguestfs is just a kernel booting under qemu running stuff and as we can do that already we may as well dogfood, so I need to be able to output different formats. Apart from dogfooding, this removes the circularity of the `moby` tool for building `docker` requiring `docker` to do so; it will only fall back to `docker` if you don't have `qemu` installed natively.

justincormack:
(also will remove most of `tools/` in the process)

PR #1873, opened 2017-05-22
Clean up of Qemu run

justincormack:
For all output formats except kernel+initrd, recommend that people just
add the full path of the file they want to run. If you do this, generally
make the options automatic.

Split the uefi option to mean "use uefi firmware" not be ISO specific.

Allow specifying a bootable disk image, so we can test disk image output
formats with qemu too.

Specifying a prefix for anything other than kernel+initrd is deprecated
and will be removed at some point.

top-10-dirty-animals-covered-in-mud-l-cbxk4f

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -19,7 +19,7 @@ const QemuImg = "linuxkit/qemu:17f052263d63c8a2b641ad91c589edcbb8a18c82"
 
 // QemuConfig contains the config for Qemu
 type QemuConfig struct {
-	Prefix         string
+	Path           string
 	ISO            bool
 	UEFI           bool
 	Kernel         bool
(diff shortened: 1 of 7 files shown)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -80,19 +80,57 @@ func runQemu(args []string) {
...
 		flags.Usage()
 		os.Exit(1)
 	}
-	prefix := remArgs[0]
+	path := remArgs[0]
 
-	// Print warning if conflicting UEFI and ISO flags are set
-	if *uefiBoot && *isoBoot {
-		log.Warnf("Both -iso and -uefi have been used")
+	// if user specifed a path that exists and ends in .iso they meant an ISO
```
riyazdf: typo nit `specified`

src/cmd/linuxkit/run_qemu.go
```diff
@@ -80,19 +80,57 @@ func runQemu(args []string) {
...
+	_, stat := os.Stat(path)
```
riyazdf: it's weird to not just have this variable be `err`

src/cmd/linuxkit/run_qemu.go
justincormack: fixed

src/cmd/linuxkit/run_qemu.go
justincormack: well, thats as I was using it later, but reworked

src/cmd/linuxkit/run_qemu.go
```diff
@@ -80,19 +80,59 @@ func runQemu(args []string) {
...
+				log.Fatalf("Cannot boot from disk and specify a disk as well at present")
+			}
+			*disk = path
+			*kernelBoot = false
+		}
+	}
+
+	// override default kernel boot if ISO specified
+	if *isoBoot {
+		*kernelBoot = false
```
rneugeba: I'm not sure I like this complicated scheme. It's really hard to describe in words, so probably not right...if we just ditch the old behaviour when it comes to ISO or UEFI boot it'll be simpler.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -80,19 +80,59 @@ func runQemu(args []string) {
...
 		flags.Usage()
 		os.Exit(1)
 	}
-	prefix := remArgs[0]
+	path := remArgs[0]
 
-	// Print warning if conflicting UEFI and ISO flags are set
-	if *uefiBoot && *isoBoot {
-		log.Warnf("Both -iso and -uefi have been used")
+	// if user specified a path that exists and ends in .iso they meant an ISO
```
fntlnz: This part determining the actual kind of bootloader to apply would fit better in a separate function , will make easier to maintain, test and debug the whole runner and result in less headaches to future ourselves :eyes:

fntlnz: Would be a good idea to do an in-depth inspection and check for the boot record and the primary volume descriptor? The first would also tell us if is bios or efi

rneugeba: I think it would be overkill to start prodding around inside the images we boot.

fntlnz: For performance reasons or because it could not be a owe of the runner ?

justincormack: I don't believe its really possible to tell, and thats also too much magic for a simple tool like this.

justincormack:
Ok I simplified this a lot, you now must specify the full path for anything except kernel boot. The autodetection should work for almost all cases so you don't need to specify the flags unless you are doing something odd like booting kernel mode with uefi, but you should be able to do that.

justincormack:
Also added a test for qcow2 boot.

src/cmd/linuxkit/run_qemu.go
fntlnz: Ok understandeable :angel:

justincormack:
Green now...

PR #1877, opened 2017-05-22
kernel-config project: add a real implementation of kernel configs

tych0:
Hi all,

I've added a "real" version of what the kernel configs would look like. I've included the script I used to generate these diffs so people can see how I generated them. I also made it so that check-kernel-config now runs before we build a kernel, so folks wouldn't build kernels without first ensuring their config has all of the options we want.

Thoughts welcome!


tych0:
Ping, anyone have any thoughts on this?

justincormack:
well lets merge anyway as its projects and then its easier to try out. I havent had a chance to look at it yet as I was out of office

tych0:
Sounds good. I think this is mostly done for now, so anyone interested in trying it out should do so. Do we have a process for figuring out how to promote stuff from projects yet?

justincormack:
Well the process is to open a PR merging it into base and removing it from projects.

I would quite like to see how it would work with a hardened config and with arm64 config...

rneugeba:
@tych0 sorry, I also hadn't had chance to look at this. good it's merged. Hopefully soon i have time to play with this

tych0:
I think how it works for arm64 should be essentially the same as x86 (with arm64-specific config options in their respective file extensions).

For another channel, we need to do a bit more work to make projects/ uniform, so we can figure out what config options are needed and how to auto-merge them and stuff. I have some thoughts on this but haven't poked at it yet.

PR #1881, opened 2017-05-23
examples: switch dhcpd to onboot in redis-os example

rneugeba:


justincormack:
LGTM

PR #1882, opened 2017-05-23
VMware vCenter run capability

thebsdbox:
**- What I did**
Added a VMware vCenter backend for LinuxKit

**- How I did it**
Using the VMware GO SDK in order to allow linuxkit to push ISOs to VMware and run them.

**- How to verify it**
```
./bin/linuxkit run vcenter -url=https://u:p@vcenter/sdk -datastore=vSphereNFS \
-network=ProdNetwork -hostname=esxi01.local -powerOn -persistentSize=1024 \
 linuxkit

Creating new LinuxKit Virtual Machine
Uploading LinuxKit file [linuxkit.iso]
Adding ISO to the Virtual Machine
Adding a persistent disk to the Virtual Machine
Adding VM Networking
Powering on LinuxKit VM
```

VMware vCenter `run` functionality


src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	controller, err := devices.FindDiskController("scsi")
+	if err != nil {
+		log.Fatalf("Unable to find SCSI device from VM configuration\n%v", err)
+	}
+	// The default is to have all persistent disks named linuxkit.vmdk
+	disk := devices.CreateDisk(controller, dss.Reference(), dss.Path(fmt.Sprintf("%s/%s", *newVM.vmName, "linuxkit.vmdk")))
+
+	disk.CapacityInKB = *newVM.persistent * 1024
```
thebsdbox: As per #1888 we create disks in MB here also.

src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,292 @@
...
+
+	newVM.vmName = flags.String("vmname", "", "Specify a name for virtual Machine")
+	newVM.path = flags.String("path", "", "Path to a specific image")
+	newVM.persistent = flags.Int64("persistentSize", 0, "Size in MB of persistent storage to allocate to the VM")
+	newVM.mem = flags.Int64("mem", 1024, "Size in MB of memory to allocate to the VM")
+	newVM.vCpus = flags.Int("cpus", 1, "Amount of vCPUs to allocate to the VM")
+	newVM.poweron = flags.Bool("powerOn", false, "Power On the new VM once it has been created")
+
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s push vcenter [options] [name]\n\n", invoked)
```
justincormack: should be `run` not `push`

src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,292 @@
...
+		fmt.Printf("'name' specifies the full path of an image file which will be uploaded\n")
```
justincormack: "image file to run"

src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,292 @@
...
+		flags.PrintDefaults()
+	}
+
+	if err := flags.Parse(args); err != nil {
+		log.Fatalln("Unable to parse args")
+	}
+
+	remArgs := flags.Args()
+	if len(remArgs) == 0 {
+		fmt.Printf("Please specify the prefix to the image to push\n")
```
justincormack: run not push; also this should not be the prefix it should be the full path

src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,292 @@
...
+		flags.Usage()
+		os.Exit(1)
+	}
+	prefix := remArgs[0]
```
justincormack: Please make this be the path, and extract a prefix if you need one. See the changes in qemu yesterday. We are phasing out use of prefix other than for hyperkit kernel_initrd format

src/cmd/linuxkit/run_vcenter.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func checkFile(file string) {
+	if _, err := os.Stat(file); err != nil {
+		if os.IsPermission(err) {
+			log.Fatalf("Unable to read file [%s], please check permissions", file)
+		}
+		if os.IsNotExist(err) {
+			log.Fatalf("File [%s], does not exist", file)
+		}
```
riyazdf: can we add a catch all for a non-nil err that isn't one of these two cases?

ex: `log.Fatalf("Unable to stat file [%s]: %v", file, err)`

thebsdbox: Sure thing :)

Not sure if we could make this a general function into `util.go` at some point. But in the interim I've updated for a catch-all.

PR #1883, opened 2017-05-23
Add a simple qemu guest agent package

ijc25:
We don't actually build it here, we just use the alpine package.

Can be instantiated with a service stanza such as:
```
  - name: qemu_ga
    image: "linuxkit/qemu_ga:ffc2c35a1126aeab5a324254ce83c4b302eeaf04"
    binds:
      - /dev/vport1p1:/dev/vport1p1
    command: ["/usr/bin/qemu-ga", "-p", "/dev/vport1p1"]
    net: host
```

Note that `net: host` is required since one function (and the only one I've
tested) is to report IP addresses to the host (e.g. via `virsh domifaddr`).

The mdev tool appears to not create the symlinks which udev would provide
`/dev/virtio-ports/org.qemu.guest_agent.0` hence the need to hardcode the
potentially unstable `/dev/vportNp1`, `N` seems to depend on the number and
order of virtio devices in use. I don't know if it is possible to get mdev to
create these links. For reference the udev rule is:

    SUBSYSTEM=="virtio-ports", KERNEL=="vport*", ATTR{name}=="?*", SYMLINK+="virtio-ports/$attr{name}"

See https://wiki.libvirt.org/page/Qemu_guest_agent for more info on the guest
agent.

**- What I did**

Added a new pkg containing the qemu guest agent

**- How I did it**

**- How to verify it**

Configure a VM with the appropriate channel. I used the following in a libvirt domain's `channels` list:

```
<channel type='unix'>
   <target type='virtio' name='org.qemu.guest_agent.0'/>
</channel>
```

With this commands such as `virsh domifaddr --source agent <domain>` ought to work, you may need to adjust the `/dev/vportNp1` to match your VM (I'm not 100% sure how `N` is arrived at, although it appears to be somewhat stable for a given VM configuration).

Added a Qemu Guest Agent package.

### DIFF
--- a/pkg/qemu-ga/Dockerfile
+++ b/pkg/qemu-ga/Dockerfile
@@ -1,0 +1,12 @@
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS build
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN mkdir -p /out/var/run
+RUN apk add --no-cache --initdb -p /out \
+    qemu-guest-agent \
+    musl
+FROM scratch
+WORKDIR /
+ENTRYPOINT []
+COPY --from=build /out /
+CMD ["/usr/bin/qemu-ga", "-p", "/dev/vport0p1"]
+LABEL org.mobyproject.config='{"net": "host"}'
--- a/pkg/qemu-ga/Makefile
+++ b/pkg/qemu-ga/Makefile
@@ -1,0 +1,14 @@
+.PHONY: tag push
+
+ORG?=linuxkit
+IMAGE=qemu-ga
+
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+
+default: push
+
+tag: Dockerfile
+	docker build --network=none -t $(ORG)/$(IMAGE):$(HASH) .
+
+push: tag
+	docker pull $(ORG)/$(IMAGE):$(HASH) || docker push $(ORG)/$(IMAGE):$(HASH)
--- a/projects/swarmd/swarmd.yml
+++ b/projects/swarmd/swarmd.yml
@@ -20,6 +20,10 @@ onboot:
   - name: metadata
     image: "linuxkit/metadata:b053fd6a945829bb626bb3546cad69aae0ff7803"
 services:
+  - name: qemu-ga
+    image: "linuxkit/qemu-ga:e5fbcf55926f6e1a96b3e49a392e547e8be1022c"
+    binds:
+      - /dev/vport0p1:/dev/vport0p1
   - name: rngd
     image: "linuxkit/rngd:b67c3151a52b05db50e6207b40876900f2208d14"
   - name: ntpd


justincormack:
`binds` and `net:host` should not be added to the image label so they do not need to be in the yaml.

There should be a test (eg that you can get IP address).

ijc25:
> binds _[...]_ should not be added to the image label so they do not need to be in the yaml

(ITIYM "should" rather than "should not")

As I discussed in the commit message, the name of the device is unpredictable, at least to me so far, so there isn't anything I could reasonably put there as a default, I have seen both `/dev/vport0p1` and `/dev/vport1p1` in VMs for which I cannot discern the reason for the difference.

> There should be a test (eg that you can get IP address).

Apart from the above (which might be OK since the device name seems fairly consistent for a given VM template) I'll need to see if I can find a client other than libvirt, since standing up a full libvirt stack in the tests is probably a bit heavy weight. https://github.com/qemu/qemu/blob/master/scripts/qmp/qemu-ga-client looks promising on first glance. It'll probably take me a bit to get round to this.

pkg/qemu-ga/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM alpine:3.5
```
rneugeba: could you use the new(ish) alpine base image for this? I've spent some time last week to convert all the packages to use it and it would be good to not introduce new package which don't...

pkg/qemu-ga/Dockerfile
```diff
@@ -0,0 +1,3 @@
...
+RUN apk add --update qemu-guest-agent
+CMD ["/usr/bin/qemu-ga"]
```
rneugeba: could you also add a `LABEL org.mobyproject.config` similar to the one in `./pkg/sysctl/Dockerfile` as this makes the YAML files a lot simpler

pkg/qemu-ga/Dockerfile
ijc25: Where is that? If it is `tools/alpine` then isn't that a compile image (with compilers and a mirror in it). Since this is a runtime container I want it to be pretty minimal and not pull all that in. If `qemu-ga` was statically linked this would say `FROM scratch`, but it has library dependencies and working from an alpine base is the easiest way to resolve those.

pkg/qemu-ga/Dockerfile
ijc25: I can, although see my reply to Justin about the device names.

pkg/qemu-ga/Dockerfile
rneugeba: yes, `tools/alpine` basically has a `apk` mirror of the union of all packages we build in `linuxkit`. You can create a multi-stage build from it and install just the packages you want. `pkg/dhcpcd/Dockerfile` has a good(?) example on how to use it. It installs a Alpine base system plus `dhcpcd`. Using the alpine base image allows us to create images in a more reproducible way as there are no network dependencies.

Since the qemu-ga is not in the base, you'd first have to add it to `./tools/alpine/packages`, build/push a new base image and then use it for your package.

This stuff has move a lot since

ijc25: OK, it's a bit roundabout but I reckon I can follow the pattern in the `dhcpcd` package. Might be a while though

rneugeba:
FYI, I'm currently pushing a new base image which has a bucnh of VM integration tools, including the qemu ones. Will PR once the push finished

ijc25:
Updated to use the `linuxkit/alpine` image as a base and to add `net: host` via the label.

I also added an entry to `projects/swarmd/swarmd.yml` to reflect what I actually tested (and what I want it for).

ijc25:
I missed https://github.com/linuxkit/linuxkit/pull/1883#pullrequestreview-39773240, will look into that now.

pkg/qemu-ga/Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: tag push
+
+ORG?=linuxkit
+IMAGE=qemu_ga
```
rneugeba: minor nit, but the package is called `qemu_qa` but the directory is `qemu-qa`

ijc25: Ah yes, that's because I created the directory but then found $something which didn't allow `-`.

ijc25: OTOH we have `pkg/open-vm-tools` which doesn't have this problem and is much the same from a packaging perspective. I'm going to try switching it all back to `qemu-ga` and see what breaks.

ijc25: Nothing broke, pushed the change.

ijc25:
Looks like one can interact with the guest agent via a host side socket and socat (some examples in http://wiki.qemu.org/Features/GuestAgent). However `linuxkit run` doesn't currently expose a way to expose a guest agent. I suppose either a specific option (`--guest-agent-endpoint-path`) or some sort of more general trap door mechanism to add arbitrary things to the QEMU command line is the way to go.

Not sure how either would interact with running qemu inside a container though.

justincormack:
Its fine if the tests are only run with native qemu not containerised.

rneugeba:
yeah, find with running native only. There is a test in `test/cases/010_platforms/000_qemu/100_container/test.sh` for local install

ijc25:
Should the test fail, cancel or skip if there is no host qemu available?

rneugeba:
return `RT_CANCEL` should do the right thing

ijc25:
So, while you can interact with the GA via `socat` it really doesn't work well, it's very async and if you run `socat` in a loop then answers can turn up several iterations later. See http://wiki.qemu.org/Features/GuestAgent#QEMU_Guest_Agent_Protocol for more info.

I've looked but I can't seem to find a simple client suitable for testing other than libvirt and I don't have time to either produce such a client or stand up libvirt in the test framework right now, so I guess this'll have to go without tests for now.

justincormack:
Ok, yes it looks pretty horrible according to the docs...

ijc25:
I especially like the need to sometimes inject 0xff into the byte stream, chosen because it can't be a valid UTF-8 character, in order to avoid deadlocks.

rneugeba:
let's merge without a test then

Issue #1886, opened 2017-05-24
External disk not work as expected


HackToday:
**Description**

**Steps to reproduce the issue:**
1. use moby build  docker example (https://github.com/linuxkit/linuxkit/tree/master/examples)
 docker kernel + initrd  images
2. then follow doc from 
https://github.com/linuxkit/linuxkit/blob/master/docs/external-disk.md

It not specify any requirements. so I ran following command

`linuxkit run qemu -disk foo.img -disk-size 1024 docker`

And if you in the VM, and check as below:

```
/ # runc list
ID          PID         STATUS      BUNDLE                         CREATED                          OWNER
dhcpcd      599         running     /run/containerd/linux/dhcpcd   2017-05-24T05:30:51.153828733Z   root
docker      635         running     /run/containerd/linux/docker   2017-05-24T05:30:53.598604676Z   root
ntpd        663         running     /run/containerd/linux/ntpd     2017-05-24T05:30:55.980477397Z   root

/ # ls -l /dev/sd*
ls: /dev/sd*: No such file or directory
/ # ls -l /dev/vd*
ls: /dev/vd*: No such file or directory
/ # 

```

**Describe the results you received:**

not any dev/sd*  dev/vd*  available

**Describe the results you expected:**

should work as the doc https://github.com/linuxkit/linuxkit/blob/master/docs/external-disk.md
dev/vd*  or  dev/sd* any available

deitch:
What platform are you running it on?

deitch:
And what version of `linuxkit`?

HackToday:
Hi  @deitch  

For your questions, info as below:

```
Linuxkit: linuxkit version 0.0
commit: 72ad8ac3a3f3a79904c9eaf585dd3d4cff8b678c

Platform:  It is one VM,  OS is: CentOS Linux release 7.2.1511 (Core) 
```

deitch:
I don't have immediate access to a CentOS platform to try. Someone from linuxkit team have that?

HackToday:
Hi @deitch , Per my understanding, linuxkit is platform independent. It said support Linux platform.(It seems not  care about OS distribution)

deitch:
True. The kernel you built is independent. But you are running it in `qemu`, which can vary between platforms, and via the `linuxkit` binary, which might as well. I had no problems on Mac, you did on CentOS. Might be irrelevant, might not be.

[...]

PR #1888, opened 2017-05-24
cmd: Unify disk command line options for local hypervisors

rneugeba:
- '-disk-size' is now always in MB
- The disk will be created if it doesn't exist (didn't happen in qemu)

Update the documentation.

resolves: #1886

image

### DIFF
--- a/docs/external-disk.md
+++ b/docs/external-disk.md
@@ -5,38 +5,18 @@
 2. Mount the disk
 
 ## Make Disk Available
-In order to make the disk available, you need to tell `linuxkit` where the disk file or block device is. This depends on your run method.
+In order to make the disk available, you need to tell `linuxkit` where the disk file or block device is.
 
-**Note:** With all of the options below, *always* give disk options before the _prefix_.
+All local `linuxkit run` methods (currently `hyperkit`, `qemu`, and `vmware`) take two arguments:
 
-### Mac HyperKit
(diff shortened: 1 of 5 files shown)


thebsdbox:
@justincormack yes those changes, worked as expected.

```
./vmware-vdiskmanager
...
     The capacity can be specified in sectors, KB, MB or GB.
```

justincormack:
Can we support `-disk-size 1G` as well - that currently works with qemu, and megabytes are quite small for most use cases.

rneugeba:
Call me lazy, but I'm not too keen to implement string parsing (and error handling) for something as trivial as disk sizes and then transpose them into `G`, `GB`, `M` etc. I'd be fine though to up the unit from `MB` to `GB`

justincormack:
I have use cases for MB and GB. Its only a couple of lines of code, I can add it later I guess...

ijc25:
https://github.com/docker/go-units contains handling for MB and GB etc, so you don't need to manually implement the string parsing.

justincormack:
@ijc but that requires you write "MiB" so I don't class it as human writeable.

ijc25:
It supports MB just fine

rneugeba:
Yes, MB is not power of 2. I kinda expect my disk sizes to be power of two...

I'll add `G` (and `M`) though it defaults to `M`

rneugeba:
Updated it to default to MB but also accept `M` and `G` as units

PR #1891, opened 2017-05-24
Add -start-vpnkit flag to linuxkit run hyperkit

MagnusS:
Adds a flag to launch a new vpnkit instance. This enables port
forwarding to be configured via a 9p mount (tag:port) if the guest 
supports it. Vsock port 62373 is also opened so vpnkit can forward 
connections back to the guest.

If -start-vpnkit is not specified HyperKit will connect to the
vpnkit instance managed by Docker for Mac, as before.

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
@@ -33,6 +34,7 @@ func runHyperKit(args []string) {
 	ipStr := flags.String("ip", "", "IP address for the VM")
 	state := flags.String("state", "", "Path to directory to keep VM state in")
 	vsockports := flags.String("vsock-ports", "", "List of vsock ports to forward from the guest on startup (comma separated). A unix domain socket for each port will be created in the state directory")
+	startVPNKit := flags.Bool("start-vpnkit", false, "Launch a new VPNKit instead of reusing the instance from Docker for Mac. The new instance will be on a separate internal network. This enables IP port forwarding from the host to the guest if the guest supports it.")
 
 	if err := flags.Parse(args); err != nil {
 		log.Fatal("Unable to parse args")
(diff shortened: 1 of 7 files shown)


MagnusS:
This also enables network access from the VM without a running DfM

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -6,13 +6,75 @@ import (
...
 	"os"
+	"os/exec"
 	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/moby/hyperkit/go"
 	"github.com/satori/go.uuid"
 )
 
+var defaultVPNKits = []string{"vpnkit",
```
riyazdf: nit: can we newline for the individual entries and closing brace?  It'll make for cleaner diffs if we ever edit this

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -6,13 +6,75 @@ import (
...
+	"/usr/local/bin/vpnkit",
+	"/Applications/Docker.app/Contents/Resources/bin/hyperkit"}
+
+func createListenSocket(path string) (*os.File, error) {
```
rneugeba: could we move the vpnkit functions further to the end. I seem to recall that in go you are supposed to have the exported/main functions first and have the internal ones later in the file. To me as a C programmer that's weird, but hey...I certainly wouldn't put them at the start as the main purpose of this file to run hyperkit

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -6,13 +6,75 @@ import (
...
+	"os/exec"
 	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/moby/hyperkit/go"
 	"github.com/satori/go.uuid"
 )
 
+var defaultVPNKits = []string{"vpnkit",
+	"/usr/local/bin/vpnkit",
```
justincormack: is this needed? lookpath will find it anyway from "vpnkit"

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -116,6 +202,12 @@ func runHyperKit(args []string) {
 	h.Memory = *mem
 	h.DiskSize = *diskSz
 
+	// Add 9p and vsock for port forwarding if VPNKit is launched automatically
+	if *startVPNKit {
+		h.Sockets9P = []hyperkit.Socket9P{{Path: vpnKitPortSocket, Tag: "port"}}
+		h.VSockPorts = append(h.VSockPorts, 62373) // forwarding
```
justincormack: whats `62373`? Could do with a comment as it is non obvious.

MagnusS:
@rneugeba Yes, if you want to have multiple VMs in the same network you would need an option for specifying the vpnkit socket. It could perhaps be useful if we want to do automated testing of network services with several nodes without having to install full DfM (you would only need hyperkit + vpnkit binaries). But it should be easy to add support for that later if we need it by adding an option to specify a custom vpnkit socket instead of using the DfM one by default. 

Only the node that was started with `-start-vpnkit` would be able to forward ports from the host though.

src/cmd/linuxkit/run_hyperkit.go
MagnusS: Hm no, I'll remove it - if you have a relatively new DfM installed it should be in path, and older versions wouldn't have the features we need.

src/cmd/linuxkit/run_hyperkit.go
MagnusS: It's the vsock port used to forward network traffic from host->guest. I'll add a comment

src/cmd/linuxkit/run_hyperkit.go
MagnusS: Good point, I'll do that in the future

src/cmd/linuxkit/run_hyperkit.go
MagnusS: Yes, I'll move them to the end - not used to that either

MagnusS:
Thanks for the reviews. Force pushed a new version with the following changes:
- Only look for "vpnkit" in path
- Move vpnkit helpers to the end
- More descriptive comments on the automatically added 9p mount/vsock port and the helper functions

justincormack:
@MagnusS there is interest in having local swarm clusters so I think adding an option to specify the socket would be useful (and we should stop relying on d4m as we wont have that in CI)

PR #1896, opened 2017-05-25
Add a mkimage package

justincormack:
This is a package that creates bootable disk images. This will be used to replace the `mkimage-*` packages. It is generally expected that you will use `qemu` to make these, so it can make `raw` or `qcow2` images, or other supported disk formats. Currently it will make `ext4` bootable `syslinux` based images, but I plan to add more configuration options later.

first-cygnet_1395980c

### DIFF
--- a/test/cases/040_packages/001_mkimage/mkimage.yml
+++ b/test/cases/040_packages/001_mkimage/mkimage.yml
@@ -1,0 +1,26 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0"
+init:
+  - linuxkit/init:cbd7ae748f0a082516501a3e914fa0c924ee941e
+  - linuxkit/runc:24dfe632ed3ff53a026ee3fac046fd544434e2d6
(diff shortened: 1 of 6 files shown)


test/cases/040_packages/001_mkimage/mkimage.yml
```diff
@@ -0,0 +1,27 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0"
+init:
+  - linuxkit/init:cbd7ae748f0a082516501a3e914fa0c924ee941e
+  - linuxkit/runc:24dfe632ed3ff53a026ee3fac046fd544434e2d6
+  - linuxkit/containerd:f1130450206d4f64f0ddc13d15bb68435aa1ff61
+  - linuxkit/ca-certificates:4e9a83e890e6477dcd25029fc4f1ced61d0642f4
```
rneugeba: probably don't need these

justincormack:
Ok initial CI was broken as I hadnt rebased over the qemu changes. Removed the certs.

PR #1897, opened 2017-05-25
RFC: Enable shellcheck for format package

majst01:
**- What I did**

**- How I did it**
This is the first step of verifying that all shell scripts are at least passing shellcheck.
**- How to verify it**
It is a integrated step in the Makefile.
First release to get feedback from maintainers if this approach is feasible. If agreed, i will do this for all pkg applications.
hamster

### DIFF
--- a/pkg/format/Makefile
+++ b/pkg/format/Makefile
@@ -7,7 +7,10 @@ DEPS=Dockerfile format.sh
 
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 
-tag: $(DEPS)
+shellcheck:
+	docker run --rm -it -v $(PWD):/mnt nlknguyen/alpine-shellcheck *.sh
+
+tag: $(DEPS) shellcheck
 	docker build --squash --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
 
 push: tag
--- a/pkg/format/format.sh
+++ b/pkg/format/format.sh
@@ -8,21 +8,21 @@
 do_fsck()
 {
 	# preen
-	/sbin/e2fsck -p $*
+	/sbin/e2fsck -p "$@"
 	EXIT_CODE=$?
 	# exit code 1 is errors corrected
 	[ "${EXIT_CODE}" -eq 1 ] && EXIT_CODE=0
 	# exit code 2 or 3 means need to reboot
-	[ "${EXIT_CODE}" -eq 2 -o "${EXIT_CODE}" -eq 3 ] && /sbin/reboot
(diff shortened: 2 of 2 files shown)


justincormack:
I would rather we checked this globally as part of CI rather than on a per package basis. We really want to remove the per package Makefiles completely at some point.

majst01:
So should i add these checks into to toplevel Makefile? I do not know how the datakit ci works.

justincormack:
@majst01 if it is in the top level Makefile then we can add to tests yes. The other option that occured to me though is we could add it to the Dockerfile, like we do for the `go lint` etc tests, so if the shellcheck is in the base `linuxkit/alpine` image we can just run it after doing `COPY blah.sh /`. That way the Makefiles stay simple and it just gets done where needed.

rneugeba:
@justincormack @majst01 I think I can add shellcheck to the alpine base. From `koalaman/shellcheck` we have:
```
/mnt # ldd /usr/local/bin/shellcheck
	/lib/ld-musl-x86_64.so.1 (0x557cc6ed6000)
	libgmp.so.10 => /usr/local/lib/libgmp.so.10 (0x7f67f2bbe000)
/mnt # ls -l /usr/local/bin/
total 15148
-rwxr-xr-x    1 root     root      15511160 May 25 02:36 shellcheck
/mnt # ls -l /usr/local/lib/
total 952
-rwxr-xr-x    1 root     root        563680 May 25 02:37 ld-musl-x86_64.so.1
-rwxr-xr-x    1 root     root        407272 May 25 02:37 libgmp.so.10
```
so we could just copy `/usr/local` from there into the base

justincormack:
I would rather compile it from source than extract a binary from a random hub repo...

rneugeba:
then we have to do this:
https://github.com/koalaman/shellcheck/blob/master/Dockerfile_builder

justincormack:
Oh, I didnt realise it was written in Haskell, yeah maybe using the binary is better

rneugeba:
OK, I have it added to the Alpine base. Should be in a PR later today

majst01:
I think using the binary from the docker image from koalaman is ok because he is the author of shellcheck. Building it by our own is a bit difficult and only haskell guys know how :-)

After it is added to linuxkit/alpine i will change my PR to use it.

rneugeba:
I've pushed the image to hub and you can use it as: `linuxkit/alpine:366c1f4a3244572e8acd5569a212a1c932495401`. I still need to update all the other packages to use it

rneugeba:
actually, had to change the base image a bit more and the hash image is now:
`linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9`

justincormack:
This no longer applies, so closing. We want to have this as a top level standalone option I think.

majst01:
Of course, once i get some spare time again i will prepare another PR therefore.

justincormack:
Thanks!

PR #1907, opened 2017-05-26
linuxkit: add -vpnkit-socket flag

MagnusS:
Add a new flag to set the VPNKit ethernet socket path. Defaults to auto,
which will try to use the existing socket from Docker for Mac.

This flag can be used to connect to the VPNKit instance started by
another linuxkit VM launched with `-start-vpnkit`. VMs connected to the
same VPNKit instance will be on the same virtual internal network.

(related to discussion in #1891)


MagnusS:
Hm - not sure why CI failed. Seems to be unrelated:
```
ssh: connect to host 35.189.232.213 port 22: Connection refused
```

justincormack:
This code is a bit confusing, how is the value `auto` actually used? I would expect the default value to be the actual path of the socket...

MagnusS:
@justincormack "auto" is used by the hyperkit API: https://github.com/moby/hyperkit/blob/master/go/hyperkit.go#L134

But I can use the path as the default instead of auto. The HyperKit API just looks for the socket in the standard location anyway.

MagnusS:
Pushed an update which defaults to the DfM standard path

Issue #1901, opened 2017-05-25
Update alpine base to alpine 3.6


rneugeba:
Alpine 3.6 got released yesterday. We should update the base image to use it.

#1900 also enabled content trust on all packages so we should do a complete update of all packages and remove the sha256 from the `FROM` lines as they re no longer needed.

@justincormack I'm happy to do that tomorrow (Friday) as I have further changes to most packages anyway

PR #1909, opened 2017-05-26
Update Alpine base, add config labels, and trust

rneugeba:
- Switch Alpine base to 3.6
- Add shellcheck to Alpine base (see also #1897)
- Update most packages to use the config label
- Update YAML files to use:
  - New package versions
  - Remove config info now supplied by the config label
  - Add more images to the trust section

Also:
- Update documentation
- Fix Alpine base image hash calculation

resolves: #1901

image

### DIFF
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -2,49 +2,31 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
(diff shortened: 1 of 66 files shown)


examples/docker.yml
```diff
@@ -60,5 +42,15 @@ files:
 trust:
   image:
```
riyazdf: this is awesome!  I added an organization level key so I'll edit this to make it less wieldy

PR #1910, opened 2017-05-26
rebuild bin/moby as necessary

### LINKED ISSUE #1880: Makefile does not rebuild bin/moby when it should
I just tripped over this because the reduced `sysctl` stanza in my swarmd.yml wasn't working after 702ad5d9d9646826a158774c9bd2cde259723856 when I thought it should be, it's not the first time this sort of thing has tripped me (and others) up but I couldn't find an issue for it

Looking at the `Makefile` I expect `bin/rtf` has the same issue.

I'm not sure what the right answer is. It's a `docker run` and not a `docker build` so I guess caching can never help us. We could write the commit to a stamp file and update only if it changed and have `bin/moby` depend on that, something like:
```

.PHONY: .bin.moby.stamp
.bin.moby.stamp:
        echo $(MOBY_COMMIT) > .bin.moby.stamp.new
        if cmp -s .bin.moby.stamp .bin.moby.stamp.new ; then mv .bin.moby.stamp.new .bin.moby.stamp else rm .bin.moby.stamp.new ; fi

bin/moby: .bin.moby.stamp | bin
         <build bin/moby, etc>
```
(NB sense of `cmp` may be wrong, also behaviour if one input isn't present needs confirming, also I'm sure I need either more or less `;`s in that oneliner)

That arranges (or should arrange) that if the commit doesn't change we don't change the stamp file and hence don't change its timestamp which avoids the rebuild. Possibly the stamp should container other stuff, like the git URL.

Or perhaps we could always build to `bin/moby-$(MOBY_COMMIT)` and create symlink at `bin/moby` from the same rule, so it would always point to the latest.

tych0:
Closes #1880

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ endif
 PREFIX?=/usr/local/
 
 MOBY_COMMIT=1cb9fab3e13c8d3931c6f989c5d36087382e8710
-bin/moby: | bin
+bin/moby: Makefile | bin
 	docker run --rm --log-driver=none $(CROSS) $(GO_COMPILE) --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --commit $(MOBY_COMMIT) --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=$(GIT_COMMIT) -X main.Version=$(VERSION)" -o $@ > tmp_moby_bin.tar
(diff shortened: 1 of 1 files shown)


justincormack:
Can you add as a dependency of `rtf` as well as it is the same

tych0:
Yep, done.

PR #1914, opened 2017-05-26
Add ima kernel images

tych0:
`linuxkit/kernel-ima` and `linuxkit/ima-utils` exist now, so let's use images that have been uploaded there.

### DIFF
--- a/projects/ima-namespace/ima-utils/Makefile
+++ b/projects/ima-namespace/ima-utils/Makefile
@@ -7,6 +7,7 @@ DEPS=Dockerfile
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 
 tag: $(DEPS)
+	docker pull linuxkit/$(IMAGE):$(HASH) || \
 	docker build --no-cache -t linuxkit/$(IMAGE):$(HASH) .
 
 push: tag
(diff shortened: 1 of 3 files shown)


projects/ima-namespace/ima-utils/Makefile
```diff
@@ -7,6 +7,7 @@ DEPS=Dockerfile
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 
 tag: $(DEPS)
+	docker pull linuxkit/$(IMAGE):$(HASH) || \
```
rneugeba: Could you use the ORG override pattern we use in all the official packages now? Sorry, things have changed a bit frantically this week

tych0: I'm happy to do that, but I'm not sure what you mean :)

tych0: Ah, I see, I'll push an update, thanks.

rneugeba: all the Makefiles now have:
```
ORG?=linuxkit
```
and then
```
push: tag
        docker pull $(ORG)/$(IMAGE):$(HASH) || \
        docker push $(ORG)/$(IMAGE):$(HASH)
```
allows you to override to push to private registry or hub org

PR #1913, opened 2017-05-26
trust: use org key in yml

riyazdf:
Enforce content trust across the `linuxkit` hub org instead of enumerating individual images where we can, since many more packages are signed.  Also enforce our official images are signed.

Signed-off-by: Riyaz Faizullabhoy <riyaz.faizullabhoy@docker.com>

</img>

PR #1915, opened 2017-05-26
Wireguard fixups

tych0:
The wireguard project was using very old ways to build the init and kernel. Let's modernize those (mostly done in prep for a PoC for "probational" channel).


projects/wireguard/tools/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine:edge as tools
```
justincormack: can you use `linuxkit/alpine` - dont see why we should not add to base now, as it doesnt affect anything.

rneugeba: I agree.

projects/wireguard/wireguard.yml
```diff
@@ -0,0 +1,37 @@
...
+     - CAP_CHOWN
+     - CAP_SETUID
+     - CAP_SETGID
+     - CAP_DAC_OVERRIDE
+    net: host
+files:
+  - path: etc/docker/daemon.json
+    contents: '{"debug": true}'
+trust:
+  image:
```
rneugeba: minor nit, but you can just do what's in here: https://github.com/linuxkit/linuxkit/pull/1913 ie use the org (I didn't know that either until I saw the PR)

justincormack: not unless they are all signed (including wireguard-utils)

rneugeba: ah, missed the wireguard bit

riyazdf: I can get these signed within the next 10 minutes if you can wait @tych0 :)

riyazdf: And I'll do it for the ima images as well - will report back once this is done

justincormack: @riyazdf it might be easier to have a `linuxkit-projects` repo which is not signed instead.

tych0: Oh, I'm uploading them all as signed now so we can make this change. Should I not?

tych0: Ah, I can't, I guess :)

```Failed to sign "docker.io/linuxkit/kernel-wireguard":4.9.15-2ca28b7589b673373a33274023ca870a3a77e081 - no valid signing keys for delegation roles
no valid signing keys for delegation roles
```

justincormack: @riyazdf its going to be really annoying if all project users have to sign everything, so making linuxkit-projects and moving everything there seems better

riyazdf: @tych0 looks like I beat you to the punch with setting up the maintainer/CI keys :)  - the `org` key should work now for this

@justincormack: I agree that would be better, so that we don't have to manage external keys if project collaborators want to sign their own repos.  We can also do more elaborate checks, like pinning keys for `linuxkit` org to check that it's signed by maintainers

riyazdf: @justincormack looks like I didn't see your reply before posting mine but 100% agree 👍

PR #1925, opened 2017-05-28
Tweak package tests and add more

rneugeba:
- Simplify the sysctl test by using a standard alpine container to execute the shell script
- add tests for binfmt/dhcpcd/ca-certificates


test/cases/040_packages/002_binfmt/test-binfmt.yml
```diff
@@ -0,0 +1,27 @@
...
+    command: ["sh", "./check.sh"]
+  - name: poweroff
+    image: "linuxkit/poweroff:a8f1e4ad8d459f1fdaad9e4b007512cb3b504ae8"
+    command: ["/bin/sh", "/poweroff.sh", "10"]
+files:
+  - path: check.sh
+    source: ./check.sh
+trust:
+  org:
+    - alpine
```
justincormack: alpine is not the org, its `library/alpine` technically. @riyazdf is there some easy way of saying we want to trust all of `library/`?

test/cases/040_packages/004_dhcpcd/test.sh
```diff
@@ -0,0 +1,25 @@
...
+clean_up() {
+	find . -iname "test-dhcpcd*" -not -iname "*.yml" -exec rm -rf {} \;
+}
+trap clean_up EXIT
+
+# Test code goes here
+moby build -output kernel+initrd test-dhcpcd
+RESULT="$(linuxkit run qemu -kernel test-dhcpcd)"
+echo "${RESULT}"
+echo "${RESULT}" | grep -q "leased"
```
justincormack: why? The test script itself seems sufficient

test/cases/040_packages/002_binfmt/test-binfmt.yml
justincormack: you can do `image: alpine:3.6` not sure if the code supports `image` without a tag though.

rneugeba:
pushed updates:
- use `image: alpine:3.6` for trust
- removed additional check from dhcpcd

test/cases/040_packages/002_binfmt/test-binfmt.yml
riyazdf: from slack: it'll require a little bit of code shuffling but should be easy enough.  I'll put in a PR when I get some cycles this week 👍

PR #1926, opened 2017-05-29
hyperkit: Add --networking option

### LINKED ISSUE #1912: hyperkit: allow having no networking
I have use cases where I don't want to run vpnkit with the hyperkit driver at all, so I have no networking.

Can we have an option for not doing any vpnkit at all, and not configuring a network?

cc @MagnusS

MagnusS:
Adds a new bool option to disable networking with `linuxkit run hyperkit`.
Networking is enabled by default.

Fixes #1912 

web_small-animal1

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -8,10 +8,17 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/moby/hyperkit/go"
 	"github.com/satori/go.uuid"
+)
+
+const (
+	networkingNone         string = "none"
+	networkingDockerForMac        = "docker-for-mac"
+	networkingVPNKit              = "vpnkit"
 )
 
 // Process the run arguments and execute run
@@ -34,9 +41,7 @@ func runHyperKit(args []string) {
 	ipStr := flags.String("ip", "", "IP address for the VM")
 	state := flags.String("state", "", "Path to directory to keep VM state in")
 	vsockports := flags.String("vsock-ports", "", "List of vsock ports to forward from the guest on startup (comma separated). A unix domain socket for each port will be created in the state directory")
-	startVPNKit := flags.Bool("start-vpnkit", false, "Launch a new VPNKit instead of reusing the instance from Docker for Mac. The new instance will be on a separate internal network. This enables IP port forwarding from the host to the guest if the guest supports it.")
-	vpnKitDefaultSocket := filepath.Join(os.Getenv("HOME"), "Library/Containers/com.docker.docker/Data/s50")
-	vpnKitEthernetSocket := flags.String("vpnkit-socket", vpnKitDefaultSocket, "Path to VPNKit ethernet socket. The Docker for Mac socket is used by default. Overridden if -start-vpnkit is set.")
+	networking := flags.String("networking", networkingDockerForMac, "Networking mode. Valid options are 'docker-for-mac', 'vpnkit[,socket-path]' and 'none'. 'docker-for-mac' connects to the network used by Docker for Mac. 'vpnkit' connects to the VPNKit socket specified. If socket-path is omitted a new VPNKit instance will be started and 'vpnkit_eth.sock' will be created in the state directory. 'none' disables networking.`")
 
 	if err := flags.Parse(args); err != nil {
 		log.Fatal("Unable to parse args")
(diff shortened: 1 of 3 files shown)


src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -49,6 +50,13 @@ func runHyperKit(args []string) {
 	}
 	prefix := remArgs[0]
 
+	if *startVPNKit && !*networking {
+		log.Fatalln("Can't start VPNKit when networking is disabled")
+	}
+	if *vpnKitEthernetSocket == "" && *networking {
+		log.Fatalln("VPNKit socket path can't be empty when networking is enabled.")
+	}
+
```
rneugeba: could this be grouped closer to the rest of the networking code below?

justincormack:
Booleans that default to true are a bit odd in cli flags, `--disable-networking` is more idiomatic. But actually having `--networking=vpnkit` might be a better default and support `--networking=none` so could add other network options later.

rneugeba:
I would also prefer `--networking=`. Could we maybe also then simplify the other VPNKit options. Default to re-use Docker for Mac, but use an option to start your own VPNKit process and disable networking completely?

MagnusS:
@rneugeba Hm, do you mean having e.g "--networking=docker-for-mac", and that a different "vpnkit" mode would start vpnkit? We also need an option to specify a custom socket (or an existing state dir with the socket in it), so we can build networks with the new vpnkit instance

justincormack:
I think I would have
- `--networking=docker-for-mac` (default)
- `--networking=vpnkit` which has a socket option, maybe defaults to create, then either do `--networking=vpnkit,/path/to/sock` or `--vpnkit-socket=path`
- `--networking-none`

rneugeba:
yes, something along those lines. I think we want a primary networking option to select the mode (re-use DfM (default), start own VPNKit, none). The `vpnkit-socket` option is really secondary and only makes sense when we start our own VPNKit *and* want to have multiple HyperKits sharing it, which is probably a bit niche.

Ah, just saw @justincormack's update. I think we propose the same thing. I prefer the `--networking=vpnkit,/path/to/sock` variation as it is one less command line options

MagnusS:
I like that too - much cleaner. But I guess we actually have two vpnkit modes, one for starting your own instance that uses a hardcoded socket path (`[state]/vpnkit_eth.sock`), and one for joining an existing one which currently takes a path to a socket. We could simplify that, e.g.:

`--networking=vpnkit`
`--networking=vpnkit-join,state-dir-of-first-node`

...and remove other vpnkit options

rneugeba:
or could you check if the socket exists and then assume join?

justincormack:
We already have a `state-dir` option, so I think omitting the path can imply that you start your own instance in state-dir, while specifying it means join an existing one at the specified path.

rneugeba:
`state-dir` also contains per VM state so it's different and can't be shared between VMs

justincormack:
No, I meant if you want to set where the path is for the master you have state-dir, so no need to specify the socket location explicitly, while for the case of using existing socket you can specify the path.

rneugeba:
sorry, misread. the above makes sense

MagnusS:
@rneugeba @justincormack thanks for comments, I pushed a new version that supports the following options:
`--networking=docker-for-mac` -- default, connects to DfM
`--networking=vpnkit` -- launch new vpnkit
`--networking=vpnkit,/.../other-vm-state-dir` -- connect to existing vpnkit
`--networking=none`

Also removed the `--vpnkit-socket` option and moved the networking code to the same place

MagnusS:
Ah, small bug with setting up the ports from the `vsock-ports` option too late so they overwrote the vpnkit forwarding port - fixed in latest update

justincormack:
The option should point at the socket file not the state dir I think.

MagnusS:
@justincormack ah ok, then I misread too :-) I can change it to point to socket. You can't override it now so it will always have the same name relative to the state dir though (unless you launch vpnkit in some other way)

justincormack:
@MagnusS yes thats fine, but you might well start vpnkit yourself in which case there is no concept of state dir, so its less confusing. We just need to document the socket name when you use `linuxkit` to create the socket (and make sure its output in verbose mode).

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -147,20 +164,18 @@ func runHyperKit(args []string) {
...
 		log.Fatalf("Cannot run hyperkit: %v", err)
 	}
 }
 
+const (
+	networkingNone         string = "none"
+	networkingDockerForMac        = "docker-for-mac"
+	networkingVPNKit              = "vpnkit"
+)
+
```
rneugeba: I don't know what's the idiomatic go way is but I find it odd to only define these constants after prominent use further up. I would define them after the import section at the beginning of the file...

MagnusS: Ok - moved to the top

PR #1928, opened 2017-05-29
Remove remaining uses of tools/go-compile from packages

### LINKED ISSUE #1919: Convert remaining go packages from go-compile
Some packages still use the `go-compile` packages and they should use the alpine base:
- [ ] `binfmt`
- [ ] ` metadata`
- [ ] `virtsock`

rneugeba:
This removes the use of the go-compile package for package builds and converts the metadata package to use a multi-stage build

resolves #1919

image


pkg/metadata/Makefile
```diff
@@ -1,45 +1,15 @@
...
-
-SHA_IMAGE=alpine:3.5@sha256:dfbd4a3a8ebca874ebd2474f044a0b33600d4523d03b0df76e5c5986cb02d7e8
-
-METADATA_BINARY=usr/bin/metadata
+.PHONY: tag push
+default: push
 
 ORG?=linuxkit
 IMAGE=metadata
+DEPS=Dockerfile Makefile $(wildcard *.g)
```
justincormack: type `*.go`

rneugeba: fixed, thanks

PR #1932, opened 2017-05-30
Use the sinit conventions for init

justincormack:
This should make it easier to switch out `init` for other versions, although the `getty` config still needs to be removed (will do in future PR) so that we are not using any of busybox `init` config or features.
    
- use `/bin/rc.init` for start
- use `/bin/rc.shutdown` for stop
- make `rc.init` run other startup scripts
- merge `containers` and `containerd` startup code


justincormack:
cc @mato should make it easier to switch out busybox `init` (will switch out `getty` code soon)

mato:
@justincormack Thanks for the heads up.

PR #1934, opened 2017-05-30
Default ns change

justincormack:
Update everything for the new `moby` default of net=host, ipc=host, uts=host

cute_field_mouse_on_a_wheat_head__40438780


justincormack:
ok will merge now then..

Issue #1613, opened 2017-04-12
./tools/perf does not compile with newer kernels


rn:
It currently use 4.9.3 but when updating to 4.9.21I noticed a compile error

nwt:
This is the same issue described at https://lkml.org/lkml/2016/10/24/546. With that patch, perf from 4.9.21 compiles without error.

rneugeba:
@nwt thanks for the pointer. I will give that a try after I reworked the general kernel build

PR #1937, opened 2017-05-30
Convert remaining packages to multi-stage, git tree hash, and Alpine base

### LINKED ISSUE #1921: Convert remaining packages to multi-stage builds
The following packages still use "old" style `tar` to/from stdin/stdout builds and should be converted to multi-stage builds plus using the git tree hash:
- [x] `metadata`
- [ ] `test/docker-bench`
- [ ] `test/ltp`
- [ ] `test/poweroff`
- [ ]  `tools/go-compile`
- [ ] `tools/guestfs`
- [ ] `tools/mkimage-*` (note, these are likely going away)
- [ ] `tools/perf` (note, this may go away, also see #1613)
- [ ] `tools/qemu`

rneugeba:
These are the remaining packages/tools which were not using multi-stage builds and/or git tree hashes and/or Alpine base.

This resolves #1921, though I didn't convert the `guestfs`/`mkimage-*` tools as they should go away. I also didn't change the `perf` package as this needs to be fixed properly first (see #1613)

### DIFF
--- a/test/pkg/docker-bench/Dockerfile
+++ b/test/pkg/docker-bench/Dockerfile
@@ -1,22 +1,31 @@
-FROM alpine:3.5
-RUN apk update && apk upgrade && apk add --no-cache bash curl
-ADD . ./
+FROM linuxkit/alpine:451603daf499e3a40308dbf5571dcffed2343ffa AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    busybox \
+    bash \
+    curl
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
 
-# Also add docker
+FROM scratch
+WORKDIR /
+COPY --from=mirror /out/ /
+
+# Add docker
 ENV DOCKER_BUCKET get.docker.com
-ENV DOCKER_VERSION 17.04.0-ce
-ENV DOCKER_SHA256 c52cff62c4368a978b52e3d03819054d87bcd00d15514934ce2e0e09b99dd100
+ENV DOCKER_VERSION 17.05.0-ce
+ENV DOCKER_SHA256 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a95d94316d9
 
-# Downloads docker but only installs the client
+# Install just the client
 RUN set -x \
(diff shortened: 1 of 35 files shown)


test/pkg/docker-bench/Dockerfile
```diff
@@ -1,22 +1,31 @@
...
+
+# Add docker
 ENV DOCKER_BUCKET get.docker.com
-ENV DOCKER_VERSION 17.04.0-ce
-ENV DOCKER_SHA256 c52cff62c4368a978b52e3d03819054d87bcd00d15514934ce2e0e09b99dd100
+ENV DOCKER_VERSION 17.05.0-ce
+ENV DOCKER_SHA256 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a95d94316d9
 
-# Downloads docker but only installs the client
+# we could avoid installing client here I suppose
```
justincormack: this only needs the client I think, installing everything is a bit much...

test/pkg/ltp/Dockerfile
```diff
@@ -17,3 +16,10 @@ RUN cd /ltp \
     && ./configure \
     && make -j "$(getconf _NPROCESSORS_ONLN)" all \
     && make install
+
+FROM debian:jessie-slim
+COPY --from=build /opt/ltp/ /opt/ltp/
+ADD check.sh ./check.sh
+WORKDIR /opt/ltp
+ENTRYPOINT ["/bin/sh", "/check.sh"]
+LABEL org.mobyproject.config='{"net": "host", "pid": "host", "capabilities": ["all"]}'
```
justincormack: net: host no longer needed

test/pkg/ltp/Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM debian:jessie
-
-ARG LTP_VERSION
+FROM debian:jessie AS build
```
justincormack: the tree hash will not work if you just use `debian:jessie` with no sha256

tools/qemu/Dockerfile
```diff
@@ -1,15 +1,20 @@
-FROM alpine:edge
+FROM linuxkit/alpine:5240cbd9cf371c8211c8f1968e57c51a32098c8f AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    busybox \
+    libarchive-tools \
+    qemu-img \
+    qemu-system-arm \
```
justincormack: I don't know that we need `qemu-system-arm` in here now - we are not using it anywhere now.

rneugeba:
OK, should have addressed all the issue

PR #1933, opened 2017-05-30
Add linuxkit run azure

radu-matei:
**- What I did**
Add `linuxkit run azure` functionality

**- How I did it**
Integrate with the Azure Go SDK

PR #1939, opened 2017-05-30
Azure support

justincormack:
cc @radu-matei


docs/azure.md
```diff
@@ -0,0 +1,85 @@
...
+export AZURE_CLIENT_ID=<your_client_id>
+export AZURE_CLIENT_SECRET=<your_client_secret>
+```
+
+Now you should be ready to deploy resources using the LinuxKit command line.
+
+## Build an image
+
+> This is a preliminary example image with SSHD and Docker services. In the future, there will be an `azure.yml` file in the `examples` directory
+
```
rneugeba: there now is a `examples/azure.yml`

examples/azure.yml
```diff
@@ -0,0 +1,101 @@
...
+    net: host
+    mounts:
+     - type: cgroup
+       options: ["rw","nosuid","noexec","nodev","relatime"]
+    binds:
+     - /var/run/:/var/run/
+     - /var/lib/docker:/var/lib/docker
+     - /lib/modules:/lib/modules
+  - name: dev
+    image: "radumatei/linuxkit-dev:dev"
```
rneugeba: can we make the examples so that they only require LinuxKit images or standard images from the library?

examples/azure.yml
```diff
@@ -0,0 +1,101 @@
+kernel:
```
rneugeba: could this YAML file be update to ressemble the other ones. The hashes don't match the latest and neither do the configs for the containers.

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"time"
+
+	"path/filepath"
```
rneugeba: could we include the standard imports in line and not as a separate sections. None of the other code does this.

src/cmd/linuxkit/push_azure.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+	"path/filepath"
+)
+
+// Process the run arguments and execute run
+func pushAzure(args []string) {
+	flags := flag.NewFlagSet("azure", flag.ExitOnError)
+	invoked := filepath.Base(os.Args[0])
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s push azure [options] [name]\n\n", invoked)
```
rneugeba: `[name]` seems to indicate that the name is optional. Should also be `path` instead of name, like in the qemu runner. Just noticed the GCP backend has the same, need to fix that

src/cmd/linuxkit/push.go
```diff
@@ -33,6 +33,8 @@ func push(args []string) {
 		os.Exit(0)
 	case "gcp":
 		pushGcp(args[1:])
+	case "azure":
```
rneugeba: `azure` also needs to be added to the help message above

src/cmd/linuxkit/push_azure.go
```diff
@@ -0,0 +1,47 @@
...
+		fmt.Printf("'name' specifies the path (absolute or relative) of a\n")
+		fmt.Printf("VHD image be uploaded to an existing Azure Storage Account\n")
+		fmt.Printf("Options:\n\n")
+		flags.PrintDefaults()
+	}
+
+	resourceGroupName := flags.String("resourceGroupName", "", "Name of resource group to be used for VM")
+	accountName := flags.String("accountName", "linuxkitstorage", "Name of the storage account")
```
rneugeba: is this going to be created if it doesn't exist. Not sure if it's a good idea to hard code the default of `linuxkitstorage`

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+// AZURE_TENANT_ID: contains your Azure Active Directory tenant ID or domain
+// AZURE_SUBSCRIPTION_ID: contains your Azure Subscription ID
+// AZURE_CLIENT_ID: contains your Azure Active Directory Application Client ID
+// AZURE_CLIENT_SECRET: contains your Azure Active Directory Application Secret
+
+func runAzure(args []string) {
+	flags := flag.NewFlagSet("azure", flag.ExitOnError)
+	invoked := filepath.Base(os.Args[0])
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s run azure [options] [name]\n\n", invoked)
```
rneugeba: see above

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+		fmt.Printf("'imagePath' specifies a VHD file which will be\n")
```
rneugeba: `imagePath` is not in the help message.

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+	accountName := flags.String("accountName", "linuxkitstorage", "Name of the storage account")
+
+	subscriptionID := getEnvVarOrExit("AZURE_SUBSCRIPTION_ID")
+	tenantID := getEnvVarOrExit("AZURE_TENANT_ID")
+
+	clientID := getEnvVarOrExit("AZURE_CLIENT_ID")
+	clientSecret := getEnvVarOrExit("AZURE_CLIENT_SECRET")
+
+	if err := flags.Parse(args); err != nil {
+		log.Fatal("Unable to parse args")
```
rneugeba: should this print the error?

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	imagePath := remArgs[0]
+
+	rand.Seed(time.Now().UTC().UnixNano())
+	virtualNetworkName := fmt.Sprintf("linuxkitvirtualnetwork%d", rand.Intn(1000))
+	subnetName := fmt.Sprintf("linuxkitsubnet%d", rand.Intn(1000))
+	publicIPAddressName := fmt.Sprintf("publicip%d", rand.Intn(1000))
+	networkInterfaceName := fmt.Sprintf("networkinterface%d", rand.Intn(1000))
+	virtualMachineName := fmt.Sprintf("linuxkitvm%d", rand.Intn(1000))
+
```
rneugeba: I don't know Azure well enough, so I'm not sure if hardcoding `linuxkit` here is a good idea.

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	virtualMachinesClient = compute.NewVirtualMachinesClient(subscriptionID)
+	virtualMachinesClient.Authorizer = autorest.NewBearerAuthorizer(token)
+
+}
+
+func getOrCreateResourceGroup(resourceGroupName, location string) *resources.Group {
+	var resourceGroup resources.Group
+	resourceGroup, err := groupsClient.Get(resourceGroupName)
+	if err != nil {
+		log.Fatalf("\nError in getting resource group")
```
rneugeba: there shouldn't be a `\n` at the start of there error message.

Would it make sense to also print the error?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+func createResourceGroup(resourceGroupName, location string) *resources.Group {
+	fmt.Printf("\nCreating resource group in %s", location)
+
+	resourceGroupParameters := resources.Group{
+		Location: &location,
+	}
+	group, err := groupsClient.CreateOrUpdate(resourceGroupName, resourceGroupParameters)
+	if err != nil {
+		fmt.Println(err.Error())
+		log.Fatalf("Unable to create resource group")
```
rneugeba: can you add the printing of the error to the same `log.Fatalf()`. Having two lines of error message for one error is odd

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	return &group
+}
+
+func createStorageAccount(accountName, location string, resourceGroup resources.Group) {
+	fmt.Printf("\nCreating storage account in %s, resource group %s\n", location, *resourceGroup.Name)
```
rneugeba: could you remove the `\n` from the start

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+			break
+		}
+	}
+}
+
+func uploadVMImage(resourceGroupName string, accountName string, imagePath string) {
+	accountKeys, err := accountsClient.ListKeys(resourceGroupName, accountName)
+	if err != nil {
+		fmt.Println(err.Error())
+		log.Fatalf("Unable to retrieve storage account key")
```
rneugeba: see my comment above

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	keys := *(accountKeys.Keys)
+
+	absolutePath, err := filepath.Abs(imagePath)
+	if err != nil {
+		fmt.Println(err.Error())
+		log.Fatalf("Unable to get absolute path")
```
rneugeba: and here as well

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+
+	dockerMount := fmt.Sprintf("%s:/vhds", directory)
+	storageAccountNameArg := fmt.Sprintf("STORAGE_ACCOUNT_NAME=%s", accountName)
+	storageAccountKeyArg := fmt.Sprintf("STORAGE_ACCOUNT_KEY=%s", *keys[0].Value)
+	vhdPath := fmt.Sprintf("VHD_PATH=/vhds/%s", image)
+
+	output, err := exec.Command("docker", "run", "-v", dockerMount, "-e", vhdPath, "-e", storageAccountNameArg, "-e", storageAccountKeyArg, "radumatei/azure-vhd-upload:alpine").CombinedOutput()
+
+	if err != nil {
+		os.Stderr.WriteString(err.Error())
```
rneugeba: is this an error and should be `log.Fatalf()`?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	fmt.Println(string(output))
```
rneugeba: Is this for logging? If so maybe add this as a `log.Info()` or somesuch. we use logrus in the rest of the code

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/arm/compute"
+	"github.com/Azure/azure-sdk-for-go/arm/network"
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/azure-sdk-for-go/arm/storage"
+	simpleStorage "github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/to"
```
rneugeba: use logrus for logging, like the rest of the code

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	vhdPath := fmt.Sprintf("VHD_PATH=/vhds/%s", image)
+
+	output, err := exec.Command("docker", "run", "-v", dockerMount, "-e", vhdPath, "-e", storageAccountNameArg, "-e", storageAccountKeyArg, "radumatei/azure-vhd-upload:alpine").CombinedOutput()
+
+	if err != nil {
+		os.Stderr.WriteString(err.Error())
+	}
+	fmt.Println(string(output))
+
+	fmt.Printf("OS Image uploaded at https://%s.blob.core.windows.net/linuxkitcontainer/linuxkitimage.vhd\n", accountName)
```
rneugeba: see elsewhere re hardcoding linuxkit here

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+
+	return &virtualNetwork
+}
+
+func createSubnet(resourceGroup resources.Group, virtualNetworkName, subnetName string) *network.Subnet {
+	fmt.Printf("\nCreating subnet %s in resource group %s, within virtual network %s", subnetName, *resourceGroup.Name, virtualNetworkName)
+
+	subnetParameters := network.Subnet{
+		SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+			AddressPrefix: to.StringPtr("10.0.0.0/24"),
```
rneugeba: is 10.0.0.0 a Azure wide thing? or is this hardcoded for convenience?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	subnet, err := subnetsClient.Get(*resourceGroup.Name, virtualNetworkName, subnetName, "")
+	if err != nil {
+		log.Fatalf("Unable to retrieve subnet")
+	}
+
+	return &subnet
+}
+
+func createPublicIPAddress(resourceGroup resources.Group, ipName, location string) *network.PublicIPAddress {
+	fmt.Printf("\nCreating public IP Address in resource group %s, with name %s", *resourceGroup.Name, ipName)
```
rneugeba: please remove the `\n` from the start

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		if ipAddressChannel == nil && errorChannel == nil {
+			break
+		}
+	}
+	time.Sleep(time.Second * 5)
+	publicIPAddress, err := publicIPAddressesClient.Get(*resourceGroup.Name, ipName, "")
+	if err != nil {
+		fmt.Println(err.Error())
+		log.Fatalf("Unable to retrieve public IP address")
```
rneugeba: please merge into one log.Fatalf()

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+			}
+		}
+		if networkInterfaceChannel == nil && errorChannel == nil {
+			break
+		}
+	}
+
+	networkInterface, err := interfacesClient.Get(*resourceGroup.Name, networkInterfaceName, "")
+	if err != nil {
+		log.Fatalf("Unable to retrieve network interface")
```
rneugeba: would it be worth logging the error?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func createVirtualMachine(resourceGroup resources.Group, storageAccountName string, virtualMachineName string, networkInterface network.Interface, publicIPAddress network.PublicIPAddress, location string) {
+	fmt.Printf("\nCreating virtual machine in resource group %s, with name %s, in location %s", *resourceGroup.Name, virtualMachineName, location)
```
rneugeba: please remove the `\n` from the start.

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+			break
+		}
+	}
+
+}
+
+func getEnvVarOrExit(varName string) string {
+	value := os.Getenv(varName)
+	if value == "" {
+		log.Fatalf(fmt.Sprintf("Missing environment variable %s\n", varName))
```
rneugeba: the `fmt.Sprintf()` should not be necessary

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func createVirtualNetwork(resourceGroup resources.Group, virtualNetworkName string, location string) *network.VirtualNetwork {
+	fmt.Printf("Creating virtual network in resource group %s, in %s", *resourceGroup.Name, location)
+
+	virtualNetworkParameters := network.VirtualNetwork{
+		Location: &location,
+		VirtualNetworkPropertiesFormat: &network.VirtualNetworkPropertiesFormat{
+			AddressSpace: &network.AddressSpace{
+				AddressPrefixes: &[]string{"10.0.0.0/16"},
```
riyazdf: same comment here re: what 10.0.0.0 is and if we can const-ify it?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	if &resourceGroup != nil {
+		return &resourceGroup
+	}
+
+	return createResourceGroup(resourceGroupName, location)
+}
+
+func createResourceGroup(resourceGroupName, location string) *resources.Group {
+	fmt.Printf("\nCreating resource group in %s", location)
```
riyazdf: same as @rneugeba's comment re: removing \n

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+		if virtualNetworkChannel == nil && errorChannel == nil {
+			break
+		}
+	}
+
+	return &virtualNetwork
+}
+
+func createSubnet(resourceGroup resources.Group, virtualNetworkName, subnetName string) *network.Subnet {
+	fmt.Printf("\nCreating subnet %s in resource group %s, within virtual network %s", subnetName, *resourceGroup.Name, virtualNetworkName)
```
riyazdf: same as @rneugeba's comment re: removing \n

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+		case _, ok := <-errorChannel:
+			if !ok {
+				errorChannel = nil
+			}
+		}
+		if ipAddressChannel == nil && errorChannel == nil {
+			break
+		}
+	}
+	time.Sleep(time.Second * 5)
```
riyazdf: what is this needed for?

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+func setVirtualMachineParameters(storageAccountName string, networkInterfaceID, location string) compute.VirtualMachine {
+	return compute.VirtualMachine{
+		Location: &location,
+		VirtualMachineProperties: &compute.VirtualMachineProperties{
+			HardwareProfile: &compute.HardwareProfile{
+				VMSize: compute.StandardDS1,
+			},
+			OsProfile: &compute.OSProfile{
+				ComputerName:  to.StringPtr("linuxkit"),
+				AdminUsername: to.StringPtr("dummyusername"),
```
riyazdf: this username and password shouldn't be in code, nor should we ever hardcode a password for users

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+	group := createResourceGroup(*resourceGroupName, *location)
+	createStorageAccount(*accountName, *location, *group)
+	uploadVMImage(*group.Name, *accountName, imagePath)
+	createVirtualNetwork(*group, virtualNetworkName, *location)
+	subnet := createSubnet(*group, virtualNetworkName, subnetName)
+	publicIPAddress := createPublicIPAddress(*group, publicIPAddressName, *location)
+	networkInterface := createNetworkInterface(*group, networkInterfaceName, *publicIPAddress, *subnet, *location)
+	go createVirtualMachine(*group, *accountName, virtualMachineName, *networkInterface, *publicIPAddress, *location)
+
+	fmt.Printf("\nStarted deployment of virtual machine %s in resource group %s", virtualMachineName, *group.Name)
```
riyazdf: same comment re: leading `\n`

src/cmd/linuxkit/run_azure.go
```diff
@@ -0,0 +1,77 @@
...
+
+	time.Sleep(time.Second * 5)
+
+	fmt.Printf("\nNOTE: Since you created a minimal VM without the Azure Linux Agent, the portal will notify you that the deployment failed. After around 50 seconds try connecting to the VM")
+	fmt.Printf("\nssh -i path-to-key root@%s\n", *publicIPAddress.DNSSettings.Fqdn)
```
riyazdf: same comment re: leading `\n` for these two lines

src/cmd/linuxkit/azure.go
ddebroy: Instead of creating all the resources (network, public IP, VM, etc) along with parameters through Go code, how about creating a JSON ARM template to create the VM and associated resources? You can instantiate the ARM template with parameters specified in JSON using the Go APIs to create a resource group and get the same result. Please see example here: https://github.com/Azure/azure-quickstart-templates/tree/master/101-vm-from-user-image

I think the ARM template will be much easier to maintain than the Go code.

src/cmd/linuxkit/azure.go
ddebroy: Maybe make the OS Profile take in a SSH public key instead of username/password? And bubble up the SSH public key as a parameter?

ddebroy:
Instead of creating all the resources (network, public IP, VM, etc) along with parameters through Go code, how about creating a JSON ARM template to create the VM and associated resources? You can instantiate the ARM template with parameters specified in JSON using the Go APIs to create a resource group and get the same result. Please see example here: https://github.com/Azure/azure-quickstart-templates/tree/master/101-vm-from-user-image

I think the ARM template will be much easier to maintain than the Go code.

src/cmd/linuxkit/azure.go
radu-matei: In my opinion, at this moment it seems much easier to debug the creation of all the items through the REST API rather than through the ARM templates (considering that for the next step I also want to expose ports on the network security groups). Just seems more natural through the REST API for the moment.

After we have everything working maybe convert it to ARM?

src/cmd/linuxkit/azure.go
radu-matei: It got lost through the commit messages. Basically, just after I submit the creation of the public IP address I need to retrieve it through the `publicIPAddressesClient`. Most of the times, there isn't enough time so that it gets registered, so I added a little delay before retrieving it.

Maybe should try to await the creation.

src/cmd/linuxkit/azure.go
radu-matei: Will try to add `Disable Password Auth`

radu-matei: The thing is unless you manually add the SSHD service when building the image, you cannot SSH into the machine.

radu-matei:
@justincormack Could you add me to your fork, please?

justincormack:
Closed in favour of #1933 the original one.

src/cmd/linuxkit/azure.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	directory, image := filepath.Split(absolutePath)
+
+	dockerMount := fmt.Sprintf("%s:/vhds", directory)
+	storageAccountNameArg := fmt.Sprintf("STORAGE_ACCOUNT_NAME=%s", accountName)
+	storageAccountKeyArg := fmt.Sprintf("STORAGE_ACCOUNT_KEY=%s", *keys[0].Value)
+	vhdPath := fmt.Sprintf("VHD_PATH=/vhds/%s", image)
+
+	output, err := exec.Command("docker", "run", "-v", dockerMount, "-e", vhdPath, "-e", storageAccountNameArg, "-e", storageAccountKeyArg, "radumatei/azure-vhd-upload:alpine").CombinedOutput()
```
radu-matei: Here is the thing I would like you to review: there is already a very competent tool that uploads VHDs to Azure. Because the Go SDK changed in the meantime, it meant I should have had to add two versions of it. Didn't want to do it, so I think this is the easiest way for the moment. 

For the first version, until the SDK stabilizes, do you think we can have this way of uploading the image?

radu-matei: [This is the Dockerfile for the image](https://github.com/radu-matei/dockerfiles/blob/master/azure-vhd-utils/Dockerfile)

src/cmd/linuxkit/azure.go
radu-matei: Basically, this prints the output of the `docker run`

src/cmd/linuxkit/azure.go
justincormack: Seems fine to me, happy to commit reasonable working code...

src/cmd/linuxkit/azure.go
justincormack: Thats fine, if you don't configure login in the image there should be no way to login.

src/cmd/linuxkit/azure.go
justincormack: Not really happy with that. We are trying to remove the `docker` dependencies in LinuxKit (now mostly gone) as it is a pain to have to have Docker installed in CI for example.

src/cmd/linuxkit/azure.go
radu-matei: Just to be clear: those parameters are only specified so that the deployment gets validated. Since there is no Azure Linux Agent, those values will **not** be enforced and users can only connect if they included the SSHD service.

What do you think should be the way to go here?

radu-matei:
Updated now in #1939

PR #1940, opened 2017-05-30
logging project: fix example

MagnusS:
Example had been updated to the standard init that didn't include logging on startup, so no logs were recorded.

To use memlogd you currently need a custom init. It has to start the `memlogd` container before running onboot/service containers and then call `ctr`/`runc` through `logwrite` to capture stderr/stdout.

### DIFF
--- a/projects/logging/examples/logging.yml
+++ b/projects/logging/examples/logging.yml
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037 # with runc, logwrite, startmemlogd
+  - linuxkit/init:b5c88b78cd9cc73ed83b45f66bc9de618223768a # NOTE: custom init with logwrite ctr/runc, startmemlogd
   - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
   - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
   - linuxkit/ca-certificates:75cf419fb58770884c3464eb687ec8dfc704169d


projects/logging/examples/logging.yml
```diff
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.x"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037 # with runc, logwrite, startmemlogd
+  - linuxkit/init:b5c88b78cd9cc73ed83b45f66bc9de618223768a # NOTE: custom init with logwrite ctr/runc, startmemlogd
```
justincormack: this will be overwritten again probably. Projects should not be pushing dev code to the base hub repos, it should have a different name.

riyazdf: how about `linuxkit/init-logging`?  Or we can have this be the first project to use a separate org e.g. `linuxkit-projects/init-logging`?

riyazdf: we made `linuxkitprojects` earlier this week so I've made `https://hub.docker.com/r/linuxkitprojects/init-logging/` - @MagnusS  please let me know your hub ID and I'll add you when I get a chance

MagnusS: @riyazdf thanks! I sent you my hub ID on slack. Maybe we could create `linuxkitprojects/logging-memlogd` as well? It's currently in `linuxkit/memlogd`.

riyazdf: you should now have write access to both `linuxkitprojects/init-logging` and `linuxkitprojects/logging-memlogd` 👍

justincormack: ping @MagnusS

MagnusS: Thanks @riyazdf and @justincormack. I think I'll close this PR for now and make a new one that points to the new repos, as I also have to update memlogd to use the latest version of the init scripts

PR #1942, opened 2017-05-31
Add editions project

MagnusS:
This PR adds some of the open source components used to build Docker for Mac under `/projects/editions`. See README.md for component descriptions and examples.

The image built by `dfm.yml` supports running docker containers w/port forwarding (no volume sharing yet). With `vsudd.yml` you can control linuxkit containerd and browse logs.

I've included memlogd from the logging project + the memlogd docker plugin (https://github.com/cpuguy83/memlogd-docker) as this allows us to tail the logs from the host with `logread` (incl container logs). The plugin is added to the `install-dfm` init container and copied to /var/lib/docker when setting up a new disk - there may be better ways to do this...

To make it simpler to work on this I've currently pushed the packages to a test organisation `edtestorg`, but once we agree on names etc we could move them to `linuxkit`. 

To resurrect the packages from history I've checked out the last version before they were deleted and moved them to the new location in a separate commit. Commit logs can be seen with --follow, but blame doesn't work.

baby-bunny-easter


projects/editions/examples/dfm.yml
```diff
@@ -0,0 +1,101 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 console=tty0 page_poison=1"
```
justincormack: I dont think you want `console=tty0` on hyperkit, it will end up discarding a bunch of messages from the console

projects/editions/examples/dfm.yml
```diff
@@ -0,0 +1,101 @@
...
+  - linuxkit/init:062e57b1d1e017e44c6339fc2b4cd41f3f10b2a9 # with memlogd
+  - linuxkit/runc:2649198589ef0020d99f613adaeda45ce0093a38
+  - linuxkit/containerd:cf2614f5a96c569a0bd4bd54e054a65ba17d167f
+  - linuxkit/ca-certificates:3344cdca1bc59fdfa17bd7f0fcbf491b9dbaa288
+  - linuxkit/memlogd:9b5834189f598f43c507f6938077113906f51012
+  - edtestorg/install-dfm:256c1e9a344ca9f68b4131759a4c8f868a3e790b # adds /dfm
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:2cf2f9d5b4d314ba1bfc22b2fe931924af666d8c"
+    net: host
```
justincormack: none of these config options are needed any more they are baked into the image label.

projects/editions/examples/dfm.yml
```diff
@@ -0,0 +1,101 @@
...
+       options: ["rw","nosuid","noexec","nodev","relatime"]
+    binds:
+     - /var/lib/docker:/var/lib/docker
+     - /lib/modules:/lib/modules
+     - /tmp/memlogd.sock:/tmp/memlogd.sock
+     - /dfm/port:/port
+     - /var/run:/var/run
+files:
+  - path: etc/docker/daemon.json
+    contents: '{"debug": true}'
```
justincormack: this isnt being used, should be removed

projects/editions/examples/dfm.yml
```diff
@@ -0,0 +1,101 @@
...
+trust:
+  image:
+    - linuxkit/kernel
```
justincormack: should be `org: linuxkit` now

projects/editions/pkg/configure-dfm/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:edge@sha256:99588bc8883c955c157d18fc3eaa4a3c1400c223e6c7cabca5f600a3e9f8d5cd
+ENTRYPOINT []
```
justincormack: we should use the standard way we do this elsewhere, from the linuxkit/alpine base

projects/editions/pkg/docker-dfm/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM linuxkit/slirp-proxy:f1eec56e1a7196f6a786223baa30e2f5e026a27d as slirp-proxy
+
+FROM docker:17.05.0-ce-dind
+COPY --from=slirp-proxy /slirp-proxy /usr/bin/slirp-proxy
+RUN addgroup -S docker
+CMD [ "/usr/local/bin/dockerd", "-H", "unix:///var/run/docker.sock", "--swarm-default-advertise-addr=eth0", "--userland-proxy-path", "/usr/bin/slirp-proxy", "--storage-driver", "overlay2", "--log-driver", "cpuguy83/memlogd-docker"]
```
justincormack: -h unix:///var/run/docker.sock is the default anyway

projects/editions/pkg/docker-dfm/Makefile
```diff
@@ -0,0 +1,16 @@
+GO_COMPILE=linuxkit/go-compile:4513068d9a7e919e4ec42e2d7ee879ff5b95b7f5@sha256:bdfadbe3e4ec699ca45b67453662321ec270f2d1a1dbdbf09625776d3ebd68c5
```
justincormack: we are not using `go-compile` any more, please update as per recent commits

projects/editions/pkg/install-dfm/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
justincormack: not needed

projects/editions/pkg/install-dfm/Makefile
```diff
@@ -0,0 +1,33 @@
+default=tag
+ORG?=edtestorg
+IMAGE=install-dfm
+DEPS=Dockerfile build/dfm/port build/dfm/docker
+
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+
+# Create empty dir for /port mount
+build/dfm/port: Makefile
+	mkdir -p build/dfm/port
```
justincormack: don't create host directories, ownership will be all wrong, create them in the Dockerfile

projects/editions/pkg/install-dfm/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	mkdir -p build/dfm/docker/plugins
+	docker run --rm --privileged \
+		-v $(PWD)/build/dfm/docker:/var/lib/docker \
+		--name $(IMAGE)-dind \
+		-d docker:17.05.0-ce-dind \
+		--storage-driver=overlay2
+	docker run --rm --link $(IMAGE)-dind:docker docker:edge plugin ls | grep -q memlogd-docker || ( \
+		docker run --rm --link $(IMAGE)-dind:docker docker:edge plugin install cpuguy83/memlogd-docker --grant-all-permissions || ( \
+			docker stop $(IMAGE)-dind ; exit 1 ) )
+	docker stop $(IMAGE)-dind
```
justincormack: we really can't have this in the build script, there must be a better way to do this.

projects/editions/pkg/proxy/Makefile
```diff
@@ -0,0 +1,30 @@
+GO_COMPILE=linuxkit/go-compile:4513068d9a7e919e4ec42e2d7ee879ff5b95b7f5@sha256:bdfadbe3e4ec699ca45b67453662321ec270f2d1a1dbdbf09625776d3ebd68c5
```
justincormack: Don't use this, compile it like the other Go code is compiled in the Dockerfile

projects/editions/pkg/vsudd/Makefile
```diff
@@ -0,0 +1,24 @@
+.PHONY: clean push
+GO_COMPILE=linuxkit/go-compile:4513068d9a7e919e4ec42e2d7ee879ff5b95b7f5@sha256:bdfadbe3e4ec699ca45b67453662321ec270f2d1a1dbdbf09625776d3ebd68c5
+IMAGE=vsudd
```
justincormack: dont use go compile

projects/editions/pkg/install-dfm/Makefile
MagnusS: Yeah, that was a hack to get this working. Maybe we could somehow pull the plugin automatically when building the service container for dockerd. We could also pull it if it's missing on first boot, but would be nice if it was included in the image.

projects/editions/pkg/proxy/Makefile
MagnusS: Oops, this was written some time ago - will update

projects/editions/pkg/install-dfm/Makefile
justincormack: We have an issue open for how to handle plugins nicely. The vast majority of people want plugins just installed once not dynamically, so it would be nice if we could support that properly.

MagnusS:
@justincormack Ok - I'll close this and make smaller PRs for the components.

PR #1944, opened 2017-05-31
Support multiple disks in qemu

justincormack:
I needed multiple disks for `mkimage` to make it easier to pass data in and out as using the network under `qemu` was a pain. Still need to change the other backends to use the same CLI syntax, just seeing if any objections...

This changes the CLI specification for disks, as it needs to be able to
be repeated.

```
linuxkit run qemu -disk name,size=1G,format=qcow2 ...
```

Options may be omitted.

### DIFF
--- a/docs/external-disk.md
+++ b/docs/external-disk.md
@@ -7,14 +7,15 @@
 ## Make Disk Available
 In order to make the disk available, you need to tell `linuxkit` where the disk file or block device is.
 
-All local `linuxkit run` methods (currently `hyperkit`, `qemu`, and `vmware`) take two arguments:
+All local `linuxkit run` methods (currently `hyperkit`, `qemu`, and `vmware`) take a `-disk` argument:
 
-* `-disk-size <size>`: size of disk. The default is in MB but a `G` can be appended to specify the size in GB, e.g. `-disk-size 4096` or `disk-size 4G` will provide a 4GB disk.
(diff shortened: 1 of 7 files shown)


rneugeba:
I like the new CLI syntax. The hyperkit and vmware backends should be fine to only support one disk for now (with the new syntax) and `./docs/external-disk.md` probably also needs updating

justincormack:
Yes it doesnt look like the Hyperkit Go APi supports multiple disks yet so I will error out if more than 1

thebsdbox:
The `VMware` backend supports multiple (well 2) disks currently, but would require some work on the SCSI controller `vmx` config to have dynamic amounts of disks all possible. Looks good to me.

justincormack:
Ok, updated to make the cli consistent for all the local drivers, ie qemu, hyperkit, vmware. I left the cloud ones unchanged as it is not really applicable, although gcp could probably benefit from multiple disk-size options at some point.

@thebsdbox I did a little passing tidying of `run_vmware` and I created the multiple disks but didnt change the templates to actually support adding more than one, but it looks like it would be fairly trivial to add the scsi device numbers for a second disk say... feel free to add a PR to do so...

justincormack:
Going to merge this now I fixed the test issues - @thebsdbox PTAL at vmware and see if you want to make additional changes...

PR #1945, opened 2017-05-31
projects: implement a probational channel

tych0:
From the main patch notes:

The name for this can change, it's just the one that was first suggested.

The goal of the probational channel is to collect some subset of projects
together in the same linuxkit image. For now, I've implemented this for the
ima-namespacing and wireguard projects, but once we flesh out the idea, I'm
happy to do the others.

See the readme for details on how to build the probational channel. We
could automate this more, but this seemed sufficient for now.

cc @riyazdf


projects/probational/Makefile
```diff
@@ -0,0 +1,14 @@
+GO_COMPILE=linuxkit/go-compile:3ba94f14de51b73551417e769d122815ec917ee7
```
riyazdf: we're moving away from `go-compile` to using a base `alpine` image, example here: https://github.com/linuxkit/linuxkit/blob/master/pkg/sysctl/Dockerfile#L7

projects/kernel-config/Makefile
```diff
@@ -11,7 +11,7 @@
 # Git tree hash of this directory. Override to force build
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 # Name on Hub
-IMAGE:=kernel
+IMAGE?=kernel
```
riyazdf: maybe by default we should use `kernel-config`?

projects/kernel-config/makeconfig.sh
```diff
@@ -67,10 +59,6 @@ function check_config()
...
 
-check_config "/config/kernel_config.base"
-check_config "/config/kernel_config.${ARCH}"
-check_config "/config/kernel_config.${KERNEL_SERIES}"
-check_config "/config/kernel_config.${ARCH}.${KERNEL_SERIES}"
-if [ -n "${DEBUG}" ]; then
-    check_config "/config/kernel_config.debug"
-fi
+for config in "$@"; do
+  check_config "$config"
```
riyazdf: this is neat 👍

projects/probational/cmd/probational.go
```diff
@@ -0,0 +1,249 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Moby is the type of a Moby config file
+// stolen from github.com/moby/tool/cmd/moby, perhaps we should make handling
+// this into a lib?
```
riyazdf: we could probably vendor the type from the package itself, and then you could manually ignore the fields you aren't interested in (ex: `Trust`)

projects/probational/cmd/probational.go
```diff
@@ -0,0 +1,249 @@
...
+	if err := yaml.Unmarshal(content, &projects); err != nil {
+		return err
+	}
+
+	content, err = ioutil.ReadFile(linuxkitPath)
+	if err != nil {
+		return err
+	}
+
+	os.Remove("kernel-config/kernel_config.probational")
```
riyazdf: why do we need to remove this?

projects/probational/cmd/probational.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return err
+	}
+
+	projectCfg := Moby{}
+	if err := yaml.Unmarshal(content, &projectCfg); err != nil {
+		return err
+	}
+
+	// merge config
```
riyazdf: perhaps (in a future PR?) we can split out this config merging and kernel patches logic into separate, exported functions that can be vendored by other projects?  Might be useful for moby tool and other related code

projects/probational/cmd/probational.go
tych0: Because it's generated, and we always append to it. I suppose we could remove it as part of the initial build process too.

projects/probational/cmd/probational.go
tych0: Yes, in fact it probably could (?) all live in the moby tool repo. I'll make some PRs there to vendor it and such.

projects/kernel-config/Makefile
tych0: Can you create that project? I'll change it here then.

projects/probational/Makefile
tych0: Cool, will fix it up.

projects/kernel-config/Makefile
rneugeba: maybe `linuxkit/kernel-probational`?

projects/kernel-config/kernel_config.debug
```diff
@@ -24,3 +24,4 @@ CONFIG_KGDB_SERIAL_CONSOLE=y
 CONFIG_KGDBOC=y
 CONFIG_DEBUG_RODATA_TEST=y
 CONFIG_DEBUG_WX=y
+# CONFIG_PANIC_ON_OOPS is not set
```
rneugeba: any reason this is commented out for debug kernels?

tych0:
Yes, originally I had it building in the root of projects/ (with its makefile there), but I moved it underneath probational/ to make the diff less controversial. I'd be happy to move it back.

projects/kernel-config/kernel_config.debug
tych0: I assume that it's because you may want to inspect some state after the panic, but I don't really know. This change is just a minor simplification of what we were already doing: https://github.com/linuxkit/linuxkit/pull/1945/commits/fb0113481a5aa4d8fd7d1ef4f731096ad9fae63f

tych0: Er, after the oops :)

projects/kernel-config/Makefile
riyazdf: @tych0 just created and add you to `linuxkit/kernel-probational` https://hub.docker.com/r/linuxkit/kernel-probational

`linuxkit/kernel-config` already exists and hasn't been pushed for a while, added you to that as well:  https://hub.docker.com/r/linuxkit/kernel-config

riyazdf:
I agree with moving probational out of `/projects` to its own top-level `/probational`, since it's more official

rneugeba:
I'm not sure it `./probational` is the right place. For probational kernel configs maybe something in `./kernel` is better but don't have a good idea how to best place it

tych0:
I'll try moving it up to /projects again, and we'll see what that looks like.

rneugeba:
maybe wait till @justincormack chimes in

justincormack:
The discussion on the call last week suggested that projects (wireguard was mentioned, but applies to others) that met criteria such as (this is my interpretation of the discussion, which was not precise):

1. do not change behaviour if you do not use the feature (eg a new module)
2. have been reviewed to check that they do not open up exploits - eg the userspace API exposed has been fuzz tested (eg syzkaller or similar) and reviewed, especially carefully if it is available to less privileged users and so increases the overall kernel attack surface.
3. were plausible to go upstream in similar-ish, even if there might be interface changes or other changes.

If they met such criteria then the suggestion was we could just add to the standard config.I would prefer this to having another layer of probation. I think both the projects here probably meet 1 and 3, I don't know the status of 2 for either of them.

riyazdf:
@justincormack: I see value in a stage between projects and the standard config in order to validate your enumerated criteria, in a curated way as to avoid kernel and configuration rot.

@tych0 and I are working on a document to supplement this PR that will detail goals/non-goals/proposed process - it will be ready to present along this PR for the security SIG.

justincormack:
I don't see why aggregating projects helps check if they meet the criteria,
it would seem to make it harder.

tych0:
Based on feeback today from the SIG, I'll close this for now. My understanding is that:

- we should use the new merging stuff in the moby tool to do the userspace init merging
- there should be no tooling in projects/ so e.g. the stuff that merges kernel command lines and such needs to be integrated into the moby tool proper
- we should try to reduce bitrot by doing makefile includes instead of copy/paste everywhere

I'll see about respinning this with the above in mind.

PR #1947, opened 2017-06-01
Add /pkg/vsudd

MagnusS:
Adds `vsudd` - a daemon for forwarding unix domain sockets via VSOCK from the guest. It is used by Desktop Editions to control `dockerd`.

I've added a minimal example and a README.md for controlling `containerd` from the host with `ctr`.

Since `go-compile.sh` has `CGO_ENABLED=0` and `cgo` is required to compile `linuxkit/virtsock` I had to call `go` directly to build the code. Happy to change this if there's a better way. 

To resurrect vsudd from history I checked out the files from before they were deleted and moved them to /pkg in a separate commit. Commit history works with `log --follow`, but not `git blame`.

I've pushed to a temporary hub user so the example works.

rabbit


rneugeba:
I can try to make virtsock cgo free next week. The virtio socket path already is. The hv sock path not yet

justincormack:
I wonder if `vsudd` should go into `github.com/linuxkit/virtsock/cmd/vsudd` instead and we just clone that to build (and it could have the necessary `Makefiles` included)? That would remove a vendor, and also make sure it was kept in sync with any API changes, and it could have tests and so on there.

rneugeba:
Sounds like a good idea

MagnusS:
Yes, that seems like a good place for it - I'll prepare a PR for virtsock

justincormack:
Closing as now in virtsock so this PR not so useful...

PR #1949, opened 2017-06-01
Update mkimage to use a tarball as source for files

justincormack:
This uses the qemu multi disk support for the `mkimage` package - instead of baking in the images into the VM, it passes them as a tarball, disguised as a raw disk. This means you can use the same `mkimage` VM again and again without rebuilding it each time, plus it is faster as the image is smaller. This was nicer than using the network in the end.


justincormack:
merging as tests pass and doesnt affect anything other than work on `moby/tool`

PR #1950, opened 2017-06-01
projects/swarmd updates

ijc25:
**- What I did**

Various updates to projects/swarmd:
* Allow all capabilities (pending rational review of the requirements, the current list was pretty random)
* Modernize the build of the swarmd package to current standards
* Build from [my containerd-wip](https://github.com/ijc25/swarmkit/tree/containerd-wip) rather than [my containerd](https://github.com/ijc25/swarmkit/tree/containerd) branch. The latter contains only the containerd executor, while the former adds other in progress features such as CNI networking support which makes things more useful.
* Drop now unneeded `net: host`
* Enable `ip_forward`

**- How I did it**

Blood, sweat & toil.

**- How to verify it**

Run it. 

N/A

### DIFF
--- a/projects/swarmd/swarmd.yml
+++ b/projects/swarmd/swarmd.yml
@@ -9,6 +9,8 @@ init:
 onboot:
   - name: sysctl
     image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
+    binds:
+     - /etc/sysctl.d/01-swarmd.conf:/etc/sysctl.d/01-swarmd.conf
   - name: dhcpcd
     image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
@@ -39,3 +41,6 @@ services:
       - /var/lib/containerd:/var/lib/containerd
       - /var/lib/swarmd:/var/lib/swarmd
       - /etc/resolv.conf:/etc/resolv.conf
+files:
+  - path: /etc/sysctl.d/01-swarmd.conf
+    contents: 'net.ipv4.ip_forward = 1'
--- a/projects/swarmd/swarmd.yml
+++ b/projects/swarmd/swarmd.yml
@@ -34,7 +34,6 @@ services:
     capabilities:
      - all
     pid: host
-    net: host
     binds:
       - /run/containerd/containerd.sock:/run/containerd/containerd.sock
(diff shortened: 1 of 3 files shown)


projects/swarmd/swarmd.yml
```diff
@@ -29,27 +31,16 @@ services:
...
-     - CAP_SETGID
-     - CAP_SETUID
-     - CAP_SETFCAP
-     - CAP_SETPCAP
-     - CAP_NET_BIND_SERVICE
-     - CAP_SYS_CHROOT
-     - CAP_KILL
-     - CAP_AUDIT_WRITE
+     - all
     pid: host
```
justincormack: why does it want pid=host?

projects/swarmd/swarmd/Dockerfile
```diff
@@ -1,36 +1,48 @@
...
+    util-linux
 
-RUN cp -r /etc/ssl /build/dist/etc/ssl
+RUN mkdir -p /out/usr/bin/ /out/etc
+RUN cp bin/swarmd bin/swarmctl /out/usr/bin/
+RUN strip /out/usr/bin/swarmd /out/usr/bin/swarmctl
 
-FROM scratch
+FROM alpine:3.5
+RUN apk add -U iptables util-linux ca-certificates
```
justincormack: you already added these to `/out` above

justincormack: so you can remove this and still do `FROM scratch`

projects/swarmd/swarmd.yml
ijc25: With CNI networking you need to be able to construct `/proc/PID/ns/net` to pass to the CNI plugin which is forked by swarmd (and hence in the same namespace as swarmd), but `PID` is from containerd's pid namespace since it is what is launching the containers so the swarmd and containerd pid namespaces must be the same. Since I'm still using the host containerd that means using pid=host.

Eventually I want to run my own containerd instance for swarmd, which will likely change this.

projects/swarmd/swarmd/Dockerfile
ijc25: Oh yes, missed that in my various rebasing and reworkings to get here.

projects/swarmd/swarmd.yml
justincormack: The alternative is to bind the namespace to a file and pass that instead, then you no longer need the same `/proc`. I think you just `mount --bind /proc/PID/ns/net /somewhere/to/keep/it` and use that as equivalent. You can also pass filenames to the `net: ...` in the linuxkit config.

Issue #748, opened 2016-11-17
enable Prometheus metrics


justincormack:
See https://github.com/docker/docker/pull/27862/files for docs

Would be nice to enable these, need to choose a port, and work out how we are going to collect them, and secure the port if necessary.

riyazdf:
From discussion on https://github.com/linuxkit/linuxkit/pull/1951:
- disabling metrics by default for now, but can override in yml
- need to get more information about the metrics: does it a UNIX socket that we can use?
- how desired/used is this?

PR #1951, opened 2017-06-01
Test that no ports are open

dave-tucker:
**- What I did**
Disabled the containerd metrics port and added a test to check we have no ports open.

**- How to verify it**
Run the tests!

image


riyazdf:
taking a look now!  I can help with building/pushing/signing, tests will fail until it's on hub

dave-tucker:
Note: I'm having trouble pushing the `containerd` image
```
Successfully tagged linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f
DOCKER_CONTENT_TRUST=1 docker pull linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f || \
	DOCKER_CONTENT_TRUST=1 docker push linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f
No trust data for e3832e460e0565d94141a59bc750e6d6523b271f
The push refers to a repository [docker.io/linuxkit/containerd]
7d1567db06ad: Preparing
940643e59c38: Preparing
denied: requested access to the resource is denied
```

justincormack:
Maybe you could add a `files` section in the top level `linuxkit.yml` that adds back the config with the metrics port in, as it is nice for demos to show the prom api endpoint for system containers?

justincormack:
cc @riyazdf re denied push. You have to sign stuff nowadays or its not going to work...

riyazdf:
@dave-tucker: I've only added maintainer and CI keys to the signing setup at present.  I'll ping you offline so we can sort this out

riyazdf:
@dave-tucker:  I've signed `linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f` and restarted the tests

rneugeba:
I think the prometheus metrics for containerd should be enabled by default. If we don't have a mechanism to disable it on a per YAML basis then the test should whitelist that port

test/cases/030_security/010_ports/test.yml
```diff
@@ -0,0 +1,23 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:aa8549c2ea757b3daa727bc5a4f484c41952fc12
```
riyazdf: this test is failing because this image doesn't exist and isn't signed, should be `linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f`

riyazdf:
seems like there's still some discussion re: leaving port open and best way to express in the yml.  cc @justincormack @rneugeba are you happy with this?

justincormack:
Yeah, there doesnt seem to be agreement on this.

rneugeba:
It's a question of wether the containerd prom port should be open by default or not.

Just disabling it for the purpose to make a test simpler does not seem a sufficient enough reason to me to disable it by default. I haven't seen another reason to disable it...

riyazdf:
It is an open port on the underlying host, so I'd like to make sure we have good justification for leaving it open.  I'm not too familiar with the metrics usage, so naively I think it would be wise to default to off from the image and leaving the example file in `linuxkit.yml` for showing how to re-enable.  We can potentially add back the default metrics port to the containerd image once we have better documentation/understanding and if this is a more popular usecase.

I could only really find the [development reports](https://github.com/containerd/containerd/blob/e119f1dc0301796189bd8cdafd8fcaf37caf0d3b/reports/2017-03-24.md#container-level-metrics) and original PR for info on the metrics in containerd - if someone has access to more documentation that would be much appreciated.

dave-tucker:
FWIW, I agree with @riyazdf here.
I would default to closed unless there is sufficient justification for leaving it open.
We're using a arbitrary port number as AFAIK there isn't a registered port for prometheus, it rang alarm bells when `netstat` reported it open and had no record of which process it belonged to.  I had to `grep` the source code to find where it was coming from.

rneugeba:
OK, I agree with that reasoning. Needs a rebase, though...

ijc25:
Is a prometheus endpoint on a Unix socket any use at all, if it is that might make a better default than a TCP port? It could e.g. be bound into some sort of aggregator container for further forwarding to the net or something.

dave-tucker:
Rebased but someone with trust setup will need to push the image as I forgot to send @riyazdf my key 🤦‍♂️

justincormack:
@ijc25 I was wondering about that too. I don't think `containerd` currently has any way of configuring that but it could be useful.

test/cases/030_security/010_ports/test.yml
```diff
@@ -0,0 +1,23 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:e3832e460e0565d94141a59bc750e6d6523b271f
```
riyazdf: should this bump to `linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a`?

riyazdf:
@dave-tucker ack on receiving your key, will set it up soon.  Pushed and signed `linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a` in the interim.

riyazdf:
seems like we've gotten to an agreement to turn off for now, and that we need more research to answer questions about containerd metrics.  I've noted that here: https://github.com/linuxkit/linuxkit/issues/748

Will merge this PR.

PR #1952, opened 2017-06-01
projects/miragesdk: add a frontend yaml format for dhcp-client

avsm:
The intention with the yaml fragment is that it specifies the
set of processes that form the daemon, with minimal privileges
for each component and each running inside a separate container.

In addition to the normal container capabilities, there is also
a new field which lets a startup process establish an RPC channel,
based on a Capnp specification. This allows for extremely
unprivileged components to be started, such as the `dhcp-engine`
in this example which can only communicate with the outside world
via the `dhcp-network` (to transmit) or `dhcp-actuator` (to alter
the state of the local Linux distribution).

This is a first cut at the yaml interface and the capnp, with the
intention to refine it as we combine it with the rest of the existing
prototype (which currently doesnt have an RPC layer).  Expect
more changes...

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/projects/miragesdk/yaml/dhcp-client.yml
+++ b/projects/miragesdk/yaml/dhcp-client.yml
@@ -1,0 +1,47 @@
+# There are three processes running in the DHCP daemon:
+# - dhcp-network: handles L2 network traffic
+# - dhcp-engine: protocol state machine
+# - dhcp-actuator: sets interface state
+# And a dhcp-client which starts the three above and exits.
+#
+# dhcp-actuator: can be written in Rust and only contain
+#   minimal support for setting syscalls.
+# dhcp-engine: requires no system access beyond the input
+#   from `dhcp-network` and output to `dhcp-actuator`.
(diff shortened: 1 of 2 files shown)


avsm:
cc @samoht @talex5 for the capnp pieces and the yaml fragment

samoht:
Thanks, that look great. I think it would be even better if we didn't have a `dhcp-client` parent process. Ideally, the moby tool would have a way to tell `containerd` to open secure channels between the containers it manages -- I need to investigate more to see if this is possible.

Note: today, the MirageSDK privileged process calls `runc run -preserve-fd ...` to start the calf process with some control channels already opened. This is why we need to do all that nested runc dance (by giving read-write access to cgroup and other caps). It would be ideal if someone else (an init process, containerd, or someone else) would do it for us.

justincormack:
@avsm you need to appease the 🐢

riyazdf:
I realize this is WIP but since this is in `/projects` I will go ahead and merge

PR #1954, opened 2017-06-01
Added AWS provider to metadata package

kencochrane:
**- What I did**
Added an AWS provider to the metadata package as referenced here: #1918

**- How I did it**
Added a new provider to metadata package and wired it up.

**- How to verify it**
I built the image and ran it on an EC2 instance and verified the correct data was added to /var/config

**- Description for the changelog**
Added an AWS provider to the metadata package

Signed-off-by: Ken Cochrane <kencochrane@gmail.com>

Issue #1918, opened 2017-05-27
AWS support


justincormack:
We need support for AWS. Includes:

- `push` to AMI
- `run` an AMI

Plus CI.

allingeek:
And Metadata support.

kencochrane:
@justincormack how can I help with this?

justincormack:
@kencochrane I am just cleaning up the code that can generate raw disk images (currently we have `img-gz` which is a compressed raw image). These should work fine on AWS. @dave-tucker was looking at the VM import code to see if we could use that, but it seems buggy, so we probably need to do the disk snapshotting method we were using before, which should be fine with the current images. We basically want a `linuxkit push` and `linuxkit run` for AWS so we have easy ability to test in CI, see eg the `gcp` code in https://github.com/linuxkit/linuxkit/tree/master/src/cmd/linuxkit for example.

dave-tucker:
I don't think the VM import method will work without some support from AWS.

See [here](http://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-image-import.html) for their docs and prerequisites. I have no idea why AWS can only support import of some OSs.

Following the example in their docs with a raw LinuxKit image gives you `ClientError: Unknown OS / Missing OS files.`

kencochrane:
@dave-tucker that stinks. Looking at this page http://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html it looks like they have only a small list of Linux flavors that they support. We would either need to have them support a LinuxKit version, or we need a different way to do it.

#### Linux/Unix (64-bit)

- Ubuntu 12.04, 12.10, 13.04, 13.10, 14.04, 14.10, 15.04, 16.04, 16.10
- Red Hat Enterprise Linux (RHEL) 5.1-5.11, 6.1-6.6, 7.0-7.1 (6.0 lacks required drivers)
- SUSE Linux Enterprise Server 11 with Service Pack 1 and kernel 2.6.32.12-0.7
- SUSE Linux Enterprise Server 11 with Service Pack 2 and kernel 3.0.13-0.27
- SUSE Linux Enterprise Server 11 with Service Pack 3 and kernel 3.0.76-0.11, 3.0.101-0.8, or 3.0.101-0.15
- SUSE Linux Enterprise Server 11 with Service Pack 4 and kernel 3.0.101-63
- SUSE Linux Enterprise Server 12 with kernel 3.12.28-4
[...]

PR #1958, opened 2017-06-02
Adding an AWS example

DieterReuter:
**- What I did**
Adding an AWS example (as clone of the GCP example), refs to: #1918

**- How I did it**
Fixed the GCP example, had to add `binds:` keyword to `sshd` service
Updated the `linuxkit/metadata` to latest version which supports AWS (introduced with #1954)
Simplified `sshd` service, no need to specify the bind mount of /etc/resolv.conf as this is already the default

**- How to verify it**
Just built the raw image and imported it into AWS as snapshot, created an AMI from the snapshot, ran a new EC2 instance and provided my own ssh keys to the instance. Verified that I can login via ssh from the outside network. 

Added an AWS example

Signed-off-by: Dieter Reuter <dieter.reuter@me.com>


justincormack:
We should not have to add the bind mount, if it needs to be there it should be in the image metadata, but I am not sure whats changed as it should not have been needed.

DieterReuter:
The bind mount is already provided as a label/default within the image `linuxkit/sshd`.

justincormack:
@DieterReuter not `/var/config/ssh/authorized_keys:/root/.ssh/authorized_keys` which you added

DieterReuter:
The bind mount `/var/config/ssh/authorized_keys:/root/.ssh/authorized_keys` comes directly from the GCP example. 

But if you'd prefer to use the standard `~/.ssh/authorized_keys` location instead, we should change the storage of the ssh keys in metadata pkg. WDYT?

justincormack:
Not sure need to take a look.

DieterReuter:
I just found some more details. sshd service is looking for additional keys at /etc/ssh with the following search `KEYS=$(find /etc/ssh -name 'ssh_host_*_key')`, but metadata pkg stores the key as `/var/config/ssh/authorized_keys` for both GCP and AWS providers.

justincormack:
Also, as the two examples are identical, not sure we need both...

kencochrane:
They might be identical right now, but maybe not in the future. Also, it might make it easier for newbies to get started, if they see an aws.yml, it is more obvious that it is for aws, vs knowing that the gcp.yml is also good for aws.

justincormack:
ok. We should move them to `blueprints/` soon though.

DieterReuter:
Maybe I could include some other things as well like node_exporter, but my intention is to have a separate AWD example to get easy started for newbies. We should write some small tutorials, I think I'll start with a short blogpost on the weekend.

justincormack:
The plan was to have https://github.com/linuxkit/linuxkit/tree/master/blueprints have really detailed "this is how you might really want to run it" examples, although we may decide this is not how we want to do it.

PR #1959, opened 2017-06-02
Update moby tool and qemu fixes

justincormack:
- no longer uses several of the `mkimage-*` tools in favour of dogfooding
with `linuxkit` and using the `mkimage` package.

- fix the qemu docker container fallbacks to work better when multiple paths are used for disks and the image.

ardea_modesta


justincormack:
Some issues with paths when using containers to build

justincormack:
I disabled the changes for making GCP images as GCP refuses to recognise them, with no debug info, even though they run fine. I tried making the partitioning exactly the same as teh guestfs build but that doesnt help, so needs more investigation.

justincormack:
I am going to merge this as the AWS support needs it. Will continue to investigate the GCP issues with the original version.

PR #1960, opened 2017-06-02
Add /pkg/vsudd

MagnusS:
Adds new package`vsudd` that clones and builds `linuxkit/virtsock/cmd/vsudd`. This is #1947 with the vsudd code removed.

I've added a minimal example and a README.md for controlling containerd from the host with ctr.

Pushed to a temporary hub user for now so the example works.

### DIFF
--- a/examples/vsudd.yml
+++ b/examples/vsudd.yml
@@ -1,0 +1,22 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:deaf5bf838bf7f131c2287ecff3ed9835b0497e2
+onboot:
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+services:
+  - name: vsudd
+    image: "linuxkit/vsudd:a66df914201aac784195d5b78262f622fe7f910c"
+    binds:
+        - /run/containerd/containerd.sock:/run/containerd/containerd.sock
+    command: ["/vsudd", 
+        "-inport", "2374:unix:/run/containerd/containerd.sock"]
+
+trust:
+  org:
+    - linuxkit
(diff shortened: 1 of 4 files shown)


pkg/vsudd/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
+
+RUN apk add --no-cache go musl-dev git build-base
+ENV GOPATH=/go PATH=$PATH:/go/bin 
+
+RUN git clone https://github.com/linuxkit/virtsock.git /go/src/github.com/linuxkit/virtsock && \
```
justincormack: you need to checkout an exact commit here, else the `git tree` hash wont work.

MagnusS: good point - updated

justincormack:
@riyazdf can you set up Magnus with access and signing key for `linuxkit/vsudd`?

riyazdf:
@MagnusS: I've pinged you offline to get your signing key set up.

In the meantime, I've created the hub repo with signing and you should be able to use `linuxkit/vsudd:a66df914201aac784195d5b78262f622fe7f910c`

examples/vsudd.yml
```diff
@@ -0,0 +1,22 @@
...
+  - linuxkit/init:1b8a7e394d2ec2f1fdb4d67645829d1b5bdca037
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:deaf5bf838bf7f131c2287ecff3ed9835b0497e2
+onboot:
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+services:
+  - name: vsudd
+    image: "edtestorg/vsudd:50f52267d8ef9d61ecbba1fe9f896fa413477295"
```
riyazdf: `linuxkit/vsudd:a66df914201aac784195d5b78262f622fe7f910c`

examples/vsudd.yml
```diff
@@ -0,0 +1,22 @@
+kernel:
```
riyazdf: shall we add a build test for this example to CI?  Can be in a followup

pkg/vsudd/Makefile
```diff
@@ -0,0 +1,15 @@
+default: tag
```
riyazdf: `push` should be the default

examples/vsudd.yml
MagnusS: Thanks - updated

MagnusS:
@riyazdf @justincormack thanks!

I've updated the PR with the new vsudd hash and made `push` the default.

examples/vsudd.yml
justincormack: it makes more sens to do full tests when we have macos CI

riyazdf:
makes sense. This is fine as-is

PR #1964, opened 2017-06-05
Add AWS Support to LinuxKit Push and Run

dave-tucker:
Fixes: #1918 

**- What I did**
Implemented `linuxkit push aws` and `linuxkit run aws`

**- How I did it**
- Added the AWS Go SDK to vendor
- Wrote code to call the AWS API to add upload a raw image to Amazon S3, Import this as an EBS snapshot and to create an AMI from it.
- Wrote code to run the AMI (with an optional additional EBS disk)
- Wrote code to output the serial console logs and perform clean up after the instance has stopped.

**- How to verify it**
```
$ moby build -output img -size 200M examples/aws.yml
$ linuxkit push aws -bucket bucketname aws.img
$ linuxkit run aws aws
# Note: You will need to manually shutdown the machine unless you add a poweroff container to the yml
```

Add push/run support for AWS


docs/aws.md
```diff
@@ -0,0 +1,44 @@
...
+```
+export AWS_REGION=eu-west-1
+```
+
+## Build an image
+
+AWS requires a `RAW` image. To create one:
+
+```
+$ moby build -output img-gz examples/aws.yml
```
justincormack: I am just about to merge support for `img`

justincormack:
@dave-tucker looks like you have an out of date version of `vndr` and its removing some license files and subvendoring which should not change.

dave-tucker:
@justincormack ohnoes! 😲 
i'll re-`vndr` and see if I can fix it!

dave-tucker:
ok I think that fixed it...

vendor.conf
```diff
@@ -23,3 +23,7 @@ golang.org/x/oauth2 1611bb46e67abc64a71ecc5c3ae67f1cbbc2b921
 golang.org/x/sys 99f16d856c9836c42d24e7ab64ea72916925fa97
 google.golang.org/api 1202890e803f07684581b575fda809bf335a533f
 google.golang.org/grpc 0713829b980f4ddd276689a36235c5fcc82a21bf
+github.com/aws/aws-sdk-go fa107560b5f3528a859a1a1511086646731bb1a8
+github.com/go-ini/ini afbc45e87f3ba324c532d12c71918ef52e0fb194
+github.com/jmespath/go-jmespath bd40a432e4c76585ef6b72d3fd96fb9b6dc7b68d
+cloud.google.com/go 0bee953ffbf6fda3be3f56f1e9578f16a1e978a4
```
justincormack: Why `cloud.google.com/go` for this PR?

justincormack: It doesnt seem to have imported anything...

ijc25: The latest `vndr` warns you about unused entries in `vendor.conf` IIRC.

src/cmd/linuxkit/push_aws.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	content, err := ioutil.ReadAll(f)
+	if err != nil {
+		log.Fatalf("error reading file: %s", err)
+	}
+
+	putParams := &s3.PutObjectInput{
+		Bucket:        aws.String(bucket),
+		Key:           aws.String(fmt.Sprintf("%s.%s", name, filepath.Ext(src))),
```
DieterReuter: this should be "%s%s" instead of "%s.%s", otherwise you'll get `Key: "aws..img"` with double dots.

src/cmd/linuxkit/push_aws.go
```diff
@@ -0,0 +1,165 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+const timeoutVar = "LINUXKIT_UPLOAD_TIMEOUT"
+
+// Process the run arguments and execute run
```
riyazdf: nit: comment doesn't match the function

src/cmd/linuxkit/push_aws.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	f, err := os.Open(src)
+	if err != nil {
+		log.Fatalf("Error opening file: %s", err)
+	}
+	defer f.Close()
+
+	if name == "" {
+		name = src[0 : len(src)-len(filepath.Ext(src))]
```
riyazdf: super nit: I slightly prefer `name = strings.TrimSuffix(src, filepath.Ext(src))` so we don't have to reason about string slicing but this is functionally fine as-is

src/cmd/linuxkit/push_aws.go
```diff
@@ -0,0 +1,165 @@
...
+	compute := ec2.New(sess)
+
+	importParams := &ec2.ImportSnapshotInput{
+		Description: aws.String(fmt.Sprintf("LinuxKit: %s", name)),
+		DiskContainer: &ec2.SnapshotDiskContainer{
+			Description: aws.String(fmt.Sprintf("LinuxKit: %s disk", name)),
+			Format:      aws.String("raw"),
+			UserBucket: &ec2.UserBucket{
+				S3Bucket: aws.String(bucket),
+				S3Key:    aws.String(src),
```
DieterReuter: here you should change `src` into the new file name created with `<name>.img`, otherwise the file names differ and you're referencing the wrong S3 file

src/cmd/linuxkit/run_aws.go
```diff
@@ -0,0 +1,186 @@
...
+	invoked := filepath.Base(os.Args[0])
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s run aws [options] [name]\n\n", invoked)
+		fmt.Printf("'name' is the name of an AWS image that has already been\n")
+		fmt.Printf(" uploaded using 'linuxkit push'\n\n")
+		fmt.Printf("Options:\n\n")
+		flags.PrintDefaults()
+	}
+	machineFlag := flags.String("machine", defaultAWSMachine, "AWS Machine Type")
+	diskSizeFlag := flags.Int("disk-size", 0, "Size of system disk in GiB")
```
riyazdf: Should we make this`GB`[ consistent with GCP](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_gcp.go#L44)?  Or instead it seems like GCP should get updated to `GiB` since that is more correct...

justincormack:
It should be GB not GiB everywhere. We already agreed this...

riyazdf:
@justincormack 👍  sgtm

docs/aws.md
```diff
@@ -0,0 +1,43 @@
...
+$ moby build -output img examples/aws.yml
+```
+
+## Push image
+
+Do `linuxkit push aws -bucket bucketname aws.img` to upload it to the
+specified bucket, and create a bootable image from the stored image.
+
+Alternatively, you can use the `AWS_BUCKET` environment variable to specify the bucket name.
+
```
kencochrane: Does this step require the `vmimport` IAM role? If so, we should put a link to the docs so that people will know how to get this setup. http://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-image-import.html#vmimport-iam-permissions

If it doesn't require that step, then awesome, that was a real pain to do that first.

DieterReuter: This IAM role for `vmimport` is strongly required! We just checked it with a new AWS account while our tests for the blog post.

vendor.conf
dave-tucker: I had some issues with vendoring. By just adding the AWS deps I managed to remove 90% of the other dependencies 😢 Perhaps it was an old `vndr` or perhaps it was issues running locally vs in a container. Either way, I've fixed it now and will push an update

src/cmd/linuxkit/run_aws.go
dave-tucker: AWS only support GiB per http://docs.aws.amazon.com/cli/latest/reference/ec2/create-volume.html#options

dave-tucker:
@justincormack not sure how to proceeed on the `disk-size` front.

Making it consistent with other platforms would be weird as the minimum supported `-disk-size` parameter for AWS would be `1.073741824G`

src/cmd/linuxkit/run_aws.go
justincormack: Yes but we refer to them as "GB" as we are not pedants and it is the customary usage for storage.

justincormack:
@dave-tucker no, the GiB pedantry must stop. Just change the message to "GB".

dave-tucker:
@justincormack done.

dave-tucker:
Addressed all comments in previous reviews...

docs/aws.md
justincormack: @dave-tucker do you want to address this?

rneugeba:
re `disk-size` I'd be fine with making GB the default and requiring `M` to make it MB for small disks. Basically switch around the current behaviour for all backends

docs/aws.md
dave-tucker: I did by adding it to the pre-reqs here: https://github.com/linuxkit/linuxkit/pull/1964/files#diff-bde0f4c987d5959961ba607474757f2aR12

PR #1977, opened 2017-06-07
Add getty pkg

deitch:

**- What I did**

Created a `pkg` called `linuxkit/getty` that replaces the getty functionality from `linuxkit/init`.

**- How I did it**

* Create `pkg/getty/` that processes `/proc/cmdline` options and starts a getty for each console tty.
* respawn logic in the wrapper `rungetty.sh` because getty `exec`s its login (or `/bin/sh`) and expects some wrapper to respawn it when the user exits
* uses `tini -s` is the container root process so that it will reap zombie children
* added `examples/getty.yml`
* `README.md` describing how to use it

The child of `tini` is an `sh` script. It probably could be replaced with a golang binary, but it was easier to get started this way, especially with some of the existing logic taken from `pkg/init/init`.

**Note:** this does *not* remove the logic from `linuxkit/init`; that will be a separate PR to be merged in only after this PR is done.

**- How to verify it**

1. Create an init container that does not have the getty logic in it - sample available on docker hub at `deitch/linuxkit-init:nogetty`
2. Modify `examples/getty.yml` to use that `linuxkit/init` - replace the `init` container line `linuxkit/init:...` with the betty-less init container, e.g. `deitch/linuxkit-init:nogetty`
3. Try each of the 3 function modes and see that it works:
    * no getty service: comment out the entire getty service, see that no login or shell is available
    * insecure: leave the getty service in place, uncomment the env var for `INSECURE`, see that it drops straight to `sh` on console (the old logic)
    * password-secured: leave getty service in place and `INSECURE` commented, see that the console asks for a login, and that `root`/`abcdefgh` (per the example) work to get a shell

Sample getty.yml available at this gist https://gist.github.com/deitch/a24c291321adc6433061618d9fbaf746

Create standalone getty container image

croc

OK, it isn't really cute... 

Per discussions with @justincormack and @rneugeba


docs/security.md
```diff
@@ -73,9 +73,12 @@ containers unintentionally expose themselves to attack vectors, immutability of
 host attack.
 
 
+## Login
+By default, linuxkit has no login available: not on console, not via ssh, nowhere. You have the _option_ of enabling login on console using a `linuxkit/getty` service container, but it is not created by default.
```
justincormack: I think you should mention the `sshd` container here too.

examples/getty.yml
```diff
@@ -0,0 +1,31 @@
...
+  - name: getty
+    image: "linuxkit/getty:8305d9a564bfbe38b7ea6d6f5bccd95ae59b03d2"
+    # to make insecure with passwordless root login, uncomment following lines
+    #env:
+    # - INSECURE=true
+files:
+  - path: etc/getty.shadow
+    # sample sets password for root to "abcdefgh" (without quotes)
+    contents: 'root:$6$6tPd2uhHrecCEKug$8mKfcgfwguP7f.BLdZsT1Wz7WIIJOBY1oUFHzIv9/O71M2J0EPdtFqFGTxB1UK5ejqQxRFQ.ZSG9YXR0SNsc11:17322:0:::::'
+  - path: root/.ssh/authorized_keys
```
justincormack: whats this doing here?

pkg/getty/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT ["/sbin/tini","-s","-v","--"]
+WORKDIR /
+COPY --from=mirror /out/ /
+COPY usr/ /usr/
+COPY etc/ /etc/
+CMD ["/usr/bin/rungetty.sh"]
+LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/tmp/ctr:/tmp/ctr", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/var:/var","/containers:/containers","/dev:/dev"], "capabilities": ["all"]}'
```
justincormack: should prbably add `/usr/bin/dist` too. At some point these will have to move into this container as they won't be in the rootfs, but ok for now.

pkg/getty/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    util-linux \
+    && true
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT ["/sbin/tini","-s","-v","--"]
+WORKDIR /
+COPY --from=mirror /out/ /
+COPY usr/ /usr/
+COPY etc/ /etc/
```
justincormack: why are these two lines here? there is nothing in the context in usr or etc?

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -0,0 +1,65 @@
...
+	#echo "$tty::respawn:/sbin/getty -n -l /bin/sh $line $speed $tty $term" >> "$inittab"
+	if ! grep -q -w "$tty" "$securetty"; then
+		echo "$tty" >> "$securetty"
+	fi
+	# respawn forever
+	infinite_loop setsid -w /sbin/getty $loginargs $line $speed $tty $term &
+}
+
+# check if we have /etc/getty.shadow
+ROOTSHADOW=/proc/1/root/etc/getty.shadow
```
justincormack: No. We can't require that the user is in the host pid namespace to run this. For many use cases you will run the getty in its own pid namespace and it should still work. This file should be bind mounted in.

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ -f $ROOTSHADOW ]; then
+	cp $ROOTSHADOW /etc/shadow
+	# just in case someone forgot a newline
```
justincormack: ugh

deitch: Better idea? I originally bind-mounted it, but then if the file does not exist (e.g. user does not want to set a password or it just does not exist), the `ctr run` fails.

pkg/getty/usr/bin/rungetty.sh
deitch: See the other comment. Is there a better method? Bind-mounting fails if the file does not exist.

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -0,0 +1,65 @@
...
+	# are we secure or insecure?
+	loginargs=
+	if [ "$INSECURE" == "true" ]; then
+		loginargs="-n -l /bin/sh"
+	fi
+
+	# skip consoles already in inittab
+	#grep -q "^$tty:" "$inittab" && return
+
+	#echo "$tty::respawn:/sbin/getty -n -l /bin/sh $line $speed $tty $term" >> "$inittab"
```
justincormack: this comment is confusing, remove?

pkg/getty/Dockerfile
deitch: Probably because I pulled the original template over from `pkg/ssh`. There is `usr/` though.

examples/getty.yml
deitch: Oh, right. Originally had sshd in it. Removing.

docs/security.md
deitch: OK

pkg/getty/usr/bin/rungetty.sh
deitch: Yeah.

pkg/getty/Dockerfile
deitch: No, there is `etc/` too. We have motd and need a basic passwd and shadow.

pkg/getty/usr/bin/rungetty.sh
justincormack: hmm, I wonder if it is just easier to pass the hashed root password in an environment variable then and construct the shadow file?

pkg/getty/usr/bin/rungetty.sh
ijc25: Maybe we need a "conditional bind" syntax which only happens if the source exists? Or maybe this should be specified in the assembly yml iff you are using shadow, which in turn might need a way to add to the binds in the labels rather than replacing (assuming that's how it does work now).

BTW, why `/etc/getty.shadow` vs `/etc/getty`?

pkg/getty/Dockerfile
justincormack: Yeah sorry I saw them after and I thought I deleted that comment

docs/security.md
deitch: Updated

examples/getty.yml
deitch: Updated

pkg/getty/usr/bin/rungetty.sh
deitch: Removed

pkg/getty/usr/bin/rungetty.sh
deitch: Heh, that was my original plan (we discussed it on slack). But it is harder to control the credentials, easier to see via inspection, etc. 

I like the mounting. I really just would prefer a "bind-mount if available"

pkg/getty/usr/bin/rungetty.sh
justincormack: That would mean constructing (or modifying) the `config.json` at runtime rather than at build time. I have considered this, but it has many disadvantages.

pkg/getty/Dockerfile
justincormack: should add `/usr/bin/dist`

pkg/getty/usr/bin/rungetty.sh
deitch: > Maybe we need a "conditional bind" syntax which only happens if the source exists

I would have used it in a heartbeat. Gotten so used to docker - which does that - that I was surprised when `ctr` gave `ENOENT`.

>  Or maybe this should be specified in the assembly yml iff you are using shadow, which in turn might need a way to add to the binds in the labels rather than replacing (assuming that's how it does work now).

You cannot add mounts in the yml without losing everything in the label. I originally did that and spent forever trying to figure out why things failed. In truth, it makes sense. Label provides _default_, but you should be able to override. Some optional additive function would be nice.

> BTW, why /etc/getty.shadow vs /etc/getty?

Because it is a `shadow` file. I didn't want anyone to get confused.

pkg/getty/Dockerfile
deitch: Good point. Doing now.

pkg/getty/README.md
```diff
@@ -0,0 +1,64 @@
...
+2. Password login
+3. Open access
+
+
+#### Login Disabled
+Login disabled prevents any console login. This is the most secure option and recommended for production deployments.
+
+To disable login entirely:
+
+1. Ensure you are running a version of `linuxkit/init` that has getty disabled.
```
justincormack: I don't think we need to say this.

pkg/getty/Dockerfile
deitch: Updated

pkg/getty/README.md
deitch: A little paranoid but a good thing. Since there are versions with getty, someone relying on it to not exist for security reasons should explicitly check.

pkg/getty/usr/bin/rungetty.sh
deitch: > That would mean constructing (or modifying) the config.json at runtime rather than at build time.

Or just have `containerd` intelligent enough to say at runtime, "if I see a config.json, mount it, else not." Actually, that would be `runc`, right?

justincormack: You can actually mix binds in the yaml with ones in the label I think by specifying the ones in the label in the more verbose "mounts" format. I guess we could have two otherwise identical `binds` fields for the same reason.

deitch: So we do the verbose `mounts` format in the label and leave `binds` alone, and let people use `binds` for mapping in the `shadow` file?

pkg/getty/usr/bin/rungetty.sh
justincormack: Yes, but looking at it maybe its a better fit...

pkg/getty/Dockerfile
justincormack: did you push the update?

pkg/getty/usr/bin/rungetty.sh
deitch: ?

pkg/getty/Dockerfile
deitch: I thought I did. Will check again.

deitch: Arg. Forgot to force the update. Try now.

pkg/getty/usr/bin/rungetty.sh
justincormack: Its a sha256, I don't think it should be desperately secret if its a decent password. Also arguably access to the root pid namespace and the root mount namespace should be equally regarded as security issues, which is in effect what the issue is here.

deitch: So inspection is less of an issue? That assumes no one else can run `ctr` commands and see it. Whereas if it is in `/etc/getty.shadow`, owned by root perms 0600, there is some protection.

The bigger issue was where it exists in the source directory (where `.yml` are). If it is an env var, I _must_ put it in the `.yml` file. If it is a `files`, then I can source it, and it is a separate file, can be `.gitignore`d, etc.

deitch: > No. We can't require that the user is in the host pid namespace to run this. For many use cases you will run the getty in its own pid namespace and it should still work. 

@justincormack not convinced you are correct. The relevant namespace is not the one `ctr` is run from, but the one `containerd` is run from. Since `containerd` always is in host, and `getty` has a label `pid:host`, shouldn't this work?

Not arguing it isn't ugly....

justincormack: we could allow setting env vars from files (like Compose does).

If you can run `ctr` you can bind mount `/etc/getty.shadow` into a container and inspect it.

deitch: > If you can run ctr you can bind mount /etc/getty.shadow into a container and inspect it.

If you can run `ctr` *and* have access to the host namespace *and* are root. Basically anyone with `pid:host` can see it (which is true anyways via `/proc/1/root/etc/getty.shadow` as you pointed out).

So let's leave the security of `/etc/shadow` out of it, and focus on the user flow.

Never really liked `--env-file` type things much if they couldn't be recorded into the "code" (i.e. yml)....

Is there no saner way to bind-mount this? What if we bind-mounted `/etc:/hostetc`, and then we don't need to bind-mount, just see if the file exists? `cp /hostetc/getty.shadow /etc/shadow`?

justincormack: If you can run `ctr` you are controlling `containerd` that runs in the host namespace and will give you root access and allow you to bind mount from root, whichever namespace you are in. Same issue with Docker (will fix this one day...).

Yes I guess you can bindmount `/etc`

deitch:
As I see it, the remaining open issue is how to get the file into the container as `/etc/shadow`. Options are:

1. Assume running from host namespace and getty has host pid, which is what we did.
2. Always bind-mount `/etc/getty.shadow` and make it required to exist or it fails (it can be empty)
3. Always bind-mount `/etc:/hostetc` and check if `/hostetc/getty.shadow` exists (same as 1, but avoids the ugly `/proc/1/root/...` ugliness
4. In the label, do the verbose mounts and let the user do the simpler one for `/etc/getty.shadow:/etc/shadow`

1 is ugly and makes assumptions that may not hold
2 requires some infra in place (existence of the file `/etc/getty.shadow`) for the getty container to run at all
3 is a little broader than I would like, but reasonably ok
4 makes the Dockerfile a little longer (but not so bad), but allows the user to decide where on the host the shadow file should live and whether to mount it, but risks someone mixing up mounts (and wiping out from the label

Did I miss any? Preferences?

justincormack:
I think 3 is best for now, can change later.

deitch:
Like this?

Don't merge yet, I need to rerun tests; want to make sure we are on the same page first, though.

justincormack:
Yeah I need to try this out as well, havent had a chance yet, and make sure there are no usability issues.

deitch:
I am running through them now - only have three scenarios anyways - and will comment here, then off to you.

deitch:
All three cases work for me. On to you, @justincormack

docs/security.md
```diff
@@ -73,9 +73,12 @@ containers unintentionally expose themselves to attack vectors, immutability of
 host attack.
 
 
+## Login
```
riyazdf: thank you for adding to this :+1:

justincormack:
Ok this seems to work pretty well.

I have a slight concern that there will be a difficulty in debugging if `containerd` does not start say, and wonder if for that kind of debug it might be possible to also support it as an `init` container as well. I think that all that would be needed would be a `/etc/init.d/001-getty` script in this container so that if used in `init` it would actually setup the getty. Obviously the bind mounts won't work but it will be in the root mount namespace anyway, and it would always need to use insecure mode because of that.

Alternatively we could create a `getty-debug` package that does the same, but reusing the same one would seem to make sense as its basically the same.

Obviously this use would not be for production, but otherwise not sure how I will diagnose weird startup conditions...

deitch:
> I have a slight concern that there will be a difficulty in debugging if containerd does not start say, and wonder if for that kind of debug it might be possible to also support it as an init container as well

Heh, I had to spin up `sshd` to debug why the `getty` wasn't working as I was building it. Hard to debug when you have no shell.

> Alternatively we could create a getty-debug package that does the same, but reusing the same one would seem to make sense as its basically the same.

No, I think your thought makes sense. Essentially, we have two use cases:

1. Run as a `service` via `containerd`: the `/etc/init.d/001-getty` exists inside the container, and thus is ignored completely. All is good.
2. Used as an `init` container: `tini` and `rungetty.sh` are ignored (although copied over), and `/etc/init.d/001-getty` is placed and thus run. 

Makes a lot of sense. I think that the `init.d` format should always be `INSECURE` since it isn't only not just for production, it is for one-time debugging and then gets removed. 

I am good with that, adding it now.

deitch:
Added `/etc/init.d/001-getty` and updated `README.md` to describe the use case.

justincormack:
Ok this looks good. We need to push the images and add them to (most/all) the examples.

deitch:
Let's do it this way

1. Push out the new linuxkit/getty image to hub
2. Approve and merge this PR
3. Push out new init image to hub
4. Approve and merge that PR
5. I'll open a PR updating the examples.

justincormack:
@deitch have you had signing keys set up? cc @riyazdf

riyazdf:
@deitch I gave this a run and it's great!  I haven't set up signing keys with @deitch, look for a DM where we can go over this.  For now, @justincormack or I can sign/push this out

riyazdf:
given the time difference, I'll go ahead and set up the repo and signing

riyazdf:
`linuxkit/getty:6a9fabaa705c5dddba58d1b8fe56c7f9ee4ab813` is [pushed](https://hub.docker.com/r/linuxkit/getty/tags/) and signed

justincormack:
ok, merging.

riyazdf:
@justincormack: hmm the yml doesn't match what I pushed, must've changed tags from addressing review feedback?  I'll update in a followup PR.

deitch:
@riyazdf and @justincormack I believe I have had keys set up, when I pushed `linuxkitprojects/compose`. I will DM with you, Riyaz.

> the yml doesn't match what I pushed
Do you mean the hash? The `pkg` dir Makefile has a `make hash` target to give the latest hash. 

Maybe it is a later commit?

deitch:
Yeah, I still get 

```
$ make hash
b94cd2441cd7402a0071909b502ebf880127b1e1
```

I wonder where your `6a9fab...` tag came from.

PR #1978, opened 2017-06-07
Remove all getty from init

deitch:
**- What I did**

Removed the `getty` functionality from `linuxkit/init`. The only tty-related code left is to `cat /etc/issue` to the given tty, and even that probably could be removed or replaced.

**- How I did it**

Remove most of the code from `setup_console()` in `pkg/init/init`

**- How to verify it**

1. Build the container image
2. Build a linuxkit image using this new `init` image instead of the original one
3. See that no getty is started: you cannot login on console, nor do you get a shell

Remove getty functionality from init in favour of `linuxkit/getty`

Another croc, because that is what I did for the getty PR:

croc

Some important points:

1. This should be merged in only *after* the PR #1977 creating the `linuxkit/getty` image is in place (and pushed out to docker hub).
2. After both of these are merged in (and pushed out to docker hub), a separate PR should update all of the examples.
3. After both are in place, we might want to think of a "merged" login container that can be configured for sshd, getty, or both. But I think we want each working well independently before going there.

### DIFF
--- a/pkg/init/init
+++ b/pkg/init/init
@@ -1,45 +1,9 @@
 #!/bin/sh
 
-setup_console() {
-	tty=${1%,*}
-	speed=${1#*,}
-	inittab="$2"
-	securetty="$3"
-	line=
-	term="linux"
-	[ "$speed" = "$1" ] && speed=115200
-
-	case "$tty" in
-	ttyS*|ttyAMA*|ttyUSB*|ttyMFD*)
-		line="-L"
-		term="vt100"
-		;;
-	tty?)
-		line=""
-		speed="38400"
-		term=""
-		;;
-	esac
-	# skip consoles already in inittab
-	grep -q "^$tty:" "$inittab" && return
-
-	echo "$tty::once:cat /etc/issue" >> "$inittab"
-	echo "$tty::respawn:/sbin/getty -n -l /bin/sh $line $speed $tty $term" >> "$inittab"
-	if ! grep -q -w "$tty" "$securetty"; then
-		echo "$tty" >> "$securetty"
-	fi
-}
-
 /bin/mount -t tmpfs tmpfs /mnt
 
 /bin/cp -a / /mnt 2>/dev/null
 
 /bin/mount -t proc -o noexec,nosuid,nodev proc /proc
-for opt in $(cat /proc/cmdline); do
-	case "$opt" in
-	console=*)
-		setup_console ${opt#console=} /mnt/etc/inittab /mnt/etc/securetty;;
-	esac
-done
 
 exec /bin/busybox switch_root /mnt /sbin/init
--- a/pkg/init/etc/init.d/000-issue
+++ b/pkg/init/etc/init.d/000-issue
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+if [ -f /etc/issue ]; then
+	for opt in $(cat /proc/cmdline); do
+		case "$opt" in
+		console=*)
+			fulltty=${opt#console=}
+			tty=${fulltty%,*}
+			cat /etc/issue >> /dev/$tty
+		esac
+	done
+fi


pkg/init/init
```diff
@@ -2,32 +2,11 @@
...
-	tty?)
-		line=""
-		speed="38400"
-		term=""
-		;;
-	esac
 	# skip consoles already in inittab
 	grep -q "^$tty:" "$inittab" && return
 
 	echo "$tty::once:cat /etc/issue" >> "$inittab"
```
justincormack: lets just delete this and remove this whole function. Maybe the getty package can print the whale?

deitch: I thought about that. What happens if someone has no logins and therefore no `linuxkit/getty` package? Do we want no whale at all?

justincormack: yeah i think thats ok. We can always add a `whale` package for the desperate 🐳. The assumption is if you don't have its probably not something you are going to be looking at much.

justincormack: It makes it much easier to rework the base init stuff if there is less here.

deitch: I like the idea of the whale there; real branding. 

Without it in init, it makes it optional for people. I think the branding is important and should be part of init. Make people work to _remove_ it (which they can be using their own init or mapping a different `/etc/issue`) rather than _add_ it.

> It makes it much easier to rework the base init stuff if there is less here.

I agree with that.

justincormack: maybe it can be moved to an init script rather than here? A small script that runs before `containerd` should do the trick

deitch: You mean in this `pkg/init`? Or another line in `init:` of `.yml`?

justincormack: either would work, its fine to put it in here I think.

deitch: I am confused. DIdn't you just say to pull it out of here?

justincormack: No I meant out of this file, and remove the modifications to `inittab`. Then this file just deals with the sadly necessary rootfs copying., plus we remove any need to modify or use `inittab`

deitch: Understood.

Why pull it out entirely? I get killing `inittab`, but does it matter if the `car /etc/issue` is in this file, or if this file calls another script?

deitch: Just updated. Is this what you meant?

justincormack: This `/init` is only there to do a specific weird function that hopefully we can remove one day and then exec the real init. The less in it the better,

No, I meant add a `/etc/init.d/00/whale` that prints the whale on each console. This should then be run before `containerd` automatically.

deitch: I like that.

deitch: Like this?

justincormack:
Looks good but we can't merge yet needs the other PR merged and the images pushing

justincormack:
Yes, I think we need getty in most example images, but the idea was to do in followup PR. Some dont need it (eg AWS has no console, ssh images probably dont etc).

deitch:
The `linuxkit/getty` merge is done. @riyazdf pushed out an image (but there appears to be a hash tag issue). Once we get the right tag out for `linuxkit/getty` and then merge this one and push the image out, we can update the examples.

justincormack:
I updated the tag as I cleanup up the Makefile.

deitch:
Is this one good to go?

justincormack:
Sure, I can build a new `init` once its merged

deitch:
Excellent! 

Nice to collaborate together on those needs that overlap with those of my clients. Wish I could spend more time on some of the core. 

I don't have rights to sign and push to `linuxkit/init`, so up to you, sir.

justincormack:
`linuxkit/init:502b12be51753f7910e7362dad3e5f4ea77646c5`

deitch:
Speedy!

deitch:
And I see that the `linuxkit/getty` image is cleaned up with the right hash.

Just need to update the examples now, right?

Curious why you removed the `make hash` target from `pkg/getty`? I know it was just a convenience, but it was a nice one.

justincormack:
Consistency. If its useful we should add it everywhere. We used to have tiny little variations in the `Makefile`s that were significant, and it was annoying but they are very close now (modulo a few that have network access still), so it is easier to see where they vary on purpose...

PR #1982, opened 2017-06-07
Convert some tests to run on hyperkit when run on OSX

### LINKED ISSUE #1923: Fix hyperkit runner stdio handling
The main issue with using the hyperkit runner for tests is that stdio can't easily be redirected because  hyperkit expect a TTY.

`linuxkit.platforms.hyperkit.run_kernel` gets around this by using `expect` which makes it hard to write tests which work with different other backends.

The `script` utility also creates a fake TTY and can be used to save the output to a file.

The idea, then, is to provide a generic shell function in the test library which save the output of a command to a file (using `script` on macOS and redirect with qemu) and use that in all tests which should be run with whatever native runner exists on the platform.

rneugeba:
The newly vendored HyperKit go bindings allow for redirection of stdout (resolves #1923) and thus allows us to use HyperKit on OSX for the kernel and most of the package tests.

Ideally we would set labels on which platforms these tests are run, but there seems to be an [issue](https://github.com/linuxkit/rtf/issues/16) with labels and `rtf`.

### DIFF
--- a/test/cases/020_kernel/000_config_4.4.x/test.sh
+++ b/test/cases/020_kernel/000_config_4.4.x/test.sh
@@ -10,13 +10,13 @@ set -e
 . "${RT_PROJECT_ROOT}/_lib/lib.sh"
 
 clean_up() {
-	find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \;
+	find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \; || true
 }
 trap clean_up EXIT
 
 # Test code goes here
 moby build -output kernel+initrd test-kernel-config.yml
-RESULT="$(linuxkit run qemu -kernel test-kernel-config)"
+RESULT="$(linuxkit run test-kernel-config)"
 echo "${RESULT}" | grep -q "suite PASSED"
 
 exit 0
--- a/test/cases/020_kernel/003_config_4.11.x/test.sh
+++ b/test/cases/020_kernel/003_config_4.11.x/test.sh
@@ -10,13 +10,13 @@ set -e
 . "${RT_PROJECT_ROOT}/_lib/lib.sh"
 
 clean_up() {
-	find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \;
+	find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \; || true
 }
 trap clean_up EXIT
 
 # Test code goes here
 moby build -output kernel+initrd test-kernel-config.yml
-RESULT="$(linuxkit run qemu test-kernel-config)"
+RESULT="$(linuxkit run test-kernel-config)"
 echo "${RESULT}" | grep -q "suite PASSED"
 
 exit 0
--- a/test/cases/020_kernel/010_kmod_4.9.x/test.sh
+++ b/test/cases/020_kernel/010_kmod_4.9.x/test.sh
@@ -13,7 +13,7 @@ IMAGE_NAME="kmod-test"
 
 clean_up() {
 	docker rmi ${IMAGE_NAME} || true
-	find . -iname "kmod*" -not -iname "*.yml" -exec rm -rf {} \;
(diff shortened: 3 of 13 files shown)


rneugeba:
hmm, weird, the CI fails with:
```
# github.com/linuxkit/linuxkit/vendor/github.com/moby/hyperkit/go
vendor/github.com/moby/hyperkit/go/pty_util.go:16: undefined: unix.TIOCGETA
vendor/github.com/moby/hyperkit/go/pty_util.go:20: undefined: unix.TIOCSETA
vendor/github.com/moby/hyperkit/go/pty_util.go:61: cannot use uint64(uint32(t.Iflag) & ^uint32(unix.IGNBRK | unix.BRKINT | unix.PARMRK | unix.ISTRIP | unix.INLCR | unix.IGNCR | unix.ICRNL | unix.IXON)) (type uint64) as type uint32 in assignment
vendor/github.com/moby/hyperkit/go/pty_util.go:62: cannot use uint64(uint32(t.Oflag) & ^uint32(unix.OPOST)) (type uint64) as type uint32 in assignment
vendor/github.com/moby/hyperkit/go/pty_util.go:63: cannot use uint64(uint32(t.Lflag) & ^uint32(unix.ECHO | unix.ECHONL | unix.ICANON | unix.ISIG | unix.IEXTEN)) (type uint64) as type uint32 in assignment
vendor/github.com/moby/hyperkit/go/pty_util.go:64: cannot use uint64(uint32(t.Cflag) & ^uint32(unix.CSIZE | unix.PARENB)) (type uint64) as type uint32 in assignment
Makefile:38: recipe for target 'bin/linuxkit' failed
```
but this works fine when build locally....and it's build in docker so unclear why this fails

rneugeba:
Ah, this is compiling it for Linux

test/cases/020_kernel/000_config_4.4.x/test.sh
```diff
@@ -4,19 +4,20 @@
 # REPEAT:
 
 set -e
+set -x
```
dave-tucker: I think this is unnecessary as IIRC `rtf` adds this when run with `rtf -x`

test/cases/020_kernel/003_config_4.11.x/test.sh
```diff
@@ -4,19 +4,20 @@
...
 set -e
+set -x
 
 # Source libraries. Uncomment if needed/defined
 #. "${RT_LIB}"
 . "${RT_PROJECT_ROOT}/_lib/lib.sh"
 
 clean_up() {
-	find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \;
+    find . -iname "test-kernel-config*" -not -iname "*.yml" -exec rm -rf {} \; || true
```
dave-tucker: nit: Are we replacing tabs with 4 spaces?
I thought I recall agreement on tabs but I may be mistaken

justincormack: https://www.youtube.com/watch?v=SsoOG6ZeyUI

justincormack:
sorry, bit of a merge conflict issue with me removing support for omitting the file extension from the yml file

test/cases/020_kernel/003_config_4.11.x/test.sh
riyazdf: @justincormack that is perfect 🤣

rneugeba:
I need to revendor anyway. I submitted a [PR for hyperkit](https://github.com/moby/hyperkit/pull/138) to make it compile on Linux (and other OSes) and then will update this PR hopefully tomorrow

rneugeba:
Update with new HyperKit, remove `set -x`, and added tabs back (which is wrong ;). 

Modified test pass on OSX with hyperkit

test/cases/020_kernel/010_kmod_4.9.x/test.sh
```diff
@@ -4,6 +4,7 @@
 # REPEAT:
 
 set -e
+set -x
```
justincormack: you left one `set -x`

rneugeba:
thanks for spotting. force pushed update

PR #1983, opened 2017-06-07
Update Moby tool

justincormack:
- rename img→raw and gcp-img→gcp
- support creating `docker` outputs, see https://github.com/moby/tool/tree/master/examples
- less memory usage via streaming outputs
- allow specification of multiple yaml files in a single command line, improves modularity
- notary fixes and local cache, so does not pull as much
- you now have to specify the full filename of the yaml file, you cannot omit the suffix

heronfish

### DIFF
--- a/test/cases/040_packages/019_sysctl/test.sh
+++ b/test/cases/040_packages/019_sysctl/test.sh
@@ -16,7 +16,7 @@ clean_up() {
 trap clean_up EXIT
 
 # Test code goes here
-moby build -output kernel+initrd test-sysctl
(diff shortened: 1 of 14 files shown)


test/cases/040_packages/019_sysctl/test.sh
```diff
@@ -16,7 +16,7 @@ clean_up() {
 trap clean_up EXIT
 
 # Test code goes here
-moby build -output kernel+initrd test-sysctl
+moby build -output kernel+initrd test-sysctl.sh
```
riyazdf: `test-sysctl.yml`?

PR #1999, opened 2017-06-08
Update AUTHORS

justincormack:

owlet

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,7 +8,7 @@ Avi Deitcher <avi@deitcher.net>
 Carlton-Semple <carlton.semple@ibm.com>
 Dan Finneran <dan@thebsdbox.co.uk>
 Daniel Hiltgen <daniel.hiltgen@docker.com>
-Dave Scott <dave.scott@docker.com>
+Daniel Nephin <dnephin@gmail.com>
 Dave Tucker <dt@docker.com>
 David Gageot <david.gageot@docker.com>
 David Scott <dave.scott@docker.com>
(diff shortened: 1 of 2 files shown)


justincormack:
There were two Dave/David Scotts - which do you prefer @djs55

justincormack:
The script is not very good at normalising name differences, only email differences...

AUTHORS
```diff
@@ -36,7 +36,9 @@ Mindy Preston <mindy.preston@docker.com>
 Natanael Copa <natanael.copa@docker.com>
 Nathan Dautenhahn <ndd@seas.upenn.edu>
 Nathan LeClaire <nathan.leclaire@docker.com>
+Nico Di Rocco <dirocco.nico@gmail.com>
 Phil Estes <estesp@linux.vnet.ibm.com>
+radu-matei <matei.radu94@gmail.com>
```
thaJeztah: If you want to have this nicer looking; name looks to be `Radu Matei`; https://github.com/radu-matei

justincormack: updated

djs55:
Could you stick me down as `David Scott <dave.scott@docker.com>`?  It's what I've been using elsewhere.

justincormack:
@djs55 amended (you could change your github profile to match...)

djs55:
@justincormack ha, good point. I've done that too. Marginally more consistent now

AUTHORS
thaJeztah: Oh, that needed an entry in `.mailmap` so that it automatically maps the e-mail to the preferred name. Let me do a follow-up

PR #2000, opened 2017-06-08
build: Make sure the tools compile on Linux/macOS/Windows

rneugeba:
This patch adds cross compilation steps for Linux, macOS and,
Windows to the CI targets to ensure that the all commands cross
compile nicely.

Although the CI runs on Linux we still explicitly compile for
Linux in case developers use the CI targets locally on a
non-Linux host.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,10 @@ LINUXKIT?=bin/linuxkit
 GOOS?=$(shell uname -s | tr '[:upper:]' '[:lower:]')
 GOARCH?=amd64
 ifneq ($(GOOS),linux)
-CROSS=-e GOOS=$(GOOS) -e GOARCH=$(GOARCH)
+CROSS+=-e GOOS=$(GOOS)
+endif
+ifneq ($(GOARCH),amd64)
+CROSS+=-e GOARCH=$(GOARCH)
 endif
 
 PREFIX?=/usr/local/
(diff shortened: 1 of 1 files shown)


justincormack:
You could add `linux/aarch64` I guess too as we have a WIP port while you are at it

rneugeba:
good idea. It required a small fix to the Makefile, which I force pushed.

However, the cross compile failed, so I didn't add it to the CI targets.
```
make GOOS=linux GOARCH=aarch64
docker run --rm --log-driver=none -e GOARCH=aarch64 linuxkit/go-compile:6579a00b44686d0e504d513fc4860094769fe7df --clone-path github.com/moby/tool --clone https://github.com/moby/tool.git --commit e0aac90f4476c7dadfec9ab4116cf1363e51ce77 --package github.com/moby/tool/cmd/moby --ldflags "-X main.GitCommit=3e5797a911c39a286fc85ce224131bccefe4a132 -X main.Version="0.0" " -o bin/moby > tmp_moby_bin.tar
Cloning into '.'...
Note: checking out 'e0aac90f4476c7dadfec9ab4116cf1363e51ce77'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at e0aac90... Merge pull request #74 from justincormack/raw-name
gofmt...
govet...
vet: import failed: can't find import: net/http
vet: import failed: can't find import: net/http
vet: import failed: can't find import: fmt
vet: import failed: can't find import: fmt
sh: import: unknown operand
make: *** [bin/moby] Error 2
```

justincormack:
Thats odd, will take a look later.

avsm:
cc @mor1 re: the arm64 failure

mor1:
Odd, though I recall having to do something a bit odd to build the tool in the past. Will try to take a look tomorrow.

PR #2003, opened 2017-06-08
[WIP] Build and Sign Packages in CI

dave-tucker:
This PR is a work in progress.

So far it adds:
- Make target to check if package is on hub, if not, build it
- Integration of above target to the top-level Makefile

What's not understood yet is:
- Is this sufficient to test changes to a package or will they fail as the locally built image isn't signed?
- Is CI configured with a key and delegation for the linuxkit repos
- Are we happy to run a `make -C push` on `master` and `tag` CI jobs

### DIFF
--- a/pkg/Makefile
+++ b/pkg/Makefile
@@ -1,5 +1,7 @@
 DIRS = $(shell find . -type d -depth 1)
-.PHONY: clean dirs $(DIRS)
+export OVERRIDES = $(CURDIR)/override.yml
+
+.PHONY: clean dirs check hash $(DIRS)
 
 push:
 	@set -e; for d in $(DIRS); do make -C "$$d"; done
@@ -7,4 +9,11 @@ push:
 tag:
 	@set -e; for d in $(DIRS); do make -C "$$d" tag; done
 
-clean: ;
+check:
+	@set -e; for d in $(DIRS); do make -C "$$d" check; done
+
+hash:
+	@set -e; for d in $(DIRS); do make -C "$$d" hash; done
+
+clean:
+	rm -f $(OVERRIDES)
--- a/Makefile
+++ b/Makefile
@@ -82,20 +82,20 @@ collect-artifacts: artifacts/test.img.tar.gz artifacts/test-ltp.img.tar.gz
 ci: test-cross
 	$(MAKE)
 	$(MAKE) install
+	$(MAKE) -C pkg check
 	$(MAKE) -C test all
-	$(MAKE) -C pkg tag
 
 ci-tag: test-cross
 	$(MAKE)
 	$(MAKE) install
+	$(MAKE) -C pkg check
 	$(MAKE) -C test all
-	$(MAKE) -C pkg tag
 
 ci-pr: test-cross
 	$(MAKE)
 	$(MAKE) install
+	$(MAKE) -C pkg check
 	$(MAKE) -C test pr
-	$(MAKE) -C pkg tag
 
 .PHONY: clean
 clean:
--- a/pkg/.gitignore
+++ b/pkg/.gitignore
@@ -1,0 +1,1 @@
+override.yml
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -2,8 +2,17 @@
 default: push
 
 ORG?=linuxkit
+STAGE_ORG?=linuxkit.datakit.ci:5000
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 BASE_DEPS=Dockerfile Makefile
+OVERRIDES?=override.yml
+
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
+	TIMEOUT=timeout
+else
+	TIMEOUT=gtimeout
+endif
 
 tag: $(BASE_DEPS) $(DEPS)
 ifndef $(NETWORK)
@@ -15,3 +24,11 @@ endif
 push: tag
 	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
 	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+
+hash:
+	@echo "$(ORG)/$(IMAGE):$(HASH)"
+
+check:
+	@$(TIMEOUT) 5 notary -s https://notary.docker.io lookup docker.io/$(ORG)/$(IMAGE) $(HASH) > /dev/null || \
+	$(MAKE) ORG=$(STAGE_ORG) push && \
+	echo "- source: $(ORG)/$(IMAGE)\n  subsitute: $(STAGE_ORG)/$(IMAGE):$(HASH)\n" >> $(OVERRIDES)


riyazdf:
@dave-tucker: I set up a CI delegation with the key you gave to me a while ago, on all linuxkit packages.  Please let me know immediately if you are planning to use a different key

pkg/binfmt/Makefile
```diff
@@ -13,3 +13,11 @@ tag: $(DEPS)
 push: tag
 	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
 	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+
+check:
+	@if curl -sSL https://registry.hub.docker.com/v2/repositories/$(ORG)/$(IMAGE)/tags/$(HASH) | grep -q 'Not found'; then\
```
justincormack: you should not use `curl` for this. Especially as it ties it to docker hub. whats wrong with using `pull` to see if it exists? Especially as we want to test it exists and is signed anyway

justincormack:
ideally we want to test from a staging repo, so we dont pollute the main repo with failed packages or things that were never merged.

We could have a way to substitute a file/files while testing, either on the `moby` cli or in an additional yaml file? So the tests run 
```
moby build test.yml $(TEST_YAML)
```
where TEST_YAML is the path to a yaml file that is generated with a map of something like:
```
- source: linuxkit/dhcpcd
  substitute: linuxkitci/dhcpcd:testbuildhash
```
and that causes the specified images to be swapped out.

Not sure, just thinking...

pkg/binfmt/Makefile
rneugeba: for small images `pull` is ok, but for larger images it might take quite a while. But i agree re `curl`. Maybe we can add something to the `moby` to do a quick check (including signing)?

justincormack: it wont pull anything if previously pulled

rneugeba: sure, but in CI we start from a clean slate (normally) so every package will be pulled

dave-tucker: yes I didn't use `pull` because there is no `pull --dry-run` to check if it's available and signed but not *actually* pull it.

> Especially as we want to test it exists and is signed anyway
Do we want to test that it's signed? See comment below for further thoughts on this.

dave-tucker:
Thanks @riyazdf - I think you gave it to @talex5 and I assume it's made it's available in CI

@justincormack continuation of my thought above. I was thinking about making it easier for CI to push images on behalf of contributors that are not part of the linuxkit org...
Otherwise, someone has to get them to generate a key and set up delegation to allow them to push in order for a PR to be testable - ideally this happens *before* they make a PR and the tests fail.

Maybe for PR's it's *ok* for images not be signed if we built them locally.
That would allow for a PR that, for example, makes a change to `pkg/containerd` and updates all of the YAML files to be testable without requiring push access to hub.

This can all be handled through `rtf` and the use of labels. e.g when `-l trust` is present, export `DOCKER_CONTENT_TRUST=1`

pkg/binfmt/Makefile
justincormack: you can use `notary` to list signed packages in a repo without pulling.

justincormack:
Nothing should be pushed to `linuxkit/` or signed there to run tests, because we mustn't sign anything that is not merged. Thats why I suggested having a way to switch the repo being used to a test one during the build process.

dave-tucker:
> Nothing should be pushed to linuxkit/ or signed there to run tests, because we mustn't sign anything that is not merged.

100% agree - I'm not suggesting otherwise. I'm suggesting only that we ignore the lack of signature for locally built images when testing PR's. Once merged, the `master` CI job can handle sign/push and running tests.

I've made the assumption that `moby` will use a locally built image (i.e from `docker build` ).

justincormack:
We won't be able to use a local image once we switch to `containerd`, and also pushing images to a test repo could be useful for debug. Also if you eg tag an image `linuxkit/blah:tag` locally, you still have to disable signing checks for that image in the tests.

dave-tucker:
I guess then we should consider adding a registry to the CI infrastructure.
Then we could push to the staging registry and I assume sign with a staging key.
We can also, occasionally, blitz the registry and start from scratch to remove stale images.

@riyazdf how do you think we should manage this? should we promote images from staging->prod or should we rebuild and sign/push on `master`? what controls should we implement to protect the signing key used for staging? I think at a minimum we'd have to wait until we have the whitelist implemented from #1853 unless we have a smart way of protecting against key compromise on PR builds.

justincormack:
ideally PR builds would be on a totally different machine without the signing keys at all.

dave-tucker:
they are. but wouldn't they need a signing key to push changes images to the `test` registry in order for the tests to be able to run with content trust enabled? or would we have to do that as a step *before* the PR is tested on a machine with the keys?

justincormack:
Not sure the test registry needs signing, but maybe it could generate an
ephemeral key, and maybe each test run would use a different repo? Not sure.

rn:
I *think* that if the hub repo is not configured for Content Trust you can just push and it'll work even if you push with `DOCKER_CONTENT_TRUST=1`. I, for example, overwrite the hub org to `rneugeba` when compiling test kernels and it works fine. So if we have a local registery we "just" need to override the `ORG` and fix up the YAML files

justincormack:
The reason I was suggesting an additional section in the yaml that specifies overrides was to avoid actually changing the yaml, you just add an override yaml file if building in CI, and leave the file unchanged

dave-tucker:
ok, then I'll do the following:

- spin up a registry as part of our CI infra
- adjust the check target to use `notary` and output a list of images that need building
- if the list is non-zero, for each image run `make -C pkg/$name push` overriding `ORG` to the staging repo for PR's
- generate a `override.yml` that contains a map like https://github.com/linuxkit/linuxkit/pull/2003#issuecomment-307161596
- configure the tests to use the `override.yml` if it's present

Note: I'm on vacation next week so this will probably be done when I get back.

pkg/binfmt/Makefile
riyazdf: for now we use hub so this has the Docker Hub notary server and registry tied to it, but this command should work for just checking one tag:
`notary -s https://notary.docker.io lookup docker.io/$(ORG)/$(IMAGE) $(HASH)`

As @justincormack points out, we can list all tags too (and grep):
`notary -s https://notary.docker.io list docker.io/$(ORG)/$(IMAGE)`

riyazdf:
lots of good discussion - I think it's a I just wanted to summarize and confirm my understanding about the current proposal:

For PR builds:
- first check if the PR author is on the whitelist or if the PR has been approved otherwise.  Abort if not
- check if all specified images exist in the production registry (hub) and notary
  - if not, build and push with `ORG` set to our test registry
  - generate `override.yml` to reflect the test registry images and run the tests
  - periodically (or immediately) delete the images from the test registry

For master builds:
- same as above, but promote any `override` images to the production registry by pushing with content trust using the CI key for signing

I agree that PR builds should not be given access to the CI key.


I'm also thinking of another way to structure this, in which we enforce policy by signature thresholding.  This may require major changes to the moby tool to understand more complex signing policy or we could migrate to DDC which has this (requires major changes to how we distribute images).  I'll comment here if I think of a concrete proposal, though I'm also 👍  with the setup we've discussed.

justincormack:
@riyazdf I think it would be nice to have the capability to support the DDC style workflows, as others may want to use them, and we may want to add further gating eg for releases, but having all the PR images in the main repo seems ugly, so I think we should still have a `ci` repo anyway.

riyazdf:
@justincormack: good point, we could support DDC style multi-signature thresholding on the production registry with the currently outlined design

pkg/Makefile
```diff
@@ -1,10 +1,16 @@
 DIRS = $(shell find . -type d -depth 1)
-.PHONY: clean dirs $(DIRS)
+.PHONY: clean dirs check hash $(DIRS)
```
fntlnz: The target named `dirs` is needed in that .PHONY ?

pkg/containerd/Makefile
```diff
@@ -13,3 +13,11 @@ tag: Dockerfile
 push: tag
 	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
 	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+
+check:
+	@if curl -sSL https://registry.hub.docker.com/v2/repositories/$(ORG)/$(IMAGE)/tags/$(HASH) | grep -q 'Not found'; then\
```
fntlnz: WDYT about using the built-in `ifneq` construct, like:

```make
check:
ifneq ($(curl -sSL https://registry.hub.docker.com/v2/repositories/$(ORG)/$(IMAGE)/tags/$(HASH) | grep -q 'Not found'),0)
  $(MAKE) tag
endif

```

dave-tucker:
Closing as I'm not actively working on this

PR #2008, opened 2017-06-09
vpnkit-forwarder: add /pkg/vpnkit-forwarder

MagnusS:
Adds /pkg/vpnkit-forwarder which downloads and builds `vpnkit-forwarder`
and `vpnkit-expose-port` from moby/vpnkit. Also includes an example for
forwarding `sshd` and updates the documentation for `hyperkit` to
include a `vpnkit` section.

`vpnkit` still requires a 9P mount for coordination, so the `socat`
forwarding is simpler to set up at the moment.

I've pushed to a temporary location so the example works.

cuteanimal05

### DIFF
--- a/docs/platform-hyperkit.md
+++ b/docs/platform-hyperkit.md
@@ -72,14 +72,14 @@ docker build -t ssh .
 docker run --rm -ti -v ~/.ssh:/root/.ssh  ssh ssh <IP address of VM>
 ```
 
-
 ### Forwarding ports to the host
 
-While VPNKit has the general tooling to expose any VMs port on the
-localhost (just like it does with containers in Docker for Mac), we
-are unlikely to expose this as a general feature in `linuxkit run` as
-it is very specific to the macOS. However, you can use a `socat` container to proxy between LinuxKit VMs ports and localhost.  For example, to expose the redis port from the [RedisOS example](../examples/redis-os.yml), use this Dockerfile:
+Ports can be forwarded to the host using a container with `socat` or with VPNKit which comes with Docker for Mac.
 
+#### Port forwarding with `socat`
+A `socat` container can be used to proxy between the LinuxKit VM's ports and
+localhost.  For example, to expose the redis port from the [RedisOS
+example](../examples/redis-os.yml), use this Dockerfile:
 ```
 FROM alpine:edge
 RUN apk add --no-cache socat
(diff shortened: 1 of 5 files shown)


ijc25:
Is it necessary/desirable for all pkg to be built out of the linuxkit repository? It would seem fairly natural to me for `moby/vpnkit` to build and publish these.

examples/vpnkit-forwarder.yml
```diff
@@ -0,0 +1,44 @@
...
+init:
+  - linuxkit/init:2599bcd5013ce5962aa155ee8929c26160de13bd
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a
+onboot:
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+  - name: mount-vpnkit
+    image: "linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9"
```
justincormack: you don't want to use the `linuxkit/alpine` image, it is huge as it as an `apk` mirror in it. Use an upstream `alpine` image

justincormack:
@ijc25 I think its fine to be here for now, as we are working on build and signing infrastructure, tests etc here. In future we could move them once thats built into tooling. Its still plumbing, so its fine being here.

examples/vpnkit-forwarder.yml
ijc25: I wonder if perhaps `linuxkit/alpine` ought to be called something like `linuxkit/alpine-build-env` or something to avoid this sort of confusion?

MagnusS: Updated to use alpine:3.6. I'll also move the image from `edtestorg` to `linuxkit` when I can sign the images (hopefully in a few hours)

riyazdf: @MagnusS I added your signing key to the `linuxkit/vpnkit-forwarder` repo, could you try pushing and signing?

MagnusS:
Pushed update with signed linuxkit/vpnkit-forwarder and updated docs to resolve merge conflict with #2017

docs/platform-hyperkit.md
```diff
@@ -91,6 +91,22 @@ docker build -t socat .
 docker run --rm -t -d -p 6379:6379 socat tcp-listen:6379,reuseaddr,fork tcp:<IP address of VM>:6379
 ```
 
+#### Port forwarding with `vpnkit`
+
+VPNKit has the general tooling to expose any guest VM port on the host (just
+like it does with containers in Docker for Mac). To enable forwarding, a
+`vpnkit-forwarder` container must be running in the VM. The VM also has to be
+booted with `linuxkit run hyperkit -networking=vpnkit`. A `vpnkit-expose-port`
+container has to run for each exposed port. The containers currently use a 9P
```
justincormack: is there some way in future we can use a single container to expose multiple ports?

MagnusS: @justincormack yes, you could run it multiple times in the same container or modify it to open multiple ports. It just configures /ports correctly - it's the vpnkit-forwarder process that does the actual forwarding. It's intended to run as a userland proxy for `dockerd`

justincormack: ok maybe change the wording as it sounds like each needs a container

MagnusS: @justincormack yes - updated, also added a link to vpnkit docs with more details about how the forwarding setup via 9p works

PR #2009, opened 2017-06-09
cmd: Default disk "size" parameter to GB

rn:
Most cloud providers allow disk size allocation on in units
of GB. Make it the default for linuxkit disk "size" arguments.
Users can override the unit by appending a M to the disk size.


justincormack:
Did you check all the tests explicitly specify a size? I think they do.

rn:
Yes, I grepped for `,size` and all instances explicitly specify the unit

PR #2010, opened 2017-06-09
Update MAINTAINERS

rn:
Both @ijc25 and @rneugeba changed their GH username


ijc:
LGTM

PR #2014, opened 2017-06-09
Enhanced the AWS provider in the metadata package, to include more data.

kencochrane:
Signed-off-by: Ken Cochrane <kencochrane@gmail.com>

**- What I did**
There is a lot of metadata available from the AWS metadata service. We originally were only capturing a few fields, this PR adds some of the more common data as well to make it easily available when you need it.

New fields added:
- public ipv4 address
- private ipv4 address
- availability zone
- instance id
- instance type
- local hostname

**- How I did it**
changed the metadata package to query the data from AWS metadata service and write the data to the /var/config directory.

**- How to verify it**
build the package and run on aws, and then look inside of `/var/config` you will notice more data available.

Enhanced the AWS provider in the metadata package, to include more data.

### DIFF
--- a/pkg/metadata/provider_aws.go
+++ b/pkg/metadata/provider_aws.go
@@ -47,6 +47,25 @@ func (p *ProviderAWS) Extract() ([]byte, error) {
 		return nil, fmt.Errorf("AWS: Failed to write hostname: %s", err)
 	}
 
+	// public ipv4
+	awsMetaGet("public-ipv4", "public_ipv4", 0644)
+
+	// private ipv4
+	awsMetaGet("local-ipv4", "local_ipv4", 0644)
+
+	// availability zone
+	awsMetaGet("placement/availability-zone", "availability_zone", 0644)
+
+	// instance type
+	awsMetaGet("instance-type", "instance_type", 0644)
+
+	// instance-id
+	awsMetaGet("instance-id", "instance_id", 0644)
+
+	// local-hostname
+	awsMetaGet("local-hostname", "local_hostname", 0644)
+
+	// ssh
 	if err := p.handleSSH(); err != nil {
 		log.Printf("AWS: Failed to get ssh data: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/metadata/provider_aws.go
```diff
@@ -47,6 +47,25 @@ func (p *ProviderAWS) Extract() ([]byte, error) {
...
 	}
 
+	// public ipv4
+	awsMetaGet("public-ipv4", "public_ipv4", 0644)
+
+	// private ipv4
+	awsMetaGet("local-ipv4", "local_ipv4", 0644)
+
+	// availability zone
+	awsMetaGet("placement/availability-zone/", "availability_zone", 0644)
```
riyazdf: not too familiar with AWS, but is the trailing `/` intended here?

pkg/metadata/provider_aws.go
```diff
@@ -61,6 +80,19 @@ func (p *ProviderAWS) Extract() ([]byte, error) {
 	return userData, nil
 }
 
+// lookup a value (lookupName) in aws metaservice and store in given fileName
+func awsMetaGet(lookupName string, fileName string, fileMode os.FileMode) {
+	lookupValue, err := awsGet(metaDataURL + lookupName)
```
rn: not sure if this is more idiomatic, but I've seen:
```
if lookupValue, err := awsGet(...); err == nil {
} else {
}
```
saves one line and `lookupValue` is slightly tighter scoped

kencochrane:
@riyazdf either works, but to keep it consistent I removed the slash

pkg/metadata/provider_aws.go
```diff
@@ -61,6 +80,19 @@ func (p *ProviderAWS) Extract() ([]byte, error) {
...
+	if err == nil {
+		err = ioutil.WriteFile(path.Join(ConfigPath, fileName), lookupValue, fileMode)
```
rn: Would it make sense to write the AWS specific meta data into its own subdir?

kencochrane: I thought about that, but didn't know what the current thinking was around this. Some of the data is AWS specific (availability zone, instance id, instance type) some isn't (public ip, local ip, local hostname). 

If you want I can change it to `/var/config/aws/<blah>` for the aws stuff, and leave the other fields the way they are.

pkg/metadata/provider_aws.go
kencochrane: sure thing, it does look cleaner.

pkg/metadata/provider_aws.go
rn: let's leave it as is for now. I guess it depends on how it will be used and we don't know that yet

rn:
can you squash the two commits?

kencochrane:
@rn I made the changes you suggested and merged my commits.

PR #2021, opened 2017-06-12
Only use console=ttyS0 for tests

justincormack:
If you also use console=tty0 you lose the output of commands running from
the system, as they run on the second console which is discarded.

Also drop page-poison as not relevant for tests.


ijc:
IIRC the final `console=` becomes the "primary" one. So you could also just switch the order, if you wanted.

justincormack:
@ijc yes, but seems simpler just to set the one we want...

PR #2022, opened 2017-06-12
mount /sys in getty container and add and fix kernel tests

### LINKED ISSUE #2019: mount securityfs on boot failed
**Description**

when run "mount | grep "securityfs", " it is null. So there maybe is some issues for the rc.local, or else.

justincormack:
Mount /sys in the `getty` container to fix #2019

Add some tests for the filesystems we expect to find.

Fix the kernel config tests so they actually fail when they fail...


justincormack:
Failures a real as I fixed the tests, there are some 4.11 module changes

justincormack:
ok, two config options were renamed in 4.11, and one now seems to be off as presumably has preconditions, will do a different PR for these and rebase

rn:
Ah, should have maybe caught those when doing the 4.11 port. Not sure I did a detailed diff when I updated

justincormack:
Its actually very confusing, trying to work out exactly what did change...

justincormack:
Ok, this has passed as I pushed the new 4.11

test/pkg/kernel-config/check-kernel-config.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
+function fail {
```
riyazdf: 👍  thanks for this!

test/pkg/kernel-config/check-kernel-config.sh
```diff
@@ -19,59 +24,116 @@ kernelMinor="${kernelMinor%%.*}"
...
 fi
 
 if [ "$kernelMajor" -ge 4 -a "$kernelMinor" -ge 10 ]; then
-  echo $UNZIPPED_CONFIG | grep -q CONFIG_BUG_ON_DATA_CORRUPTION=y || (echo "CONFIG_BUG_ON_DATA_CORRUPTION=y" && exit 1)
+  echo $UNZIPPED_CONFIG | grep -q CONFIG_BUG_ON_DATA_CORRUPTION=y || fail "CONFIG_BUG_ON_DATA_CORRUPTION=y"
+fi
+
+if [ "$kernelMajor" -ge 4 -a "$kernelMinor" -le 10 ]; then
+  echo $UNZIPPED_CONFIG | grep -q CONFIG_DEBUG_RODATA=y || fail "CONFIG_DEBUG_RODATA=y"
+  echo $UNZIPPED_CONFIG | grep -q CONFIG_DEBUG_SET_MODULE_RONX=y || fail "CONFIG_DEBUG_SET_MODULE_RONX=y"
```
riyazdf: 👍

PR #2025, opened 2017-06-12
Remove initrd library, it has moved to moby tool

justincormack:
As we are not using it here, makes more sense for it to go there.

Moved with history in https://github.com/moby/tool/pull/78

kitten


riyazdf:
Actually, before merging this, should we also bump the moby commit to after this?
 https://github.com/moby/tool/pull/78

justincormack:
It doesnt matter, it was vendored into moby/tool anyway

riyazdf:
@justincormack ah good point.  Let's merge this then 👍

PR #2027, opened 2017-06-12
Use the user's ssh key automatically in ssh examples

justincormack:
This requires moby tool update to support `~` in paths, but
makes everything much nicer.

### DIFF
--- a/examples/sshd.yml
+++ b/examples/sshd.yml
@@ -22,7 +22,9 @@ services:
     image: "linuxkit/sshd:abc1f5e096982ebc3fb61c506aed3ac9c2ae4d55"
 files:
   - path: root/.ssh/authorized_keys
-    contents: '#your ssh key here'
+    source: ~/.ssh/id_rsa.pub
+    mode: "0600"
+    optional: true
 trust:
   org:
     - linuxkit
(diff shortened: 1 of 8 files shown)


examples/sshd.yml
```diff
@@ -22,7 +22,8 @@ services:
     image: "linuxkit/sshd:abc1f5e096982ebc3fb61c506aed3ac9c2ae4d55"
 files:
   - path: root/.ssh/authorized_keys
-    contents: '#your ssh key here'
+    source: ~/.ssh/id_rsa.pub
```
riyazdf: I think CI is failing to build this example because it can't find this on its host?

justincormack: hmm, not sure whats best to do about that...

justincormack: maybe will fake it for the tests

riyazdf: faking for the test SGTM

ijc:
Perhaps support something like `only-if-present: true` or maybe `conditional-source: ~/.ssh/foo`?

justincormack:
Added `optional` flag in https://github.com/moby/tool/pull/82 and amended

justincormack:
Ok, this is passing tests now

PR #2030, opened 2017-06-12
docs: update security events with new kernels

riyazdf:
update these docs to reflect that the kernel bumps fix them.  Changed around categories to better fit the scope of LinuxKit.  cc @tych0 

</img>


tych0:
LGTM, I verified that the fixes for these all went into stable.

Issue #2031, opened 2017-06-13
project/kubernetes: master always NotReady, failure to configure CNI/weave?


ijc:
Running d310ebf44ff000c9b9e90542bbd0a62d386be3e1 with a patch to avoid stale local images:
```diff
diff --git a/projects/kubernetes/Makefile b/projects/kubernetes/Makefile
index 05280281..cb1c297b 100644
--- a/projects/kubernetes/Makefile
+++ b/projects/kubernetes/Makefile
@@ -20,10 +20,10 @@ push-container-images: build-container-images cache-images
 build-vm-images: kube-master-initrd.img kube-node-initrd.img
 
 kube-master-initrd.img: kube-master.yml
-	../../bin/moby build -name kube-master kube-master.yml
+	../../bin/moby build --pull -name kube-master kube-master.yml
 
 kube-node-initrd.img: kube-node.yml
-	../../bin/moby build -name kube-node kube-node.yml
+	../../bin/moby build --pull -name kube-node kube-node.yml
 
 clean:
 	rm -f -r \
```

I am running, from `projects/kubernetes`:
```
$ git clean -fdx . && make build-container-images && make build-vm-images && ./boot-master.sh 
```
(I think `build-container-images` is redundant given the use of `--pull`, but I left it in in case I am wrong)

Then on the serial console I am running:
```
# runc exec --tty kubelet ash -l
```
(I'm doing this in preference to using `ssh_into_kubelet.sh` since this works on both MacOS and Linux)

Then:
```
# kubeadm-init.sh
[kubeadm] WARNING: kubeadm is in beta, please do not use it for production clusters.
[init] Using Kubernetes version: v1.6.1
[init] Using Authorization mode: RBAC
[preflight] Skipping pre-flight checks
[preflight] No supported init system detected, won't ensure kubelet is running.
[certificates] Generated CA certificate and key.
[certificates] Generated API server certificate and key.
[certificates] API Server serving cert is signed for DNS names [moby kubernetes kubernetes.default kubernetes.default.svc kubernetes.default.svc.cluster.local] and IPs [10.96.0.1 192.168.65.5]
[certificates] Generated API server kubelet client certificate and key.
[certificates] Generated service account token signing key and public key.
[certificates] Generated front-proxy CA certificate and key.
[certificates] Generated front-proxy client certificate and key.
[certificates] Valid certificates and keys now exist in "/etc/kubernetes/pki"
[...]

PR #2032, opened 2017-06-13
Small improvements to projects/kubernetes

ijc:
**- What I did**

Some small updates to `projects/kubernetes`:
* kubernetes: update to latest init
* kubernetes: Combine boot-{master,node}.sh into a single script
* kubernetes: give each instance a specific state directory
* kubernetes: Create disk in state dir

and some improvements to the qemu runner:
* Add state dir support to Qemu runner.
* Add metadata support to Qemu runner.

The scripting can now be used to boot master & nodes on both Linux and MacOS.

**- How I did it**

**- How to verify it**

After building `./boot.sh` and `./boot.sh 0 -join wibble` should boot on both MacOS and Linux, although due to #2031 I've not been able to fully test that. Note that on Linux the node is not yet expected to be able to contact the master, since I haven't sorted out the networking side yet.

Minor updates to projects/kubernetes, improved support in LinuxKit qemu runner.

/cc @errordeveloper

### DIFF
--- a/projects/kubernetes/README.md
+++ b/projects/kubernetes/README.md
@@ -13,7 +13,7 @@ make build-vm-images
 
 Boot Kubernetes master OS image using `hyperkit` on macOS:
 ```
-./boot-master.sh
+./boot.sh
 ```
 
 Get IP address of the master:
@@ -36,12 +36,12 @@ and try `kubectl get nodes` from within the master.
 
 To boot a node use:
 ```
-./boot-node.sh <n> [<join_args> ...]
+./boot.sh <n> [<join_args> ...]
 ```
 
 More specifically, to start 3 nodes use 3 separate shells and run this:
 ```
-shell1> ./boot-node.sh 1 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
-shell2> ./boot-node.sh 2 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
-shell3> ./boot-node.sh 3 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
+shell1> ./boot.sh 1 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
+shell2> ./boot.sh 2 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
+shell3> ./boot.sh 3 --token bb38c6.117e66eabbbce07d 192.168.65.22:6443
 ```
--- a/projects/kubernetes/boot-master.sh
+++ b/projects/kubernetes/boot-master.sh
@@ -1,5 +1,0 @@
-#!/bin/bash -eu
-disk="kube-master-disk.img"
-set -x
-rm -f "${disk}"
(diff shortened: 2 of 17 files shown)


justincormack:
@ijc the `mkimage` failure looks like the state directory cleanup needs fixing.

justincormack:
Other than that LGTM when tests green

ijc:
@justincormack that sounds plausible, but OOI how did you determine that? [The logs](https://linuxkit.datakit.ci/log/saved/linuxkit-build-of-13a94c7a3c2a5777e9bc48764f2f67e9109f999d-default-ci-pr/5665ca4ece435a83ea678cc0487916d73c865c02) don't seem to contain anything other than the `[FAIL    ] linuxkit.packages.mkimage 194.11s`. Am I failing to find some more enlightening source of information?

justincormack:
Full logs are in the link at the top of the logs page once the run finishes, in this case at https://github.com/linuxkit/linuxkit-logs/commit/5665ca4ece435a83ea678cc0487916d73c865c02

ijc:
Ah, I read "Loaded from commit 5665ca4ece435a83ea678cc0487916d73c865c02" as meaning the test run had started with that state, rather than that being where the log was loaded from!

ijc:
Hmph. `linuxkit.packages.mkimage` works for me locally and I can't really see how:
```
+[STDERR  ] 2017-06-13T11:02:07.489584158Z: + find . -iname mkimage* -not -iname *.yml -exec rm -rf {} ;
+[STDERR  ] 2017-06-13T11:02:07.495564602Z: find: `./mkimage-state': No such file or directory
```
Could ever happen, unless there is some other process racing with it.

Unlike in the logs the quoting is correct in the actual test `find . -iname "mkimage*" -not -iname "*.yml" -exec rm -rf {} \;`.

Something quite exciting must be going on I think.

justincormack:
shouldnt it be `find. -iname 'mkimage*'` with single quotes?

ijc:
I think either sort of quoting ought to be ok for this case, globs aren't expanded in either.

I may have forgotten to type `make` after switching back to this branch and before running the tests, so maybe I will be able to repro after all...

ijc:
Yes, I can repro after all, will figure it out...

rn:
I also noticed this when I moved some of the tests to work with HyperKit and "fixed" it by doing a `|| true` after the find in the clean up function #1982

ijc:
`find` is trying to recurse into `mkimage-state` after it has traversed (and remove it)...

I think the right answer is to add `-depth` to the `find`, which I'll do here.

I think it might also be worth making the qemu runner only make the state dir if it has something to put there, in this case it's just creating an empty dir, it's also leaking `qcow2.disk-state`...

justincormack:
Ah that makes sense

ijc:
I'm just testing a patch to not create the dir if it isn't needed, I expect I'll push it before the CI finishes, but if not whoever is thinking of merging might want to wait for one more commit to show up.

PR #2036, opened 2017-06-13
copy setsid to setsidu so it does not get overwritten by busybox when…

deitch:
… used in init

**- What I did**
* Ensure correct `setsid` is available to `rungetty.sh`
* Remove unused but breaking `/etc/inittab`

**- How I did it**

1. `cp /usr/bin/setsid /usr/bin/setidu`
2. In `rungetty.sh`, use `setsidu`
3. In `Dockerfile`, do `/bin/rm -f /out/etc/inittab`

In `services` mode, this is unneeded but still works. In `init` mode, this is needed because any other container may have the original alpine `setsid` which is `setsid -> /bin/busybox`, which does not work.
**- How to verify it**
Run it in each of `services` mode and `init` mode. For `init`, check it as first or any later order.

Lock down correct `setsid` binary.

As discussed in thread with @justincormack

### DIFF
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -9,6 +9,17 @@ RUN apk add --no-cache --initdb -p /out \
     util-linux \
     && true
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+#
+# We require a version of `setsid(1)` which supports the `-w`
+# option, which is not available in all implementations (e.g. the
+# `busybox` implementation does not support it). When this is run
+# as part of a LinuxKit `init` image (rather than as a standalone
+# container) we cannot guarantee which version of `setsid` will
+# be present once the layers are combined, so we take a copy of
+# our own, known good, version for use later.
+RUN cp /out/usr/bin/setsid /out/usr/bin/setsid.getty
+# we really do not want a rogue inittab here
+RUN rm -rf /out/etc/inittab
 
 FROM scratch
 ENTRYPOINT ["/sbin/tini","-s","-v","--"]
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -7,12 +7,11 @@ infinite_loop() {
 	done
 }
 
-# run getty on all known consoles - except those already in inittab
+# run getty on all known consoles
 start_getty() {
 	tty=${1%,*}
 	speed=${1#*,}
-	inittab="$2"
-	securetty="$3"
+	securetty="$2"
 	line=
 	term="linux"
 	[ "$speed" = "$1" ] && speed=115200
@@ -39,7 +38,7 @@ start_getty() {
 		echo "$tty" >> "$securetty"
 	fi
 	# respawn forever
-	infinite_loop setsid -w /sbin/getty $loginargs $line $speed $tty $term &
+	infinite_loop setsid.getty -w /sbin/getty $loginargs $line $speed $tty $term &
 }
 
 # check if we have /etc/getty.shadow
@@ -53,7 +52,7 @@ fi
 for opt in $(cat /proc/cmdline); do
 	case "$opt" in
 	console=*)
-		start_getty ${opt#console=} /etc/inittab /etc/securetty
+		start_getty ${opt#console=} /etc/securetty
 	esac
 done
 


pkg/getty/Dockerfile
```diff
@@ -9,6 +9,9 @@ RUN apk add --no-cache --initdb -p /out \
     util-linux \
     && true
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN cp /out/usr/bin/setsid /out/usr/bin/setsidu
```
justincormack: can you just add a comment about why this is.

We should be able to remove it once `init` no longer uses `busybox` which is not that far off.

deitch: > can you just add a comment about why this is

Added lines of comment. :-)

> We should be able to remove it once init no longer uses busybox which is not that far off

Maybe. People can compose `init:` out of any containers they want. If any other uses busybox - not unreasonable - and it is later in the sequence than `getty`, it will kill it. It might be a good idea to leave it no matter what. We are guaranteed reasonably fairly this way that it will stay around. 

I think there is a bigger issue that `init:` images can trounce each other, but container images are built with the assumption that it is entirely self-contained. Probably should be a better way to handle how inits merge.

pkg/getty/Dockerfile
```diff
@@ -9,6 +9,18 @@ RUN apk add --no-cache --initdb -p /out \
     util-linux \
     && true
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+#
+# when run standalone as a regular or services container, the below is unnecessary
+#  when run as a linuxkit init, all files are copied over
+#  if a different image is later in the sequence, that one may use busybox,
+#    which symlinkx setsid -> /bin/busybox. That would override setsim and cause this to fail.
```
ijc: "setsim" is a typo I think. Is "symlinkx" too?

ijc:
Where does the name `setsidu` come from? Is the `u` suffix supposed to stand for something? Perhaps `setsid.getty` would make it clearer this is part of the `getty` container?

Do I understand correctly that the underlying issue here an incompatibility between the `busybox` and `linux-utils` versions of `setsid`? Specifically is it the lack of the `-w` in the `busybox` version? It would be good to call out `-w` as the issue in the comment. Or maybe it is possible to either get `-w` added to the `busybox` alpine version or to rework the user to not need it?

pkg/getty/Dockerfile
deitch: Not at all. Are you not familiar with that obscure dialect of Medieval English wherein words are pluralized by adding an "x" instead of an "s"? And this is a little-known variant on `setsid` that also sets your SIM card number....

deitch: Fixed. :-)

deitch:
> Where does the name setsidu come from? Is the u suffix supposed to stand for something? Perhaps setsid.getty would make it clearer this is part of the getty container?

Yes. I cannot recall where I got adding `u`. `setsid.getty` is fine, happy to change.

> Do I understand correctly that the underlying issue here an incompatibility between the busybox and linux-utils versions of setsid? Specifically is it the lack of the -w in the busybox version?

Correct, although I don't know if there are other incompatibilities.

> It would be good to call out -w as the issue in the comment. 

Good. Doing it now.

> Or maybe it is possible to either get -w added to the busybox alpine version or to rework the user to not need it?

We might, although someone always could stack some `init` images that have some other variant. If we are using a particular command that has specific requirements, we need to isolate it. Normally, containers do that, but in `init`, it stacks them, so we cannot rely on it, unless `moby build` gains some way of resolving conflicts.

deitch:
All issues raised by @ijc fixed.

pkg/getty/Dockerfile
```diff
@@ -9,6 +9,19 @@ RUN apk add --no-cache --initdb -p /out \
...
+# when run standalone as a regular or services container, the below is unnecessary
+#  when run as a linuxkit init, all files are copied over
+#  if a different image is later in the sequence, that one may use busybox,
+#    which symlinks `setsid -> /bin/busybox`. That would override `setsid` and cause this to fail,
+#    since `setsid` from busybox lacks the `-w` option, which `linux-utils` has.
+#  we cp setsid to a normally unused location so we are guaranteed it sticks around,
+#    and then reference it by the unique `setsid.getty` name in `rungetty.sh`.
+#  if linuxkit.yml init were composed entirely of images that do not override `setsid`, we
+#    could remove this line and use just `setsid` in `rungetty.sh`
+RUN cp /out/usr/bin/setsid /out/usr/bin/setsid.getty
```
ijc: I think you can simplify this explanation a bit e.g. :
```
# We require a version of `setsid(1)` which supports the `-w`
# option, which is not available in all implementations (e.g. the
# `busybox` implementation does not support it). When this is run
# as part of a LinuxKit `init` image (rather than as a standalone
# container) we cannot guarantee which version of `setsid` will
# be present once the layers are combined, so we take a copy of
# our own, known good, version for use later.
```

deitch: Sure, I don't care. Just copy-paste.

deitch: Changed

ijc:
LGTM.

deitch:
@justincormack or @riyazdf can you push out the updated image? Or do I have delegated rights?

riyazdf:
@deitch I can push a new image and open a PR for updating ymls 👍

PR #2037, opened 2017-06-13
add a static usermode helper

### LINKED ISSUE #1760: write/use a safe usermodehelper
Linux 4.11 has a much safer mechanism for the user mode helpers. We should use this.

```
read-only usermodehelper

A common way attackers use to escape confinement is by rewriting the user-mode helper sysctls
 (e.g. /proc/sys/kernel/modprobe) to run something of their choosing in the init namespace. To 
reduce attack surface within the kernel, Greg KH introduced CONFIG_STATIC_USERMODEHELPER,
 which switches all user-mode helper binaries to a single read-only path (which defaults to 
/sbin/usermode-helper). Userspace will need to support this with a new helper tool that can 
demultiplex the kernel request to a set of known binaries.
```

cc @riyazdf @tych0

tych0:
The binary is used in tandem with CONFIG_STATIC_USERMODEHELPER=y in 4.11+,
see the big comment in the binary for the current whitelist of binaries.

Should fix #1760.


pkg/init/Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM alpine:edge AS build
```
rn: this should use `linuxkit/alpine:<latest hash>` just like the next `FROM` line. the alpine base image should have all the tools you need to compile it.

pkg/init/Makefile
```diff
@@ -8,7 +8,7 @@ DEPS=Dockerfile init $(wildcard etc/*) $(wildcard etc/init.d/*)
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 
 tag: $(DEPS)
-	docker build --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
```
rn: If you use the alpine base image you don't need network

pkg/init/usermode-helper.c
```diff
@@ -0,0 +1,36 @@
...
+		 *    (but e.g. systemd needs, if anyone ever tries to boot
+		 *    that on linuxkit)
+		 *  - /sbin/request-key, which we don't provide
+		 *  - on x86, machine check
+		 *
+		 * Today we only call mdev and modprobe, but as we add more
+		 * features to linuxkit this whitelist may need changing (or a
+		 * policy, like always allow stuff in /sbin).
+		 */
+		exit(2);
```
rn: should we log the binary name (or is this already logged?)

kernel/kernel_config-4.11.x
```diff
@@ -3560,7 +3560,8 @@ CONFIG_HAVE_HARDENED_USERCOPY_ALLOCATOR=y
 CONFIG_HAVE_ARCH_HARDENED_USERCOPY=y
 CONFIG_HARDENED_USERCOPY=y
 # CONFIG_HARDENED_USERCOPY_PAGESPAN is not set
-# CONFIG_STATIC_USERMODEHELPER is not set
+CONFIG_STATIC_USERMODEHELPER=y
+CONFIG_STATIC_USERMODEHELPER_PATH="/sbin/usermode-helper"
```
rn: does this need to be hard coded here or is there a sysctl/syfs entry we can use? Seems weird to have this in the kernel config...

pkg/init/Dockerfile
tych0: Right, I'll update thanks.

pkg/init/usermode-helper.c
```diff
@@ -0,0 +1,36 @@
...
+
+int main(int argc, char *argv[])
+{
+	if (!strcmp(argv[0], "/sbin/mdev")) {
+		/* busybox uses /sbin/mdev for early uevent bootstrapping */
+		execv(argv[0], argv);
+	} else if (!strcmp(argv[0], "/sbin/modprobe")) {
+		/* allow modprobe */
+		execv(argv[0], argv);
+	} else {
```
rn: we probably want to add `/sbin/reboot` and `/sbin/poweroff`as this is used by `orderly_reboot() ` and `orderly_poweroff()` in `kernel/reboot.c` and on Hyper-V this can be triggered from the host (see ` drivers/hv/hv_util.c`

pkg/init/usermode-helper.c
tych0: It's not logged by the kernel as far as I can tell. Where/how would you like to log it?

rn: maybe fprintf(stderr, ) for now? I don't know where that will end up :)

kernel/kernel_config-4.11.x
tych0: In fact this hard coding is the bit that's designed to prevent exploits: people used to rewrite the sysctl for dynamic binaries like `modprobe` to something that they controlled. By hard coding it, they have to overwrite the binary itself instead.

tych0: see https://outflux.net/blog/archives/2017/05/02/security-things-in-linux-v4-11/

pkg/init/usermode-helper.c
tych0: Good catch, thanks!

rn: some residual memory from a day or two of debugging/cursing a year or so ago :)

kernel/kernel_config-4.11.x
rn: Cool, yes, that makes a lot of sense

pkg/init/Dockerfile
tych0: Hmm. It doesn't seem to have make or gcc?

```
~/packages/linuxkit/pkg/init usermode-helper make
docker build --no-cache --network=none -t linuxkit/init:8ea972f0537944e1cf96d4f914d82bea0f9fe62c .
Sending build context to Docker daemon  18.94kB
Step 1/16 : FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
 ---> 72ff54f67634
Step 2/16 : ADD usermode-helper.c .
 ---> 3fba2922bf2b
Removing intermediate container 50e00ff17303
Step 3/16 : RUN make usermode-helper
 ---> Running in 992e606cfe3f
/bin/sh: make: not found
The command '/bin/sh -c make usermode-helper' returned a non-zero code: 127
Makefile:11: recipe for target 'tag' failed
make: *** [tag] Error 127
~/packages/linuxkit/pkg/init usermode-helper 2 ls
bin/  Dockerfile  etc/  init*  Makefile  usermode-helper.c
~/packages/linuxkit/pkg/init usermode-helper vi Makefile 
~/packages/linuxkit/pkg/init usermode-helper vi Dockerfile 
~/packages/linuxkit/pkg/init usermode-helper make
docker build --no-cache --network=none -t linuxkit/init:8ea972f0537944e1cf96d4f914d82bea0f9fe62c .
Sending build context to Docker daemon  18.94kB
Step 1/16 : FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
 ---> 72ff54f67634
Step 2/16 : ADD usermode-helper.c .
 ---> a25aa94cf323
Removing intermediate container 537002280a38
Step 3/16 : RUN gcc usermode-helper.c -o usermode-helper
 ---> Running in 69826310ac85
/bin/sh: gcc: not found
The command '/bin/sh -c gcc usermode-helper.c -o usermode-helper' returned a non-zero code: 127
Makefile:11: recipe for target 'tag' failed
make: *** [tag] Error 127
```

rn: it does, but it's in a APK mirror. You still need to `apk add` them as in:
```
FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS build
RUN apk add --no-cache --initdb -p /out alpine-baselayout make gcc
...
```
[note to self that I need to rewrite the packages document...]

tych0: when I use `-p out`, `RUN` still doesn't pick up the commands; I dropped that, which I think is fine since this is just a stage in a multi-stage build, so it won't bloat the final image.

pkg/init/usermode-helper.c
tych0: It doesn't end up anywhere that I could figure out, when debugging this i just logged to a file in /tmp. I'll see if I can figure out where it goes.

pkg/init/Dockerfile
```diff
@@ -1,3 +1,10 @@
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS build
+# RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
```
riyazdf: remove unused line?  Though it seems like we use this in other images so I'm wondering if we need this here

tych0: Yep, sorry about that.

rn: sorry, partly my fault. this line is not needed. It's only needed when you build a new root fs for a package. since we are not doing this in this stage it is not needed

pkg/init/Dockerfile
```diff
@@ -13,3 +19,4 @@ COPY --from=mirror /out/ /
 COPY init /
 COPY etc etc/
 COPY bin bin/
+COPY --from=build usermode-helper /sbin/
```
rn: I would move that higher to have the `COPY` lines from other stages grouped together.

rn:
could you squash the commits?

pkg/init/Dockerfile
tych0: Fixed, thanks.

tych0:
Yep, just squashed them.

pkg/init/Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS build
+RUN apk add --no-cache --initdb alpine-baselayout make gcc musl-dev
+
+ADD usermode-helper.c .
+RUN make usermode-helper
```
rn: how does this work without Makefile in the PR?

pkg/init/usermode-helper.c
ijc: Does `syslog(3)` end up somewhere useful?

pkg/init/usermode-helper.c
```diff
@@ -0,0 +1,40 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+int main(int argc, char *argv[])
+{
+	if (!strcmp(argv[0], "/sbin/mdev")) {
+		/* busybox uses /sbin/mdev for early uevent bootstrapping */
+		execv(argv[0], argv);
```
ijc: For each of these it would be good for debugability if it would log the error if the `execv` returns.

tych0:
Yes, absolutely. I just added usermode-helper.c to the makefile deps

pkg/init/Dockerfile
tych0: using make's "implicit targets" trick, https://www.gnu.org/software/make/manual/html_node/Implicit-Rules.html

You can do it with any .c file in any directory, assuming that file doesn't need any special compiler or linker args.

pkg/init/usermode-helper.c
tych0: Not as far as I can tell. Should it go somewhere besides dmesg in linuxkit?

ijc: Oh, there doesn't seem to be a syslog, so no perhaps not. Not sure if this is what `projects/logging` is looking to address or not.

pkg/init/Dockerfile
rn: ah, I learned something today

pkg/init/usermode-helper.c
rn: Maybe we can add `fprintf(stderr,...)` for now and update to something else later.

tych0: Ok, done.

pkg/init/usermode-helper.c
```diff
@@ -0,0 +1,51 @@
...
+
+	fprintf(stderr, "usermodehelper: ");
+	for (i = 0; i < argc; i++) {
+		fprintf(stderr, "%s ", argv[i]);
+	}
+	fprintf(stderr, "\n");
+
+	if (!strcmp(argv[0], "/sbin/mdev")) {
+		/* busybox uses /sbin/mdev for early uevent bootstrapping */
+		execv(argv[0], argv);
```
rn: shouldn't we catch and log if `execv` errors? @ijc commented on this earlier...

pkg/init/usermode-helper.c
```diff
@@ -0,0 +1,51 @@
...
+		 *    (but e.g. systemd needs, if anyone ever tries to boot
+		 *    that on linuxkit)
+		 *  - /sbin/request-key, which we don't provide
+		 *  - on x86, machine check
+		 *
+		 * Today we only call mdev and modprobe, but as we add more
+		 * features to linuxkit this whitelist may need changing (or a
+		 * policy, like always allow stuff in /sbin).
+		 */
+		exit(2);
```
rn: and log the command we denied

pkg/init/usermode-helper.c
rn: just spotted the `perror` at the end and with the logging of the command line options that should provide enough info for debug

pkg/init/usermode-helper.c
riyazdf: cc @MagnusS ^

MagnusS: [projects/logging](https://github.com/linuxkit/linuxkit/tree/master/projects/logging) contains an in-memory logging daemon that could capture the logs. I've not made it syslog compatible (yet) as it captures data from fd's passed to it (typically stderr/stdout from onboot/service/dockerd containers) so we don't have to bind mount /dev/log, but if we need it we could add it. It's a bit hard to enable at the moment unfortunately as it needs to run as a service before the onboot containers, which requires a modified `init`

pkg/init/usermode-helper.c
justincormack: I think its ok to ignore logging for now, but maybe we can add a comment in saying that it may be useful to add later.

justincormack: Or log to stdout like the `perror` below, and add a note saying stdout should be redirected to a log in future.

tych0: Yep, we are logging the command (and error, in case of exec failure). I just added a comment pointing out that the output doesn't go anywhere useful now, but once we have syslog, it should probably go there.

rn:
once this and #2047 is merged I'll kick off a kernel build/push

rn:
well, and after I've updated to 4.11.5, 4.9.32, 4.4.72

tych0:
On Thu, Jun 15, 2017 at 11:04:05AM -0700, Rolf Neugebauer wrote:
> once this and #2047 is merged I'll kick off a kernel build/push

We'll need an init push too, otherwise 4.11 will probably fail to boot
(or the devices will be all screwed up, at least).

PR #2039, opened 2017-06-14
Docker for Mac blueprint

MagnusS:
Adds a blueprint for the open source components of Docker for Mac. It includes support for VPNKit networking and port forwarding to the host. Docker can be controlled via a unix domain socket in the linuxkit state directory.

### DIFF
--- a/blueprints/docker-for-mac.yml
+++ b/blueprints/docker-for-mac.yml
@@ -1,0 +1,89 @@
+# This is a blueprint for building the open source components of Docker for Mac
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:2599bcd5013ce5962aa155ee8929c26160de13bd
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
+  - name: sysfs
+    image: "linuxkit/sysfs:1244c5a86dfa2318c4e304af68d37e12367e1b7f"
+  - name: binfmt
+    image: "linuxkit/binfmt:8ac5535f57f0c6f5fe88317b9d22a7677093c765"
+  # Format and mount the disk image in /var/lib/docker
+  - name: format
+    image: "linuxkit/format:180cb2dc1de5e60373385080f8148abf10a3afac"
+  - name: mount
+    image: "linuxkit/mount:ff5338822f20375b8913f5a80f9ed4f6ea9a592b"
+    command: ["/mount.sh", "/var/lib/docker"]
+  # mount-vpnkit mounts the 9p share used by vpnkit to coordinate port forwarding
+  - name: mount-vpnkit
+    image: "alpine:3.6"
+    binds:
+        - /var/:/host_var:rbind,rshared
+    capabilities:
+        - CAP_SYS_ADMIN
+    rootfsPropagation: shared
+    command: ["sh", "-c", "mkdir -p /host_var/vpnkit/port && mount -v -t 9p -o trans=virtio,dfltuid=1001,dfltgid=50,version=9p2000 port /host_var/vpnkit"]
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+services:
+  # Enable getty for easier debugging
+  - name: getty
+    image: "linuxkit/getty:ef9d667af71089326419fb08e9cc9d567cf15748"
+    env:
+        - INSECURE=true
+  - name: rngd
+    image: "linuxkit/rngd:1fa4de44c961bb5075647181891a3e7e7ba51c31"
+  # Run ntpd to keep time synchronised in the VM
+  - name: ntpd
+    image: "linuxkit/openntpd:45deeb05f736162d941c9bf494983f655ab80aa5"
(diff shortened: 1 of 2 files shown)


MagnusS:
Updated with comments in the blueprint and fixed minor link issue in README

blueprints/docker-for-mac.yml
```diff
@@ -0,0 +1,90 @@
...
+  - name: binfmt
+    image: "linuxkit/binfmt:8ac5535f57f0c6f5fe88317b9d22a7677093c765"
+  # Format and mount the disk image in /var/lib/docker
+  - name: format
+    image: "linuxkit/format:180cb2dc1de5e60373385080f8148abf10a3afac"
+  - name: mount
+    image: "linuxkit/mount:ff5338822f20375b8913f5a80f9ed4f6ea9a592b"
+    command: ["/mount.sh", "/var/lib/docker"]
+  # mount-vpnkit mounts the 9p share used by vpnkit to coordinate port forwarding
+  - name: mount-vpnkit
```
rn: would it be worth it making this a package? Maybe even a stand-alone Go binary without alpine? Not for now but in the future?

blueprints/docker-for-mac.yml
```diff
@@ -0,0 +1,90 @@
...
+    capabilities:
+        - CAP_SYS_ADMIN
+    rootfsPropagation: shared
+    command: ["sh", "-c", "mkdir -p /host_var/vpnkit/port && mount -v -t 9p -o trans=virtio,dfltuid=1001,dfltgid=50,version=9p2000 port /host_var/vpnkit"]
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+services:
+  # Enable getty for easier debugging
+  - name: getty
```
rn: do we need getty for docker for mac?

blueprints/docker-for-mac.yml
```diff
@@ -0,0 +1,90 @@
...
+    mounts:
+     - type: cgroup
+       options: ["rw","nosuid","noexec","nodev","relatime"]
+    binds:
+        - /var/lib/docker:/var/lib/docker
+        - /lib/modules:/lib/modules
+        - /var/vpnkit:/port
+        - /var/run:/var/run
+        - /var/config/docker:/var/config/docker
+        - /containers/services/vpnkit-forwarder/rootfs/vpnkit-expose-port:/usr/bin/vpnkit-expose-port
```
rn: oh, this part is a bit ugly as it assumes something about another containers rootfs...

blueprints/docker-for-mac.yml
MagnusS: Yes, we could do that - that was actually the first version of this. But the 9p mount will probably go away at some point and be replaced with vsock connection so this seemed to be sufficient for now as we may not need a separate package in the future.

blueprints/docker-for-mac.yml
MagnusS: Not really - it's there for debugging. We could remove it

blueprints/docker-for-mac.yml
ijc: I think needing to bind something from one service/container to another is going to be quite common, more normally I'd expect it to be for control socket type things (I may have a need for that in `projects/swarmd` before too long) rather than binaries as it is here.

rn: the plan is/was to use one of the native containerd storage backends for container images and then you won't be able to do bind mounting of binaries like this anymore. Maybe add it to the init section and then you can bind mount from the root fs

blueprints/docker-for-mac.yml
rn: pre-coffee reviewing and I missed the comment above. It's fine to have it in

blueprints/docker-for-mac.yml
MagnusS: Yes... It avoids creating a separate `docker-dfm` package that just pulls in the `vpnkit-expose-port` binary in a multi-stage build. An alternative is to do the same as in the prototype: https://github.com/MagnusS/linuxkit/blob/editions-wip/projects/editions/pkg/docker-dfm/Dockerfile (slirp-proxy is now vpnkit-expose-port from vpnkit-forwarder package). But bind mounting the binary is simpler until the image format changes...

blueprints/docker-for-mac.yml
rn: thanks for clarifying

blueprints/docker-for-mac.yml
ijc: > you won't be able to do bind mounting of binaries like this anymore.

My point is that we need a solution for this use case.

rn: @ijc understood. I suspect the `init` section containers won't be in the containerd storage. So using that would allow you to use packages to add files to the rootfs

MagnusS: @rn for the binary use case it would be nice if there was a way that didn't require creating an additional container as a dependency. Maybe it could be possible to e.g. specify a bind mount from a path in another service and translate that into the actual paths when building the image? Something like `- other-service:/bin/file:/bin/file`

justincormack: These paths are not part of the API and will change, so no you can't do this. Its fine to add the `vpnkit-expose-port` to the current `docker-ce` package for now.

MagnusS: @justincormack Ok - PR for docker-ce in #2048

MagnusS:
Updated `with linuxkit/docker-ce:dda71ff9fe5ebbfa794b98c57c32df286b212848` to avoid bind mounting `vpnkit-expose-port`

justincormack:
Looks fine for a first version.

justincormack:
We should look at the clock sync - we are using chrony not openntpd, plus there is the clock reset hack...

justincormack:
Can we add a build test at least for this (we dont have OSX CI yet). Can be in a new PR.

MagnusS:
@justincormack yes, this is just an initial version - it's still missing a few things, such as better time sync, iptables workaround for swarm and trim support

PR #2040, opened 2017-06-14
sdk: replace custom transport protocol by Capnproto

samoht:
Initial patch by @talex5.

The patch replace the custom-made transport protocol by [cap'n proto](https://capnproto.org/). We are interested by Cap'n proto because of its security features backed in its [RPC protocol](https://capnproto.org/rpc.html). 

I have verified that the image builds and that the `mirage-dhcp.yml` example gets a lease pretty quickly using the hyperkit backend.


projects/miragesdk/src/Makefile
```diff
@@ -1,39 +1,25 @@
...
+DEPS=Dockerfile \
+	$(shell find . -name jbuild) \
+	$(shell find . -name '*.ml') \
+	$(shell find . -name '*.mli') \
+	$(shell find . -name '*.c') \
+	$(shell find . -name '*.json')
 
-tag: .build
-	docker tag $(IMAGE) linuxkitprojects/$(IMAGE):$(shell cat hash)
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
```
rn: strictly speaking, this is not correct as the `Dockerfile` seems to pull in things like `alpine:3.5` from the network so the package build is not entirely dependent on the content of the directory. This is more like an FYI, ie something which needs addressing at some point

PR #2041, opened 2017-06-14
Custom containerd client for use in init.

ijc:
**- What I did**

Latest `ctr` has removed the `--runtime-config` option which we rely on. `ctr` is not (considered by containerd devs) to be a supported interface (it's more of a debug tool).

This PR replaces `/etc/init.d/010-containerd`'s use of `ctr` with a simple custom client written against the containerd client library.

**- How I did it**

The interesting commit message is:

>    init: replace ctr with a custom client using the containerd client library
>    
>    Currently it supports only `service start <SERVICE>`, but it could grow e.g.
>    `stop`, `exec` etc in the future (although you can still use `ctr` for those).
>    
>    I wasn't able yet to figure out how to make containerd do the rw bind step:
>        /bin/mount --bind "$f/rootfs" "$f/rootfs"
>        mount -o remount,rw "$f/rootfs"
>    So that remains for now.
>    
>    In order to be able to use go-compile.sh the containerd build needs to move
>    from /root/go to /go as the GOPATH.
>    
>    The vendoring situation is not ideal, but since this tool wants to be an exact
>    match for the containerd it seems tollerable to reuse its vendoring.
>    
>    Signed-off-by: Ian Campbell <ian.campbell@docker.com>

**- How to verify it**

Everything should just continue working.

Switched away from `ctr` usage in init.

### DIFF
--- a/pkg/containerd/cmd/service/skanky-vendor.sh
+++ b/pkg/containerd/cmd/service/skanky-vendor.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+#
+# We only need the containerd client and its transitive dependencies
+# and we conveniently have a checkout already. We actually prefer to
+# reuse containerd's vendoring for consistency anyway.
+
+set -eu
+ctrd=$1
+cp -r $ctrd/vendor/ vendor/
+# We need containerd itself of course
+mkdir -p vendor/github.com/containerd
+cp -r $ctrd vendor/github.com/containerd/containerd
+# Stop go finding nested vendorings
+rm -rf vendor/github.com/containerd/containerd/vendor
--- a/pkg/containerd/cmd/service/start.go
+++ b/pkg/containerd/cmd/service/start.go
@@ -1,0 +1,113 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func start(args []string) {
+	invoked := filepath.Base(os.Args[0])
+	flags := flag.NewFlagSet("start", flag.ExitOnError)
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s start [service]\n\n", invoked)
+		fmt.Printf("Options:\n")
+		flags.PrintDefaults()
+	}
+
+	sock := flags.String("sock", "/run/containerd/containerd.sock", "Path to containerd socket")
+
+	dumpSpec := flags.String("dump-spec", "", "Dump container spec to file before start")
+
+	if err := flags.Parse(args); err != nil {
+		log.Fatal("Unable to parse args")
+	}
+	args = flags.Args()
+
+	if len(args) != 1 {
+		fmt.Println("Please specify the service")
+		flags.Usage()
+		os.Exit(1)
+	}
+
+	service := args[0]
+	rootfs := filepath.Join("/containers/services", service, "rootfs")
+	log.Infof("Starting service: %q", service)
+	log := log.WithFields(log.Fields{
+		"service": service,
+	})
+
+	client, err := containerd.New(*sock)
+	if err != nil {
+		log.WithError(err).Fatal("creating containerd client")
+	}
+
+	ctx := namespaces.WithNamespace(context.Background(), "default")
+
+	var spec *specs.Spec
+	specf, err := os.Open(filepath.Join("/containers/services", service, "config.json"))
+	if err != nil {
+		log.WithError(err).Fatal("failed to read service spec")
+	}
+	if err := json.NewDecoder(specf).Decode(&spec); err != nil {
+		log.WithError(err).Fatal("failed to parse service spec")
+	}
+
+	log.Debugf("Rootfs is %s", rootfs)
+
+	spec.Root.Path = rootfs
+
+	if *dumpSpec != "" {
+		d, err := os.Create(*dumpSpec)
+		if err != nil {
+			log.WithError(err).Fatal("failed to open file for spec dump")
+		}
+		enc := json.NewEncoder(d)
+		enc.SetIndent("", "    ")
+		if err := enc.Encode(&spec); err != nil {
+			log.WithError(err).Fatal("failed to write spec dump")
+		}
+
+	}
+
+	ctr, err := client.NewContainer(ctx, service, containerd.WithSpec(spec))
+	if err != nil {
+		log.WithError(err).Fatal("failed to create container")
+	}
+
+	io := func() (*containerd.IO, error) {
+		logfile := filepath.Join("/var/log", service+".log")
+		// We just need this to exist.
+		if err := ioutil.WriteFile(logfile, []byte{}, 0666); err != nil {
+			log.WithError(err).Fatal("failed to touch logfile")
+		}
+		return &containerd.IO{
+			Stdin:    "/dev/null",
+			Stdout:   logfile,
+			Stderr:   logfile,
+			Terminal: false,
+		}, nil
+	}
+
+	task, err := ctr.NewTask(ctx, io)
+	if err != nil {
+		// Don't bother to destroy the container here.
+		log.WithError(err).Fatal("failed to create task")
+	}
+
+	if err := task.Start(ctx); err != nil {
+		// Don't destroy the container here so it can be inspected for debugging.
+		log.WithError(err).Fatal("failed to start task")
+	}
+
+	log.Debugf("Started %s pid %d", ctr.ID(), task.Pid())
+}
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -28,4 +28,4 @@ COPY --from=mirror /out/ /
 COPY usr/ /usr/
 COPY etc/ /etc/
 CMD ["/usr/bin/rungetty.sh"]
-LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/etc:/hostroot/etc","/tmp/ctr:/tmp/ctr", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/usr/bin/dist:/usr/bin/dist", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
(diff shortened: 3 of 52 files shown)


ijc:
Forgot to mention that during the build I see:
```
go build...
# cmd/service
warning: unable to find runtime/cgo.a
```

But it doesn't appear to cause any harm

pkg/containerd/cmd/service/skanky-vendor.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
```
rn: "does exactly what it says on the tin"

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/api/services/containers"
+	"github.com/containerd/containerd/api/services/execution"
+	//"github.com/containerd/containerd/api/types/mount"
```
rn: could you add a one line comment as to why this is commented out (or remove it)

ijc: Oops, will remove.

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,123 @@
...
+		enc.SetIndent("", "    ")
+		if err := enc.Encode(&spec); err != nil {
+			log.Fatalf("dumping spec: %s", err)
+		}
+
+	}
+
+	ctr, err := client.NewContainer(ctx, service, containerd.WithSpec(spec),
+		func(ctx context.Context, client *containerd.Client, c *containers.Container) error {
+			c.RootFS = service // This seems odd, but mimics ctr.
```
ijc: A bit more experimentation suggests that in this mode of operation not initialising `c.RootFS` is fine (and correct).

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	logfile := filepath.Join("/var/log", service+".log")
+	// We just need this to exist.
+	if err := ioutil.WriteFile(logfile, []byte{}, 0666); err != nil {
+		log.Fatalf("touch logfile: %s", err)
+	}
+
+	tasks := client.TaskService()
+	rsp, err := tasks.Create(ctx, &execution.CreateRequest{
```
ijc: This ought to be `ctr.NewTask` I think.

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,123 @@
...
+	service := args[0]
+	rootfs := filepath.Join("/containers/services", service, "rootfs")
+	log.Infof("Starting service: %q", service)
+	log := log.WithFields(log.Fields{
+		"service": service,
+	})
+
+	client, err := containerd.New(*sock)
+	if err != nil {
+		log.Fatal("creating containerd client")
```
rn: should this log the `err`?

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	ctx := namespaces.WithNamespace(context.Background(), "default")
+
+	var spec *specs.Spec
+	specf, err := os.Open(filepath.Join("/containers/services", service, "config.json"))
+	if err != nil {
+		log.Fatalf("failed to read config.json: %s", err)
```
rn: should this be `%v` instead of `%s` (as well as in the other places below)?

pkg/containerd/cmd/service/start.go
ijc: Yes. Will fix.

pkg/containerd/cmd/service/start.go
ijc: No, `%s` is correct and causes `err.String()` to be called, `%v` prints out the raw underlying datatype which is usually not what you want.

ijc: ...at least I know who is responsible for all the use of `"%v", err` in the code base ;-)

rn: that would be @justincormack not me. He made me convert some of them back to `%v` and seems to have a strong opinion about it...

ijc: I stand corrected, I made `stdin` be `/dev/nullx` and with both `%s` and `%v` the output is:
```
FATA[0000] failed to create task: rpc error: code = NotFound desc = stat /dev/nullx: no such file or directory  service=getty
```
Closer reading of https://golang.org/pkg/fmt/ confirms that both `%v` and `%s` will, if the argument implements the `Error` interface, call the `Error()` method (which is what I meant when I said `String()` above):
>  If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
>
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
>
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any). 

I was sure I had seen garbled and unhelpful error messages from the use of `%v` on error types in the past, I must have been confused.

pkg/containerd/cmd/service/skanky-vendor.sh
```diff
@@ -0,0 +1,36 @@
...
+# and we conveniently have a checkout already. We actually prefer to
+# reuse containerd's vendoring for consistency anyway.
+
+set -eu
+ctrd=$1
+mkdir -p vendor/github.com/containerd/containerd
+ln -nfs $ctrd/* vendor/github.com/containerd/containerd
+rm -f vendor/github.com/containerd/containerd/vendor
+
+for p in github.com/Microsoft/go-winio \
```
ijc: I don't think it should be any more difficult to maintain than a `vendor.conf`, either as the client code here changes or as the `CONTAINERD_COMMIT=` in the `Dockerfile` changes -- which is roughly equivalent in this context to changing the containerd entry in a `vendor.conf` and then syncing all the other entries from containerd's `vendor.conf`, which is how I would do things if I was doing this with a `vendor.conf`.

pkg/containerd/cmd/service/start.go
ijc: I wrote this code when I was still using the default `log` package but have subsequently switched to logrus. Probably the right thing to do is switch to `log.WithError(err).Fatal("blah blah")`.

ijc:
Force pushed addressing all the issues, plus moved the remount of rootfs into the tool, dropped a now unnecessary `wait` from the initscript (not sure if it was ever necessary) and bumped to a newer (master from yesterday) containerd (which was previously blocked by the removal of `--runtime-config`).

I did try and go to today's containerd master, but it seems to be broken, will debug, but no need to block this PR I think.

ijc:
@riyazdf 
> do you have a link to the PR that removed -runtime-config? I'm curious what the rationale was

It went away in https://github.com/containerd/containerd/pull/962, but not explicitly mentioned (not a supported tool, etcetc). IIRC it's not the first time it went away.

ijc:
> I did try and go to today's containerd master, but it seems to be broken, will debug, but no need to block this PR I think.

I figured it out (`config.toml` needed an update) and have added today's containerd to the end of the series.

justincormack:
@ijc the `runtime/cgo.a` is normal, must find out why eventually.

You have a rebase conflict with Rolf's init changes

pkg/containerd/cmd/service/skanky-vendor.sh
justincormack: Why can't we just copy all of `containerd`s `vendor` directory as we have a checkout? Having more things does not matter.

ijc:
It's also going to conflict with #2044. I may as well wait for that (trivial) PR to be merged before rebasing

pkg/containerd/cmd/service/skanky-vendor.sh
ijc: We need containerd itself too, so it's not a simple `cp` of `vendor` but it might be simpler than this scheme.

pkg/containerd/cmd/service/start.go
```diff
@@ -0,0 +1,128 @@
...
+			Source:      rootfs,
+			Options:     []string{"bind"},
+		},
+		{
+			Destination: "/",
+			Type:        "",
+			Source:      "",
+			Options:     []string{"remount", "rw"},
+		},
+	}, spec.Mounts...)
```
justincormack: I really don't like this, would rather do the remounts up front, not add them to the spec. These won't be needed if we use the image store, and they are not really part of the spec which should reflect what the user specified only, they are part of our implementation. Also they are only needed due to `runc` bugs.

justincormack: (you can leave the remount in the shell script for now if you like, I will probably replace it all with a Go wrapper soon anyway)

ijc: will do this tomorrow.

pkg/containerd/cmd/service/skanky-vendor.sh
justincormack: sure it needs containerd (ex its vendor too), I but it avoids listing them

pkg/getty/Dockerfile
```diff
@@ -17,4 +17,4 @@ COPY --from=mirror /out/ /
 COPY usr/ /usr/
 COPY etc/ /etc/
 CMD ["/usr/bin/rungetty.sh"]
-LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/etc:/hostroot/etc","/tmp/ctr:/tmp/ctr", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/usr/bin/dist:/usr/bin/dist", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
+LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/etc:/hostroot/etc", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/usr/bin/dist:/usr/bin/dist", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
```
justincormack: Why did you remove `/tmp/ctr:/tmp/ctr`? You still need this if you want to run `ctr` from the getty

ijc: `/tmp/ctr` a) is not created is you aren't using `ctr` (which is the purpose of the tool) and b) isn't created even by newer versions of `ctr`.

The only thing it is/was used for AFAICT is to hold the stdio fifos from `ctr run` and `ctr exec`. There is/was nothing in there which is useful to any independent future run of `ctr`, you can run `ctr` in the `getty` container just fine without sharing `/tmp/ctr` with the host mount namespace.

ijc:
Force pushed with all comments (apart from the `/tmp/ctr` thing, still being discussed above) fixed.

pkg/getty/Dockerfile
justincormack: ok

PR #2042, opened 2017-06-14
Kernel config project docs

tych0:
Here's some minor improvements for the kernel config project, as well as some docs on how I imagine it would be used for other people to play with. cc @justincormack asked for this last week.


projects/kernel-config/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* support for arch and version specific config
+* make diffs as readable as possible
+* ensure that all of our config settings are kept after oldconfig
+
+We achieve the goals by:
+
+* having version-specific config in separate files, which are automatically
+  merged
+* only keeping track of visible symbols, only keeping track of a delta from
```
rn: what does "symbol" mean in this context. I may have a very narrow concept of symbol in my head...

further down you specifically call it "kconfig symbol" so might be worth using that on first use of symbol

projects/kernel-config/Makefile
```diff
@@ -12,7 +12,7 @@
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 # Name on Hub
 IMAGE:=kernel
-ARCH?=$(shell if [ "$$(arch)" = "x86_64" ]]; then echo x86; else $$(arch); fi)
+ARCH?=x86
```
rn: I may have asked this before, but what's the rational for `x86`? I would prefer using the kernel arch for this

projects/kernel-config/README.md
tych0: Sorry, "symbol" in this context means something like CONFIG_USB or whatever (this is what the kconfig code calls them). kconfig distinguishes between "visbile" and "invisible" symbols, "visible" symbols are ones that users can choose values for, and "invisible" symbols are ones that users can't (but have some value that's a result of a computation across other values).

Basically, since users don't control e.g. CONFIG_IRQ_WORK (it's invisible), there's no reason for us to track it in our diffs.

projects/kernel-config/Makefile
tych0: Good question. I think this is probably leftover from some other idea I had when I was playing around with this and munging defconfig in the kernel tree. I don't see any reason right now why we can't switch it, so I'll add a patch to this PR to do that.

projects/kernel-config/Makefile
```diff
@@ -12,7 +12,7 @@
 HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
 # Name on Hub
 IMAGE:=kernel
-ARCH?=$(shell if [ "$$(arch)" = "x86_64" ]]; then echo x86; else $$(arch); fi)
+ARCH?=x86_64
```
justincormack: whats the story for non x86_64 here?

justincormack:
Just going to merge this as better to discuss the project as a whole rather than on a PR

projects/kernel-config/Makefile
tych0: The idea is that it's as easy as defining `kernel_config.${ARCH}` and include all the arch-specific config there. The non-arch config lives in `kernel_config.base`, so all of that will be the same across all LinuxKit kernels.

PR #2044, opened 2017-06-15
Update getty image for setsid changes

riyazdf:
New getty images to pick up updates from #2036, I pushed and signed `linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a`

cc @deitch 

</img>

### DIFF
--- a/pkg/getty/README.md
+++ b/pkg/getty/README.md
@@ -10,7 +10,7 @@ If you want a console getty, add the following to your `moby.yml`:
 ```
 services:
   - name: getty
-    image: "linuxkit/getty:e36b5e794256c2a56c000035f73a4d46f7ffb3e6"
+    image: "linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a"
 ```
 
 The above will launch a getty for each console defined in the cmdline, i.e. `/proc/cmdline`.
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/mount.sh", "/var/lib/docker"]
 services:
   - name: getty
-    image: "linuxkit/getty:ef9d667af71089326419fb08e9cc9d567cf15748"
+    image: "linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a"
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,7 +16,7 @@ onboot:
     image: "linuxkit/metadata:31a0b0f5557c6123beaa9c33e3400ae3c03447e0"
 services:
   - name: getty
-    image: "linuxkit/getty:ef9d667af71089326419fb08e9cc9d567cf15748"
+    image: "linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a"
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/getty.yml
+++ b/examples/getty.yml
@@ -14,7 +14,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: "linuxkit/getty:ef9d667af71089326419fb08e9cc9d567cf15748"
+    image: "linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a"
     # to make insecure with passwordless root login, uncomment following lines
     #env:
     # - INSECURE=true
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -11,7 +11,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: "linuxkit/getty:ef9d667af71089326419fb08e9cc9d567cf15748"
+    image: "linuxkit/getty:886d35fe30c47750e8cfbf2f73016e9d2cc6361a"
     env:
      - INSECURE=true
 trust:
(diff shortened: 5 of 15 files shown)


pkg/getty/README.md
```diff
@@ -10,7 +10,7 @@ If you want a console getty, add the following to your `moby.yml`:
 ```
 services:
   - name: getty
-    image: "linuxkit/getty:e36b5e794256c2a56c000035f73a4d46f7ffb3e6"
```
ijc: This is interesting. By doing this you have changes the tree hash of `pkg/getty` and therefore the tag which is produced if you rebuild this directory, hence invalidating this change. Lather, rinse, repeat...

(I noticed this because I did a bulk update in a PR and then my build no longer matched! I undid this change in that case).

deitch: Not quite, @ijc . The has is not the latest commit to the entire repo, just the latest commit that affects the `pkg/getty` directory. Look at the `HASH` code from the Makefile:

```
```

It takes the head of the `pkg/getty/` dir. As long as you don't change that dir, nothing changes.

I still do find it annoying, though, e.g. if you change the `Makefile`, the hash _does_ change. Which is why, outside of this particular project (where there are consistent ways of doing it), I put everything that should go into an image into a subdir, so that only if stuff inside that dir changes, it affects the hash, but not the `README.md` or `Makefile`. So `pkg/getty` would look like:

```
pkg/
 |- getty/
          |- src/
                |- etc/
                      |- ....
                |- usr/
                      |- ....
          |- Makefile
          |- Dockerfile
          |- README.md
```

And the line in `Makefile` instead is:

```
HASH?=$(shell git ls-tree HEAD -- ./src/ | awk '{print $$3}')
```

Look at `projects/compose/` for an example of that.

ijc:
`pkg/getty/README.md` is in `pkg/getty` and is therefore part of the hash, that's the problem I was getting at (note that it is the tree-hash of the directory, it is not a commit hash)

 Moving some stuff to subdirs would be one was to resolve this, although in your proposal `Makefile` and `Dockefile` would no longer be part of the hash but they should be because they impact the output.

pkg/getty/README.md
ijc: (wrote this in the wrong place so it was unthreaded, pasting it here, will delete the other comment)

`pkg/getty/README.md` is in `pkg/getty` and is therefore part of the hash, that's the problem I was getting at (note that it is the tree-hash of the directory, it is not a commit hash)

Moving some stuff to subdirs would be one was to resolve this, although in your proposal `Makefile` and `Dockerfile` would no longer be part of the hash but they should be because they impact the output.

deitch: > note that it is the tree-hash of the directory, it is not a commit hash

Uh, yeah, good point. Oops. :-)

deitch: > pkg/getty/README.md is in pkg/getty and is therefore part of the hash, that's the problem I was getting at
> although in your proposal Makefile and Dockerfile would no longer be part of the hash but they should be because they impact the output

Maybe. I would tend more towards accepting the `Dockerfile` in there than the `Makefile`.

Truth is, I never really understood why we hash it this way. @justincormack or @rn explain it?

I think that the hash-based refs make the most sense: `linuxkit/getty@sha256:ac652...`. If you want it immutable, make it the generated contents of the image, which just works.

rn: we don't want to use the `sha256` because the build itself is not reproducible in all cases (binaries containing timestamps and such things) so we take the hash from all the input and take care that the build does not pull un-deterministic stuff from the network (either by using the alpine base image or pulling very specific version from the network).

We used to explicitly hash the contents of the input files, but that process was error prone (occasionally missed a file) and complex (tar pipelines) and settled for the git tree hash (which is the hash of the contents of the directory) as the next best thing.

Having the README with example hashes in the directory then  becomes a slight issue as we tend to update hash with `git grep ... | sed ...` and thus alter the content hash of the directory creating the circle...

In general, I would prefer having documentation in `./docs` anyway as it's easier to have it one place

riyazdf: bah - my mistakes.  I intentionally updated the docs so that they didn't point to old images but forgot about the implications of the tag hash.

I agree that keeping the `README` in `./docs` would be preferred

justincormack: The `Makefile` has to be in the hash, because it can change the thing that is built (eg different compile options). I do think we should move the READMEs to `docs/` and/or comments in the examples, and definitely remove specific hashes from them.

deitch: @rn thanks for explaining.

@justincormack got it on the `Makefile`. But how do you structure that so `Makefile` is in but `README` (and other irrelevant stuff) is out?

rn: you can't if you use the the git tree hash. As suggested, best may be to move the README to the docs directory and not have it refer to specific hashes. They will be outdated very quick anyway

deitch: In `pkg/getty/docs/` it won't work, because it is under `pkg/getty/` (or `pkg/<whatever>/`), so you mean the main `docs/`? Fine, but I do think each `pkg/<whatever>/` should have its own `README.md`. 

So is the structure:

```
|- pkg
      |- whatever 
             |- README.md
             |- image/
                  |- Makefile
                  |- Dockerfile
                  |- etc/
 ....
```

or just to keep the `README.md` in scope, but not reference an actual hash?

rn: no, put them under top-level `./docs` together with all the other docs.

deitch: Not sure about that. Each package may have its own idiosyncrasies. Each should have its own README

rn: something like the getty stuff should probably be in the top-level docs. It's fine to have packages specific docs, but I wouldn't expect them to change much (unless the code changes) so shouldn't be a problem re hash

PR #2048, opened 2017-06-15
pkg/docker-ce: add vpnkit-expose-port

MagnusS:
Adds vpnkit-expose-port to the docker-ce image.

vpnkit-expose-port is the dockerd userland proxy used to expose
forwarding ports with vpnkit. This adds the binary to the image in
/usr/bin/vpnkit-expose-port, but does not enable it by default.

ralph-nhspca-2926

### DIFF
--- a/pkg/docker-ce/Dockerfile
+++ b/pkg/docker-ce/Dockerfile
@@ -17,8 +17,10 @@ RUN apk add --no-cache --initdb -p /out \
 	xz
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
 
+FROM linuxkit/vpnkit-forwarder:883de832c2c3cb72cd9b01e3f7bd788649e0f2c2 AS vpnkit
 FROM scratch
 COPY --from=mirror /out/ /
+COPY --from=vpnkit /vpnkit-expose-port /usr/bin/vpnkit-expose-port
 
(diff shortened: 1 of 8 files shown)


pkg/docker-ce/Dockerfile
```diff
@@ -1,3 +1,4 @@
+FROM linuxkit/vpnkit-forwarder:883de832c2c3cb72cd9b01e3f7bd788649e0f2c2 AS vpnkit
```
justincormack: I think its more logical to add this just before `FROM scratch` so we consistently do the `FROM linuxkit/alpine` at the top of files.

MagnusS: Ok - moved and updated

justincormack:
This needs to update the `docker-ce` hash in the examples that use it once its pushed

MagnusS:
@riyazdf thanks!

yml files updated

MagnusS:
Ah missed one - should be all now

riyazdf:
@MagnusS: thanks!  I think there's also one more usage in `projects/kubernetes/image-cache/Dockerfile`

MagnusS:
@riyazdf that was the one I missed - should be in the last update

riyazdf:
👍  needed a refresh!  All good :)

PR #2050, opened 2017-06-15
Removed unused vendoring

justincormack:
Not required since #2025


rn:
I noticed we also vendor infrakit, but couldn't find why?

justincormack:
Hmm, I noticed this when making another vendoring update and `vndr` removed this, but it didnt complain about infrakit...

justincormack:
Oh, it was just the README etc, removed as well

rn:
infrakit is still in `vendor.conf`. should be removed

justincormack:
removed

PR #2059, opened 2017-06-16
Use library org in examples to verify nginx, other official images

riyazdf:
`library` org refers to the [official images](https://hub.docker.com/explore/) on docker hub, which includes nginx, alpine, etc

</img>

### DIFF
--- a/examples/swap.yml
+++ b/examples/swap.yml
@@ -40,3 +40,4 @@ services:
 trust:
   org:
     - linuxkit
+    - library
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -32,5 +32,4 @@ services:
 trust:
   org:
     - linuxkit
-  image:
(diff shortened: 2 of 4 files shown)


examples/swap.yml
```diff
@@ -40,3 +40,4 @@ services:
 trust:
   org:
     - linuxkit
+    - library
```
justincormack: I think we generally leave line feeds at the end of files. (I think its only your editor that doesn't, so everyone else will keep reverting this...)

riyazdf: It's weird because gogland shows an extra line in the UI :(  Thanks for noticing, I'll fix this manually now and tinker with the settings to figure out why it's doing that

riyazdf: fixed, vim to the rescue.

PR #2066, opened 2017-06-17
[Work in progress] always-pxe support for Packet

### LINKED ISSUE #2065: run_packet: persisting iPXE
**Description**

Right now, if you are running on Packet, a Linuxkit install does not survive a reboot; that means development of a custom image has to start again from scratch.

From the Packet [documentation on iPXE]:

> When provisioning the Custom iPXE operating system kicks off, we set the next boot option to PXE on first boot.  By default, this PXE process only happens once on the first boot, and subsequent reboots will default back to local disk.

> If you want to persist with PXE on subsequent boots, you can use the always_pxe=(true|false) parameter for the custom_ipxe operating system.  

[documentation on iPXE]:https://help.packet.net/technical/infrastructure/custom-ipxe

Boot linuxkit on packet; reboot the server; try to log in. 

Unable to log in to server.

If a flag `--always-pxe true` is set, subsequent reboots will restart from PXE instead of trying to boot from disk. This will allow testers to simply reboot the system in order to load a new image, rather than deleting and starting over from scratch.

I'll do a PR to implement this, at least to support the testing process. 

The syntax is aimed to exactly match the Packet API, and like the Packet API, the default will be `--always-pxe false`.

It's unclear to me at this point whether this will be the right thing for production environments, though I don't see any reason why it couldn't be if the image is smart enough to e.g. initialize disk only once and test a condition before starting over from scratch. At the moment there's no disk support yet.

vielmetti:
Work in progress.  Closes #2065.

Add a flag `--always-pxe` defaulting to `false` that if set to `true` will cause the machine to reboot from PXE.

Not yet tested, this is work in progress.

**- What I did**

Added a new flag, set default to current default behavior, if flag set to true then new behavior.

**- How I did it**

Parse arguments and update the custom iPXE boot string

**- How to verify it**

Address the question of what the right default is (it might be "true" in which case no option flag is needed).

Add persistent iPXE support for Packet

```
 _   /|
 \'o.O'
 =(___)=
    U
```

### DIFF
--- a/src/cmd/linuxkit/run_packet.go
+++ b/src/cmd/linuxkit/run_packet.go
@@ -82,10 +82,9 @@ func runPacket(args []string) {
 	}
 	hostname := getStringValue(packetHostnameVar, *hostNameFlag, "")
 	name := getStringValue(packetNameVar, *nameFlag, prefix)
-	alwaysPXE := getStringValue(packetAlwaysPXEVar, *alwaysPXEFlag, defaultAlwaysPXE)
 	osType := "custom_ipxe"
 	billing := "hourly"
-	userData := fmt.Sprintf("#!ipxe\n\ndhcp\nset always_pxe %s\nset base-url %s\nset kernel-params ip=dhcp nomodeset ro serial console=ttyS1,115200\nkernel ${base-url}/%s-kernel ${kernel-params}\ninitrd ${base-url}/%s-initrd.img\nboot", alwaysPXE, url, name, name)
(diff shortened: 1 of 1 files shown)


rn:
thanks for the patch.

Having flags which default to true seem weird. `always-pxe` should be the default for `linuxkit run`, so maybe we should have a flag to disable the default behaviour like `disable-always-pxe` or similar (or not give the option to disable it)
 
Could you also squash the commits into one

vielmetti:
ah, the Packet API behaves a little differently from how I first read it, so I opened

https://github.com/packethost/packngo/issues/22

to address the question of `packngo` library support.

rn:
closing this in favour of #2376

PR #2068, opened 2017-06-18
Verify kernel with GPG and sha256 checksum

mscribe:
**- What I did**

For security, the kernel needs to be verified after downloading it.

**- How I did it**

I changed the Dockerfile that downloads the kernel, to download the sha256sums.asc file, verify it with gpg, find the checksum for that version of the kernel in the sha256sums.asc file, verify the zipped kernel with it, then unzip the kernel and verify the unzipped tar file with gpg as well.

**- How to verify it**

You can go into the linuxkit/kernel directory and run:
```
docker build --build-arg KERNEL_VERSION=4.11.6 --build-arg KERNEL_SERIES=4.11.x --build-arg DEBUG= --no-cache -t linuxkit/kernel:4.11.6-hash .
```

However, you might need to update alpine in order to get `apk add gnupg` to work (I had to).  If **hkp://keys.gnupg.net** turns out to be too unstable in the long run, you can set a specific public key server like **pgp.mit.edu** instead.

Verify kernel with GPG and sha256 checksum after downloading.

Goose


justincormack:
Based on the experience with Docker, all the keyservers are pretty unreliable and it is better to include the actual key in the script.

mscribe:
Thanks @justincormack.  Since the kernel version is passed in as an argument, it wasn't easy to stick keys / checksums in the Dockerfile.

justincormack:
I just mean the gpg key, which does not vary by kernel version.

mscribe:
The 2 public keys that get imported are a combined 2,220 lines long, so I'm thinking they wouldn't really belong in a Dockerfile.  They could be put in .asc files (maybe in the linuxkit/kernel folder) and then referenced from the Dockerfile though...

justincormack:
Yes, putting them in files sounds good.

justincormack:
Thanks. Can you squash your commits and sign them off please.

kernel/Dockerfile
```diff
@@ -33,10 +34,23 @@ ARG KERNEL_SERIES
 ARG DEBUG
 
 ENV KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.xz
+ENV KERNEL_SHA256_SUMS=https://www.kernel.org/pub/linux/kernel/v4.x/sha256sums.asc
+ENV KERNEL_PGP2_SIGN=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.sign
 
-RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
+# PGP keys
```
riyazdf: could we elaborate on this comment about which keys this file includes?  This will be helpful as this key pinning expands/changes with the kernel release managers

mscribe: Do you want me to change the comment from:
```
# PGP keys
```
To:
```
# PGP keys for 589DA6B1 (greg@kroah.com) and 6092693E (autosigner@kernel.org)
```
Then commit, squash and sign?

justincormack: yes that sounds good

PR #2074, opened 2017-06-19
FIX Docs Link: ./kernel to ../kernel/README.md

ndauten:
Signed-off-by: Nathan Dautenhahn <ndd@cis.upenn.edu>

**- What I did**

Converted link of .kernel in docs/kernels.md to point to ../kernel/README.md

**- How I did it**

Modified it. 

**- How to verify it**

Read it and verify link works

FIX Docs Link: ./kernel to ../kernel/README.md

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -72,7 +72,7 @@ configuration.
 
 To build and test locally modified kernels, e.g., to try a different
 kernel config or new patches, the existing kernel build system in the
-[`./kernel`](./kernel) can be re-used. For example, assuming the
+[`./kernel`](../kernel/README.md) can be re-used. For example, assuming the
 current 4.9 kernel is 4.9.28, you can build a local kernel with:
 ```
 make build_4.9.28 HASH=foo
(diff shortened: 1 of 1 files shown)


ndauten:
This might be better served linking the Makefile instead because the README doesn't have any general instructions on kernel image building.

justincormack:
I think the current link makes more sense, it is just there to point out where the actual build code is, the README is that document itself.

riyazdf:
@justincormack: good point, the GH page will display the README for the directory as well.  I'm fine with closing this to leave the link as-is

ndauten:
The link as is points to "./kernel" and should minimally be changed to "../kernel". I wasn't sure if that would work fine in github, which is why I pointed it directly at ../kernel/README.md. 

If the link "../kernel" works in github then that would be my proposal.

riyazdf:
@ndauten: good point, I missed that.  I agree that we should change the link to point to `../kernel`

justincormack:
Ah ok, yes directory links work fine in github.

PR #2077, opened 2017-06-20
Update RTF version

dave-tucker:
This brings updated help text and fixes handling of labels as reported
in https://github.com/linuxkit/rtf/issues/16


justincormack:
We should add rtf to the homebrew repo too, see https://github.com/linuxkit/homebrew-linuxkit/issues/1

PR #2080, opened 2017-06-20
Simplify Makefiles for Packages

dave-tucker:
These now inherit from a top-level package.mk
Options like use of the network can be enabled on a per package basis
This removes a lot of duplicate code and make the maintenace of these
Makefiles much easier

### DIFF
--- a/pkg/binfmt/Makefile
+++ b/pkg/binfmt/Makefile
@@ -1,15 +1,4 @@
-.PHONY: tag push
-default: push
+include ../package.mk
 
-ORG?=linuxkit
 IMAGE=binfmt
-DEPS=Dockerfile Makefile main.go $(wildcard etc/binmft.d/*)
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=main.go $(wildcard etc/binmft.d/*)
--- a/pkg/dhcpcd/Makefile
+++ b/pkg/dhcpcd/Makefile
@@ -1,15 +1,4 @@
-.PHONY: tag push
-default: push
+include ../package.mk
 
-ORG?=linuxkit
 IMAGE=dhcpcd
-DEPS=Dockerfile Makefile dhcpcd.conf usr/lib/dhcpcd/dhcpcd-hooks/10-mtu
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --squash --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=dhcpcd.conf $(wildcard usr/lib/dhcpcd/dhcpcd-hooks/*)
--- a/pkg/getty/Makefile
+++ b/pkg/getty/Makefile
@@ -1,15 +1,4 @@
-.PHONY: tag push
-default: push
+include ../package.mk
 
-ORG?=linuxkit
 IMAGE=getty
-DEPS=Dockerfile usr/bin/rungetty.sh
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --squash --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=usr/bin/rungetty.sh $(wildcard etc/*) $(wildcard etc/init.d/*)
--- a/pkg/init/Makefile
+++ b/pkg/init/Makefile
@@ -1,15 +1,4 @@
-.PHONY: tag push
-default: push
+include ../package.mk
 
-ORG?=linuxkit
 IMAGE=init
-DEPS=Dockerfile init $(wildcard etc/*) $(wildcard etc/init.d/*) usermode-helper.c
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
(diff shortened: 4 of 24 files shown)


pkg/package.mk
```diff
@@ -0,0 +1,17 @@
+.PHONY: tag push
+default: push
+
+ORG?=linuxkit
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+NETWORK=0
+
+tag: $(DEPS)
```
justincormack: you may as well add `Dockerfile Makefile $(DEPS)`  here, and then you can remove the (inconsistent) use of `Dockerfile` and `Makefile` in the individual files.

pkg/package.mk
```diff
@@ -0,0 +1,17 @@
...
+ifeq ($(NETWORK), 1)
```
rn: it would be better to check for empty/not set here

pkg/package.mk
```diff
@@ -0,0 +1,17 @@
...
+	docker build --squash --no-cache -t $(ORG)/$(IMAGE):$(HASH) .
```
samoht: can you also somehow pass `--squash` and `--no-cache` in env variables? The rational is that `--squash` needs the experimental mode and `--no-cache` is annoying in dev mode.

justincormack: Do we actually need `--squash` any where now? I know we did previously.

justincormack: Also we should not really need `--no-cache` as everything in `pkg` should be reproducible.

rn:
We'd need to audit the dockerfiles if they generate temporary layers. Most prob don't but some still may. On a plane about to take off so can't check

justincormack:
Given that 17.06 stable is about to ship with experimental off, removing use of `--squash` would be a good idea.

rn:
I reviewed the Dockerfiles in the `./pkg` and none need the `--squash`

samoht:
I guess it was before we used multi-stage builds in order to reduce image sizes.

justincormack:
Also before the /out model.

On 21 Jun 2017 12:36, "Thomas Gazagnaire" <notifications@github.com> wrote:

I guess it was before we used multi-stage builds in order to reduce image
sizes.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/linuxkit/linuxkit/pull/2080#issuecomment-310182953>,
or mute
the thread

dave-tucker:
Thanks for the review comments. I'll update this tomorrow based on the current round of feedback.

dave-tucker:
Okay! PR updated.

- Removed `--squash` and `--no-cache` from the `docker build`
- Set `DEPS=Dockerfile Makefile` in `package.mk` and appended to it in the individual package `Makefiles` if necessary.

pkg/binfmt/Makefile
```diff
@@ -1,15 +1,4 @@
...
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=$(DEPS) main.go $(wildcard etc/binmft.d/*)
```
ijc: You can write `DEPS += main.go ...` here.

Might it be better to have `package.mk` define `BASE_DEPS = Makefile Dockerfile` and then have `tag: $(BASE_DEPS) $(DEPS)`, that way in the default case noone needs to worry about accidentally overwriting the base deps (but you can still zero them if you want). Or perhaps this will be more complicated/confusing for users by introducing a two tier set of deps.

rn: I too think that either `BASE_DEPS` or `+=` are better with a slight preference for the former.

dave-tucker: I didn't like using `+=` as it seemed a little weird.

- `DEPS+=main.go` breaks syntax hilighting in vim and shows DEPS+ as the variable name.
- `DEPS += main.go` and `DEPS +=main.go` have the same result. This is good as you can't screw up by adding too many spaces but seems a little odd style wise. especially as the other assignments in tdon't have spaces either side of the operator.

I think I prefer the `$(BASE_DEPS) $(DEPS)` split

dave-tucker:
Updated again to replace `DEPS=$(DEPS)` with a combination of `BASE_DEPS` and `DEPS`

pkg/dhcpcd/Makefile
```diff
@@ -1,15 +1,4 @@
...
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --squash --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=dhcpcd.conf usr/lib/dhcpcd/dhcpcd-hooks/10-mtu
```
rn: I know this was in the original, but `$(wildcard usr/lib/dhcpcd/dhcpcd-hooks/*)` would be better

pkg/getty/Makefile
```diff
@@ -1,15 +1,4 @@
...
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --squash --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=usr/bin/rungetty.sh
```
rn: I know this was in the original, but there are additional files in the getty subdir which should also be deps

pkg/binfmt/Makefile
ijc: > breaks syntax highlighting in vim 

Did you file a vim bug?

(Personally I would write spaces around all the assignments, and hence use `DEPS += ..` but each to their own)

pkg/init/Makefile
```diff
@@ -1,15 +1,4 @@
...
-
-HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
-
-tag: $(DEPS)
-	docker build --no-cache --network=none -t $(ORG)/$(IMAGE):$(HASH) .
-
-push: tag
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(HASH) || \
-	DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(HASH)
+DEPS=init $(wildcard etc/*) $(wildcard etc/init.d/*) usermode-helper.c
```
rn: super minor nit, but I think it's nicer if the individual files are listed before wildcards and not mixed like this (again, I realises this was copied from the original, but no reason to not improve it)

pkg/dhcpcd/Makefile
ijc: That'll pick up `10-mtu~` and `10-mtu.bak` and other editor detritus like that, won't it?

(that's generally why I prefer things like this to have a `.hook` suffix or something so you can match `*.hook`, but I suppose this is a dhcp implementation choice)

dave-tucker:
Updated again to address the nits identified by @rn

PR #2083, opened 2017-06-20
Add PGP key for 00411886 (torvalds@linux-foundation.org)

riyazdf:
Following up on #2068 with Linus Torvalds' key, as we have a few kernels in projects that already have used or plan to use those kernels: (ex: 4.9, 4.12-rcs, etc)

cc @mscribe 

</img>


rn:
the [build failure](https://github.com/linuxkit/linuxkit-logs/blob/linuxkit-build-of-503c08a6fb5c2f5abb7e2e91b64f94e766751a53-default-ci-pr/value/dba2e7db-41be-4c82-a69c-8f261a2f3600/linuxkit.build.examples.docker.log) seems unrelated. I've hit the rebuild button

Issue #2088, opened 2017-06-21
LinuxKit kernel build fails


zlim:
**Description**
linuxkit/kernel build fails.

**Steps to reproduce the issue:**
```
linuxkit/kernel$ make
```
**Describe the results you received:**
```
linuxkit/kernel$ make
docker pull linuxkit/kernel:4.11.5-438be97e754a8068f3bc1b7ddbbe7a9ab7c8cb97 || docker build --build-arg KERNEL_VERSION=4.11.5 --build-arg KERNEL_SERIES=4.11.x --build-arg DEBUG= --no-cache -t linuxkit/kernel:4.11.5-438be97e754a8068f3bc1b7ddbbe7a9ab7c8cb97 .
Error response from daemon: manifest for linuxkit/kernel:4.11.5-438be97e754a8068f3bc1b7ddbbe7a9ab7c8cb97 not found
Sending build context to Docker daemon  1.033MB
Step 1/28 : FROM linuxkit/alpine:4f0ddee221c46f142e5a190dd43d2e07256ef98d AS kernel-build
 ---> edbfa267b8a3
Step 2/28 : RUN apk add     argp-standalone     automake     bash     bc     binutils-dev     bison     build-base     curl     diffutils     elfutils-dev     flex     git     gmp-dev     gnupg     installkernel     kmod     libelf-dev     libressl-dev     libunwind-dev     linux-headers     ncurses-dev     sed     squashfs-tools     tar     xz     xz-dev     zlib-dev     || true
 ---> Running in 72acf02759f6
ERROR: unsatisfiable constraints:
  gnupg (missing):
    required by: world[gnupg]
 ---> 6c00f75989a5
Removing intermediate container 72acf02759f6
Step 3/28 : ARG KERNEL_VERSION
 ---> Running in 252e25ed837b
 ---> 8956e6f9bca8
Removing intermediate container 252e25ed837b
Step 4/28 : ARG KERNEL_SERIES
 ---> Running in 7823a49732b6
 ---> 2ef911da19ad
Removing intermediate container 7823a49732b6
Step 5/28 : ARG DEBUG
 ---> Running in 0bcfd0009faa
 ---> 7b2963491786
Removing intermediate container 0bcfd0009faa
Step 6/28 : ENV KERNEL_SOURCE https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.xz
 ---> Running in e1517c26492b
 ---> 10429a3e11c4
[...]

PR #2091, opened 2017-06-22
alpine: include gnupg and zfs packages

riyazdf:
Include `gnupg` package in the base alpine image for kernel signing verification (see https://github.com/linuxkit/linuxkit/issues/2088)

Also include `zfs` from other issue and discussion at moby summit


zlim:
Works locally. Thanks!

PR #2092, opened 2017-06-22
add an auditd container

### LINKED ISSUE #52: add audit logging
For debugging seccomp and related issues.

tych0:
Two things to note here: we unfortunately can't just exec audit, because it
dies if /var/log/audit doesn't exist, so we have to do a wrapper script.

Second, it also dies if it can't re-nice itself, so we have to give it
CAP_SYS_NICE as well as the audit caps.

I didn't add this to the default linuxkit.yml because the linuxkit/audit
repo doesn't exist yet, but we should probably (?) do that.

Closes #52

### DIFF
--- a/pkg/auditd/Dockerfile
+++ b/pkg/auditd/Dockerfile
@@ -1,0 +1,29 @@
+FROM linuxkit/alpine:cdb4e4d12ada4071a6c6a60bf4c14d35171ffae8 AS build
+RUN apk add abuild gcc git
+
+ADD build.sh /
+RUN adduser -D -G abuild builder && sudo -u builder /build.sh
+
+FROM linuxkit/alpine:cdb4e4d12ada4071a6c6a60bf4c14d35171ffae8 AS mirror
+COPY --from=build /home/builder/*apk /
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --initdb -p /out alpine-baselayout tini
+RUN apk add --allow-untrusted -p /out /*apk
+
+# Remove apk residuals. We have a read-only rootfs, so apk is of no use.
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY auditd.conf /etc/audit
+COPY audit.rules /etc/audit
+COPY runaudit.sh /usr/bin
+
+CMD ["/sbin/tini", "/usr/bin/runaudit.sh"]
+
+LABEL org.mobyproject.config='{"pid": "host", "binds": ["/var/log:/var/log"], "capabilities": ["CAP_AUDIT_CONTROL", "CAP_AUDIT_READ", "CAP_AUDIT_WRITE", "CAP_SYS_NICE"]}'
--- a/pkg/auditd/Makefile
+++ b/pkg/auditd/Makefile
@@ -1,0 +1,3 @@
+include ../package.mk
+
+IMAGE?=auditd
--- a/pkg/auditd/audit.rules
+++ b/pkg/auditd/audit.rules
@@ -1,0 +1,12 @@
+## First rule - delete all
+-D
+
+## Increase the buffers to survive stress events.
+## Make this bigger for busy systems
+-b 8192
+
+## This determine how long to wait in burst of events
+--backlog_wait_time 0
+
+## Set failure mode to syslog
+-f 1
--- a/pkg/auditd/auditd.conf
+++ b/pkg/auditd/auditd.conf
@@ -1,0 +1,32 @@
+#
+# This file controls the configuration of the audit daemon
+#
+
+local_events = yes
+write_logs = no
+log_format = RAW
+flush = INCREMENTAL_ASYNC
+freq = 50
+priority_boost = 4
+disp_qos = lossy
+dispatcher = /sbin/audispd
+name_format = NONE
+##name = mydomain
+max_log_file_action = ROTATE
+space_left = 75
+space_left_action = SYSLOG
+action_mail_acct = root
+admin_space_left = 50
+admin_space_left_action = SUSPEND
+disk_full_action = SUSPEND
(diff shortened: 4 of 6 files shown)


justincormack:
Is there a way to make it log to `stdout` instead? We are going to be reworking logs for remote logging etc, and we can't really have programs owning their log files, or we can't do rotation etc.

tych0:
Yep, it's a pretty easy change, I can make it now.

ijc:
> /var/log/audit doesn't exist, so we have to do a wrapper script.

You can create directories in the `files` section of the yaml, but I guess that might be tricky to rely on and a bit of learning-curve/pain for every user who forgets/doesn't know. Perhaps the label stuff could be extended to allow making dirs?

Or if you make the binds `/var/log:/var/log` *and* `/var/log/auditd:/var/log/auditd` will it be auto created? I can't remember sadly.

There's also the option of a wishlist bug upstream to add a `mkdir`...

justincormack:
No, we can't create the directories when the bind is specified as the filesystem is read only. We could add directories that there are binds for at build time I guess.

tych0:
Yep, and if we run it logging to stdout, we don't even need that dir, so I removed that bit.

Also note that events like USER_AUTH and USER_LOGIN are recorded from userspace (via pam or the actual thing itself), and busybox doesn't seem to send these signals, but the "real" versions of them do. So if you spend a few hours trying to figure out why the USER_ events don't seem to be logged, that's why :).

Anyway, above is the updated patch, with logging to stdout and the same set of default rules as ubuntu uses.

tych0:
Hmm. SIGBUS? Seems like the ci might be angry?

pkg/auditd/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY auditd.conf /etc/audit
+COPY audit.rules /etc/audit
+COPY runaudit.sh /usr/bin
+
+CMD ["/sbin/tini", "/usr/bin/runaudit.sh"]
```
justincormack: I don't think we would need `tini` as I would not expect `auditd` to even fork?

tych0: It depends on how you have it configured. If you use a dispatcher with it, it forks the dispatcher and that itself forks whatever you tell it to. IIRC it also forks when parsing config for some reason.

justincormack: ah ok.

justincormack:
I re-ran the CI as it seemed to hang right at the end shutting down GCP this time

justincormack:
The efi ISO test is failing, possibly for unrelated reasons... but it is consistent.

justincormack:
Can you rebase, CI should be fixed after.

pkg/auditd/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
```
riyazdf: should we just remove this line?  Do we know when mainline alpine might get the package?

pkg/auditd/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: tag push
```
riyazdf: can we use the new style Makefile?  Should be pretty simple, see https://github.com/linuxkit/linuxkit/blob/master/pkg/rngd/Makefile as an example

pkg/auditd/Dockerfile
tych0: I have no idea what the promotion process is, unfortunately. I suppose we could talk to @ncopa about it :)

tych0: Anyway, yes, sorry about that, I dropped the line.

pkg/auditd/Makefile
tych0: Yep, done.

pkg/auditd/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM alpine:edge AS mirror
```
justincormack: This is not reproducible, so we can't use this like this with tree hashes. The options are either to pin this to a specific sha256 hash, or to build auditd from source, or not use the standard Makefile.

riyazdf: ah I missed this - good point.  Let's use `alpine:edge@sha256:b45de39913783fcc1ee31e15bc9a3df4226a0a51111e00ca74ea8531c099d4df`

justincormack: That is still not reproducible as it installs packages.

tych0: Building from source won't be reproducible either.

I don't see a way to do a reproducible build without essentially replicating what we have for main for this. In any case, per a talk with @ncopa I've started the discussion about main inclusion for audit: https://github.com/alpinelinux/aports/pull/1776

justincormack: building from source is reproducible if you check out an exact git commit or downloaded version, thats what we do elsewhere.

ijc: I think we don't get exact binary reproducibility with this method, unless the code base happens to support it, in general we only get the same functional output given the same inputs i.e. without variation in components due to network mirrors etc changing. (I suspect you are each talking about a different level of "reproducible" here).

justincormack: Sure, if it embeds build timestamps or whatever, but the requirement at present is that all the inputs are covered by the tree hash. We will work on the other parts later.

tych0: Ok, I set it up to build from source. We need a few of audits deps (and abuild) in the base image, so there's a commit that adds those. I think CI will fail until that's merged; I'll PR it separately assuming this approach works for people.

justincormack:
I think the base alpine package has been updated again since this, so the hash is probably different?

tych0:
I'm including some of the auditd build deps in the base image (added in the first commit in the series), so the base image hash will always be different. That's why I was thinking ci would fail, because it tries to build everything in pkg/.

In any case, you're right that the underlying package list has changed, so I rebased and updated the hash so that it should be right.

PR #2095, opened 2017-06-22
Kernel and kernel config update

### LINKED ISSUE #2061: Use modules for hardware drivers
With module loading now working, we should move some device drivers to be compiled as modules.

As a rule of a thumb, we should only compile in device which are commonly used on Hypervisors (mostly virtio and hyper-v) and move all other device drivers to be compiled as modules.

This should speed up boot times on local hypervisors where we care most about these things

rn:
- Fix kernel dockerfile to error on `apk add` (see #)
- Update kernels to 4.11.6/4.9.33/4.4.73
- Move network drivers, networking options, NFS, and others from kernel to modules (resolves #2061)

image


rn:
I've tested this with 4.11 and it removes a bunch of the network interfaces and boots on hyperkit without requiring a module load. We still have the `tunl0` and `ip6tnl0`interface because I couldn't completely move the tunnel support into modules, primarily because I left VXLAN to be compiled in

kernel/Dockerfile
```diff
@@ -26,8 +25,7 @@ RUN apk add \
     tar \
     xz \
     xz-dev \
-    zlib-dev \
-    || true
```
ijc: 👍

pkg/getty/profile
```diff
@@ -0,0 +1,4 @@
+echo "You are running inside a container!"
+echo "You can access the host root filesystem under /proc/1/root"
+
+PS1="getty \$ "
```
ijc: I suspect you didn't intend to include this.

rn: Oops, yes. should have done a `git clean`

kernel/Dockerfile
justincormack: the `|| true` is there so  that you don't get whitespace changes when changing the last item or adding a new one ta the end.

rn: but it also masks errors

justincormack: ah, ok yes

rn:
Hmm failed on an unrelated GCP issue. I've compiled/pushed/signed the kernel images and hit a rebuild

rn:
kernel tests needed updating as well to remove NFS and rpc_pipefs as they have been moved to modules. Force pushed an update

rn:
@tych0 I hadn't had a chance to update the kernel-config project, but might be good to track these changes once merged

justincormack:
Please don't remove these from the tests, they should be there. You can add a set of things to `modprobe` before running the tests.

justincormack:
(or just add a script to modprobe everything pre-test)

rn:
Added the modprobe for selected modules and the test passes locally.

@justincormack if you are happy with it and the test is green, please merge as I'm off soon

test/hack/test.yml
```diff
@@ -12,7 +12,7 @@ onboot:
     image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: check-kernel-config
-    image: "linuxkit/test-kernel-config:2acaa564c1801dd2ae1546c70c472dc58ac030a1"
+    image: "linuxkit/test-kernel-config:9f08e3b99f8ac2f422251b3e8c94ce874ee34119"
     readonly: true
```
justincormack: did you mean to delete this `readonly: true` too?

justincormack:
ok, one tiny comment but looks fine

test/hack/test.yml
rn: yes, missed that one, thanks for spotting

justincormack:
the `run_efi` test is also failing here, I don't think its related but unclear why it started failing recently (or exactly when).

tych0:
@rn yes, I can take a look at that. We should probably also update kernel-config for the user mode helper changes.

kernel/Dockerfile
ijc: If we keep them sorted then the chances of something coming after `zlib` are at least slim...

Or the list could end
```
    zlib-dev \
    # end of list
```
although I'm not sure that is nicer than the occasional whitespace-fu on update.

ijc: Either `&& true` or `|| false` would also give the desired result, wouldn't it?

It looks like (from `git grep -B1 '\(||\|&&\) \(true\|false\)'`) a lot of places following this sort of pattern are using the `&& true` variant.

There are other uses of `|| true`, many of which look suspect to me. Lots of

* `rm -rf ${STUFF}* || true`. But `rm -f` ignores non-existent arguments which I think is the main reason for the use of `|| true`, so it is unneeded.
* `RUN tar cf «something» usr || true`, I can't imagine why we would want to ignore such errors, we should insist on an explanatory comment for such constructs IMHO.

justincormack:
CI failure is the ISO EFI test which needs more investigation, see #2100

PR #2098, opened 2017-06-23
Use less memory in /init

justincormack:
Previously we would copy the whole root filesystem to a new
tmpfs root, then free it up. We do this as `pivot_root` fails
on an initrd, it wants a real mounted `tmpfs`.
    
Copying it all used a lot of memory, as it uses twice as much
and each tmpfs can only be 50% of RAM, so it needs four times
as much memory as the unpacked size.
    
Instead, for `/containers` which is the largest part, move each
file one by one, so that they are freed after the move. This
should almost halve the memory requirement.

zebra

### DIFF
--- a/pkg/init/init
+++ b/pkg/init/init
@@ -2,6 +2,14 @@
 
 /bin/mount -t tmpfs tmpfs /mnt
 
+# just copying everything uses a lot of space, so use mv which frees it up
+# copy directory structure under /containers
+find /containers -type d | cpio -p /mnt 2> /dev/null
+# move files under /containers
+find /containers -not -type d -exec mv {} /mnt/{} \;
+# delete remaining directories
+rm -rf /containers
+
 /bin/cp -a / /mnt 2>/dev/null
 
 /bin/mount -t proc -o noexec,nosuid,nodev proc /proc
--- a/blueprints/docker-for-mac.yml
+++ b/blueprints/docker-for-mac.yml
@@ -3,7 +3,7 @@ kernel:
   image: "linuxkit/kernel:4.9.34"
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
+  - linuxkit/init:e7f7119d21c423d64db186fe6fe011edd7422418
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
   - linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a
 onboot:
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.34"
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
+  - linuxkit/init:e7f7119d21c423d64db186fe6fe011edd7422418
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
   - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
   - linuxkit/ca-certificates:46b59484919bfa9af700e54e042048cb06261de4
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.34"
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
+  - linuxkit/init:e7f7119d21c423d64db186fe6fe011edd7422418
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
   - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
   - linuxkit/ca-certificates:46b59484919bfa9af700e54e042048cb06261de4
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.34"
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
-  - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
+  - linuxkit/init:e7f7119d21c423d64db186fe6fe011edd7422418
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
   - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
   - linuxkit/ca-certificates:46b59484919bfa9af700e54e042048cb06261de4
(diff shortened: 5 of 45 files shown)


justincormack:
cc @kencochrane this should halve the RAM requirement...

kencochrane:
Awesome! Thanks

justincormack:
@dave-tucker @talex5 CI seems to have got very slow. This keeps timing out after 60m

justincormack:
It does seem to be consistently timing out on GCP, I think there may be a real issue, although it seems odd. Am travelling for a few days so may not be able to debug, but any input welcome.

justincormack:
@talex5 can you increase the total test timeout? I can't see anything abnormal about this PR, but our total CI time has got quite long already, I think we may just be marginal with a 1 hour timeout.

Opened an issue about general test slowness https://github.com/linuxkit/linuxkit/issues/2106

pkg/init/init
```diff
@@ -2,6 +2,14 @@
 
 /bin/mount -t tmpfs tmpfs /mnt
 
+# just copying everything uses a lot of space, so use mv which frees it up
+# copy directory structure under /containers
+find /containers -type d | cpio -p /mnt 2> /dev/null
```
ijc: Does this read and discard everything while creating the directories? Could be a source of slowness, although given everything is in a tmpfs on both ends perhaps not that likely.

pkg/init/init
```diff
@@ -2,6 +2,14 @@
...
+# move files under /containers
+find /containers -not -type d -exec mv {} /mnt/{} \;
```
ijc: This involves considerable more forking than the plain `cp -a` version.

ijc: Might a `tar | tar` (or `cpio` if you prefer) avoid that?

pkg/init/init
justincormack: no, why would it do that?

pkg/init/init
justincormack: The whole point is to remove the file from the old mount point immediately after copying it to the new one, else nothing changes.

pkg/init/init
ijc: Oops, I'd forgotten about the `-type d` by the time I got to the `cpio -p`.

pkg/init/init
justincormack: I can't detect any slowdown in running images from the CI runs. Will do some more local tests but it seems unlikely it is forking overhead.

ijc: Doh, yes, was focussing to much on the potential performance issues.

justincormack:
The whole copying process takes under a second, so this PR is not slowing things down. It does cycle through 1000 pids, which we can remove later by rewriting in Rust.

talex5:
Yes, builds are taking around 1 hour it seems. I've increased it to 2 hours.

justincormack:
@talex5 it is still timing out at 60 minutes - did you restart CI when you changed timeout?

talex5:
Looks like I restarted with the old image... trying again...

justincormack:
@talex5  this is _still_ timing out after 60 minutes, I restarted it just an hour ago!

talex5:
Hmm. I've made it log the timeout at the start now - I think it really is pushed now!

justincormack:
needs a rebase...

justincormack:
The extra forking (presumably) takes about 12s under qemu, perhaps a bit longer on the build machine. I guess this is where the overall slowdown was happening. Its non measurable on a real machine. I might look at rewriting it in a real language I guess.

riyazdf:
shall we rebase this?

justincormack:
I am thinking of rewriting in Go to avoid the forking which is slow in qemu

deitch:
I am trying to understand something. Why do we `cp` it all over from `/` to `/mnt` and then `switch_root` in the first place, and not use it as is? On boot, the entire linuxkit image is  initramdisk and copied to tmpfs in memory as read-only anyways, is it not? What do we gain by the copy?

justincormack:
You can't run runc directly from the original rootfs unless you pass
through some special options, as pivot_root fails. It's very annoying.

deitch:
> You can't run runc directly from the original rootfs unless you pass
through some special options, as pivot_root fails. It's very annoying

I was wondering. Since both are in `tmpfs`, both read-only, both with identical content, why not just run it directly? 

So we are going through this whole copy - and now doing messy "copy directories but move files in /containers/" (can we have `mv` files out of `/containers` if the whole thing is mounted `ro` by kernel?) - because of an issue with `runc` in the original mounted `/`?

What is the fail, and what are the "special options" required? Is it because the initrd itself is created with some special options that get in the way? And could we `pivot_root` inside `init` (rather than copying everything over) and thus bypass the issue? Or is it that it must be a completely separate tmpfs?

deitch:
Oh. From `man (2) pivot_root`:

> new_root and put_old must not be on the same file system as the current root

So, where does `runc` run into the problem?

justincormack:
It is this condition https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L99

deitch:
> It is this condition

`runc` here is setting up the "root" for the new container, and wants to `pivot_root` for the new process, right? I like that line [here](https://github.com/opencontainers/runc/blob/67bd2ab554bdf8a604f020e57448b37ef2102171/libcontainer/rootfs_linux.go#L635) that says, "no matter what the documentation says, this is valid..."

In any case, [here] it says:

```go
// NoPivotRoot will use MS_MOVE and a chroot to jail the process into the container's rootfs
// This is a common option when the container is running in ramdisk
```

1. What is special about running in ramdisk (vs another tmpfs that we create in `pkg/init` that it needs `NoPivotRoot`?
2. What is the downside of actually running as `NoPivotRoot` (as the docs seem say)?

I get it, could be that we have to accept all of this, but as we twist ourselves a bit to save the memory and move some stuff and copy other, I am wondering it the cleanest answer - just run the darn think in initrd - isn't doable. But could be going down a rabbit hole.

justincormack:
There used to be a kernel patch floating around that fixed this, but it never got merged. It is not clear what the issue is - `pivot_root` is very fussy. If we could persuade `runc` to always use `MS_MOVE` it might be ok, but there may be other software with the same issue, so it seems more compatible to continue to do this for now, despite it being annoying. Should document better.

deitch:
> Should document better

If nothing else, it would have saved me bothering you to explain it. :-)

deitch:
> If we could persuade runc to always use MS_MOVE it might be ok, but there may be other software with the same issue, so it seems more compatible to continue to do this for now, despite it being annoying

Unknown and unintended consequences. I get it. For now, easier to live with the `cp -a` and eventually remove it when no longer needed.

Do you have a link to that kernel patch and the discussion of the issue? Happy to do a quick PR documenting this so we have the reasons why we do this in the docs.

deitch:
In the meantime, as the current state of `runc`+kernel stands, we have no choice but to copy. 

1. Could we just `mv` everything over to the new `tmpfs`?
2. More fundamentally, does the short-term memory usage really impact? Once we move over and switch root, don't we unmount the ramdisk and recover the memory?

justincormack:
This PR was changing the "just `mv` everything behaviour" as that uses four times the memory of the filesystem size (the tmpfs can each only get 50% full, and a full copy uses twice the size). This meant that people with large images can survive with only about half the memory usage by deleting each file after it is moved. So eg a 250MB image currently requires 1GB RAM, vs 512MB with this PR.

deitch:
> This PR was changing the "just mv everything behaviour" as that uses four times the memory of the filesystem size

I thought his PR was changing _from_ `cp` everything _to_ `mv` everything?

justincormack:
The difference is previously it would `cp` then delete the old set at the end, while this deletes after each file is moved.

justincormack:
Closing for now, will reowrk

Issue #1872, opened 2017-05-22
uefi ISO will not run in qemu container


justincormack:
Just gives me exit status 139. Error unclear.

rneugeba:
it runs in tests..

justincormack:
weird. Will try on Linux.

rneugeba:
are you running qemu in a container on your mac?

justincormack:
Yes

justincormack:
Doesnt work for me on Linux either.

Getting this error which is a bit confusing, looks like it is trying to use the BIOS as a disk

```
WARNING: Image format was not specified for '/usr/share/ovmf/bios.bin' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
```

rneugeba:
that's not an error but a warning

rneugeba:
Looks like it doesn't run on the host either:
```
/usr/local/bin/qemu-system-x86_64 -device virtio-rng-pci -smp 1 -m 1024 -machine q35 -cdrom linuxkit-efi.iso -boot d -pflash /Users/rneugeba/Desktop/qemu/bios.bin
```
hangs:
![screen shot 2017-06-08 at 11 58 30](https://user-images.githubusercontent.com/3338098/26925630-669b4870-4c42-11e7-8373-21d8a90dff2f.png)

The same EFI ISO works fine on Windows

weird

justincormack:
Hmm, odd that it works in tests. qemu version issue?

rneugeba:
it also hangs when I run it `-containerized`. My mac qemu version is 2.9.0. I'm trying to find a different source for the EFI firmware, because that's something which also may have changed

justincormack:
It works for me at the moment in a container on the mac, but it segfaults on Debian Jessie. It is also failing on CI; it used to be skipped though.

PR #2100, opened 2017-06-23
Disable Qemu EFI ISO test for now

justincormack:
See #1872 this is very flaky and without a controlled known
working environment this is not helpful.

boot


rn:
Do we know what changed? The test used to pass...

PR #2101, opened 2017-06-23
Add a metadata provider for Vultr

furious-luke:
Vultr uses a very similar approach to AWS, including using the
same IP address for serving metadata. In fact, it seems
as though if AWS appears first in the list of providers, that
provider mistakenly believes to be running on AWS (hence the
insertion of `NewVultr` in between GCP and AWS). I don't believe
AWS servers will accidentally try to use the Vultr provider,
as it seems that the `/v1/` endpoint doesn't exist on AWS. It would
probably pay to come up with a better way of differentiating between
them though, just to be safe.

**- What I did**
Added a metadata provider to gather information from Vultr VPSs.

**- How I did it**
Copied the AWS provider and made modifications.

**- How to verify it**
I verified it by running a LinuxKit machine on Vultr and examined the resultant metadata.

Add a metadata provider for Vultr.

This is Pepper the ferret!
image

### DIFF
--- a/pkg/metadata/provider_vultr.go
+++ b/pkg/metadata/provider_vultr.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"time"
+)
+
+const (
+	vultrMetaDataURL = "http://169.254.169.254/v1/"
+)
+
+// ProviderVultr is the type implementing the Provider interface for Vultr
+type ProviderVultr struct {
+}
+
+// NewVultr returns a new ProviderVultr
+func NewVultr() *ProviderVultr {
+	return &ProviderVultr{}
+}
+
+func (p *ProviderVultr) String() string {
+	return "Vultr"
+}
+
+// Probe checks if we are running on Vultr
+func (p *ProviderVultr) Probe() bool {
+	// Getting the index should always work...
+	_, err := vultrGet(vultrMetaDataURL)
+	return (err == nil)
+}
+
+// Extract gets both the Vultr specific and generic userdata
+func (p *ProviderVultr) Extract() ([]byte, error) {
+	// Get host name. This must not fail
+	hostname, err := vultrGet(vultrMetaDataURL + "hostname")
+	if err != nil {
+		return nil, err
+	}
+	err = ioutil.WriteFile(path.Join(ConfigPath, Hostname), hostname, 0644)
+	if err != nil {
+		return nil, fmt.Errorf("Vultr: Failed to write hostname: %s", err)
+	}
+
+	// public ipv4
+	vultrMetaGet("interfaces/0/ipv4/address", "public_ipv4", 0644)
+
+	// private ipv4
+	vultrMetaGet("interfaces/1/ipv4/address", "private_ipv4", 0644)
+
+	// private netmask
+	vultrMetaGet("interfaces/1/ipv4/netmask", "private_netmask", 0644)
+
+	// region code
+	vultrMetaGet("region/regioncode", "region_code", 0644)
+
+	// instance-id
+	vultrMetaGet("instanceid", "instance_id", 0644)
+
+	return nil, nil
+}
+
+// lookup a value (lookupName) in Vultr metaservice and store in given fileName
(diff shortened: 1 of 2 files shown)


furious-luke:
Sorry about that, totally forgot to sign the commit.

justincormack:
Sorry, can you rebase, should fix the CI issue.

Do you add any headers to your metadata replies that could distinguish from AWS? GCP requires special headers on the requests but obviously that is an API breakage.

pkg/metadata/provider_vultr.go
```diff
@@ -0,0 +1,123 @@
...
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"time"
+)
+
+const (
+	vultrMetaDataURL = "http://169.254.169.254/v1/"
+)
```
kencochrane: does Vultr not have user-data in it's metadata service?

pkg/metadata/provider_vultr.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("Vultr: Status not ok: %d", resp.StatusCode)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("Vultr: Failed to read http response: %s", err)
+	}
+	return body, nil
+}
```
kencochrane: not really specific to this PR, just a comment in general. I wonder if we should make `vultrGet`, `awsGet`, `vultrMetaGet`, `awsMetaGet`, etc. generic and put in a common package, so we can reuse instead of having to copy into each of these providers.  `gcpGet` is similar but requires an http header, which is still doable.

I guess it only makes sense if there are going to be more like these.

justincormack: I think it is fine for now. They are not very large bits of code and the small differences will tend to make the shared code messier, so unless there are a lot more and no standards I think this is fine for now.

pkg/metadata/provider_vultr.go
furious-luke: Not exactly. There is a `userdata` field contained in it's metadata, but that's only valid when running a prebuilt VPS system. When using an ISO of any kind the server ignores it, making its value for LinuxKit rather dubious.

furious-luke:
> Do you add any headers to your metadata replies that could distinguish from AWS? GCP requires special headers on the requests but obviously that is an API breakage.

Sorry, could you elaborate a little? It's not clear to me how adding extra headers on the requests could help distinguish between AWS and Vultr...

justincormack:
@furious-luke GCP requires a special header on your metadata request or it ignores it. It is very confusing the first time you use it as you don't expect this. See https://github.com/linuxkit/linuxkit/blob/master/pkg/metadata/provider_gcp.go#L75

furious-luke:
@justincormack Ah, okay. I've tested the Vultr metadata server as is and it works as expected on Vultr VPS without any extra headers. Still not too clear on how we might use the headers to distinguish between AWS and Vultr though, can you elaborate on that part?

justincormack:
If Vultr returns some specific header on their replies we could check for that, if we can trust it not to change. There might not be one.

furious-luke:
Ah, got you; I was confused about adding the headers to the requests as opposed to them appearing in the responses. There's no mention of additional headers in the Vultr docs, but I haven't checked directly.

Perhaps we could check a more specific path in the `Probe` functions? I'm sure there must be some amount of mutual exclusivity between the metadata URLs provided by AWS and Vultr...

pkg/metadata/provider_vultr.go
kencochrane: ok, thank you, that makes sense.

pkg/metadata/provider_vultr.go
kencochrane: agreed, no reason to do it now. Just a thought for the future when we have a bunch of these, that are all the same.

riyazdf:
Merging now. @furious-luke: we can modify for a more specific path in `Probe` on a followup PR if you find anything.

riyazdf:
new image: `linuxkit/metadata:9a614b312a8d507acacc6905267c5c43589c2f01`

PR #2103, opened 2017-06-26
Makefile cleanups

justincormack:
Remove some incorrect parts, like `tar xf tmp_linuxkit_bin.tar > $@`

Split out the cross build testing, and give it some parallelism, as it is a bit slow.

monsey_tawny_owl


justincormack:
this cuts the time to do the cross tests from 3.5m to 2.5m on my laptop

PR #2102, opened 2017-06-23
Getty and ssh improvements

ijc:
**- What I did**

Updated the `sshd` and `getty` containers such that the following runs ok:
```
dist pull docker.io/library/redis:alpine
ctr run -t docker.io/library/redis:alpine test
```
and produces the usual redis start of day logging.

**- How I did it**

* Added `/etc/localtime` (configured for UTC) to host fs, since `ctr` wants to bind it.
* Added `ca-certificates` to both containers, since `dist` wants some roots
* Mounted host `/tmp` into the containers, `ctr` and `containerd` need to share `/tmp`.
* Synchronised `sshd` containers bind mounts to include all those used by `getty` (which includes the binds of `ctr` and `dist` themselves)

**- How to verify it**

```
dist pull docker.io/library/redis:alpine
ctr run -t docker.io/library/redis:alpine test
```
In either container ought to work

**- Description for the changelog**
Tools such as `ctr` and `dist` now work in both the `getty` and `sshd` containers.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2104, opened 2017-06-26
test: ensure ctr works in getty container

ijc:
Fix a typo in the main containerd test too.

**- What I did**

Added a test that `ctr run` (and `dist pull`) work from within the `getty` container, to avoid regressions in #2102.

**- How I did it**

Some `expect(1)` based insanity.

**- How to verify it**

The tests should pass.

Doesn't need anything above the #2102 entry IMHO.

[](https://en.wikipedia.org/wiki/Jar_of_Flies)

### DIFF
--- a/test/cases/040_packages/007_getty-containerd/test-ctr.yml
+++ b/test/cases/040_packages/007_getty-containerd/test-ctr.yml
@@ -1,25 +1,18 @@
 kernel:
   image: "linuxkit/kernel:4.9.x"
-  cmdline: "console=ttyS0 console=tty0 page_poison=1"
+  cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/init:17693d233dd009b2a3a8d23673cb85969e1dce80
   - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
   - linuxkit/containerd:a33cdcf50b8107ffe14c92802c460fe7ada39acd
   - linuxkit/ca-certificates:75cf419fb58770884c3464eb687ec8dfc704169d
 onboot:
-  - name: sysctl
-    image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
   - name: dhcpcd
     image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
     image: "linuxkit/getty:9f27c1272b6d128c9a09745e916f151d09cb0d27"
-    # to make insecure with passwordless root login, uncomment following lines
-    #env:
-    # - INSECURE=true
-  - name: rngd
-    image: "linuxkit/rngd:1fa4de44c961bb5075647181891a3e7e7ba51c31"
 files:
   - path: etc/getty.shadow
     # sample sets password for root to "abcdefgh" (without quotes)
--- a/test/cases/040_packages/003_containerd/test.sh
+++ b/test/cases/040_packages/003_containerd/test.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# SUMMARY: Run contianerd test
+# SUMMARY: Run containerd test
 # LABELS:
 # REPEAT:
 
--- a/test/cases/040_packages/007_getty-containerd/test.exp
+++ b/test/cases/040_packages/007_getty-containerd/test.exp
@@ -1,0 +1,96 @@
+#!/usr/bin/env expect
+spawn linuxkit run test-ctr
+set pid [exp_pid]
+set timeout 60
+set prompt ":~# "
+
+expect {
+    timeout {
+        puts "FAILED boot"
+        exec kill -9 $pid
+        exit 1
+    }
+    "Welcome to LinuxKit" {
+        puts "SUCCESS boot"
+    }
+}
+expect {
+    timeout {
+        puts "FAILED login user"
+        exec kill -9 $pid
+        exit 1
+    }
+    "ogin: " {
+	send "root\n"
+    }
+}
+expect {
+    timeout {
+        puts "FAILED login pass"
+        exec kill -9 $pid
+        exit 1
+    }
+    "assword: " {
+	send "abcdefgh\n"
+    }
+}
+
+expect {
+    timeout {
+        puts "FAILED dist pull"
+        exec kill -9 $pid
+        exit 1
+    }
+    $prompt {
+	send "dist pull docker.io/library/redis:alpine\n"
+    }
+}
+expect {
+    timeout {
+        puts "FAILED dist pull"
+        exec kill -9 $pid
+        exit 1
+    }
+    $prompt {
+	puts "SUCCESS dist pull"
+	send "ctr run -t docker.io/library/redis:alpine test\n"
+    }
+}
+expect {
+    timeout {
+        puts "FAILED ctr run"
+        exec kill -9 $pid
+        exit 1
+    }
+    "The server is now ready to accept connections on port 6379" {
+	puts "SUCCESS ctr run"
+	# Ctrl-C
+	send "\003"
+    }
+}
+expect {
+    timeout {
+        puts "FAILED kill ctr"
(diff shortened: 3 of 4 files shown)


ijc:
There's a sort of implicit assumption that testing `getty` means `sshd` is less likely to regress too, although that's not a given. Arranging to test via ssh as well seem like it will be more fiddly than I can cope with.

It'd be good in general to keep an eye out for PRs which touch `getty` but don't make a similar change to `sshd` I think.

justincormack:
An ssh test that is similar should be ok-ish on some platforms at least. hyperkit+vsudd would allow forwarding the ssh port to a local unix socket, and you can generate a key and use that in the config file. qemu will only forward to tcp which is a bit more annoying as you have to find a free local port, I couldnt see any easy way to script that.

justincormack:
failed with `/usr/bin/env: expect: No such file or directory`

ijc:
@talex5 could you add `expect` to the CI environment for now please.

talex5:
@ijc: just to check: you want me to update the Debian GCP image used to run Docker on the build VMs?

ijc:
@talex5 TBH, I don't know, I need `expect` in whatever environment is invoking `make test` or `make test-pr`. I don't know if that is the same machine as the build machines or some separate test machine.

talex5:
OK, I've added it and triggered a rebuild.

justincormack:
Still failing with missing `expect`

ijc:
https://github.com/linuxkit/linuxkit-logs/blob/cc69862834afee3bd69990f51b698fca915eb081/value/d83750a3-8138-4598-add9-c925f254aa23/linuxkit.packages.containerd-ctr-getty.log suggests the issue is with the console or something else, it seems to be running enough expect to boot at least.

justincormack:
Oh maybe I was reading the wrong log.

justincormack:
Looks like a timeout?

ijc:
It gets to `clocksource: Switched to clocksource tsc` in about 20s and then is silent until the 30s timeout, I would have expected the welcome whale next I think, instead it appears to hang.

justincormack:
Oh, the `linuxkit.yml` file uses `tty0` which means your output does not go to serial console. You should not use that for tests anyway, you should add a local yml file.

ijc:
It works for me with the current `linuxkit.yml`, I think the use of `tty0` just means the init process's stdio goes there, but we run a `getty` on `ttyS0` so it gets the whale and a prompt etc which is what the expect script needs.

I wanted to test the actual thing most people will try first, i.e. the toplevel `linuxkit.yml` which is why it is testing that.

justincormack:
I don't think you should test the top level file, there is no guarantee it stays like that, or even that we don't remove it.

ijc:
Bumped the timeout from 30s to 60s, on my laptop the boot takes ~2s, whereas in CI it is taking more like 20s it seems, so it is plausible that there might be a 10s gap before the next thing.

justincormack:
It doesnt use kvm so it will be much slower.

ijc:
Right now `README.md` at the top level tells people to use that yml as their first attempt. If the top-level yml changes in a way which breaks this test then CI will tell us and we can decide what we want to do, which might be change the test.

ijc:
If I nobble KVM (I should add a qemu runner option for that...) then there is a noticeable pause after the `tsc` log line, so I like this theory.

justincormack:
I would still much rather you had your own yaml file, without nginx in, basically examples/getty but without the password. This is basically a test of the getty container so it should be based on those examples.

ijc:
OK, I'll wait and see if this run works due to the timeout fix and then adjust.

ijc:
Increasing the t/o seems to have worked, I'll update to use its own yml now.

rn:
Kinda obscure, but the numbering of tests in `test/cases/040_packages` is alphabetically base on the first letter of the package (a = 001, b = 002, etc). I would prefer of we group tests of the same package, so creating a new group with `test/cases/040_packages/003_containerd` containing two sub directories of tests for containerd might be nicer.., but other LGTM

justincormack:
Its more a test for `getty` than `containerd` arguably...

ijc:
The important thing IMHO is that `ctr` and `dist` work in the `getty` environment, happy to move to whatever you two agree on.

Using the index of the first letter as the number seems a bit silly to me, it'll sort the same way without out it so it's just a source of make work for the 9/10 contributors who don't spot the pattern, but whatever.

rn:
Agree, the numbering is a bit silly, but it's the numbering which determines the order (and not necessarily the alphabetical order). Anyway...

how about `007-getty-containerd`?

ijc:
`git mv cases/040_packages/020_containerd-ctr-getty cases/040_packages/007-getty-containerd` => DONE.

I note that `rtf list` produces the same output both before and after running ` ( cd cases/040_packages/ ; for i in ???_* ; do git mv "${i}" "${i#???_}"; done; )` though...

rn:
Thanks
could you squash the commits? No need to have 5 commits for one new test

test/cases/040_packages/007-getty-containerd/test-ctr.yml
```diff
@@ -0,0 +1,26 @@
...
+  - linuxkit/init:17693d233dd009b2a3a8d23673cb85969e1dce80
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:a33cdcf50b8107ffe14c92802c460fe7ada39acd
+  - linuxkit/ca-certificates:75cf419fb58770884c3464eb687ec8dfc704169d
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:7d2b8aaaf20c24ad7d11a5ea2ea5b4a80dc966f1"
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
```
justincormack: doesnt need `sysctl` or `dhcpcd` - may as well keep the tests minimal.

test/cases/040_packages/007-getty-containerd/test-ctr.yml
```diff
@@ -0,0 +1,26 @@
...
+services:
+  - name: getty
+    image: "linuxkit/getty:9f27c1272b6d128c9a09745e916f151d09cb0d27"
+  - name: rngd
```
justincormack: or rngd`

test/cases/040_packages/007-getty-containerd/test-ctr.yml
ijc: It needs `dhcpcd`  for the `dist pull`, I dropped `rngd` and `sysctl`.

test/cases/040_packages/007_getty-containerd/test-ctr.yml
```diff
@@ -0,0 +1,22 @@
+kernel:
+  image: "linuxkit/kernel:4.9.x"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:17693d233dd009b2a3a8d23673cb85969e1dce80
+  - linuxkit/runc:3a4e6cbf15470f62501b019b55e1caac5ee7689f
+  - linuxkit/containerd:a33cdcf50b8107ffe14c92802c460fe7ada39acd
+  - linuxkit/ca-certificates:75cf419fb58770884c3464eb687ec8dfc704169d
+onboot:
+  - name: dhcpcd
```
justincormack: we dont need dhcpcd either...

rn: `dist` needs it, no?

ijc: We do: https://github.com/linuxkit/linuxkit/pull/2104#discussion_r124247508

justincormack: ah of course yes

ijc:
CI timed out after 1hr, wasn't the limit just extended to 2hrs? If that's true could you restart this one please? (who can?  @talex5?)

talex5:
Restarted.

justincormack:
Ah, tis green

PR #2110, opened 2017-06-27
qemu: Add -enable-kvm to force (attempted) use (or not) of KVM

ijc:
The default is set by probing for /dev/kvm. Use of KVM can be forcibly disabled
with `-enable-kvm=false`. Note that passing `-enable-kvm=true` when `/dev/kvm` is
not present still boots but complains:

    Could not access KVM kernel module: No such file or directory
    failed to initialize KVM: No such file or directory
    Back to tcg accelerator.

**- What I did**

Added `-enable-kvm` option to `linuxkit run qemu`.

**- How I did it**

Plumbed in some new golang `flags` usage.

**- How to verify it**

Try passing `-enable-kvm` or `-enable-kvm=false` before and after renaming `/dev/kvm` to something else.

LinuxKit run using QEMU backend can now opt-in/out to use of KVM. Default remains to autodetect.

[](https://www.metal-archives.com/albums/A_Forest_of_Stars/Beware_the_Sword_You_Cannot_See/478451)

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -37,6 +37,11 @@ type QemuConfig struct {
 	PublishedPorts []string
 }
 
+func haveKVM() bool {
+	_, err := os.Stat("/dev/kvm")
+	return !os.IsNotExist(err)
+}
+
 func runQemu(args []string) {
 	invoked := filepath.Base(os.Args[0])
 	flags := flag.NewFlagSet("qemu", flag.ExitOnError)
@@ -68,6 +73,7 @@ func runQemu(args []string) {
 	fw := flags.String("fw", "/usr/share/ovmf/bios.bin", "Path to OVMF firmware for UEFI boot")
 
 	// VM configuration
+	enableKVM := flags.Bool("kvm", haveKVM(), "Enable KVM acceleration")
 	arch := flags.String("arch", "x86_64", "Type of architecture to use, e.g. x86_64, aarch64")
 	cpus := flags.String("cpus", "1", "Number of CPUs")
 	mem := flags.String("mem", "1024", "Amount of memory in MB")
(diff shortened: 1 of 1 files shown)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -68,6 +73,7 @@ func runQemu(args []string) {
 	fw := flags.String("fw", "/usr/share/ovmf/bios.bin", "Path to OVMF firmware for UEFI boot")
 
 	// VM configuration
+	enableKVM := flags.Bool("enable-kvm", haveKVM(), "Enable or disable QEMU")
```
justincormack: comment says "qemu" not kvm

ijc: Sigh. I keep making the same mistake. Will fix.

justincormack: may as well just call the flag `--kvm`

ijc: The "QEMU" option is `-enable-kvm` although I confess that wasn't a consideration when I typed it above and it doesn't have the same `=false` ability, so compat is probably not the biggest concern.

I'll change.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -68,6 +73,7 @@ func runQemu(args []string) {
 	fw := flags.String("fw", "/usr/share/ovmf/bios.bin", "Path to OVMF firmware for UEFI boot")
 
 	// VM configuration
+	enableKVM := flags.Bool("kvm", haveKVM(), "Enable or disable KVM acceleration")
```
justincormack: I think the comment should read like a boolean, ie "Enable KVM acceleration"

ijc: Done

PR #2112, opened 2017-06-27
RFC: add "trim after image delete" to the Docker for Mac blueprint

djs55:
Previously in Docker for Mac we run `fstrim /var` every 15 minutes from `cron` to reclaim space. If an image is deleted then it can take a long time for the space to be freed on the host. This PR adds a helper which watches for Docker events  and triggers an `fstrim` after an image delete. If a batch of image deletes happen within 10s (e.g. as part of a `docker system prune`) then only one `fstrim` will be issued.

**- What I did**

Caused `fstrim /var` to be run after Docker images are deleted.

**- How I did it**

Added a new package `pkg/trim-after-delete` which contains the Docker event watcher and modified `blueprints/docker-for-mac.yml` to hook it up.

**- How to verify it**

Viewing the log file `/var/log/trim-after-delete.log` from a `linuxkit run docker-for-mac`. Log lines are created when the Docker daemon connection is established, when an event has been observed and when `fstrim /var` has been triggered.

- Docker for Mac: trigger TRIM to reclaim disk space when images are deleted

**- Other notes**

- The Docker event watcher needs more code to be vendored. Since this is specific to `trim-after-delete` I created another `vendor.conf` in the `pkg/trim-after-delete` directory based on the same `docker/docker` hash as the root `vendor.conf`. Would it be better to put this code in another git repository and reference it? Or should I merge this with the root `vendor` directory and then copy this into the build context somehow?
- The `make` runes attempt to tag and push to `linuxkit/trim-after-delete` which obviously doesn't exist.

Comments appreciated!

### DIFF
--- a/blueprints/docker-for-mac.yml
+++ b/blueprints/docker-for-mac.yml
@@ -79,6 +79,9 @@ services:
             "--swarm-default-advertise-addr=eth0",
             "--userland-proxy-path", "/usr/bin/vpnkit-expose-port",
             "--storage-driver", "overlay2" ]
+  # Monitor for image deletes and invoke a TRIM on the container filesystem
+  - name: trim-after-delete
+    image: "linuxkit/trim-after-delete:6cc6131300c287fcd40041a28119fee2fc874539"
 
 files:
     - path: /var/config/docker/daemon.json
--- a/pkg/trim-after-delete/Dockerfile
+++ b/pkg/trim-after-delete/Dockerfile
@@ -1,0 +1,26 @@
+# We need the `fstrim` binary:
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    busybox \
+    util-linux
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+# We also need the Go binary which calls it:
+RUN apk add --no-cache go musl-dev
+ENV GOPATH=/go PATH=$PATH:/go/bin
+
+COPY .  /go/src/trim-after-delete
+RUN go-compile.sh /go/src/trim-after-delete
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+COPY --from=mirror /go/bin/trim-after-delete /usr/bin/trim-after-delete
+CMD ["/usr/bin/trim-after-delete", "--", "/sbin/fstrim", "/var/lib/docker"]
+LABEL org.mobyproject.config='{"binds": ["/var/run:/var/run", "/var/lib/docker:/var/lib/docker"], "capabilities": ["CAP_SYS_ADMIN"]}'
--- a/pkg/trim-after-delete/main.go
+++ b/pkg/trim-after-delete/main.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+// Listen for Docker image delete events and run a command after a delay.
+
+// Event represents the subset of the Docker event message that we're
+// interested in
+type Event struct {
+	Type   string
+	Action string
+}
+
+// String returns an Event in a human-readable form
+func (e Event) String() string {
+	return fmt.Sprintf("Type: %s, Action: %s", e.Type, e.Action)
+}
+
+// DelayedAction runs a function in the future at least once after every call
+// to AtLeastOnceMore.
+type DelayedAction struct {
+	c chan interface{}
+}
+
+// NewDelayedAction creates a delayed action which guarantees to call f
+// at most d after a call to AtLeastOnceMore.
+func NewDelayedAction(d time.Duration, f func()) *DelayedAction {
+	c := make(chan interface{})
+	go func() {
+		for {
+			<-c
+			time.Sleep(d)
+			f()
+		}
+	}()
+	return &DelayedAction{
+		c: c,
+	}
+}
+
+// AtLeastOnceMore guarantees to call f at least once more within the originally
+// specified duration.
+func (a *DelayedAction) AtLeastOnceMore() {
+	select {
+	case a.c <- nil:
+		// Started a fresh countdown
+	default:
+		// There is already a countdown in progress
+	}
+}
+
+func main() {
+	// after-image-deletes --delay 10s -- /sbin/fstrim /var
+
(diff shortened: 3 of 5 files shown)


blueprints/docker-for-mac.yml
```diff
@@ -79,6 +79,15 @@ services:
...
             "--storage-driver", "overlay2" ]
+  # Monitor for image deletes and invoke a TRIM on the container filesystem
+  - name: trim-after-delete
+    image: "linuxkit/trim-after-delete:3e8920e29f84bfe30d15d482760a8a3b1d06ffb2"
+    capabilities:
+    binds:
+        - /var/:/var
+    capabilities:
+        - CAP_SYS_ADMIN
+    command: ["/usr/bin/trim-after-delete", "--", "/sbin/fstrim", "/var"]
```
justincormack: The mountpoint here (not on the old D4M) is `/var/lib/docker` not `/var`.

pkg/trim-after-delete/.gitignore
```diff
@@ -0,0 +1 @@
+trim-after-delete
```
justincormack: there won't be a binary built, so you don't need this.

MagnusS:
@djs55 Nice! Thanks for adding this. Since `fstrim` is the only cron job in DfM this may also save us a package in the final blueprint.

pkg/trim-after-delete/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN go-compile.sh /go/src/trim-after-delete
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+COPY --from=mirror /go/bin/trim-after-delete /usr/bin/trim-after-delete
+CMD ["/usr/bin/trim-after-delete", "--", "/sbin/fstrim", "/var"]
+LABEL org.mobyproject.config='{"binds": ["/var:/var"], "capabilities": ["CAP_SYS_ADMIN"]}'
```
justincormack: I think this should be `/var/lib/docker:/var/lib/docker`

blueprints/docker-for-mac.yml
justincormack: Also as these are in the label on the contaienr, they are not needed here

pkg/trim-after-delete/main.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+// DelayedAction runs a function in the future at least once after every call
+// to AtLeastOnceMore.
+type DelayedAction struct {
+	d time.Duration // maximum time to wait before calling f
+	t *time.Timer   // if non-nil we can be sure that f is going to be called again
+	f func()        // the function to call in the future
+	m sync.Mutex
```
justincormack: ugh, using a Mutex. In Go. Surely there is a nicer way to do this with a channel and a goroutine. Goroutine waits for an item on the channel, sleeps max time, then wakes up and consumes all the events in the channel and calls the function?

pkg/trim-after-delete/vendor.conf
```diff
@@ -0,0 +1,13 @@
+github.com/Sirupsen/logrus 3d4380f53a34dcdc95f0c1db702615992b38d9a4
+github.com/moby/moby 4fc2710dc7a8fc87849e94770fcf33eececae2fc
```
justincormack: you should not be vendoring `moby/moby` should only be `docker/docker`.

pkg/trim-after-delete/main.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
```
justincormack: I would really be inclined to not vendor `docker/docker` at all, and just write a pure http version. The events protocol, especially given what you want from it, is pretty trivial I would think. Definitely less than 160,000 lines of code...

pkg/trim-after-delete/main.go
ijc: This reminded me of http://www.jtolds.com/writing/2016/03/go-channels-are-bad-and-you-should-feel-bad/

justincormack: Actually, you only need one event in the channel at a time, if there is busy you can discard. I think it might be a sane use case for channels...

djs55:
Thanks -- I've
- removed the duplicate `bind`/`capability` from the `docker-for-mac.yml`
- bind-mounted `/var/run:/var/run` (for the `/var/run/docker.sock` which is being created lazily so I don't think I can reliably bind mount it) and `/var/lib/docker:/var/lib/docker` for the `fstrim`
- written a simple HTTP Docker event fetcher (without using the official client)
- completely removed all the vendoring (the only thing left was `logrus` which didn't seem worth the hassle)
- replaced the `Mutex` and reference with a channel and a non-blocking send

It's certainly smaller :) hopefully it's also better!

blueprints/docker-for-mac.yml
```diff
@@ -79,6 +79,9 @@ services:
             "--swarm-default-advertise-addr=eth0",
             "--userland-proxy-path", "/usr/bin/vpnkit-expose-port",
             "--storage-driver", "overlay2" ]
+  # Monitor for image deletes and invoke a TRIM on the container filesystem
+  - name: trim-after-delete
+    image: "test"
```
justincormack: Just a note that this of course needs to be replaced with the real image!

djs55: Good point -- what's the best way of handling the image? I think someone with access would need to create `linuxkit/trim-after-delete` and then I (or someone else) can run `make` which will `tag` and `push`?

pkg/trim-after-delete/main.go
```diff
@@ -0,0 +1,145 @@
...
+			if event.Action == "delete" && event.Type == "image" {
+				log.Printf("The delayed action will happen at least once more")
+				action.AtLeastOnceMore()
+			}
+		}
+	}
+
+}
+
+// Event represents the subset of the Docker event message that we're interested
```
justincormack: prefer types at the top of the file really, easier to read.

blueprints/docker-for-mac.yml
justincormack: cc @riyazdf

justincormack:
Looks much nicer!

blueprints/docker-for-mac.yml
riyazdf: Adding hub and notary repo now

riyazdf: everything's set up - @djs55 since I don't have your key yet (we can sort this out later), just ping me when you've addressed all review comments for the package and I'll push/sign

djs55:
@justincormack I've moved the types to the top, hopefully that is more readable.

justincormack:
LGTM once package is pushed

riyazdf:
`docker.io/linuxkit/trim-after-delete:6cc6131300c287fcd40041a28119fee2fc874539` pushed and signed

djs55:
I've updated the hash in the `docker-for-mac.yml` and force-pushed. Hopefully CI will be green this time.

PR #2113, opened 2017-06-27
Improve package and kernel builds

### LINKED ISSUE #1812: Too easy to push images built from a dirty tree tagged as if clean.
Many of the linuxkit package builds use
```
```
and build and _by default_ push `linuxkit/$(IMAGE):$(HASH)`. However if the tree is dirty i.e. you are changing the package and testing your changes before committing then `HASH` will be unchanged from the latest commit so the build will be tagged with that and a push will potentially overwrite the proper version.

We should arrange somehow for `HASH` to be suffixed (e.g. with `-dirty` similar to `git describe`) to indicate that the tree is dirty, that would mitigate most of the problem with accidental pushing too.

`git ls-tree` doesn't seem to be able to tag things as dirty, automatically. It might be possible to use `git diff --quiet` to detect and produce a useful error code.

rn:
- If a package or the kernel is build from a dirty git repository, add `-dirty` to the tag and disallow pushing to hub (resolves #1812).

- If the latest commit has a tag, also push an image to hub as `<org>/<image>:<tag>`. This should allow us to simplify YAML files once we introduce releases

- For packages, don't build the package if it already exists on hub. This was already done for the kernel, but for packages, the check was only performed on `push`. This should make it easier for CI to just attempt to build all packages (wip #1074).

- For kernel build, merge the `sign` target into `push`. This brings it in line with the package and other builds which do not have a `sign target`.

- fix rngd build (see https://github.com/moby/moby/issues/33849)

image


rn:
Note, this is marked as RFC. If people are happy with the changes, I'll append this PR with further commits to update the YAML files

justincormack:
Looks good.

riyazdf:
Design LGTM!

rn:
I've updated the YAML files. Since this required a rebuild of all packages I've also updated the alpine base to the latest and fixed the `rngd` build.

Removed the `[RFC]`prefix

rn:
I noticed that I had also update `./pkg/getty/README.md` and `./pkg/swap/README.md` using a `git grep ... | sed ...` variation and this changes the tree hash. I've now added a commit to remove the explicit hashes from the READMEs and updated the two packages.

rn:
@justincormack @riyazdf since I pushed another commit (see https://github.com/linuxkit/linuxkit/pull/2113#issuecomment-311422847), could either of you add another LGTM if you are fine with me merging when green?

justincormack:
LGTM

rn:
had to fix a typo in the 4.4. kernel config test, but now green, to merging

PR #2120, opened 2017-06-28
Make it more obvious that getty/sshd are namespaced

### LINKED ISSUE #2078: Non-host filesystem for getty and sshd containers is confusing everyone
Multiple people (nearly everyone AFAICT) have been confused by the getty/sshd containers not being in the host mount namespace in the last week (usually manifests as "where has this file I'm sure should be in the fs gone?").

There are good reasons for this isolation but perhaps we can mitigate somehow. Some things which spring immediately to mind are an `/etc/motd` displayed on login with a reminder (and perhaps some cut-and-pastable runes to get into the root namespace) or an indication in `$PS1` (or maybe both).

Maybe someone has a better idea.

dave-tucker:
This PR fixes #2078 by implementing the suggestions from the discussion there

- Add a reminder in the MOTD
- Add a reminder in the $PS1

To do this I used `agetty` from `util-linux` instead of the provided busybox getty

Note: this will fail CI as I don't have signing permissions for getty

### DIFF
--- a/blueprints/docker-for-mac.yml
+++ b/blueprints/docker-for-mac.yml
@@ -34,7 +34,7 @@ onboot:
 services:
   # Enable getty for easier debugging
   - name: getty
-    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    image: "linuxkit/getty:6d35e3fe138aaeaf099b5b4f31b3f12ba725cb49"
     env:
         - INSECURE=true
   - name: rngd
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -18,7 +18,7 @@ services:
   - name: rngd
     image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
   - name: sshd
-    image: "linuxkit/sshd:f095b62ddca658e99d5751872c933ef3d5d18cec"
+    image: "linuxkit/sshd:7535658a08a01c61e2851e4bc1f6346e89412de8"
     binds:
      - /var/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -15,7 +15,7 @@ services:
   - name: dhcpcd
     image: "linuxkit/dhcpcd:6c1ca76dbf808d5c27d10cbf22a8d4399be5c8ae"
   - name: sshd
-    image: "linuxkit/sshd:f095b62ddca658e99d5751872c933ef3d5d18cec"
+    image: "linuxkit/sshd:7535658a08a01c61e2851e4bc1f6346e89412de8"
 files:
   - path: root/.ssh/authorized_keys
     source: ~/.ssh/id_rsa.pub
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/mount.sh", "/var/lib/docker"]
 services:
   - name: getty
-    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    image: "linuxkit/getty:6d35e3fe138aaeaf099b5b4f31b3f12ba725cb49"
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,13 +16,13 @@ onboot:
     image: "linuxkit/metadata:09781c8a8097e5aa3ae522c74856d80e1da3b915"
 services:
   - name: getty
-    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    image: "linuxkit/getty:6d35e3fe138aaeaf099b5b4f31b3f12ba725cb49"
     env:
      - INSECURE=true
   - name: rngd
     image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
   - name: sshd
-    image: "linuxkit/sshd:f095b62ddca658e99d5751872c933ef3d5d18cec"
+    image: "linuxkit/sshd:7535658a08a01c61e2851e4bc1f6346e89412de8"
     binds:
      - /var/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
--- a/examples/getty.yml
+++ b/examples/getty.yml
@@ -14,7 +14,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    image: "linuxkit/getty:6d35e3fe138aaeaf099b5b4f31b3f12ba725cb49"
     # to make insecure with passwordless root login, uncomment following lines
     #env:
     # - INSECURE=true
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -11,7 +11,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    image: "linuxkit/getty:6d35e3fe138aaeaf099b5b4f31b3f12ba725cb49"
     env:
      - INSECURE=true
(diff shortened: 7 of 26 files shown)


deitch:
This is a very good idea. Particular reason for switching from `getty` (busybox) to `agetty`? I don't much care, but curious why? Is it just for `agetty`'s `-a root` autologin option, so that it becomes a full login shell with `motd` and root's `.profile` (and hence `PS1`)?

deitch:
Separately, and probably good to get feedback, I have felt that changing the `examples/` and `projects/` should be a distinct second stage PR from changing the `pkg/` itself. So you would have PR 1 enables new feature, PR 2 takes advantage of new features. I feel it keeps the PRs cleaner and easier to review and follow. 

Am I in the minority here? Happy to have a discussion around it.

ijc:
@deitch `agetty` is partly because you can't seem to add options to the shell with the busybox getty, so you can't pass the `-l` to make it a login shell and hence don't get to parse profile and therefore don't get `PS1`. The fact that `agetty` actually supports `-a root` is a nice bonus on top of that.

WRT multiple PRs, I think there it is already hard enough to get one such PR merged with racing against other changes and rebasing (and therefore regenerating hashes etc). Also if you don't update the hashes then you don't have the CI actually testing the changes to `pkg/` until you do the update.

The amount of noise in the diffs caused by updating the yml files is a real problem though and maybe something we should look for other ways to address/mitigate, I think that deserves and issue of its own rather than using the comments on this largely unrelated PR.

pkg/getty/root/.profile
```diff
@@ -0,0 +1 @@
+PS1="(getty ns) \w \$ "
```
ijc: Would you mind doing the same to the sshd container too?

Perhaps we should put this in `/etc/profile` or some other suitable global file, since the `getty` container (at least in theory, I've not tried it) supports non-root logins. It looks like alpine's `/etc/profile` sources `/etc/profile.d/*.sh` which makes it nice and easy to drop a file in.

Debian's default profile automatically adds the contents of `/etc/debian_chroot` to `$PS1` (which can be usefully abused for non-chroot purposes too!), although `/etc/profile.d/*.sh` makes it mostly moot I suppose.

rn:
In my opinion changing a package and updating the YAMLs should be *one* PR but ideally the YAML update should be a separate commit (as done here).

deitch:
> The fact that agetty actually supports -a root is a nice bonus on top of that

:-) 

> The amount of noise in the diffs caused by updating the yml files is a real problem

This is what I was referring to, primarily. My instinct was multiple PRs, but as you said:

> hard enough to get one such PR merged with racing against other changes and rebasing (and therefore regenerating hashes etc)

Is it? When the changes are primarily in a single `pkg/<foo>/`, will rebasing be needed at all, and will hashes change? Sure, with the examples, but that is an argument in favour of the other direction: keep `examples/` changes on their own, separate from functional package changes.

> you don't have the CI actually testing the changes to pkg/ until you do the update

Yeah, that is an issue. 

Maybe all of this pushes us towards splitting `linuxkit/linuxkit` into multiple repositories? `linux kit/linuxkit` for the tool, `linuxkit/sshd` for the sshd package, `linuxkit/examples` for the examples, etc.? OK, radical solution, but even so?

pkg/getty/root/.profile
dave-tucker: 👍 I'll see about moving it to a more global location so it can be used for both sshd and getty.
it'd be very handy to have some runes in `$PS1` updates based on which namespace you are in.

dave-tucker:
Re: Updating *.yml we should probably move discussion to an issue of it's own and try to keep comments here to the change at hand. Otherwise we risk losing pertinent comments in the quagmire of closed PR's once this is merged/closed.

I'm working on an update based on @ijc's comments and hope to have an amended PR ready shortly

deitch:
> Re: Updating *.yml we should probably move discussion to an issue of it's own and try to keep comments here to the change at hand. Otherwise we risk losing pertinent comments in the quagmire of closed PR's once this is merged/closed.

Which would be pretty ironic, given that the issue raised was specifically around getting lost in lots of changes in the commit(s)! :-)

I will open a new issue.

dave-tucker:
Updated `sshd` and `getty` to use `/etc/profile`
This is based on alpine's `/etc/profile` with the `PS1` line changed to reflect which namespace you are in.

pkg/getty/etc/profile
```diff
@@ -0,0 +1,13 @@
+export CHARSET=UTF-8
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+export PAGER=less
+export PS1='(getty ns):\w\$ '
+export PS1='\h:\w\$ '
```
ijc: This is overwriting your carefully crafted `$PS1` from above.

Why not use a file in `/etc/profile.d/*.sh ` instead of duplicating the whole file?

Also I'm not sure about dropping `\h` since knowing the host is still useful in the getty container, so maybe your snipped should be `export PS1=(getty ns) $PS1"` to just extend the basic one?

justincormack:
Looks like this will print `You are not in the root namespace.` even in the case you add the `getty` to `init` and you are in the root namespace...

pkg/sshd/etc/profile
```diff
@@ -0,0 +1,13 @@
+export CHARSET=UTF-8
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+export PAGER=less
+export PS1='(sshd ns):\w\$ '
+export PS1='\h:\w\$ '
```
ijc: Same as for the getty case.

(I'm wondering more and more if getty and sshd couldn't be partially combined.

deitch: They probably should @ijc. @justincormack raised that early on in getty's life. But we didn't want starting one to cause both to kick off. Someone might want just `getty`, another just `sshd`. And putting it in `init:` should not kick off `sshd` unless so desired.

pkg/getty/etc/profile
deitch: Seeing Justin's point from below (init case), should `PS1=` be more intelligent?

pkg/sshd/etc/profile
ijc: I was proposing to merge the sources (i.e. `pkg/getty` and `pkg.sshd`) but to still produce two distinct output containers which could be added independently.

deitch: Only differences would be the `CMD` line in `Dockerfile`? I do like it.

deitch: What would we call the image? `gettysshd`? `sshetty`? Short form might be inappropriate to use in polite company! :-)

ijc: I would imagine there would be differences in the default binds, the command line, the set of installed packages, but similarities in the login arrangements and in things like PS1 munging.

The names would be `getty` and `sshd`, just like they are today.

deitch: Actually, I think the only differences would be `CMD` and the addition of `openssh-server` apk package.

dave-tucker:
@justincormack maybe the prompt in the motd should read.
`Please pay attention to the namespace indicator in the prompt.`

@deitch @ijc ack on comment above.
I think the `getpid` syscall might work in determining which ns we're in.
Then the prompt would have read something like `(ns: 43) linukit22314:~$` and possibly even update when you `nsenter`

justincormack:
That assumes that the getty is in the host pid namespace, which is the default but is not required.

dave-tucker:
ok, i'll leave it static for now. adding a smarter PS1 could be done as an enhancement later

dave-tucker:
next round of updates force pushed. we're now using `etc/profile.d/namespace.sh` to add a namespace indicator to the `$PS1` and the `motd` has been rewritten to be a reminder to be conscious of namespaces instead of explicitly indicating that you aren't in the root ns.

justincormack:
"you are in a maze of twisty little namespaces, all alike"

deitch:
Zork!

dave-tucker:
I will happily change the MOTD to that and implement add `north`, `south`, `east`, `west` and `look` to `/bin`

rn:
looks like getty is not pushed (maybe ssh as well)

ijc:
@rn Dave said in the top comment that he does have the perms to push. Do we normally have someone pull the PR and do it manually?

justincormack:
@ijc yes

dave-tucker:
I can push, but I can't sign 😢

rn:
pushing/signing. will hit rebuild

deitch:
LGTM. We still have the open question if we want to map `/:/hostroot` just to make it straightforward to access everything? Another few chars on the `Dockerfile`?

PR #2124, opened 2017-06-29
DfM: Add support for exposing swarm service ports

MagnusS:
Add support for forwarding swarm service ports to the host in the Docker for Mac blueprint. This should now work: 

`docker service create --publish 8080:80 nginx`

This PR adds a new package for binaries that are not part of the `pkg/vpnkit-forwarder` service called `pkg/vpnkit-expose-port`. The two binaries in this package are `vpnkit-expose-port` (userland proxy) and `vpnkit-iptables-wrapper` (iptables wrapper for swarm, new in this PR). 

These binaries have to be either copied into the dockerd container (e.g. in a multistage build) or bind-mounted into it by using `pkg/vpnkit-expose-port` as an init-container. I've updated `docker-for-mac.yml` to use it as an init-container, as we can then use a standard `docker-ce` package (yml hashes are currently temporary).

`docker-ce` already includes the `vpnkit-expose-port` binary at the moment, but it can be removed from that package if this is merged.

There may be alternative ways to do this, for example by adding the binaries directly in `docker-ce`. However, since the new `vpnkit-iptables-wrapper` is very Docker for Mac specific it seemed better to not include it directly. If we later get support for service composition, that could probably be used too: https://github.com/moby/tool/issues/92

(this also updates to latest containerd, which was missed in https://github.com/linuxkit/linuxkit/pull/2113)

6193147deee2e648713e2b65ec84da8f

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -3,6 +3,7 @@ kernel:
   image: "linuxkit/kernel:4.9.34"
   cmdline: "console=ttyS0 page_poison=1"
 init:
+  - linuxkit/vpnkit-expose-port:b9bbd9b79c4682daec991c71934341b50772de00 # install vpnkit-expose-port and vpnkit-iptables-wrapper on host
   - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
   - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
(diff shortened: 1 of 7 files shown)


justincormack:
Reminds me we need a better long term solution than wrapping iptables!

justincormack:
@riyaz can you sort out the repos and give Magnus access to all the vpnkit ones?

riyazdf:
added @MagnusS and created `linuxkit/vpnkit-expose-port`

MagnusS:
@riyazdf thanks! pushed an update to vpnkit-forwarder (linuxkit/vpnkit-forwarder:0fc6a1d9a3d7a677d3c165b8f5f6922a5f558bc8), but got access denied for vpnkit-expose-port.

riyazdf:
@MagnusS: my mistake!  Looks like hub perms weren't correct but should work now for `vpnkit-expose-port`

MagnusS:
@riyazdf works now! thanks

I've rebased the PR and updated to the real hashes

PR #2117, opened 2017-06-28
Fixes/improvements to `pkg/package.mk`

ijc:
**- What I did**

Various fixes/improvements to `pkg/package.mk`:
* Added a convenience target to print out the tag
* Simplified implementation of `-dirty` suffix and restricted it to only consider the package directory (same as for the hash itself)
* Consolidated the uses of `$(ORG)/$(IMAGE):$(HASH)`
* Fixed hash calculation for Jessie era git (loosely based on https://github.com/linuxkit/linuxkit/pull/2063#issuecomment-310465640)

**- How I did it**

Typing.

**- How to verify it**

I used:
```
$ for pkg in init containerd ; do make -C pkg/$pkg --no-print-directory show-tag; ( cd pkg/$pkg && make show-tag ); done
linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389-dirty
linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389-dirty
linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
```

On both my Stretch (ish, git 2.11.0) and in a Jessie container (git 2.1.4).

**- Description for the changelog**

Improvements to package build, including compatibility with Jessie era git tooling.

**- A picture of a cute animal (not mandatory but encouraged)**
[](https://en.wikipedia.org/wiki/The_Dream_Calls_for_Blood)

PR #2126, opened 2017-06-29
Refresh projects/swarmd and add networking support

ijc:
**- What I did**

A basic containerd executor has now been merged into swarmkit, but there are additional enhancements in my WIP branch, including CNI networking, so continue to track that for now.

This also integrates a weave container and adds the weave and reference CNI plugins to the swarmd container.

Also updates the build and resyncs with mainline linuxkit.yml.

**- How I did it**

**- How to verify it**

Buidl an image from projects/swarmd/swarmd.yml and marvel in the swarmy goodness.

Updated projects/swarmd.

[](https://www.metal-archives.com/albums/Bison_B.C./Quiet_Earth/208724)

### DIFF
--- a/projects/swarmd/README.md
+++ b/projects/swarmd/README.md
@@ -3,27 +3,33 @@
 This adds a `swarmd` package for Moby which contains the standalone
 swarmkit orchestration daemon (`swarmd`) and CLI tool (`swarmctl`).
 
-The package tracks [docker/swarmkit#1965][PR1965] which
-is a WIP PR adding a containerd executor to swarmkit.
+The package tracks [ijc's `container-wip` branch][containerd-wip].
+Compared with mainline swarmkit (which container a basic containerd
+executor merged in [PR1965]) this reworks the executor to use the
+container client library and adds support for CNI networking.
 
-With a suitable moby image (such as `swarmd.yml` from this directory)
-something like this should work:
+With a suitable LinuxKit image (such as `swarmd.yml` from this
+directory) something like this should work:
 
-    runc exec swarmd swarmctl service create --image docker.io/library/nginx:alpine --name nginx
-    runc exec swarmd swarmctl service ls
+    ctr exec -- swarmd swarmd swarmctl service create --image docker.io/library/nginx:alpine --name nginx
+    ctr exec -- swarmd swarmd swarmctl service ls
+
+Note that `swarmd` uses the "swarmd" containerd namespace, so to see
+swarmd managed containers you will need to use `-n swarmd` on all
+`ctr` commands e.g.:
+
+    ctr -n swarmd containers ls
+
+Alternatively you may export `CONTAINERD_NAMESPACE=swarmd`.
 
 ### TODO
 
-Currently the swarm state directory needs to be at a path which is
-identical from the PoV of both the `containerd` and `swarmd`
-processes. For now this means that the swarmkit state is put in
-`/var/lib/containerd/swarmd`.
-
-Bootstrapping a cluster needs more invesigation. Tokens and join
+Bootstrapping a cluster needs more investigation. Tokens and join
 addresses can currently only be passed on the `swarmd` command line
 which is inconvenient for automated image deployment.
 
 Swarmkit [PR 1965][PR1965] also contains a number of TODOs which are not
 separately listed here.
 
-[PR1665]: https://github.com/docker/swarmkit/pull/1965
+[PR1965]: https://github.com/docker/swarmkit/pull/1965
+[containerd-wip]: https://github.com/ijc/swarmkit/tree/containerd-wip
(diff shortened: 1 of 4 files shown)


projects/swarmd/README.md
```diff
@@ -3,21 +3,26 @@
...
 
-Currently the swarm state directory needs to be at a path which is
-identical from the PoV of both the `containerd` and `swarmd`
-processes. For now this means that the swarmkit state is put in
-`/var/lib/containerd/swarmd`.
+Alternatively you may export `CONTAINERD_NAMESPACE=swarmd`.
+
+### TODO
 
 Bootstrapping a cluster needs more invesigation. Tokens and join
```
justincormack: typo "invesigation"

projects/swarmd/swarmd.yml
```diff
@@ -30,17 +34,45 @@ services:
     image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
   - name: ntpd
     image: "linuxkit/openntpd:a4c642d52e985922fcd97db52e471db123cc6841"
+  - name: weave
+    image: weaveworks/weave:e8fdf3e549348592f5802a730562e175e05bff8a63d99051440ebb933f6e7ef2 # Must match swarmd/Dockerfile
+    command: ["/bin/sh", "/home/weave/weaver-wrapper"]
+    capabilities:
+     - all
+    pid: host
+    net: host
```
justincormack: no longer needed

projects/swarmd/swarmd/Dockerfile
```diff
@@ -1,22 +1,45 @@
...
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    busybox \
+    ca-certificates \
+    iptables \
+    musl \
+    util-linux
+
```
rn: you probably want to add a `RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache` here to remove the temporary `apk` files

ijc: Yes, I seem to have managed to drop that during one of my rebases.

projects/swarmd/swarmd/Dockerfile
```diff
@@ -1,22 +1,45 @@
+FROM weaveworks/weave:e8fdf3e549348592f5802a730562e175e05bff8a63d99051440ebb933f6e7ef2 AS weave
```
rn:
```
$ docker pull weaveworks/weave:e8fdf3e549348592f5802a730562e175e05bff8a63d99051440ebb933f6e7ef2
Error response from daemon: manifest for weaveworks/weave:e8fdf3e549348592f5802a730562e175e05bff8a63d99051440ebb933f6e7ef2 not found
```
maybe there is a `sha256` missing?

projects/swarmd/swarmd.yml
```diff
@@ -30,17 +34,44 @@ services:
     image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
   - name: ntpd
     image: "linuxkit/openntpd:a4c642d52e985922fcd97db52e471db123cc6841"
+  - name: weave
+    image: weaveworks/weave:e8fdf3e549348592f5802a730562e175e05bff8a63d99051440ebb933f6e7ef2 # Must match swarmd/Dockerfile
```
rn: see below. `sha256` missing?

projects/swarmd/swarmd/Dockerfile
ijc: I guess I missed this because I had it locally already...

projects/swarmd/swarmd/Makefile
```diff
@@ -1,15 +1,4 @@
-.PHONY: tag push
-all: push
+ORG?=linuxkitprojects
+include ../../../pkg/package.mk
```
riyazdf: I think you'll need `NETWORK=1` because the `docker build` requires networking for git cloning

ijc: `pkg/package.mk` has:
```
ifndef $(NETWORK)
NET_OPT=
else
NET_OPT=--network=none
endif
```
So setting `NETWORK=1` would appear to disable the network. Which is a bit confusing because e.g. `pkg/runc` sets `NETWORK=1` and does a clone, and appears to work fine.

Ah, I see the error to work as the author (@dave-tucker I think?) expected it would need to be `ifndef NETWORK`, as it is it is expanding to either `ifndef` (i.e. no argument at all) or `ifndef 1` (if `NETWORK=1`) both of which apparently evaluate to true (if it were possible ot assign to a variable named `1` things might be different).

Rather than fix the syntax here thought I'd suggest we fix the semantics and use `ifneq ($(NETWORK),1)` (or `y`) which I think will match what people expect. I'll add that to #2117 I guess, I'll also add the NETWORK=1 here in preparation.

ijc: Actually, fixing it in this way required the `NETWORK=1` to be moved above the `include` in all the package makefiles. So rather than attaching to #2117 (which is already a bit fraught), I'll open an issue.

justincormack:
Network fix merged so you can update

PR #2127, opened 2017-06-29
Update Docker CE to 17.06.0

justincormack:
baby-owls

### DIFF
--- a/pkg/docker-ce/Dockerfile
+++ b/pkg/docker-ce/Dockerfile
@@ -31,13 +31,14 @@ RUN set -x \
 	&& echo 'dockremap:165536:65536' >> /etc/subuid \
 	&& echo 'dockremap:165536:65536' >> /etc/subgid
 
-ENV DOCKER_BUCKET get.docker.com
-ENV DOCKER_VERSION 17.05.0-ce
-ENV DOCKER_SHA256 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a95d94316d9
+# DOCKER_TYPE is stable, edge or test
+ENV DOCKER_TYPE stable
+ENV DOCKER_VERSION 17.06.0-ce
+ENV DOCKER_SHA256 e582486c9db0f4229deba9f8517145f8af6c5fae7a1243e6b07876bd3e706620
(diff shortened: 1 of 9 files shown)


pkg/docker-ce/Dockerfile
```diff
@@ -31,13 +31,14 @@ RUN set -x \
...
-ENV DOCKER_SHA256 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a95d94316d9
+# DOCKER_TYPE is stable, edge or test
+ENV DOCKER_TYPE stable
+ENV DOCKER_VERSION 17.06.0-ce
+ENV DOCKER_SHA256 e582486c9db0f4229deba9f8517145f8af6c5fae7a1243e6b07876bd3e706620
 
 # we could avoid installing client here I suppose
 RUN set -x \
-	&& curl -fSL "https://${DOCKER_BUCKET}/builds/Linux/x86_64/docker-${DOCKER_VERSION}.tgz" -o docker.tgz \
+	&& curl -fSL "https://download.docker.com/linux/static/${DOCKER_TYPE}/$(uname -m)/docker-${DOCKER_VERSION}.tgz" -o docker.tgz \
```
rn: not sure I like the `uname -m` here...

justincormack: I wanted it to work for arm64 in future (although we are not currently releasing arm64 binaries but hey), rather than hard coding `x86_64` - the paths have always been designed to be `uname -m` compatible deliberately I believe.

riyazdf: confirmed that the linuxkit/alpine image `uname -m` returns `x86_64`

rn: I was more concerned about cross-building images and then `uname -m` won't work. But really a minor issue

rn:
it's green, so merging

pkg/docker-ce/Dockerfile
justincormack: At the moment we are aiming for native build not cross. Cross with `binfmt` should set the right `uname` anyway

Issue #2045, opened 2017-06-15
RFC: opening point-to-point channels between containers


samoht:
Hi there,

There are some use-cases (for instance the MirageSDK) where we need to open secure point-to-point channels between containers managed using the minimal amount of privileges. The usual way to do this is to use a technique called [fd passing](https://sumitomohiko.wordpress.com/2015/09/24/file-descriptor-passing-with-sendmsg2-and-recvmsg2-over-unix-domain-socket/) where we use `sendmsg / recvmsg` to pass open fds on a unix domain socket between two local processes: this can let a privileged process deal with the logic of opening the fds before sending them to an unprivileged process for doing the read/write.

Here a proposal to do that for managing secure channels between containers, baked in the moby yaml file format. This needs patches to containerd client (and/or init), containerd, and maybe need to add a shim (I am not very familiar with the internals of containerd, so please correct me if I am wrong).

## format change

I propose to define a new section for secure channels, where we could (in the future) add more options about the transport (for instance using encryption, etc). We can even imagine allowing broadcast (e.g. allowing more than 2 participants), using a remote endpoint (with only 1 local participant), etc.

```
channels:
   - name: blah
```

this can be used by containers using the `channels` keyword:

```
services:
   - name: foo
      channels: blah, very-secure-one
```

## containerd client (some of this stuff could be done by the init process too)

For every container needing at least one channel:

- bind and listen on a unix domain socket (for instance `/tmp/stdin/<container-id>.sock`)
- create an empty paylod queue of file descriptors
- run a thread which accept only one connection (to that unix domain socket) and exit.
- When a connection is accepted, send the contents of the payload queue and exit.
- use `/tmp/stdin/<container-id>.sock` in the container's `config.json` as the `stdin` entry

For every top-level channel:

- open a socketpair, add the created fds to the payload queue of both ends of the channel

## containerd

- modify containerd to accept unix domain sockets as inputs for stdin/stdout/stderr (instead of just fifo files). In the case of sockets, should use `connect` instead of opening the fifo.
- an/or modify containerd to accept extra fd/files instead of overwriting the semantics of `stdin` (not sure if it is a good idea, as https://github.com/containerd/containerd/pull/592 was not very well received upstream)

## container-shim(?) or inside the container

_note: I am not sure about this part, would be great to have some feedback or alternative ideas_

- before the container start (if using a shim, or after it started otherwise), use recvmsg to receive the payload of fds. Dup them at the right place if needed.

[...]

PR #2129, opened 2017-06-29
sdk: add a file-descriptor sharing server

samoht:
```
$ fdd init &
$ fdd share /tmp/foo # serve a fresh socketpair on that path
$ fdd test /tmp/foo  # read both sides of the socketpair and test that it works
```

Instead of `fdd test` (which is only useful for testing), users are expected to
connect to the unix domain socket and call `recvmsg(2)`. They will get one side
of the socketpair. Two different processes can do this and they will be able to
talk to each other.

### DIFF
--- a/projects/miragesdk/src/fdd/.gitignore
+++ b/projects/miragesdk/src/fdd/.gitignore
@@ -1,0 +1,4 @@
+*.install
+.merlin
+_build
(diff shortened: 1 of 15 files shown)


samoht:
This is related to #2045

PR #2130, opened 2017-06-29
Bind /etc/docker/daemon.json in examples/docker.yml 

caminada:
**- What I did**
bind /etc/docker/daemon.json so dockerd can read it

**- How to verify it**
With following setting the dockerd is also listening on TCP:

    files:
         - path: etc/docker/daemon.json
           contents: '{"debug": false, "hosts":["tcp://0.0.0.0:2375","unix:///var/run/docker.sock"]}'

red-ara

### DIFF
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -40,6 +40,7 @@ services:
     binds:
      - /var/lib/docker:/var/lib/docker
      - /lib/modules:/lib/modules
+     - /etc/docker/daemon.json:/etc/docker/daemon.json
 files:
   - path: etc/docker/daemon.json
     contents: '{"debug": true}'


ijc:
The `containers/services/docker/rootfs/` paths are not an ABI i.e. not supported. The right fix is to bind /etc/docker/daemon.json from the host into the docker container I think.

examples/docker.yml
```diff
@@ -41,7 +41,7 @@ services:
      - /var/lib/docker:/var/lib/docker
      - /lib/modules:/lib/modules
 files:
-  - path: etc/docker/daemon.json
+  - path: containers/services/docker/rootfs/etc/docker/daemon.json
```
rn: this is not the right way to do this. The container rootfs is off limits as it may change once we switch to the containerd storage. it's best to bind mount the json file into the docker container

caminada:
Thanks for the input.
I changed it to bind the file instead of using the wrong path in my patch.

examples/docker.yml
```diff
@@ -40,8 +40,9 @@ services:
     binds:
      - /var/lib/docker:/var/lib/docker
      - /lib/modules:/lib/modules
+     - /etc/docker/daemon.json:/etc/docker/daemon.json
 files:
-  - path: etc/docker/daemon.json
+  - path: containers/services/docker/rootfs/etc/docker/daemon.json
```
justincormack: you should remove this change, with the bind it is not needed

caminada: of course...

rn:
could you also squash the commits into one?

justincormack:
Can you also add the bind in https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac.yml

MagnusS:
@justincormack the bind is already in docker-for-mac.yml, it's in /var/config/docker: https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac.yml#L76

Plan is to replace /var/config with config from host

justincormack:
@MagnusS ah ok, missed that!

PR #2119, opened 2017-06-28
Add Docker for Mac host time sync daemon

djs55:
**- What I did**

Cause the Docker for Mac VM clock to be resynchronised when the host awakes from sleep.

**- How I did it**

Host-side code on Docker for Mac `connect`s to an `AF_VSOCK` socket (port `0xf3a4`) whenever it detects that the Mac has resumed from sleep. This connection is received by a daemon called `host-timesync-daemon` in the VM which reads `/dev/rtc0` and calls `settimeofday` (this requires `CAP_SYS_TIME`)

The first version of this code was originally written in C by @MagnusS -- I've ported it to Go in this PR.

**- How to verify it**

- `moby build docker-for-mac.yml`
- `linuxkit run docker-for-mac`
- `tail -f /var/log/host-timesync-daemon.log `
- on the Host in the `blueprints` directory: `echo "00000003.0000f3a4" | nc -U ./docker-for-mac-state/connect `
- observe in the log entries like
```
2017/06/28 16:48:58 Connection to: ffffffff:f3a4 from: 00000002.04000003
2017/06/28 16:48:58 Setting system clock to 2017-06-28 16:48:59 +0000 UTC
2017/06/28 16:49:01 Connection to: ffffffff:f3a4 from: 00000002.04000004
2017/06/28 16:49:01 Setting system clock to 2017-06-28 16:49:01 +0000 UTC
```

PR #2132, opened 2017-06-29
Add pkg/chronyd and update DfM blueprint

MagnusS:
`chronyd` is used by Docker for Mac to synchronise time. This adds a package for starting `chronyd`, adds `chrony` to the alpine image and updates the Docker for Mac blueprint.

(Alpine and pkg/chronyd hashes are temporary as I can't update them)

tincancat

### DIFF
--- a/blueprints/docker-for-mac.yml
+++ b/blueprints/docker-for-mac.yml
@@ -5,7 +5,7 @@ kernel:
 init:
   - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
   - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
-  - linuxkit/containerd:b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a
+  - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
 onboot:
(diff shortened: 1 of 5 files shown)


rn:
what's the advantage of `chronyd` over `openntp`? I recall that `chronyd` had issues synching with the host NTP server while openntp seemed more stable

MagnusS:
@rn We could experiment with another ntp server, but right now `chronyd` with the current config + the timesync daemon from #2119 seems to work ok based on issues reported (although probably still a bit broken, a proper fix would be to implement pvclock). If we change it we should make sure that we at least not make it worse. I’m not sure if `openntpd` was also tested when `chronyd` was chosen (maybe @justincormack remembers more about this).

justincormack:
We originally went with `chronyd` for d4m as it was supposed to be better for laptop type use cases, but if we force a sync like we do now I would think there would be little difference. Maybe run it with `openntpd` for a bit letting it shutdown and check it works ok.

MagnusS:
Ok, I'll run some tests with `openntpd` when #2119 is ready

MagnusS:
Closing this PR for now as it will conflict with #2137. Will reopen if we can't use openntpd

PR #2133, opened 2017-06-30
sdk: add a yml example on how to use fdd to create container channels

samoht:
Lots of boilerplate for now on, will work on upstreaming that in the tool
properly if needed later.

### DIFF
--- a/projects/miragesdk/examples/fdd.yml
+++ b/projects/miragesdk/examples/fdd.yml
@@ -1,0 +1,37 @@
+kernel:
+  image: "linuxkit/kernel:4.9.34"
+  cmdline: "console=ttyS0 page_poison=1"
+init:
+  - linuxkit/init:36c56f0664d49c5a6adc1120d1bf5ba6ac30b389
+  - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
+  - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
+  - linuxkit/ca-certificates:46b59484919bfa9af700e54e042048cb06261de4
+  - samoht/fdd
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
+services:
+  - name: getty
+    image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
+    env:
+     - INSECURE=true
+  - name: rngd
+    image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:6c1ca76dbf808d5c27d10cbf22a8d4399be5c8ae"
+files:
+  - path: etc/init.d/020-fdd-init
+    mode: "0700"
+    contents: |
+      #!/bin/sh
+      /bin/fdd init -vv -s /run/fdd.sock
+  - path: etc/init.d/030-fdd-share
+    mode: "0700"
+    contents: |
+      #!/bin/sh
+      /bin/fdd share -vv -s /run/fdd.sock /tmp/channel-net-eng &&
+      /bin/fdd share -vv -s /run/fdd.sock /tmp/channel-conf-net &&
+      /bin/fdd share -vv -s /run/fdd.sock /tmp/channel-conf-act
+trust:
+  org:
+    - linuxkit
--- a/projects/miragesdk/src/fdd/Dockerfile
+++ b/projects/miragesdk/src/fdd/Dockerfile
@@ -3,10 +3,10 @@ RUN sudo apk add m4
 RUN opam install jbuilder lwt fd-send-recv logs fmt cmdliner
 ADD . /src
 RUN opam pin add fdd /src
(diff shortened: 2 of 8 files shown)


projects/miragesdk/examples/fdd.yml
```diff
@@ -0,0 +1,40 @@
...
+    env:
+     - INSECURE=true
+  - name: rngd
+    image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:6c1ca76dbf808d5c27d10cbf22a8d4399be5c8ae"
+files:
+  - path: etc/init.d/020-fdd
+    mode: "0700"
+    contents: "/bin/fdd init -vv -s /var/run/fdd.sock"
```
samoht: Actually it's a bit overkilling to use `onboot` service to create the channels, maybe it should only be in the `fdd` init script.

projects/miragesdk/examples/fdd.yml
```diff
@@ -0,0 +1,40 @@
...
+  - linuxkit/runc:291131ec026430371e7c36165c3f43734fbc2541
+  - linuxkit/containerd:1e3e8f207421de8deac8cedc26a138d6b1661a0d
+  - linuxkit/ca-certificates:46b59484919bfa9af700e54e042048cb06261de4
+  - samoht/fdd
+onboot:
+  - name: sysctl
+    image: "linuxkit/sysctl:3aa6bc663c2849ef239be7d941d3eaf3e6fcc018"
+  - name: channel-net-eng
+    image: samoht/fdd
+    binds: [ "/var/run/fdd.sock:/var/run/fdd.sock" ]
```
justincormack: please don't use `/var/run` its just `/run`, `/var/run` is just a symlink.

samoht:
I've fixed the yaml example. I also added a `fdd exec` to wrap command in an environment where file-descriptor are pre-open. For instance, this now works:

```
$ fdd init &
$ fdd share /tmp/foo
$ fdd exec -m /tmp/foo:1 -- echo hello world!
$ fdd exec -m /tmp/foo:0 -- cat
hello world!
```

samoht:
At the moment, it's just wrapper script to be able to prototype stuff. I don't expect end user to see anything of this -- I plan to push some of these stuff to the moby tool (for instance, use a new `channel` section instead of having to modify init scripts) and I will experiment by changing `container-shim` instead of having to call `fdd exec`.

Which part of the current design do you find ugly?

justincormack:
The whole point of anonymous socketpairs is to control exactly which process gets the two ends of the channel, and you totally lose this if you just get it via the filesystem instead, when some other process could race to get the other file descriptor.

samoht:
But if you control the init scripts and the packaging, you can decide which process can see the unix domain socket (via bind mounts for instance) and in which order. Do you have a better idea?

The good thing about the current approach is that it's easy enough to test that all this (rather complex) is working..

samoht:
e.g. it's easy to extend to something like that:

```
 $ fdd init &
$ fdd share /tmp/foo --secret foo
$ fdd exec -m /tmp/foo:1 --secret foo -- echo hello world!
$ fdd exec -m /tmp/foo:0 -- cat
<error>
$ fdd exec -m /tmp/foo:0 --secret foo -- cat
hello world!
```

projects/miragesdk/examples/fdd.yml
```diff
@@ -0,0 +1,34 @@
...
+    image: "linuxkit/rngd:b50b22dd574c5377708977af769f053009fff6d5"
+  - name: dhcpcd
+    image: "linuxkit/dhcpcd:6c1ca76dbf808d5c27d10cbf22a8d4399be5c8ae"
+files:
+  - path: etc/init.d/020-fdd-init
+    mode: "0700"
+    contents: "/bin/fdd init -vv -s /run/fdd.sock"
+  - path: etc/init.d/030-fdd-share
+    mode: "0700"
+    contents: |
```
justincormack: these files need to start with `#!/bin/sh`

justincormack:
Thats no better than just having one program listen on a unix socket in the filesystem while the other one connects to it, but much more complicated.

samoht:
Yes, this is indeed more complicated but because I am trying to limit the number of syscalls and the number of operations that the 2 processes have to do on startup.

If you have 2 processes and A needs to listens and accept on `/tmp/foo` and B connects on `/tmp/foo`. A needs to be started before B and/or B needs to wait for A to be ready to accept connections. A and B also needs to both see the same unix domain socket in their namespace.

With `fdd`, A and B can just read/write on the given file-descriptor (for instance fd 3), with:
```
fd exec /tmp/foo:3 -- ./A
fd exec /tmp/foo:3 -- ./B
```

samoht:
Also this let change the way the two side of the socketpair are created, for instance using a shared pages + ring buffer, or network connection, etc. E.g. the process won't care as it will still read from fd 3, but the init scripts (done by `fdd` or something similar) could handle a more complex connection startup logic.

projects/miragesdk/examples/fdd.yml
samoht: thx, force-pushed a fix

justincormack:
Going to merge as just a project example.

PR #2137, opened 2017-06-30
Split DfM blueprint in base and docker yml

MagnusS:
This splits the Docker for Mac blueprint into a base and docker image. This allows the base to be built and tested with different versions of Docker in the future.

Bonus updates:
- vsock port now same as in Docker for Mac (was 2375, now 2376)
- update containerd to the version missed in https://github.com/linuxkit/linuxkit/pull/2113


MagnusS:
Updated to fix the test (quotes missing)

PR #2138, opened 2017-06-30
getty: Mount all of hostroot and use host profile.d

dave-tucker:
This commit does two things:

- Changes the mount from /etc:/hostroot/etc to /:/hostroot
  This allows for easier to access to other binaries that we added
  during the init section.
- The getty container will now add any files in /etc/profile.d to its
  own profile.d directory. This allows the user to add /hostroot/usr/bin
  to their path automatically.

A concrete example of where this is useful is with my CI worker I'm building with linuxkit.
Before this PR, I had to re-add **ALL** of the getty binds and add an additional bind for every binary and library that I added in my `init` containers.

e.g

```
name: getty
  image: "linuxkit/getty:6cbeee0392b0670053ce2bf05a5a0d67ec2bce05"
  env:
    - INSECURE=true
  binds:
    # defaults
    - "/run:/run"
    - "/tmp:/tmp"
    - "/etc:/hostroot/etc"
    - "/usr/bin/ctr:/usr/bin/ctr"
    - "/usr/bin/runc:/usr/bin/runc"
    - "/usr/bin/dist:/usr/bin/dist"
    - "/var:/var"
    - "/containers:/containers"
    - "/dev:/dev"
    - "/sys:/sys"
    # our stuff
    - "/usr/local/bin/make:/usr/local/bin/make"
    # I can't even begin to add a line for every git binary!
```

Now I use this:
```
services:
  name: getty
    image: "davetucker/getty:localdev"
    env:
      - INSECURE=true
files:
  - path: etc/profile.d/hostroot.sh
    contents: |
      export PATH="$PATH:/hostroot/bin:/hostroot/sbin:/hostroot/usr/bin:/hostroot/usr/sbin"
      export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/hostroot/usr/lib"
      export GIT_TEMPLATE_DIR="/hostroot/usr/share/git-core/templates"
```

### DIFF
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -29,4 +29,4 @@ COPY --from=mirror /out/ /
 COPY usr/ /usr/
 COPY etc/ /etc/
 CMD ["/usr/bin/rungetty.sh"]
-LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/tmp:/tmp", "/etc:/hostroot/etc", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/usr/bin/dist:/usr/bin/dist", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
+LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/tmp:/tmp", "/:/hostroot", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/usr/bin/dist:/usr/bin/dist", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -49,6 +49,15 @@ if [ -f $ROOTSHADOW ]; then
 	echo >> /etc/shadow
 fi
 
+# check for scripts that should be added to profile.d
+PROFILED=/hostroot/etc/profile.d/
+for f in ${PROFILED}*.sh; do
+	filename="$(basename ${f})"
+	if [ ! -f "/etc/profile.d/${filename}" ]; then
+		cp "${f}" /etc/profile.d/
+	fi
+done
+
 for opt in $(cat /proc/cmdline); do
 	case "$opt" in
 	console=*)


justincormack:
If you want to be in the host namespace you should just be there, mounting everything in pretty much destroys the point. Also the point is not to have all sorts of stuff in the host namespace at all, for example `ctr` and `dist` will be removed at some point as the host namespace should only have the absolute minimum needed to boot.

deitch:
I definitely agree with the `/:/hostroot` mapping.

I am unsure of the `profile.d` mounting. 

1. Do we want everything from the host's `/etc/profile.d/` in the container's? Philosophically/architecturally speaking?
2. Whether we do or do not want it, is using `/etc/profile.d/` on the host as a "transfer station" to get it into `getty` container a good idea? 

In the CI example above, I get *why* you put in `/etc/profile.d/hostroot.sh`, but it puts things in host that properly need be only in the `getty` container.

this may be part of a larger question of, "how do I get files into just one or several images"? I see a few ways:

1. Using `files:` as a "waystation" (which is what the above does).
2. Extending the image, so you might have `FROM linuxkit/getty` and build your own that you use.
3. Some other "file transit" mechanism.

Number 2 probably is the "purest" (containerally-speaking), but I see why people wouldn't want to build special images for a common config use case like files.

Number 1 is problematic in polluting the root system, but maybe that is what `files:` is for?

Number 3 would be nice if we could think of a straightforward way. Again, maybe I am misunderstanding, this is exactly what `files:` is for, and we accept the pollution (and possibly conflict issues)?

deitch:
> for example ctr and dist will be removed at some point as the host namespace should only have the absolute minimum needed to boot.

We probably should have a `getty` and `sshd` PR that puts the client binaries for `runc` and `ctr` directly in the image.

We probably first should have a PR that merges `sshd` and `getty`!

justincormack:
The plan with `dist` and `ctr` was to build them in `alpine` (which now has the `containerd` hash, so that any container could easily add them.

deitch:
> The plan with dist and ctr was to build them in alpine (which now has the containerd hash, so that any container could easily add them

Makes sense. Will wait for the PR until then.

dave-tucker:
> Number 2 probably is the "purest" (containerally-speaking), but I see why people wouldn't want to build special images for a common config use case like files.

If 2 is the answer I'd be fine with that and I can go ahead and close this PR.

deitch:
> If 2 is the answer I'd be fine with that and I can go ahead and close this PR.

Even if 2 is the answer, please don't close the PR. It is worth it for the `/:/hostroot` in any case.

"Purest" does not necessarily equal "best". I agree that it is a burden/overhead to create a special getty image and use it just to get some config files in; personally, I would not want to do it.

Let's wait to get some feedback from the rest of the crew here. My view is just one, and (as I wrote above), I am not intimately familiar with the history and original purpose of `files:`.

On that front... how about if we adopted the concept of "data volume containers" here? What if we could have a `files:` equivalent that went not into the host filesystem, but into a special data-volume container that is then shared to the desired other containers (`services`, `onboot`, etc.)?

That certainly would obviate the pollution issue, and would eliminate any chance of a file upload conflicting with core OS functionality.

justincormack:
A volume would just be a special path prefix, but it would at least separate the things for each container.

justincormack:
But I really don't think `hostroot` is a good idea. You already have it at `/proc/1/root` if you have the  host pid namespace, so you can use that if you really need to, but you should not be depending on it.

But you should explain your CI worker. What does it really need to do? Why does CI have a getty at all?

riyazdf:
I'm also -1 for `/:/hostroot` for the reasons that @justincormack has enumerated above - I agree it'd be most helpful to walk through the CI worker usecase so we can drill down on what it needs and the best way to solve them

dave-tucker:
> But you should explain your CI worker. What does it really need to do? Why does CI have a getty at all?

`getty` is just for local debugging and would be disabled in prod I think but as I had it enabled this is where I first found the issue.

In production, it's the same problem but with `sshd` instead of `getty`.
The linuxkit controller `ssh`'s to our worker and runs a script called `test.sh` which accepts a tarball of the source on stdin: https://github.com/linuxkit/linuxkit-ci/blob/ab99f610823a27112f76a577a488555760f3ccae/src/linuxkit_build.ml#L245

```
#!/bin/bash
set -eux
rm -rf /tmp/build
mkdir /tmp/build
cd /tmp/build
tar xf -
if [ -x ci-setup.sh ]; then ./ci-setup.sh; fi
make ci-pr
```
This then calls `make ci-pr` and we're in familiar territory once again: https://github.com/linuxkit/linuxkit/blob/master/Makefile#L81

In order for `make` to work, we need `git`, `make` and `docker` available to `sshd`.

My initial approach was to package each seperately and add them to `init` and bind them in to `sshd` or `getty`.

Since opening this PR, I've now created my own `sshd` package as follows:
```
FROM linuxkit/alpine:c608d404c1cb36cce0c7d9303e30b52c9d81ccf0 as alpine
RUN apk add git make

FROM linuxkit/docker-ce:9e125aa533108731d11f6d7ec17aba6ded0cb4eb as docker

FROM linuxkit/sshd:f095b62ddca658e99d5751872c933ef3d5d18cec

COPY --from=alpine /usr/bin/make /usr/bin/
COPY --from=alpine /usr/lib/libssh* /usr/lib/libcurl* /usr/lib/libexpat* /usr/lib/libpcre* /usr/lib/
COPY --from=alpine /usr/libexec/git-core /usr/libexec/git-core/
COPY --from=alpine /usr/share/git-core /usr/share/git-core/
COPY --from=alpine /usr/bin/git* /usr/bin/
COPY --from=docker /usr/bin/docker /usr/bin
```

rn:
for CI we should have separate dev/ci package which is based off alpine base, which may have the CIs public ssh keys burned in etc.  The approach you take above is a bit weird and fragile (trying to guess and copy all the relevant bits installed from `apk add git make` will likely break at some point. You also have a dependency in the ssh package which I don't like as updating requires another step.

I would just copy the ssh setup and add `make` and `git` to the packages.

deitch:
> But I really don't think hostroot is a good idea. You already have it at /proc/1/root if you have the host pid namespace, so you can use that if you really need to, but you should not be depending on it.

When we first did `getty`, I needed something from the host `/etc` (do not recall at the moment), and used `/proc/1/root`, but the general sense was that was a bad idea; if we need something mounted, do it explicitly. Why are we now encouraging going via `/proc/1/root` vs an explicit mount of `/`?

deitch:
> In order for make to work, we need git, make and docker available to sshd.

I think this raises the same issues as before. Sometimes people log in (via getty or ssh) because they want to use what is there. Other times (like @dave-tucker 's CI case) they need additional tooling. What is the best way to get additional "files" (in this case lots of executables and libs, especially with `git`) installed. What is the best way to do it?

I lean towards the docker-inspired patterns:

* "data-volume pattern" for getting config files into a container (some work would need to be done here)
* extend the base container (`sshd` or `getty`) to make a custom container if you need more things in it

That having been said (especially the second point) moby and linuxkit really are about _composing_ parts together (like Lego) to build better and bigger things. Is the "extend" pattern the best and easiest? Or is there a better pattern?

rn:
> When we first did getty, I needed something from the host /etc (do not recall at the moment), and used /proc/1/root, but the general sense was that was a bad idea; if we need something mounted, do it explicitly. Why are we now encouraging going via /proc/1/root vs an explicit mount of /?

The difference is that `/` from the root is not *needed* and is just a convenience. So, if you want to poke around the root fs use `/proc/1/root`. IIRC in your case you *needed* something specific from the root fs and in that case it's better to be explicit

rn:
re what should go in the ssh/getty containers. I think there are two distinct use case.

- Some folks want a login and then debug/play etc. For these we should enable `apk` in the `getty` and `sshd` containers and then they can install whatever they like.
- for CI we need/want to tightly control what's installed and which versions. So for these we should bite the bullet and provide separate/distinct pacakges, ideally derived from the base sshd/getty container, but if that's too tedious duplicate the little bit of setup

deitch:
> IIRC in your case you needed something specific from the root fs and in that case it's better to be explicit

Yeah, that makes sense. But keeping in mind that the primary reason for logging in (getty or ssh) is to do things on the system, I would think accessing stuff from `/` would be pretty common. Admittedly, `/containers` and `/etc` are the most common, and hence we mount them, but even so.

My preference to expand it to `/:/hostroot` is largely driven by Dave's use case. When someone _needs_ something else from `/`, they are going to have to redo the entire set of labels. Seems easier to just have `/` mounted in and be done with it.

Then again, once we get to the point where there is almost *nothing* in `/`, we really won't need it./ The important areas - `/etc`, `/var`, `/containers`, `/dev` - already are mounted in.

deitch:
>  I think there are two distinct use case.

@rn I see the use cases, although I think they are more "use the system" and "login to do things" (like Dave's case). 

I didn't understand your second bullet point, though:

> bite the bullet and provide separate/distinct pacakges

Separate packages? Of what?

rn:
> Separate packages? Of what?

CI should have separate package which has SSH and the tools needed for CI installed and nothing else. This is different to a SSH container where a user can do `apk add` to add what they need

justincormack:
I think something like "volumes" (although its an overloaded term, maybe "data" is better) eg

`/data/etc` has stuff that is traditionally in host `/etc/` (eg timezones, say)
`/data/docker` might have docker related stuff that needs to be shared to some other containers

You can pretty much do that now, rather than just dumping everything in the host root, although it might be nice to have something that is like `init` but lets you unpack at a different path.

deitch:
> CI should have separate package which has SSH

Oh! You mean for LinuxKit CI (I thought you meant using linuxkit as part of some other user-0built toolchain). I get it now. Thanks.

justincormack:
Yes we want to self host CI, a currently it uses an indeterminate Debian machine which makes dealing with and replicating CI issues harder.

deitch:
> I think something like "volumes" (although its an overloaded term, maybe "data" is better) eg

Sure. I like that.

> You can pretty much do that now, rather than just dumping everything in the host root, although it might be nice to have something that is like init but lets you unpack at a different path

We can, nothing preventing people from "doing the right thing." But it isn't immediately obvious and the path of least resistance, so people just don't. If we had something like:

```yml
data:
   - path: /foo/bar
      source: /some/path
```

And then mounted it via:

```yml
services:
  - name: mycontainer
     data:
       - /foo/bar
```

That might work nicely. It also distinguishes between "traditional" mounts, which are rootfs-based, and "data", which comes from the special place (the fact that we store it at `/data` is irrelevant./

It has the added benefit of keeping `data:` separate from `binds:`, which may be set in the image `LABEL`, and thus eases the "add just one file" case Dave raised above.

justincormack:
We should probably open an issue for design around this...

dave-tucker:
I'm going to go ahead and close this as my use case has been addressed in the comments above

PR #2147, opened 2017-07-03
move the containerd init script to containerd package

justincormack:
They tended to get out of sync otherwise, and it makes sense like this.

hh


rn:
This seems to be missing files in `./pkg/containerd/etc` and also does not remove it from the `init` package

examples/azure.yml
```diff
@@ -2,7 +2,7 @@ kernel:
   image: "linuxkit/kernel:4.9.35"
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:24942921d1356bb801b30ca6d7197d2bfdcc26f9
+  - linuxkit/init:14a38303ee9dcb4541c00e2b87404befc1ba2083
   - linuxkit/runc:2310ad9d266cf5d9c4d07613bd2135ed7eb8a21f
   - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
```
rn: does this need updating as well?

justincormack: yes, fixed

Issue #1773, opened 2017-05-05
Needs a mechanism to respond to ACPI (or other) power events


ijc:
**Description**

Neither the `Reboot` nor `Shutdown` options in `virt-manager` cause anything to happen.

I _think_ those are generating ACPI power events and that something in the userspace needs to react to these as appropriate. I tried `apk -U add acpi` (which includes `acpid`) and ran the daemon by hand but this didn't appear to help so there must be something more required.

I didn't try a three finger salute since it was greyed out (I suspect because I hadn't configured a graphical console on the VM, just serial).

Although I am using KVM/libvirt/virt-manager I think this likely affects most virtualisation and cloud solutions as well as bare metal (where the physical power button typically generates an ACPI event on moderns systems I think).

I am using `projects/swarmd/swarmd.yml` but I don't see anything in any of the other images which would implement this.

Running `poweroff` or `reboot` manually from the console works mostly as expected, except there is no status message (I'm not sure if it is actually transitioning runlevels or not) and you can continue to interact with the console shell for several seconds afterwards until it suddenly goes dead or reboots.

**Steps to reproduce the issue:**

Start a LinuxKit image, attempt to send it a power event.

**Describe the results you received:**

Nothing happened.

**Describe the results you expected:**

I expected the system to either reboot or poweroff.

**Additional information you deem important (e.g. issue happens only occasionally):**

justincormack:
It does transition runlevels on halt - thats what the 5s delay is, and it is configured in `inittab`.

I really don't know what is involved in this - it has mainly been designed as crash-only so far, but if it is not intrusive it would be nice to support shutdown I guess...

rneugeba:
IIRC, on Hyper-V there is a special VMBus message which get's handled in the kernel, which then get's passed to user-space via `orderly_poweroff()`. I don't know the normal ACPI flow, but none of the kernel ACPI code calls `orderly_poweroff()`

ijc:
Resolved by #2154!

PR #2154, opened 2017-07-05
Add pkg/acpid

MagnusS:
This adds a new service for running busybox's `acpid`. It needs `busybox-initscripts` for `/etc/acpi`, which contains the poweroff script. I've tested this with Docker for Mac and it successfully shuts down on power button events.

Alpine hash is temporary

### DIFF
--- a/pkg/acpid/Dockerfile
+++ b/pkg/acpid/Dockerfile
@@ -1,0 +1,22 @@
+FROM linuxkit/alpine:2e362f4459ba4491655061cccdd2fcc7a2de5eb3 AS mirror
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    busybox
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM linuxkit/alpine:2e362f4459ba4491655061cccdd2fcc7a2de5eb3 AS mirror2
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    busybox-initscripts
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+COPY --from=mirror /out/ /
+COPY --from=mirror2 /out/etc/acpi /etc/acpi
+
+CMD ["/sbin/acpid", "-f", "-d"]
+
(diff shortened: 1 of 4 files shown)


MagnusS:
This may also fix #1773, but I've only tested it with Hyperkit

rn:
could you maybe remove all the other, un-needed stuff `busybox-initscripts` pulls in from `/out` and just have the required `acpi` scripts in the package?

MagnusS:
@rn I updated to copy just `/etc/acpi` from a second mirror with `busybox-initscripts`.

rn:
I have pushed/signed `linuxkit/alpine:2e362f4459ba4491655061cccdd2fcc7a2de5eb3` with your changes. but some unrelated versions changed. Can you apply the attached patch and update the hash in the `acpid` package? Sorry, this part is a bit tedious...

[foo.patch.txt](https://github.com/linuxkit/linuxkit/files/1125953/foo.patch.txt)

rn:
This also LGTM, but getting this pass CI is a bit finicky with the base image update (see https://github.com/linuxkit/linuxkit/pull/2154#issuecomment-313218541). We definitely need something which monitors external power events and re-using `acpid` for now is fine with me, though I have a feeling this might be a good candidate to be replaced with a tiny binary which just watches for power events.

MagnusS:
Thanks for the patch. I've updated the PR.

rn:
could you add a test for this under `010_platforms/010_hyperkit`? Separate PR is fine

MagnusS:
@riyazdf thanks for setting up signing - I've pushed to `linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918`
@rn I'll take a look at adding a test, do the hyperkit tests also run in ci now?

rn:
the CI doesn't run macOS tests yet, but good to add tests anyway

ijc:
@MagnusS if you can provide a suitable stanza for a the service I can try this on libvirt (cf #1773).

ijc:
PS maybe a readme or example with the stanza would be good, or maybe we can justify adding this to the default since power control seems pretty useful in all sorts of situations from bare metal to various kinds of VM.

MagnusS:
@ijc it's enabled by just adding the service, there's an example now in #2157. I could add an `acpi.yml` in /examples with just the minimum?

ijc:
@MagnusS thanks, I was expecting extra privs or binds or something! I'll give this a go.

I'd think an example would be the minimum needed, yes.

ijc:
Adding `linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918` as a service with no other options worked for me. Thanks!

PR #2158, opened 2017-07-06
Add ACPI hyperkit test and update platform docs

MagnusS:
- Add minimal acpi example
- Update hyperkit platform docs

I wasn't sure where to add the documentation for this as it is most likely similar for all platforms. I added it to hyperkit specifically though, as that's the platform I'm working on at the moment. 

This only adds a minimal example, but perhaps ACPI support should be included in more of the existing examples?

(see also previous discussion in #2154)


rn:
I think it should be fine to just update the docs and point to the blueprint and add a test,  ie I don't think we need a minimal example for this. Each example is build as part of the CI but so should the blueprint

MagnusS:
@rn ok - I removed the example and added a test instead. Modified docs to point to blueprint

PR #2161, opened 2017-07-06
pkg: Update content trust handling.

ijc:
Firstly add option to disable content trust, for the use of e.g. projects which
are pushing to the linuxkitprojects org (which has no trust setup) rather than
the main linuxkit org.

Secondly, when trust _is_ enabled then enable it globally, in particular it is
now active for the `docker build` and hence containers referenced in
Dockerfiles via "FROM" will be checked.

**- What I did**

Made it possible for individual packages to opt out of docker content trust. Use this for `projects/swarmd/swarmd`.

**- How I did it**

If `DOCKER_CONTENT_TRUST` is not already set in the environment then set it by default unless the package has opted out by setting `NOTRUST`, which I set in the swarmd package.

**- How to verify it**

`make -C projects/swarmd/swarmd push` should not ask you for your notary password. While make -C `pkg/something push` should (iff you have modified and committed to that package.

Per package control of trust on push.

[](https://en.wikipedia.org/wiki/In_the_Rectory_of_the_Bizarre_Reverend)

### DIFF
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -7,16 +7,16 @@ HASH_COMMIT?=HEAD # Setting this is only really useful with the show-tag target
 HASH?=$(shell git ls-tree --full-tree $(HASH_COMMIT) -- $(CURDIR) | awk '{print $$3}')
 
 ifneq ($(HASH_COMMIT),HEAD) # Others can't be dirty by definition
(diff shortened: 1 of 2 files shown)


ijc:
I remembered after pushing this that the other change I wanted to do post #2117 was to use `:=` in a few places to reduce the number of times some things are evaluated. I'm going to append that here.

ijc:
Extra commit appended.

PR #2164, opened 2017-07-07
Allow metadata package to over write existing config files

rn:
This should enable adding defaults to the image and over write them with metadata/userdata/


MagnusS:
Thanks for fixing this. The Docker for Mac blueprint needs an update too: https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac/base.yml#L13

rn:
thanks, no idea how I missed that. pushed update

PR #2170, opened 2017-07-08
[WIP] Memorizer project initial add

ndauten:
**- What I did**

Create initial Memorizer project subdirectory.

**- How I did it**

Followed default kernel build process. There are a few things left out like perf build to get it building. 

**- How to verify it**

Standard project build, push, moby build, and linuxkit test.


ndauten:
The current status is the system builds but does not have a clean tree. So it doesn't have a docker image pushed to the hub yet. I wanted to get this PR setup so that I could start interacting and ask a few questions. 

The build works fine, but when running the minimal .yml it appears to run out of max stack size and gets stuck. This same configuration builds and boots fine outside of the linuxkit environment. 

@riyazdf a few questions:
- license: my initial license on the memorizer src was NSCA/Illinois, I presume I need to change this?
- Is there any special kernel config options that I need to be aware of when setting things up? 
- Is there a way that I can rebuild in the same docker image? It has to do a full compile every time.
- Is there a way to get a shell in the image the Dockerfile is creating so that I can manually do a few things? 

Once you let me know about the source I can add the diff patch. I want to make sure I do things as appropriate.

Overall though, I think the boot is hanging. It occurs right after the LinuxKit  welcome message and whale are printed.

ndauten:
A printout of the stick point: 

```
[   12.366585] ALSA device list:
[   12.367390]   No soundcards found.
[   12.373411] Freeing unused kernel memory: 3336K
[   12.375834] Write protecting the kernel read-only data: 26624k
[   12.379365] Freeing unused kernel memory: 1460K
[   12.395372] Freeing unused kernel memory: 1652K
[   24.676204] input: ImExPS/2 Generic Explorer Mouse as /devices/platform/i8042/serio1/input/input3
[  111.343361] x86/mm: Checked W+X mappings: passed, no W+X pages found.
[  115.431610] ip (1131) used greatest stack depth: 27240 bytes left

Welcome to LinuxKit

                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
           \______ O           __/
             \    \         __/
              \____\_______/
```

justincormack:
Sorry my comment seems to have got lost (was on a train) - I think teh stack warnings are just because you added more debug. Can you check the boot messages with `console=tty0` removed - they may be more helpful. I fixed this upstream now.

justincormack:
Also you can run a getty in the `init` section see https://github.com/linuxkit/linuxkit/tree/master/pkg/getty#linuxkit-debug - if there are some kernel issues this should still give you a console if `runc` is failing.

avsm:
Looks like this was closed in favour of a new PR at #2171

ndauten:
Yeah, sorry @justincormack this PR had some very weird commit behaviors. Learning learning on my end.

PR #2152, opened 2017-07-05
Improve mount and format packages

dave-tucker:
**- What I did**
- `format`
  - Added supports writing partition labels
  - Added support for creating `btrfs` and `xfs` partitions
- `mount`
  - Allowed for a disk to be specified by name, label or UUID
  - Fixed automatic mount logic to take already mounted disks and partitions in to account.
- Added `extend` package to extend partitions on disks
- Extending disks with the new `extend` package
- Added test cases and updated documentation

**- How I did it**

Typing.

**- How to verify it**

Run ze tests!

**- Description for the changelog**

Add multiple disk/mountpoint support for format and mount packages

**- A picture of a cute animal (not mandatory but encouraged)**

Issue #1150, opened 2017-02-04
multiple disk support


justincormack:
Need to sanely support at least the case of a new boot disk on cloud, wit hthe data disk attached.
Check the sizes,

rn:
we now have multi-disk support I think. so closing

PR #2173, opened 2017-07-09
pkg/mount: add support for mounting a specific partiion

### LINKED ISSUE #2123: RFC: Mount partition with linuxkit/mount by label
Would be nice to mount partitions (optional) by label instead of the first partition of type linux. 
mount.sh should handle a second (optional) ARG with the partition label which should be mounted

aboyett:
**- What I did**
Added support for mounting a specific partition

**- How I did it**
Modified the linuxkit/mount image to accept an (optional) additional argument that is the absolute path to the device, or a LABEL=foo or UUID=foo identifier.

**- How to verify it**
Modify an existing linuxkit image to invoke the linuxkit/mount image with one of the following forms appropriate for your external disk:

    command: ["/mount.sh", "/dev/sdb1", "/var/data"]
    command: ["/mount.sh", "LABEL=DATA", "/var/data"]
    command: ["/mount.sh", "UUID=deadbee7-cafe-dead-c0d3-1234567890ab", "/var/data"]

pkg/mount: add support for mounting a specific partiion

image

Closes #2123

### DIFF
--- a/pkg/mount/mount.sh
+++ b/pkg/mount/mount.sh
@@ -2,6 +2,13 @@
 
 set -x
 
+# read optional device indentifier
+if [ $# -eq 2 ]
+then
+	DEV_ID="$1"
+	shift
+fi
+
 MOUNTPOINT="$1"
 
 [ -z "$MOUNTPOINT" ] && echo "No mountpoint specified" && exit 1
@@ -9,6 +16,42 @@ MOUNTPOINT="$1"
 mkdir -p "$MOUNTPOINT"
 
 mount_drive()
+{
+	# if explicit device identifier was given
+	if [ -n "$DEV_ID" ]
+	then
+		mount_specified_drive && return
+	else
+		mount_detected_drive && return
+	fi
+
+	echo "WARNING: Failed to mount a persistent volume at $MOUNTPOINT (is there one?)"
+}
+
+mount_specified_drive()
+{
+	# if explicit device identifier was given
+	if [ -n "$DEV_ID" ]
+	then
+		if [ -n "$(echo $DEV_ID | grep -E '^(LABEL|UUID)=')" ]
+		then
+			# identifier is LABEL or UUID
+			DEV=$(findfs $DEV_ID)
+			mount "$DEV" "$MOUNTPOINT" && return
+		elif [ -b "$DEV_ID" ]
+		then
+			# identifier is a block-device
+			DEV="$DEV_ID"
+			mount "$DEV" "$MOUNTPOINT" && return
+		fi
+		echo "Warning: Unknown Device identifier provided: $DEV_ID"
+	fi
+
+	# unable to mount specified device
+	return 1
+}
+
+mount_detected_drive()
 {
 	# TODO fix for multiple disks, cdroms etc
 	DEVS="$(find /dev -maxdepth 1 -type b ! -name 'loop*' ! -name 'nbd*' | grep -v '[0-9]$' | sed 's@.*/dev/@@' | sort)"
@@ -29,7 +72,8 @@ mount_drive()
 		fi
 	done
 
-	echo "WARNING: Failed to mount a persistent volume (is there one?)"
+	# unable to detect and mount device
+	return 1
 }
 
 mount_drive


deitch:
Thanks. A few questions:

1. Were the issues in #2123 resolved? There was some issue with it not mounting.
2. Can you update the docs for how to mount a disk? They are here https://github.com/linuxkit/linuxkit/blob/master/docs/external-disk.md
3. It looks like it looks for and mounts the disk by label, and then also runs the `find` in `/dev` and mounts? Is that intentional?
4. Do we have tests for disk mounting?

aboyett:
> Thanks. A few questions:
> 
>     Were the issues in #2123 resolved? There was some issue with it not mounting.
I have been able to successfully mount an external disk using this code using the command-lines given in the commit and PR descriptions.

>     Can you update the docs for how to mount a disk? They are here https://github.com/linuxkit/linuxkit/blob/master/docs/external-disk.md
Sure! Would you prefer an amendment to the existing commit, or subsequent commits in the same PR?

>     It looks like it looks for and mounts the disk by label, and then also runs the find in /dev and mounts? Is that intentional?
On successful mount, `return` is called; I mirrored this pattern in the new approach. You do make a good point that it still falls back to the autodetecting default approach on failure. Do you agree it would be reasonable to abort further mounting attempts if we fail to mount the explicitly given device identifier?

>     Do we have tests for disk mounting?
I don't see any mounting tests in https://github.com/linuxkit/linuxkit/tree/master/test/cases/040_packages nor any relevant mentions of "mount" in the test output on https://linuxkit.datakit.ci/. Should I check somewhere else?

pkg/mount/mount.sh
```diff
@@ -10,6 +17,24 @@ mkdir -p "$MOUNTPOINT"
...
+		then
+			# identifier is LABEL or UUID
+			DEV=$(findfs $DEV_ID)
+			mount "$DEV" "$MOUNTPOINT" && return
+		elif [ -b "$DEV_ID" ] ; then
+			# identifier is a block-device
+			DEV="$DEV_ID"
+			mount "$DEV" "$MOUNTPOINT" && return
+		else
+			echo "Warning: Invalid Device identifier provided: $DEV_ID"
```
rn: should this return?

pkg/mount/mount.sh
```diff
@@ -10,6 +17,24 @@ mkdir -p "$MOUNTPOINT"
...
+		fi
+	fi
+
+
```
rn: super minor nit but extra line can be removed

pwFoo:
Would be a great improvement because of flexibility :)

justincormack:
@aboyett re "On successful mount, return is called; I mirrored this pattern in the new approach. You do make a good point that it still falls back to the autodetecting default approach on failure. Do you agree it would be reasonable to abort further mounting attempts if we fail to mount the explicitly given device identifier?" - definitely do not fall back if explicit instruction fails, that would be very confusing. Just fail.

deitch:
> I have been able to successfully mount an external disk using this code using the command-lines given in the commit and PR descriptions.

Great! If it works, it works.

> Can you update the docs for how to mount a disk? They are here https://github.com/linuxkit/linuxkit/blob/master/docs/external-disk.md
Sure! Would you prefer an amendment to the existing commit, or subsequent commits in the same PR?

I think amendment. I think the standard for linuxkit projects is single comment amended. I usually get asked to squash multiple commits down (but could be personal! :-) )

> On successful mount, return is called; I mirrored this pattern in the new approach. You do make a good point that it still falls back to the autodetecting default approach on failure. Do you agree it would be reasonable to abort further mounting attempts if we fail to mount the explicitly given device identifier?

Agreed with @justincormack here: "definitely do not fall back if explicit instruction fails". 

This makes me think about the syntax. We want to move towards being able to:

* mount multiple disks explicitly (name, label, dev, etc.) to specific mount points
* mount all disks as found (current behaviour)

We need a syntax to support both of those. We also should be able to fail if an _explicit_ mount is unavailable. 

May I recommend the following (do not change until @rn or @justincormack weighs in):

* options to `mount.sh` are one option per file-system, colon-separated
* wildcard option means "mount all you can find in dev" (current behaviour), but excluding any already mounted explicitly. For wildcard, probably just `/dev` with no specific subpath or mountpoint. 
* no arguments is equivalent to wildcard (I suspect this would be controversial)
* any _explicit_ request fs that is not found is a failure
* wildcards not found (_implicit_) are not a failure

Something like:

```yml
command: ["/mount.sh", "/dev/sdb1:/var/data"]
command: ["/mount.sh", "LABEL=DATA:/var/data"]
command: ["/mount.sh", "UUID=deadbee7-cafe-dead-c0d3-1234567890ab:/var/data"]
command: ["/mount.sh", "/dev/sdb1:/var/data", "/dev/sdb2:/var/foo"]
command: ["/mount.sh", "LABEL=DATA:/var/data","/dev/sdb3:/var/bar"]
command: ["/mount.sh", "LABEL=DATA:/var/data","/dev/sdb3:/var/bar","/dev"]
command: ["/mount.sh", "/dev"]
command: ["/mount.sh"]
```

I wanted to use `*` for wildcard, but that could create shell globbing issues (auto-correct tried to change "globbing" to "glowing"... radioactive `mount.sh`?)

Thoughts?

justincormack:
@deitch the current one does not mount all devices, it mounts the first one it finds and then returns. Ideally it should use a better heuristic, eg it should prefer a non boot volume, or choose largest device.

pwFoo:
I think mount a prepared partition (by label) should be fine. first or last partition tastes like mount a random  partition *g*
RancherOS uses a default label name. If there is a partition with that label, it will be mounted.

deitch:
Um yeah. Right. Ok. So do we want to keep that behaviour? Replace it with "all"? Both?

rn:
can we move the discussion about multiple mounts etc to an issue instead of hijacking a PR which, although related, is a perfectly fine improvement (once the minor nits are addressed) over what we currently have. We already have an issue for multiple disks (https://github.com/linuxkit/linuxkit/issues/1150) and the discussion should happen there.

deitch:
> can we move the discussion about multiple mounts etc to an issue instead of hijacking a PR which, although related, is a perfectly fine improvement

It came up because it affects this PR. Previously, we just ran `mount.sh` and it found the first valid disk and mounted. Now, we are adding the ability to mount a specific disk at a specific mount point (an *excellent* improvement), but that creates an options syntax for `mount.sh`. We want to get the syntax able to support multiple mounts with the next PR without having to change it (backwards-incompatible) again.

I think it would be fine for now if we had a syntax that allowed us to extend to multiple mounts in a future PR without having to change the current to this PR syntax (hence the suggested single colon-separated arg vs 2 args), and fall back to current behaviour (first disk found) with no arg.

Future PR can just add options, including a wildcard.

Work?

justincormack:
I don't think we should aim to support multiple mounts in one, the syntax is going to always be ugly, and adding multiple `mount` images will be costless eventually anyway. If we can make a nice syntax, maybe, but its low priority.

deitch:
I didn't think about that. Nothing wrong with having multiple onboot mount images, is there?

Convinced me. 

Ok then. Let's leave the syntax as is. We just need the fail-if-not-found, docs updated, and mount only that disk if args given / first disk = current behaviour if no arts given.

dave-tucker:
This conflicts with https://github.com/linuxkit/linuxkit/pull/2152

justincormack:
@dave-tucker if you remove the `mount` parts of #2152 is there anything missing here?

pkg/mount/mount.sh
aboyett: This won't return based on the current feedback.  Autodetection of mounting will not be attempted if a device is specified but cannot be found (or mounting it fails.) In that case this message will be displayed, then the lower "WARNING: Failed to mount a persistent volume (is there one?)" message will be displayed.

pkg/mount/mount.sh
aboyett: Done.

justincormack:
I think we may go for #2152 instead, which rewrites the shell scripts in Go; @dave-tucker is adapting to support labels and be in line with the decisions we made here.

aboyett:
Ah, ok. I'll close this PR and look forward to filesystem label support in #2152 :smile:

PR #2177, opened 2017-07-09
Move ttyS0 after tty0 as it is more common

justincormack:
This means more users will see console output at boot time.

Remove page poison from example CLIs, we should document this
elsewhere and put in blueprints, but it is unnecessary in every example


justincormack:
rebased after the other changes, forgot they might conflict

PR #2179, opened 2017-07-10
Use existing securetty

### LINKED ISSUE #2097: getty in init errors trying to modify securetty
If you use `getty` in the `init` section I just happened to notice it says:

```
/usr/bin/rungetty.sh: line 1: can't create /etc/securetty: Read-only file system
```

Thats probably not a terrible disaster, but maybe we should try to fix it...

deitch:

**- What I did**
Changed `pkg/getty` as follows:

* Instead of _modifying_ `/etc/securetty` to match `console=` lines, _read_ `/etc/securetty` to determine if a getty can be started for a given tty device
* If a console tty device does *not* exist in `/etc/securetty`, refuse to start a `getty` on that terminal, and report an error message
* Override default in-container `/etc/securetty` with file from rootfs, similar to `/etc/getty.shadow`

**- How I did it**
Typing. :-)

**- How to verify it**
Try the following scenarios:

1. Build with securetty matching console, see getty is available
2. Build with securetty *not* matching console, see error message and getty is not available 

Use existing tty to determine whether to start a getty, rather than override securetty just because `console=...` appears in cmdline

closes #2097 

Nazgul fell beast

Yes, a Nazgul fell beast. Really.

### DIFF
--- a/pkg/getty/README.md
+++ b/pkg/getty/README.md
@@ -16,6 +16,12 @@ services:
 The above will launch a getty for each console defined in the cmdline, i.e. `/proc/cmdline`.
 
 
+### securetty
+Every console defined in the `cmdline` **must** also already exist in `/etc/securetty` if you wish to login on that tty as root. If it does not exist, a getty will be started, but you will not be able to login as root. A warning message will be sent to that tty.
+
+If you are using a console that is not in `securetty`, you can add it by overriding the default `securetty` file in the linuxkit root filesystem using `files:` in your moby `.yml` file.
+
+
 ### Login Options
 There are 3 ways to launch a getty on a linuxkit instance:
 
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -35,7 +35,8 @@ start_getty() {
 	fi
 
 	if ! grep -q -w "$tty" "$securetty"; then
-		echo "$tty" >> "$securetty"
+		# we could not find the tty in securetty, so start a getty but warn that root login will not work
+		echo "getty: cmdline has console=$tty but does not exist in $securetty; will not be able to log in as root on this tty $tty." > /dev/$tty
 	fi
 	# respawn forever
 	infinite_loop setsid.getty -w /sbin/agetty $loginargs $line $speed $tty $term &
@@ -47,6 +48,13 @@ if [ -f $ROOTSHADOW ]; then
 	cp $ROOTSHADOW /etc/shadow
 	# just in case someone forgot a newline
 	echo >> /etc/shadow
+fi
+
+ROOTSTTY=/hostroot/etc/securetty
+if [ -f $ROOTSTTY ]; then
+	cp $ROOTSTTY /etc/securetty
+	# just in case someone forgot a newline
+	echo >> /etc/securetty
 fi
 
 for opt in $(cat /proc/cmdline); do
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/mount.sh", "/var/lib/docker"]
 services:
   - name: getty
-    image: linuxkit/getty:5ab31289889d61a5d2ecbeea8e36ce74ac54737c
+    image: linuxkit/getty:0bd92d5f906491c20e4177c57f965338fe5a8c5f
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,7 +16,7 @@ onboot:
     image: linuxkit/metadata:f122f1b4e873f1d08cd67bd9105385fd923af0cb
 services:
   - name: getty
-    image: linuxkit/getty:5ab31289889d61a5d2ecbeea8e36ce74ac54737c
+    image: linuxkit/getty:0bd92d5f906491c20e4177c57f965338fe5a8c5f
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/getty.yml
+++ b/examples/getty.yml
@@ -14,7 +14,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:5ab31289889d61a5d2ecbeea8e36ce74ac54737c
+    image: linuxkit/getty:0bd92d5f906491c20e4177c57f965338fe5a8c5f
     # to make insecure with passwordless root login, uncomment following lines
     #env:
     # - INSECURE=true
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -11,7 +11,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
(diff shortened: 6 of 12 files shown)


deitch:
FYI, I would have preferred to have a separate file override mapped into the container, rather than rootfs `/etc/securetty`, but:

* it would be used for `init` mode anyways
* mapping files in directly without overriding the labels is a pain
* `/etc/securetty` is a system-wide thing anyways

I will change the hashes on examples in a moment

deitch:
Updated with `examples/` hashes updated.

This really is a pain with the hashes. And then when we make some changes in here, the hash changes, and we have to update them all over again.

ijc:
> If a console tty device does not exist in /etc/securetty, refuse to start a getty on that terminal, and report an error message

I think that's a bit aggressive, at least in secure (i.e user & password mode rather than direct login) since AIUI all `securetty` does is gate whether _root_ can login, other users can login fine. Perhaps that distinction is slightly moot in the context of LinuxKit today.

justincormack:
Ok, well just report a message. I just don't want an unexpected failure thats non obvious, as its confusing.

deitch:
> I think that's a bit aggressive, at least in secure (i.e user & password mode rather than direct login) since AIUI all securetty does is gate whether root can login, other users can login fine

I was thinking that as well. But that seemed to be the thrust in #2097. 

As it is, `getty` has support for override a `shadow`, but not a `passwd`, so there is no way now in `getty` (and `sshd`) to login as anything *but* root.

My takeaway from it was:

* root only login now
* no `securetty` entry for the console tty = fail
* when we have time (or someone really needs it), add support for additional logins... but I wasn't sure how that played with the "life philosophy" of linuxkit.

ijc:
> Ok, well just report a message. I just don't want an unexpected failure thats non obvious, as its  confusing.

Agreed, I used to trip over this all the time before `/dev/xvc0` (and later) `/dev/hvc0` were added to all the distro securetty files...

deitch:
> well just report a message

It still will be confusing. Someone will try to login as root and fail. Of course, that happens with all linux distros now if you login as root.

I guess we can leave it as is (three points above), or let the getty start but spit out the error message.

deitch:
Of course, an error message _when_ you login would be more useful, but has security implications and not in our control anyways, is it?

justincormack:
I think all the `getty` implementations just silently fail, without a useful message.

deitch:
> I think all the getty implementations just silently fail, without a useful message

They do AFAIK. :-(

justincormack:
The CI fail seems to be about trust data, not sure why it is suddenly breaking.

deitch:
So what is the consensus on failure?

justincormack:
Which failure? I restarted the CI to see if it was transient.

deitch:
Failure = console tty not in securetty

justincormack:
Oh, just print a warning before it runs if its not there

deitch:
But still start the getty?

justincormack:
Yes, user may have set up unpriv login or something, who knows, but almost certainly they want to rebuild their image anyway so it doesnt really matter.

justincormack:
Did you push+sign the `getty` images? Thats probably why CI is failing.

deitch:
> Oh, just print a warning before it runs if its not there

OK, updating the commit right now.

> Did you push+sign the getty images? Thats probably why CI is failing

I should have good signing keys, so I can `make push` to get it out. But I wasn't sure what our flow is here. Do we approve the PR, then push them out manually with `make push`, and then merge in the commit after CI passes? 

Most projects I wait for CI to pass and then CI itself pushes them out. On the other hand, this is all hash-based, so no harm in extra images.

So what is the process?

deitch:
Updated to:

1. Only print the error, but still start getty
2. Fix hash in `examples/*.yml` (pain...)
3. Explicitly print error to `/dev/console`, since services print to their own logs by default

deitch:
Just let me know about the CI process, and we can be done with this minor one.

riyazdf:
> Most projects I wait for CI to pass and then CI itself pushes them out. On the other hand, this is all hash-based, so no harm in extra images.

@deitch: the ultimate goal is to have our CI system also push and sign, but we are still working toward that.  Since this is LGTM'd, please go ahead with pushing and signing

deitch:
> please go ahead with pushing and signing

ok. Going to be paranoid and build+test one more time...

deitch:
We never fixed that bug about trust when the image is local, did we? Or was it decided not to be a bug?

deitch:
That was dumb. I had the error message going to `/dev/console` instead of the specific `/dev/$tty`. Fixing and amending...

deitch:
That works now. Fixing hash on examples, pushing out commit, and pushing getty image.

deitch:
All done. However, the trust didn't take. I just did `make push`, which becomes `docker push linuxkit/getty:05948dffafff5a3679883e5c48b10270a72509cf`. Doesn't it need some explicit tagging or env var to sign it?

justincormack:
@deitch the default should be to set trust on: from `package.mk`:
```
ifeq ($(DOCKER_CONTENT_TRUST),)
ifndef NOTRUST
export DOCKER_CONTENT_TRUST=1
endif
endif
```

setting `DOCKER_CONTENT_TRUST=1` is what tells it to ask you to sign

justincormack:
You can pull the images and then do `DOCKER_CONTENT_TRUST=1 docker push ...` again.

pkg/getty/README.md
```diff
@@ -16,6 +16,12 @@ services:
 The above will launch a getty for each console defined in the cmdline, i.e. `/proc/cmdline`.
 
 
+### securetty
+Every console defined in the `cmdline` **must** also already exist in `/etc/securetty`. If it does not exist, a console will not be started, and an error message will be sent to the console.
+
```
ijc: This is no longer true I think.

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -35,7 +35,8 @@ start_getty() {
 	fi
 
 	if ! grep -q -w "$tty" "$securetty"; then
-		echo "$tty" >> "$securetty"
+		# we could not find the tty in securetty, so we cannot start a getty here
```
ijc: this is also no longer true.

deitch:
Nice catch on the comments, @ijc, thanks, fixed.

deitch:
> the default should be to set trust on: from package.mk... setting DOCKER_CONTENT_TRUST=1 is what tells it to ask you to sign

Right, but something is wrong apparently. I have the latest commit, I see that code in `package.mk`, and `pkg/getty/Makefile` does include it, but still it appears to run without. I even added a dummy target to see that it sets it, and it does:

```
foo:
    echo ${DOCKER_CONTENT_TRUST}
```

and

```
$ make foo
echo 1
1
```

In the meantime, I will push with explicit set of trust, but this is strange.

@riyazdf any ideas?

deitch:
OK, never mind, my mistake. I apparently ran `docker push` manually, not from the `Makefile`, even though I had no recollection of it. But bash history doesn't lie. :-)

Pushed out signed now:

```
9b87db43fd41: Pushed
09c570e1a98c: Pushed
2947f73d710c: Layer already exists
0bd92d5f906491c20e4177c57f965338fe5a8c5f: digest: sha256:4d3539223426b7418aefd3ef1f6dd84e4c56df3238e3097d9a687d596180ad3f size: 942
```

justincormack:
Oh, I just restarted CI, hopefully it will pass now these are pushed.

PR #2180, opened 2017-07-10
ARM64: Adapt the kernel Dockerfile to multi-arch support

arm64b:
The original kernel Dockerfile hardcodes the amd64 as the
only arch supported, this patch removes this kind of hardcode
and make the Dockerfile is ready to support both amd64 and
arm64 by using the runtime ARCH type.

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -17,7 +17,6 @@ RUN apk add \
     kmod \
     libelf-dev \
     libressl-dev \
-    libunwind-dev \
     linux-headers \
     ncurses-dev \
     sed \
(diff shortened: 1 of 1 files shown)


kernel/Dockerfile
```diff
@@ -50,13 +53,24 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
-COPY kernel_config-${KERNEL_SERIES} /linux/arch/x86/configs/x86_64_defconfig
+# When using COPY with more than one source file, the destination must be a directory and end with a /
+COPY kernel_config-${KERNEL_SERIES}* /linux/
 COPY kernel_config.debug /linux/debug_config
 
-RUN if [ -n "${DEBUG}" ]; then \
-    sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' /linux/arch/x86/configs/x86_64_defconfig; \
-    cat /linux/debug_config >> /linux/arch/x86/configs/x86_64_defconfig; \
-    fi
+RUN if [ $(uname -m) == x86_64 ]; then \
```
sbourlon: @arm64b A small improvement:
```
case "$(uname -m)" in
  x86_64) do_something;;
  aarch64) do_something;;
esac
```

arm64b: I will make some changes on this PR and re-submit a new one later on

PR #2185, opened 2017-07-11
ARM64: Adapt the kernel Dockerfile to multiarch support

arm64b:
The original kernel Dockerfile hardcodes the amd64 as the
only arch supported, this patch removes this kind of hardcode
and make the Dockerfile is ready to support both amd64 and
arm64 by using the runtime arch type.

**- What I did**
Remove the hardcode of x86
**- How I did it**
Get the info of the ARCH through runtime 'uname -m' command
**- How to verify it**
Build the kernel docker image on both arm64 and amd64 platform, then add this component into linuxkit.yml, run 'bin/moby build linuxkit.yml' 'bin/linuxkit run linuxkit'. you got it!

summit

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -17,7 +17,6 @@ RUN apk add \
     kmod \
     libelf-dev \
     libressl-dev \
-    libunwind-dev \
     linux-headers \
     ncurses-dev \
     sed \
@@ -25,7 +24,11 @@ RUN apk add \
     tar \
     xz \
     xz-dev \
-    zlib-dev
+    zlib-dev && \
+# libunwind-dev pkg is missed from arm64 now, below statement will be removed if the pkg is available.
+    if [ $(uname -m) == x86_64 ]; then \
+        apk add libunwind-dev; \
+    fi
 
 ARG KERNEL_VERSION
 ARG KERNEL_SERIES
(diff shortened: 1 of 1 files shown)


kernel/Dockerfile
```diff
@@ -50,13 +53,27 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
 
-COPY kernel_config-${KERNEL_SERIES} /linux/arch/x86/configs/x86_64_defconfig
+# When using COPY with more than one source file, the destination must be a directory and end with a /
+COPY kernel_config-${KERNEL_SERIES}* /linux/
 COPY kernel_config.debug /linux/debug_config
 
-RUN if [ -n "${DEBUG}" ]; then \
+RUN case $(uname -m) in \
+    x86_64) \
+    cp /linux/kernel_config-${KERNEL_SERIES} /linux/arch/x86/configs/x86_64_defconfig; \
```
rn: could you indent this and the other cases so they are easier to read?

kernel/Dockerfile
```diff
@@ -50,13 +53,27 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
     cat /linux/debug_config >> /linux/arch/x86/configs/x86_64_defconfig; \
-    fi
+    fi \
+    ;; \
+    aarch64) \
+    cp /linux/kernel_config-${KERNEL_SERIES}-aarch64 /linux/arch/arm64/configs/defconfig; \
+    if [ -n "${DEBUG}" ]; then \
+    sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' /linux/arch/arm64/configs/defconfig; \
+    cat /linux/debug_config >> /linux/arch/arm64/configs/defconfig; \
+    fi \
```
rn: there is a fair bit of replication here. It might be easier and simpler if the `case` statement just copies the config file and sets an env variable. Then the rest of the code is shared

kernel/Dockerfile
arm64b: yes, for sure!

kernel/Dockerfile
arm64b: Good point, I thought the env will be lost after the RUN command. So it's was as it looks now. Let me try to make some tweaks...

rn: The environment variable should be lost after the `RUN` but all the commands are executed in the same shell so should still be valid

arm64b: hmm, got it. Right, the env will be shared in the same layer environment. cool

kernel/Dockerfile
```diff
@@ -50,13 +54,24 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
-RUN if [ -n "${DEBUG}" ]; then \
-    sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' /linux/arch/x86/configs/x86_64_defconfig; \
-    cat /linux/debug_config >> /linux/arch/x86/configs/x86_64_defconfig; \
-    fi
+RUN case $(uname -m) in \
+    x86_64) \
+        cp /linux/kernel_config-${KERNEL_SERIES} ${KERNEL_DEF_CONF}; \
+        ;; \
+    aarch64) \
+        ${KERNEL_DEF_CONF}=/linux/arch/arm64/configs/defconfig; \
```
justincormack: This should be `KERNEL_DEF_CONF=/linux/arch/arm64/configs/defconfig`

justincormack: Also it seems strange setting this with `ENV` for amd64 but as a shell environment variable on arm64, should be consistent and just use a shell variable I think

arm64b: Shell environment LGTM, I think probably I misunderstood rn's proposal :)  I do think use a shell var will make the code be more consistent in that context.

kernel/Dockerfile
```diff
@@ -34,6 +37,7 @@ ARG DEBUG
 ENV KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.xz
 ENV KERNEL_SHA256_SUMS=https://www.kernel.org/pub/linux/kernel/v4.x/sha256sums.asc
 ENV KERNEL_PGP2_SIGN=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.sign
+ENV KERNEL_DEF_CONF=/linux/arch/x86/configs/x86_64_defconfig
```
rn: This is not needed here as the only place where the variable is used is in the section below: https://github.com/linuxkit/linuxkit/pull/2185/files#diff-e605bdf17a402b8724f3ef2ca42b291aR61

kernel/Dockerfile
rn: @arm64b I think @justincormack was referring to the fact that the case statement is asymmetric, when it should be more something like:
```
case $(uname -m) in \
  x86_64) \
    KERNEL_DEF_CONF=/path/to/x86/config \
    ;; \
 aarch64) \
   KERNEL_DEF_CONF=/path/to/arm64/config
   ;;
```

arm64b:
@justincormack and @rn  Thanks for the kindly review! Next step we will upload all the kernel configure files for arm64, after that we will end the kernel component in terms of ARM64 support :-), then move to the base alpine image...

PR #2186, opened 2017-07-11
Re-run vendoring

justincormack:
Latest `vndr` has changed its algorithm a bit. Also it vendors some extra things at the top level, so moved it down to where the command is as that is all it is for.

unshorn_alpaca_grazing


justincormack:
There are a few minor changes, but in fact the changes are due to it picking up `pkg/containerd` which we don't really want in the top level vendor.

@ijc is going to think about this.

justincormack:
Ok, moved vendoring to `src/cmd/linuxkit` so it does not get confused by packages, which should have standalone vendoring.

ijc:
Is this still effective in the root:
```
$ cat .gitattributes
vendor/* linguist-vendored
kernel/*.x linguist-language=text
```
or does the linguist-vendored one need to move down the source tree or have ** prepending or ???

Everything else LGTM, although CI seems unimpressed.

ijc:
Oh, perhaps `docs/vendoring.md` needs an update too?

justincormack:
I think that should all be addressed now.

docs/vendoring.md
```diff
@@ -2,12 +2,12 @@ Vendoring
 =========
 
 The Go code in this repo depends on a number of Go libraries.
-Theses are vendored in to the `vendor` directory using [`vndr`](https://github.com/lk4d4/vndr)
+Thes are vendored in to the `src/cmd/linuxkit/vendor` directory using [`vndr`](https://github.com/lk4d4/vndr)
```
ijc: You corrected the typo too far ;-)

docs/vendoring.md
```diff
@@ -47,11 +47,13 @@ go get -u github.com/LK4D4/vndr
...
 
 ```
+cd src/cmd/linuxkit
 vndr
 ```
 
 To update a single dependency:
 
 ```
+/src/cmd/linuxkit
```
ijc: Missing the leading `cd`

PR #2188, opened 2017-07-11
Add labels for repo and commit to packages

justincormack:
- use image spec metadata https://github.com/opencontainers/image-spec/blob/master/annotations.md
- omit commit if dirty

fix https://github.com/moby/tool/issues/20

Initial commit for discussion before rebuilding everything...

mama-baby-bear-animals-36003651-500-350

### DIFF
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -46,6 +46,9 @@ show-tag:
 tag: $(BASE_DEPS) $(DEPS)
 	docker pull $(TAG) || docker build $(LABELS) $(NET_OPT) -t $(TAG) .
 
+forcetag: $(BASE_DEPS) $(DEPS)
+	docker build $(LABELS) $(NET_OPT) -t $(TAG) .
+
 push: tag
 ifneq ($(DIRTY),)
 	$(error Your repository is not clean. Will not push package image.)
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -13,6 +13,16 @@ endif
 
 TAG:=$(ORG)/$(IMAGE):$(HASH)$(DIRTY)
 
+REPO?=https://github.com/linuxkit/linuxkit
+ifneq ($(REPO),)
+REPO_LABEL=--label org.opencontainers.image.source=$(REPO)
+endif
+ifeq ($(DIRTY),)
+REPO_COMMIT=$(shell git rev-parse HEAD)
+COMMIT_LABEL=--label org.opencontainers.image.revision=$(REPO_COMMIT)
+endif
+LABELS=$(REPO_LABEL) $(COMMIT_LABEL)
+
 BASE_DEPS=Dockerfile Makefile
 
 # Get a release tag, if present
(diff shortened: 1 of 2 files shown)


ijc:
https://github.com/opencontainers/image-spec/issues/658 was trying to address this from the image spec end. I'm not sure what the result was. The table in https://github.com/opencontainers/image-spec/issues/658#issuecomment-300336575 doesn't seem to have been implemented since [`annotations.md`](https://github.com/opencontainers/image-spec/blob/master/annotations.md) doesn't seem to cover everything listed there, in particular `source` and `revision` (corresponding to `vcs-url` and `vcs-ref` you've used here) seem to be missing.

pkg/package.mk
```diff
@@ -13,6 +13,17 @@ endif
 
 TAG:=$(ORG)/$(IMAGE):$(HASH)$(DIRTY)
 
+REPO?=https://github.com/linuxkit/linuxkit
+ifneq ($(REPO),)
+REPO_LABEL=--label org.label-schema.vcs-url=$(REPO)
+endif
+REPO_DIRTY:=$(shell git update-index -q --refresh && git diff-index --quiet HEAD -- || echo "-dirty")
```
rn: Doesn't the same logic we use for `DIRTY` apply here, ie we don't care if the repo is dirty as long as the directory is clean? It seems odd to me to apply tow different notions of dirty

ijc:
Code LGTM. Possibly you could take `HEAD` if the subtree is clean rather than worrying about the whole repo, but the way you have it is equally fine.

justincormack:
Yes, maybe will be consistent and use the same `DIRTY`

justincormack:
Ok, changed to use the same `DIRTY`

justincormack:
@ijc I think we can potentially add things to image spec metadata as well, but they are not very accessible compared to Docker image metadata at present, eg in terms of tooling, and we are generating Docker images not image spec images for now anyway.

justincormack:
we could use `org.opencontainers.image.source` and `org.opencontainers.image.revision` as the tags instead see https://github.com/opencontainers/image-spec/blob/master/annotations.md - maybe better

ijc:
Oh, those were the ones I claimed weren't specified in my comment above, but it turns out my local tree was just very out of date.

justincormack:
Updated to use image spec labels as they specifically say they obsolete label schema

ijc:
LGTM (maybe `LABEL_OPT` for consistency with `NET_OPT` but that's just pedantry on my behalf)

justincormack:
Added to kernel build

justincormack:
@ijc changed to `LABELS` in the interests of pedantry; it doesn't feel like an `OPT`

justincormack:
@rn yes, will add a rebuild once we are agreed on everything here

rn:
looks good to me, though, as we discussed, ideally we'd also include `package.mk` into the hash somehow, but that can be a separate effort

justincormack:
I am updating the images, but as their hashes will not change this can be merged once CI is green.

You can see labels if you re-pull and do `docker inspect linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918`

justincormack:
Going to leave the image updates (some done) as the base image needs updating anyway, so will re-generate with labels anyway today/tomorrow.

PR #2189, opened 2017-07-11
Remove shellcheck from base

justincormack:
We should use this from a container at top level instead; it makes it
much harder to port to other architectures if we have to have this as
it is written in Haskell making a multi arch build much harder.

400px-nautilus_species_shells


justincormack:
As I am waiting  for the Go security update in base I wont rebuild until tomorrow.

cc @arm64b this should remove one thing that is needed in port that varies by architecture.

PR #2191, opened 2017-07-12
Move onboot startup script to runc package

justincormack:
As onboot does not use containerd at all, this means you can run very
    minimal setups with just `runc` if you use no services, for example
    most of our tests do not actually use services, or if you have other
    similar very minimal use cases.
    
    Move ulimit setup to `init` which makes more sense.

frog

### DIFF
--- a/projects/compose/compose-dynamic.yml
+++ b/projects/compose/compose-dynamic.yml
@@ -2,9 +2,9 @@ kernel:
   image: linuxkit/kernel:4.9.36
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
-  - linuxkit/runc:a0f2894e50bacbd1ff82be41edff8b8e06e0b161
-  - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
+  - linuxkit/init:a626edd428b575d10b994be0d4c5b034838e9946
+  - linuxkit/runc:f5b28379fc969237e0fb11aa1ed946622150e9a0
+  - linuxkit/containerd:f1f9c27e153b510e09b2e3f221181f87ade0fe1a
   - linuxkit/ca-certificates:67acf038c44bb191ebb704ec7bb39a1524052cdf
 onboot:
   - name: sysctl
--- a/projects/compose/compose-static.yml
+++ b/projects/compose/compose-static.yml
@@ -2,9 +2,9 @@ kernel:
   image: linuxkit/kernel:4.9.36
   cmdline: "console=ttyS0 page_poison=1"
 init:
-  - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
-  - linuxkit/runc:a0f2894e50bacbd1ff82be41edff8b8e06e0b161
-  - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
+  - linuxkit/init:a626edd428b575d10b994be0d4c5b034838e9946
+  - linuxkit/runc:f5b28379fc969237e0fb11aa1ed946622150e9a0
+  - linuxkit/containerd:f1f9c27e153b510e09b2e3f221181f87ade0fe1a
   - linuxkit/ca-certificates:67acf038c44bb191ebb704ec7bb39a1524052cdf
 onboot:
   - name: sysctl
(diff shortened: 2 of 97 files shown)


projects/compose/compose-dynamic.yml
```diff
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
-  - linuxkit/runc:2310ad9d266cf5d9c4d07613bd2135ed7eb8a21f
+  - linuxkit/runc:f73ab00a4d18edf6beb056685426e9d6b5e3d254
   - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
```
rn: missed this update

projects/compose/compose-static.yml
```diff
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
-  - linuxkit/runc:2310ad9d266cf5d9c4d07613bd2135ed7eb8a21f
+  - linuxkit/runc:f73ab00a4d18edf6beb056685426e9d6b5e3d254
   - linuxkit/containerd:b1766e4c4c09f63ac4925a6e4612852a93f7e73b
```
rn: and here

projects/etcd/etcd.yml
```diff
@@ -3,8 +3,8 @@ kernel:
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
   - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
```
rn: init needs an update

projects/ima-namespace/ima-namespace.yml
```diff
@@ -3,8 +3,8 @@ kernel:
   cmdline: "console=ttyS0 console=tty0 page_poison=1 ima_appraise=enforce_ns"
 init:
   - linuxkit/init:b3740303f3d1e5689a84c87b7dfb48fd2a40a192
```
rn: and here as well

projects/okernel/examples/okernel_simple.yaml
```diff
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
   - linuxkit/init:12348442d56c2ee9abf13ff38dff2e36b515bd1e
```
rn: init and containerd updates missed here

projects/wireguard/wireguard.yml
```diff
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=ttyS0 console=tty0 page_poison=1"
 init:
   - linuxkit/init:cbd7ae748f0a082516501a3e914fa0c924ee941e
```
rn: init and containerd update missed here

justincormack:
At some times at least I have not been updating hashes in projects as they may have reasons to use things, not sure what to do there. I was about to write a script to check hashes are correct, but was going to omit projects...

rn:
then surely we shouldn't update any hashes which some of the changes do

justincormack:
Yes, projects are currently a mess, as some of them do have special requirements and some do not. The change above was just changing all the ones that were currently the same as the ones in the main example, rather than updating everything regardless of how it was set before, which is what I usually do on the grounds if it matches other things it cant be special. I dont think that works in all cases (eg logging probably broken) though.

ijc:
Yeah, my MO has been to update anything which matches the toplevel `linuxkit.yml`.

justincormack:
Added the others

rn:
not all of them, but LGTM

riyazdf:
re: projects updating, agree with @justincormack that it's been fairly adhoc and messy.  I'm fine with leaving them as-is for most updates.

justincormack:
This does mean that if you mount a persistent disk under the `containerd` directory (and don't eg mount a `tmpfs` there too) then the `containerd` image store etc will be persistent.

justincormack:
(I think this is better behaviour, previously it would probably leave a mess).

justincormack:
(Also needed if you want to eg test `containerd` on btrfs root.)

Just removing `contaienrd` from the tests where not required.

PR #2201, opened 2017-07-13
docker-ce: add git binary

djs55:
**- What I did**

Fixed the following `docker build` cases:

```
$ cat > docker-compose.yml <<EOT
version: '2'

services:
  test:
    build: "https://github.com/dnephin/docker-build-from-url.git"
EOT

$ docker-compose build
```
and
```
$ curl  -d "" --unix-socket /var/run/docker.sock http://unix/v1.22/build?remote=https%3A%2F%2Fgithub.com%2Fdnephin%2Fdocker-build-from-url.git
```

Previously you would see an error like:
```
$ curl  -d "" --unix-socket /var/run/docker.sock http://unix/v1.22/build?remote=https%3A%2F%2Fgithub.com%2Fdnephin%2Fdocker-build-from-url.git
failed to init repo at /var/lib/docker/tmp/docker-build-git645459714: : exec: "git": executable file not found in $PATH
```

**- How I did it**

I added the `git` binary to the `docker-ce` container: it seems that the docker daemon will occasionally want to fork/exec `git`.

### DIFF
--- a/blueprints/docker-for-mac/docker-17.06-ce.yml
+++ b/blueprints/docker-for-mac/docker-17.06-ce.yml
@@ -3,7 +3,7 @@ services:
   # Bind mounts /var/run to allow vsudd to connect to docker.sock, /var/vpnkit
   # for vpnkit coordination and /var/config/docker for the configuration file.
   - name: docker-dfm
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
+    image: linuxkit/docker-ce:baf282455697809fc4e2ee398567d9a4d6c061d9
     capabilities:
      - all
     net: host
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -30,7 +30,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:19370f5d9bec84eb91073b7196b732f1301d9c90
   - name: docker
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
+    image: linuxkit/docker-ce:baf282455697809fc4e2ee398567d9a4d6c061d9
     capabilities:
      - all
     net: host
--- a/pkg/docker-ce/Dockerfile
+++ b/pkg/docker-ce/Dockerfile
@@ -11,6 +11,7 @@ RUN apk add --no-cache --initdb -p /out \
 	curl \
 	e2fsprogs \
 	e2fsprogs-extra \
+	git \
 	iptables \
 	musl \
 	xfsprogs \
--- a/projects/compose/compose-dynamic.yml
+++ b/projects/compose/compose-dynamic.yml
@@ -27,7 +27,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:19370f5d9bec84eb91073b7196b732f1301d9c90
   - name: docker
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
+    image: linuxkit/docker-ce:baf282455697809fc4e2ee398567d9a4d6c061d9
     capabilities:
      - all
     net: host
--- a/projects/compose/compose-static.yml
+++ b/projects/compose/compose-static.yml
@@ -27,7 +27,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:19370f5d9bec84eb91073b7196b732f1301d9c90
   - name: docker
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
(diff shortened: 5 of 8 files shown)


justincormack:
you need to update the hashes for docker-ce...

change looks fine.

justincormack:
Hmm


Are you sure docker needs git? If it's just compose then maybe it shouldn't
be there.

justincormack:
And I didn't think we had compose installed...

ijc:
`builder/remotecontext/git/gitutils.go` in `moby/moby` shells out to the git binary.

justincormack:
Cc @friism

deitch:
Ha! I suggested to the maintainers of `library/node` image that `:alpine` should include git for the same reason: `npm` can use git urls, but the image didn't have git. 

I kind of lost that one, glad to see we are considering it here if basic operation needs it. Problem, of course, is `git` is relatively large. `apk add git` is 25 MiB in 17 (!) packages. I wonder why it isn't a single go binary?

thaJeztah:
Yes; building from a remote source can be handled either in the client (clones the repo client-side, then sends it as build context to the daemon), or daemon-side; pass the url to the daemon, and the repo is cloned there and used as build context

deitch:
> can be handled either in the client or daemon-side

@thaJeztah, how does the docker client determine if to do locally or pass to engine?

friism:
It doesn't, `docker build URL` always fetches with git client side. Only when using the API (like `docker-compose` does) will the engine-side git checkout be triggered

deitch:
> then the daemon shells out to git

Yes, I think it should be in there as well (although I don't like the extra 25MB).

Question for @thaJeztah: if the daemon supports git URLs directly, should it exec out to the shell or talk the git api directly? I don't even know if there a git client lib in go...

deitch:
So `docker build` always does it client, and `docker-compose` always does it via engine? OK. That documented anywhere?

thaJeztah:
I'm not sure if that's documented; IIRC, originally this was always handled daemon-side, but got moved to the client (see https://github.com/moby/moby/issues/14298). I think the daemon-side git was left for backward compatibility

justincormack:
It might be a good time to start to deprecate it. Given it does not have credentials it is not that useful.

thaJeztah:
yup; was thinking the same; one advantage of course, is that the content doesn't have to be transferred twice  (download repo, then upload to daemon)

ijc:
Git is only 6 packages the 17 packages / 25M figure which apk prints out is the base system+git after installing git, not just git (although 20+M _is_ git itself):

The `linuxkit/docker-ce` package ~100M, so we'd be looking at a 20% increase in size (as opposed to the 500+% increase from adding git to the alpine base system), and that's an upper bound since it looks like a few of the deps are already installed until `docker-ce` (`ca-certificates`, `libssh2`, `libcurl` are, so only `expect` , `pcre` and `git` itself are extra).

Most of the git binaries in `/usr/libexec/git-core` are hard linked to a fairly small set of actual binaries (~14 of 1-2M each) so there isn't a lot of opportunity I can see for splitting lesser-used functionality into other packages that I can. But I do see:
* 1.2M `git-remote-testsvn`
* 1.2M `git-instaweb`
* 1.3M `git-imap-send`
* 1.2M `git-http-backend`
* 1.2M `git-shell`
* 1.2M `git-fast-import`

None of those are hardlinked to other tools and all seem either pretty niche or specific to server side git usage (i.e. providing `git://` or `http://` servers) rather than client usage and splitting them out would drop 7.3M, i.e. it would cut about 40% of the base package size without sacrificing any common functionality (IMNSHO) 

The fact that all of the unique binaries (so considering only one of each hardlinked set) are very similar sizes in the 1-2M range makes me suspicious that they are all statically linking a common 1M `libgit.a` type thing, but that is true of the Debian packages too so I presume it is an git upstream (as opposed to Alpine specific) decision which there is no reason (or appetite) to go against.

//cc @ncopa

deitch:
> Git is only 6 packages the 17 packages / 25M figure which apk prints out is the base system+git after installing git, not just git (although 20+M is git itself):

Um, yeah. Oops. :-)

djs55:
I updated the hashes. I guess the image would need to be signed and pushed for the CI to pass:
```
Trusted pull for linuxkit/docker-ce:baf282455697809fc4e2ee398567d9a4d6c061d9 failed: No trust data for baf282455697809fc4e2ee398567d9a4d6c061d9
```

justincormack:
If we merge #2235 you need to do this on the upstream library image...

justincormack:
I think we will merge #2235 so you should probably do a PR against https://github.com/docker-library/docker instead

PR #2202, opened 2017-07-13
Add experimental support for LCOW

rn:
A previous commit already added the kernel support (hopefully) necessary for LCOW (Linux Containers for Windows). This PR adds a new init container based on https://github.com/Microsoft/opengcs/blob/master/docs/customosbuildinstructions.md and a blueprint to create a kernel+initrd.

We currently can't test if this actually works...

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ endif
 
 PREFIX?=/usr/local/
 
-MOBY_COMMIT=4db06aa1732b44a8cadd9c8577df0aa5c716e701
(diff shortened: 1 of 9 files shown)


rn:
Don't merge yet. fix coming in

rn:
forced pushed an update. It's getting there but now there seems to be an issue with the way `mkfs.ext4` is invoked. Still debugging

rn:
Pushed a bunch of updates and it's all working now. This also update `moby` to the latest to shrink down the initrd to 7MB

PR #2198, opened 2017-07-13
docker-for-mac: clear the default namespace on start

djs55:
**- What I did**

Fixed a bug in the `docker-for-mac` blueprint which would cause it to fail to boot if the disk isn't empty i.e. on second and subsequent boots.

**- How I did it**

I delete the directory `/var/lib/containerd/io.containerd.runtime.v1.linux/default` in an `onboot` action before starting `services`. I first tried to modify the `010-containerd` script but unfortunately this is run before the disk is mounted, so the old entries later magically reappear. 

**- How to verify it**

Boot the `docker-for-mac` example twice, without erasing the disk between runs. Without the patch you'll see errors like
```
Starting service: "acpid"
FATA[0000] failed to create task                         error="rpc error: code = Unknown desc = runtime create failed: mkdir /var/lib/containerd/io.containerd.runtime.v1.linux/default/acpid: file exists" service=acpid
```

With the patch, there are no errors and all the services start correctly.

PR #2204, opened 2017-07-14
Service start of day cleanup

ijc:
**- What I did**

Added a `system-init` command to the `service` binary to setup the system ready for running system services at start of day. Initially this contains only code to remove any stale system containers from a previous boot in the case where `/var/lib/containerd` is on a persistent disk to solve the same issue as #2198.

~~This is currently WIP since it is based on (and thus for now includes) #2191. It will need rebasing once #2191 is merged. In the meantime only "containerd: Add a service command to cleanup stale containers on boot" is worth looking at.~~ #2191 is now merged, this PR has been rebased.

**- How I did it**

Added a new command described above and called it from `020-containerd` before starting the services.

**- How to verify it**

I've been running `service system-init` by hand from an init (not service, or you will be trying pull the rug out from under yourself and hang) based getty in the top level `linuxkit.yml`.

You could also configure a persistent disk for `/var/lib` or `/var/lib/containerd` and do an unclean reboot.

Stale processes

[](https://en.wikipedia.org/wiki/From_Mars_to_Sirius)


ijc:
Rebased onto merge of #2191

PR #2205, opened 2017-07-14
Small alpine mirror improvements

ijc:
**- What I did**

Two things:
* Kept a copy of the original (upstream Alpine) `/etc/apk/repositories`. This will be useful in dependent containers which want to include a working `apk` configuration by avoiding the need to include alpine version numbers there
* Record the hash of the pushed alpine bash container, currently this relies on the committer remembering to include it in the commit message. `tools/alpine` now supports the `show-tag` target like `pkg/package.mk` based builds do.

Note that nothing here is updated to use the new alpine hash.

Note2 I had planned to use the first change to add `apk` to the `getty` and `sshd` containers but ran into a blocker, see #2206.

**- How I did it**

Updated the `Dockerfile` and `Makefile`

**- How to verify it**

This should validate both of my claims together:
```
docker run -it --rm $(make --no-print-directory -C tools/alpine/ show-tag) cat /etc/apk/repositories.upstream
```

[](https://en.wikipedia.org/wiki/Wiseblood_(Corrosion_of_Conformity_album))

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -27,8 +27,8 @@ RUN abuild-sign /mirror/$(uname -m)/APKINDEX.tar.gz
 # fetch OVMF for qemu EFI boot (this is not added as a package)
 RUN apk add -X http://dl-cdn.alpinelinux.org/alpine/edge/community ovmf
 
-# set this as our repo
-RUN echo "/mirror" > /etc/apk/repositories && apk update
+# set this as our repo but keep a copy of the upstream for downstream use
+RUN mv /etc/apk/repositories /etc/apk/repositories.upstream && echo "/mirror" > /etc/apk/repositories && apk update
 
 # add Go validation tools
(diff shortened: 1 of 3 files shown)


ijc:
Related to this, I noticed that we a) build as `$(IMAGE):build`, which prevents multiple concurrent builds on the same host and b) `docker rmi $(IMAGE):build` after `tag` or `push`.

If we were to use the usual git treesh approach instead of hardcoding `build` (but keep the hash scheme used for the pushed tag the same as it is now, based on the packages etc) maybe we could remove the `docker rmi` and benefit a bit from local caching during development, or maybe this is deliberate for some reason (perhaps due to changing network repositories?).

justincormack:
Yes, I guess using tree hash as the build time tag should be ok. Did you build id file thing get merged into 17.06? we can use that!

ijc:
Oh, yes I did!

PR #2207, opened 2017-07-14
remove -x from getty script

justincormack:
Was annoying me by being ultra verbose...

cc @deitch

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -53,7 +53,7 @@ services:
     image: linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918
   # Enable getty for easier debugging
   - name: getty
-    image: linuxkit/getty:5ab31289889d61a5d2ecbeea8e36ce74ac54737c
+    image: linuxkit/getty:deb9332e786e72591bd9be200bcc9c7a534eb754
     env:
         - INSECURE=true
   # Run ntpd to keep time synchronised in the VM
--- a/linuxkit.yml
+++ b/linuxkit.yml
@@ -16,7 +16,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:5ab31289889d61a5d2ecbeea8e36ce74ac54737c
+    image: linuxkit/getty:deb9332e786e72591bd9be200bcc9c7a534eb754
     env:
      - INSECURE=true
   - name: rngd
--- a/pkg/getty/etc/securetty
+++ b/pkg/getty/etc/securetty
@@ -1,0 +1,33 @@
+console
+tty1
+tty2
+tty3
+tty4
+tty5
+tty6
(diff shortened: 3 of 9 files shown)


deitch:
> Was annoying me

I can see that! 

LGTM.

justincormack:
hmm, the test seems to be relying on the output...

ijc:
I can't see where, the stuff in the expect script is all normal stuff.

I think the issue is this:
```
getty: cmdline has console=ttyS0 but does not exist in /etc/securetty; will not be able to log in as root on this tty ttyS0.
```
which is preventing the script from logging in. Why this is just showing up here I cannot imagine.

justincormack:
Yes, we actually need to add a full `securetty` not sure why this did not fail before

justincormack:
Ok added a reasonable `securetty` file. Very unclear why the tests passed before still...

justincormack:
@deitch I removed fetching `/etc/securetty` from host, as the host one is lacking a lot, and happy to add any more things to here (or you can bind mount).

justincormack:
All green, merging.

deitch:
@justincormack looks fine (and already merged), but with the `securetty` now I the container, what is the recommended method for a user overriding? If they just bind-mount, then it messes up all of the defaults in the `LABEL`. 

That `data:` idea is looking better...

justincormack:
yeah, we should probably do the data think soon, but you can just open a PR to add new interface types here, there arent actually that many, and most are for different architectures which will need other support anyway.

deitch:
I know why I am crazy enough to be online at this late hour - I just came back from watching an amazing Canada-USA ice hockey game - but why are you on this late? :-)

So you are thinking that soon enough the one in the getty package just will be comprehensive enough that people mostly won't need to mount it?

justincormack:
Well it's already better than the host one. The host should not be
responsible for setting up the getty container correctly, it is likely to
not have a securetty file at all soon...

deitch:
> to not have a securetty file at all soon

Soon? Why not now? See any reason to keep it?

deitch:
Actually, do you know where the host-level `/etc/securetty` comes from? We don't have it anywhere in our repo except in `pkg/getty`, yet 

```
(ns: getty) linuxkit-025000000008:~# ls -l /proc/1/root/etc/securetty
-rw-r--r--    1 root     root            65 Jun 11 06:38 /proc/1/root/etc/securetty
```

justincormack:
It's in the base alpine I think, which we are going to be able to remove
once we rewrite a few more shell scripts.

deitch:
> It's in the base alpine I think

[Here](https://github.com/linuxkit/linuxkit/blob/master/pkg/init/Dockerfile#L9) with `alpine-baselayout`? 

```
RUN apk add --no-cache --initdb -p /out alpine-baselayout busybox musl
```

justincormack:
```
/ # apk info -W /etc/securetty 
/etc/securetty is owned by busybox-1.26.2-r5
```

deitch:
Yeah. I just did the first install, but `alpine-baselayout` depend upon `busybox`, which installs is.

Issue #1297, opened 2017-03-14
checks for executables


justincormack:
check that executables in the built image are static pie, not suid etc.

PR #2211, opened 2017-07-14
Make runc a static pie executable

justincormack:
This was waiting for the runc update as I got the extra variables added a while back.

I think this is the last non static pie executable. Will try to work out a way to test this at some point see #1297 (I did test by hand)

dogpie


duglin:
definitely one of the best dog picts  :-)

PR #2216, opened 2017-07-14
fix BASE in Makefile to match Dockerfile

### LINKED ISSUE #2215: Makefile - Docker file base image mismatch
There is a mismatch between the base image in the Dockerfile, and the base defined in the Makefile for tools/guestfs

The Makefile is pulling debian:jessie
https://github.com/linuxkit/linuxkit/blob/master/tools/guestfs/Makefile#L3

The Dockerfile is using debian:stretch
https://github.com/linuxkit/linuxkit/blob/master/tools/guestfs/Dockerfile#L1

Right now this is just resulting in an extra image being pulled for this particular tool, as the BASE in the Makefile is not used anywhere else, which is not significant; however, the overriding thought is it feels a bit redundant, and a potential point for errors in the build stack to be introduced.

liqdfire:
Signed-off-by: Liqdfire <liqdfire@gmail.com>

fixes #2215 

**- What I did**
  Corrected BASE defined in Makefile

### DIFF
--- a/tools/guestfs/Makefile
+++ b/tools/guestfs/Makefile
@@ -1,6 +1,6 @@
 .PHONY: tag push
 
-BASE=debian:jessie
+BASE=debian:stretch
 IMAGE=guestfs
 
 default: push


justincormack:
Thanks!

PR #2217, opened 2017-07-14
Skanky hack to emulate udhcpc in LCOW

rn:
For some network setups, opengcs currently relies on `udhcpd` which is not included in Alpine's busybox (and is not avail as a package). So this hack, adds a `udhcpc` script which calls `dhcpcd`  in one shot mode.

Networking still does not come up in LCOW but the script is not called. I suspect some other issue, but for now we need a work around for lack of `udhcdc`.

@justincormack feel free to reject if this is too distasteful :)

### DIFF
--- a/blueprints/lcow.yml
+++ b/blueprints/lcow.yml
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=ttyS0"
   tar: none
 init:
-  - linuxkit/init-lcow:ae67df4c4cabb35aae06898cb94fe03972a172e9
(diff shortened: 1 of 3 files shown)


justincormack:
@rn it is in `/usr/sbin/udhcpd` in Alpine...

justincormack:
You may need to add `busybox-initscripts` to get the startup script (or maybe not)

rn:
Hmm, weird. It is indeed there. now idea how I missed it. closing

PR #2210, opened 2017-07-14
pkg/init: Do not run inits in the background

ijc:
We want them to run in sequence. For example we want mounts to be done (done by
`pkg/runc/etc/init.d/010-onboot`) before we start services (done by
`pkg/containerd/etc/init.d/020-containerd`). This was most likely introduced by
28b4245b122a ("Move onboot startup script to runc package").

None of the initscripts in pkg/* block, but some in projects (selinux and
logging, not updated here) do.

Signed-off-by: Ian Campbell <ijc@docker.com>

**- What I did**

Removed the `&` from the invocation in `/bin/rc.init`.

**- How I did it**

I pressed `x` in `vim`, twice IIRC.

**- How to verify it**

I booted the kubernetes example.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**
[](https://en.wikipedia.org/wiki/The_Great_Southern_Trendkill)

PR #2219, opened 2017-07-15
make init sequential

justincormack:
carry #2210 

frog


rn:
What's the rational behind removing the usermode helper? We need it for 4.11 kernels  which is configured to use it to shell out to other commands.

justincormack:
Oh, sorry, I was confusing it with something else

justincormack:
removed that commit

justincormack:
Hmm, trying a rebuild; the log didnt make it clear what in the `containerd-getty` test failed, could not actually see an error...

justincormack:
Ah Redis output message has changed slightly in 4.0...

PR #2221, opened 2017-07-17
fixes #2145 - /var/run should symlink to /run

### LINKED ISSUE #2145: /var/run not a symlink
`/var/run` should be a symlink to `/run` (which is a writeable tmpfs mount). I am sure it used to be.

dweomer:
Signed-off-by: Jacob Blain Christen <dweomer5@gmail.com>

**- What I did**
* removed `/out/var/run` from the "build" layer
* created symlink `/var/run` pointing to `/run`

**- How I did it**
Last layer in the init container is `RUN ln -s /run /var/`

**- How to verify it**
Before:
> $ docker cp $(docker create linuxkit/init:059b2bb4b6efa5c58cf53fed4d0ea863521959fc sh):/ - | tar tv | grep '/run'
> tar: Removing leading `/' from member names
> lrwxrwxrwx 0/0               0 2017-07-14 14:02 /bin/run-parts -> /bin/busybox
> drwxr-xr-x 0/0               0 2017-07-14 14:02 /run/
> __drwxr-xr-x 0/0               0 2017-07-14 14:02 /var/run/__

After:
> docker cp $(docker create linuxkit/init:059b2bb4b6efa5c58cf53fed4d0ea863521959fc-dirty sh):/ - | tar tv | grep /run
> tar: Removing leading `/' from member names
> lrwxrwxrwx 0/0               0 2017-07-16 21:21 /bin/run-parts -> /bin/busybox
> drwxr-xr-x 0/0               0 2017-07-16 21:21 /run/
> __lrwxrwxrwx 0/0               0 2017-07-16 21:21 /var/run -> /run__

`/var/run` is once again correctly symlinked to `/run`

### DIFF
--- a/pkg/init-lcow/Dockerfile
+++ b/pkg/init-lcow/Dockerfile
@@ -5,7 +5,7 @@ RUN apk add --no-cache --initdb -p /out \
     busybox \
     e2fsprogs \
     musl
-RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache /out/var/run
 
 FROM linuxkit/alpine:3744607156e6b67e3e7d083b15be9e7722215e73 AS build
 RUN apk add --no-cache build-base curl git go musl-dev
@@ -38,4 +38,5 @@ WORKDIR /
 COPY --from=mirror /out/ /
 COPY --from=build /out/ /
 COPY --from=runc /usr/bin/runc /sbin/runc
+RUN ln -s /run /var/
 
--- a/pkg/init/Dockerfile
+++ b/pkg/init/Dockerfile
@@ -9,7 +9,7 @@ RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out alpine-baselayout busybox musl
 
 # Remove apk residuals. We have a read-only rootfs, so apk is of no use.
-RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache /out/var/run
 
 FROM scratch
 ENTRYPOINT []
@@ -20,3 +20,4 @@ COPY --from=mirror /out/ /
 COPY init /
 COPY etc etc/
 COPY bin bin/
+RUN ln -s /run /var/


dweomer:
Using the Docker image from `examples/docker.yml`, before:
```
(ns: getty) linuxkit-525400123456:~# ls -alF /run/
total 0
drwxr-xr-x    4 root     root            80 Jul 17 04:51 ./
drwxr-xr-x   20 root     root           400 Jul 17 04:51 ../
drw-r-----    3 root     root           100 Jul 17 04:51 containerd/
drwx------    2 root     root            40 Jul 17 04:51 runc/
(ns: getty) linuxkit-525400123456:~# ls -alF /var/run/
total 4
drwxr-xr-x    3 root     root           120 Jul 17 04:51 ./
drwx------   11 root     root           220 Jul 14 14:02 ../
drwxr-xr-x    4 root     root            80 Jul 17 04:51 dhcpcd/
-rw-r--r--    1 root     root             2 Jul 17 04:51 dhcpcd.pid
srw-rw----    1 root     root             0 Jul 17 04:51 dhcpcd.sock=
srw-rw-rw-    1 root     root             0 Jul 17 04:51 dhcpcd.unpriv.sock=
```
... and after ...
```
(ns: getty) linuxkit-525400123456:/var# ls -alF /var/run/
total 0
drwxr-xr-x    4 root     root            80 Jul 17 04:48 ./
drwxr-xr-x   20 root     root           400 Jul 17 04:48 ../
drw-r-----    3 root     root           100 Jul 17 04:48 containerd/
drwx------    2 root     root            40 Jul 17 04:48 runc/
(ns: getty) linuxkit-525400123456:/var# ls -alF /run/
total 0
drwxr-xr-x    4 root     root            80 Jul 17 04:48 ./
drwxr-xr-x   20 root     root           400 Jul 17 04:48 ../
drw-r-----    3 root     root           100 Jul 17 04:48 containerd/
drwx------    2 root     root            40 Jul 17 04:48 runc/
```

Looks like the `dhcpd` image might need a tweak as well, or do we care?

dweomer:
Amended to replicate changes for the `init-lcow` image as well.

pkg/init-lcow/Dockerfile
```diff
@@ -38,4 +38,5 @@ WORKDIR /
 COPY --from=mirror /out/ /
 COPY --from=build /out/ /
 COPY --from=runc /usr/bin/runc /sbin/runc
+RUN ln -s /run /var/
```
rn: please don't change `init-lcow`. It does things completely differently (completely different init)

justincormack:
I think it might be better to change this in `tools/alpine` so that it applies to all packages. Otherwise there will be the same issue in each container.

ijc:
How did `/var/run` end up as not a symlink in the first place? It looks like we might just be inheriting this from the base `alpine:3.6` image, in which case the question is why is it not a symlink there and what might be broken by changing it downstream? Or maybe this is a bug/regression in `alpine:3.6` and we should report it there?

justincormack:
It does not seem to be a symlink in any of the recent Alpine Docker images. I don't currently have a physical install of Alpine due to retirement of my old Xen server, so not sure about there cc @ncopa.

rn:
from my alpine VM:
```
ls -l /var/run
lrwxrwxrwx    1 root     root             4 May  2 11:35 /var/run -> /run/
```

justincormack:
Ok, so it looks like a bug in the static filesystem generation that is used to build the Docker images at a guess cc @ncopa

ncopa:
I think it is `openrc` that creates the /var/run symlink, and openrc is not used in docker. I think it was done that way so it was possible to migrate from `/var/run` -> `/run` during bootup.

I think we should create the symlink in the official alpine docker image.

justincormack:
Ok, if it is going to be fixed upstream, I will close this PR.

PR #2228, opened 2017-07-17
kubernetes project updates

ijc:
This PR could be considered a WIP, in that there is plenty more to do, but I think it would be better to merge several tranches of changes rather than wait for the final result.

**- What I did**

Several things:
* Updated the `projects/kubernetes` build to use the more standard `pkg/package.mk` and `Dockerfile` based build scheme, dropping the use of the box tool. Cleaned up some of the previously semiauto generated scripts to become first class source (still more to do there). Image is in the `linuxkitprojects` org.
* Used image labels to simplify the yaml side of things.
* Updated the `boot.sh` to expose ssh port on the host (handy on Linux, although I'm not using the `ssh_into_kubelet.sh` script myself so I haven't touched it)
* Updated to the standard `pkg/mounts` image
* Disabled content trust for `gcr.io`, it has never supported it so the `image-cache` target can never have worked. It does now.

**- How I did it**

Emacs mostly.

**- How to verify it**

I've been using `boot.sh` on Linux, which suffices to check something boot but due to there being no network connectivity between individual QEMU VMs in the default setup I've also tested with two `virt-manager` vms.

Updates and improvements to the Kubernetes project.

[](https://en.wikipedia.org/wiki/Fallen_Angel_of_Doom)


ijc:
Rough laundry list of other stuff:
* Sort out build of image-cache images.
* Update to newer kube and weave.
* Use `pkg/metadata`.
* Replace use of alpine onboot service for a simple `mkdir` (on the persistent disk after `pkg/mount` or `pkg/format`) somehow.
* Make `ssh_into_kubelet.sh` work with `boot.sh` on Linux (without breaking MacOS!). Script may already be broken due to switch to `ctr` etc anyway
* Network connectivity between nodes when running with QEMU on Linux (perhaps `listen=` on master and `connect=` on nodes?).
* Try to reduce sizes of everything.

projects/kubernetes/Makefile
```diff
@@ -1,19 +1,10 @@
...
-push-container-images: build-container-images cache-images
-	docker image push linuxkit/kubernetes:latest
-	docker image push linuxkit/kubernetes:latest-mounts
+build-container-images:
+	make -C kubernetes tag
+
+push-container-images: cache-images
+	make -C Kubernetes push
 	docker image push linuxkit/kubernetes:latest-image-cache-common
 	docker image push linuxkit/kubernetes:latest-image-cache-control-plane
```
justincormack: shoudnt this be `linuxkitprojects`?

ijc: Was planning to do that when I reworked the build for those images.

You did make me notice `Kubernetes` in the line above though, should be lowercase...

projects/kubernetes/kube-master.yml
```diff
@@ -16,14 +16,14 @@ onboot:
...
+    image: linuxkit/mount:ac8939c4102f97c084d9ddfd445c1908fce6d768
+    command: ["/mount.sh", "/var/lib/"]
+  - name: var
+    image: library/alpine:3.6
+    command: ["mkdir", "/var/lib/kubeadm"]
     binds:
-     - /dev:/dev
-     - /var:/var:rshared,rbind
+     - /var/lib:/var/lib:rshared,rbind
+    rootfsPropagation: shared
```
justincormack: `mkdir` does not need shared root propagation, or `rshared` options, it is not mounting anything

justincormack:
needs a rebase already...

ijc:
Rebased, with a `kubernetes: Use pkg/metadata` thrown in for nothing.

Issue #2213, opened 2017-07-14
getty: Add some fancier namespace detection


dave-tucker:
Currently the `getty` and `sshd` containers `$PS1` contains `(ns: getty)` or similar to indicate that they may or may not be namespaced. There is also an accompanying message in the motd.

It would be much nicer to detect if we are namespaced and to put that in `$PS1` as the current method can be very confusing (e.g if you add `getty` to `init` in your YAML)

@justin suggested using the `NS_GET_PARENT` `ioctl` 

@ijc suggested doing a `readlink` on `/proc/1/ns/mnt` and `/proc/$$/ns/mnt` and comparing the two

I hacked together [a script](https://gist.github.com/dave-tucker/854f6a0fd024b3c4f06d1fbf3e77f969) to test the `ioctl` idea and ran this in Linuxkit as an `onboot` service with both `pid: new` and `pid: host`.
Unfortunately in both cases this returned `EPERM` which [according to the documentation](http://man7.org/linux/man-pages/man2/ioctl_ns.2.html) means:

> The requested namespace is outside of the caller's namespace
              scope.  This error can occur if, for example, the owning user
              namespace is an ancestor of the caller's current user
              namespace.  It can also occur on attempts to obtain the parent
              of the initial user or PID namespace.

Either that, or my code is wrong :trollface:

deitch:
A few other ideas:

1. Convention (file): Create a flag-file that exists only in the root, e.g. have `moby` insert a file called, e.g. `/.rootns`. Of course, anyone else could create that file in their own image, or bind-mount it, but at least there we have a convention. It is messy, I don't love it... but it should work.
2. List the containers and see if itself (getty or sshd) is running. Since we mount in the various `/run` and `/containers`, that also would work. Again, someone breaks the mounts, we have an issue. 
3. Check pid and/or PPID: If `$$ == 1`, you know you are in a container. If `$$ != 1`, you are either `host=pid` or in root so check the parent. For container, `$PPID` will be `runc` or something similar; for root host, it will be init (PID 1).

I kind of like 3, since

PR #2233, opened 2017-07-17
getty: Detect if you are in a namespace

dave-tucker:
This is an attempt at fixing #2213

If pid 1 isn't /sbin/init we're in a container
If we can see `containerd|runc` in the `pstree` then we're in the host
pid namespace but still in a container.
Otherwise, we're not in a namespace.

I"ve tested this with getty on `init`, in `services` with both a `host`
and `new` pid namespace and it seems to work as expected.

### DIFF
--- a/pkg/getty/etc/profile.d/namespace.sh
+++ b/pkg/getty/etc/profile.d/namespace.sh
@@ -1,1 +1,0 @@
-export PS1="(ns: getty) $PS1"
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -41,6 +41,14 @@ start_getty() {
 	infinite_loop setsid.getty -w /sbin/agetty $loginargs $line $speed $tty $term &
 }
 
+
+# check if we are namespaced, and, if so, indicate in the PS1
+if [ -z "$INIGETTY" ]; then
+	cat >/etc/profile.d/namespace.sh <<"EOF"
+export PS1="(ns: getty) $PS1"
+EOF
+fi
+
 # check if we have /etc/getty.shadow
 ROOTSHADOW=/hostroot/etc/getty.shadow
 if [ -f $ROOTSHADOW ]; then
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -43,7 +43,7 @@ start_getty() {
 
 
 # check if we are namespaced, and, if so, indicate in the PS1
-if [ -z "$INIGETTY" ]; then
+if [ -z "$INITGETTY" ]; then
 	cat >/etc/profile.d/namespace.sh <<"EOF"
 export PS1="(ns: getty) $PS1"
 EOF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -53,7 +53,7 @@ services:
     image: linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918
   # Enable getty for easier debugging
   - name: getty
-    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
+    image: linuxkit/getty:58620cff1b0bf8b5d144d087602115e996f18a02
     env:
         - INSECURE=true
   # Run ntpd to keep time synchronised in the VM
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -53,7 +53,7 @@ services:
     image: linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918
   # Enable getty for easier debugging
   - name: getty
-    image: linuxkit/getty:894eef1e5f62f3bc31de8ffaff2b6c0e093c4595
+    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
     env:
         - INSECURE=true
   # Run ntpd to keep time synchronised in the VM
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/usr/bin/mountie", "/var/lib/docker"]
 services:
   - name: getty
-    image: linuxkit/getty:894eef1e5f62f3bc31de8ffaff2b6c0e093c4595
+    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/usr/bin/mountie", "/var/lib/docker"]
 services:
   - name: getty
-    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
+    image: linuxkit/getty:58620cff1b0bf8b5d144d087602115e996f18a02
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,7 +16,7 @@ onboot:
     image: linuxkit/metadata:f5d4299909b159db35f72547e4ae70bd76c42c6c
 services:
   - name: getty
-    image: linuxkit/getty:894eef1e5f62f3bc31de8ffaff2b6c0e093c4595
+    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,7 +16,7 @@ onboot:
     image: linuxkit/metadata:f5d4299909b159db35f72547e4ae70bd76c42c6c
 services:
   - name: getty
-    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
+    image: linuxkit/getty:58620cff1b0bf8b5d144d087602115e996f18a02
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/getty.yml
+++ b/examples/getty.yml
@@ -14,7 +14,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:894eef1e5f62f3bc31de8ffaff2b6c0e093c4595
+    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
     # to make insecure with passwordless root login, uncomment following lines
     #env:
     # - INSECURE=true
--- a/examples/getty.yml
+++ b/examples/getty.yml
@@ -14,7 +14,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
+    image: linuxkit/getty:58620cff1b0bf8b5d144d087602115e996f18a02
     # to make insecure with passwordless root login, uncomment following lines
     #env:
     # - INSECURE=true
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -11,7 +11,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
+    image: linuxkit/getty:58620cff1b0bf8b5d144d087602115e996f18a02
     env:
      - INSECURE=true
 trust:
--- a/examples/minimal.yml
+++ b/examples/minimal.yml
@@ -11,7 +11,7 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
 services:
   - name: getty
-    image: linuxkit/getty:894eef1e5f62f3bc31de8ffaff2b6c0e093c4595
+    image: linuxkit/getty:b4cdc4101eb41ce2b8aaf1ef0209445ebdc577b4
     env:
      - INSECURE=true
 trust:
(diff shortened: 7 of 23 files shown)


pkg/getty/usr/bin/inans.sh
```diff
@@ -0,0 +1,17 @@
+#/bin/sh
+
```
justincormack: can we just make this a function in the other script?

dave-tucker: actually yes that'd be much simpler 😁

pkg/getty/etc/profile.d/namespace.sh
```diff
@@ -1 +1,23 @@
-export PS1="(ns: getty) $PS1"
+#/bin/sh
+
+in_a_ns() {
+	IN_A_NS=0
+	NOT_IN_A_NS=1
+
```
justincormack: I would add if `readlink /proc/self/ns/mnt` and `readlink /proc/1/ns/mnt` we are in a namespace, this works well with the current setup of shared pid namespace but not mount. (It doesnt work if you are in a pid namespace).

ijc:
Both the "pid == `/sbin/init`" and "can see `containerd` or `runc` in the pstree" heuristics are susceptible to false positives. I'd expect `/sbin/init` to not be that uncommon a name even in a container, running nested `containerd`+`runc` seems less likely, but is in no way out of the question.

pkg/getty/etc/profile.d/namespace.sh
```diff
@@ -1 +1,23 @@
...
+	if pstree | grep -qE "(containerd|runc)"; then
+		return $IN_A_NS
+	fi
+
+	# If none of the above are true, we're not in a namespace
+	return $NOT_IN_A_NS
+}
+
+if in_a_ns; then
+	export PS1="(namespaced) $PS1"
```
ijc: Don't want to keep the `getty` bit?

deitch:
> false positives ... no way out of the question

Agreed, but are we going to get much better than what @dave-tucker put in here, perhaps with the addition of @justincormack's suggestion?

To the best of my knowledge (please correct me), there is no absolutely reliable way for a process inside a namespace to find out that it is. 

The only other thing I could think of was a flag file on the root filesystem, a la `/.iamrootfs`, but even that could exist elsewhere. 

The question is, is this good enough for now?

ijc:
I think it is good enough, but I would prefer if the comments mentioned the possibility of false +ve explicitly. The `pstree` one already says "probably" but the `/sbin/init` sounds absolute as written. Maybe there should also be an overall intro para to the function explaining that the whole thing is just heuristics.

pkg/getty/etc/profile.d/namespace.sh
```diff
@@ -1 +1,23 @@
...
+#/bin/sh
+
+in_a_ns() {
+	IN_A_NS=0
+	NOT_IN_A_NS=1
+
+	# If PID 1 != /sbin/init then we are in a container
+	if ! pgrep /sbin/init > /dev/null; then
+		return $IN_A_NS
+	fi
```
ijc: This checks whether there is any process in the current pid namespace named `/sbin/init` but doesn't check if it is pid 1 AFAICT. An alternative to pgrep would be to consult `/proc/1/cmdline`.

pkg/getty/etc/profile.d/namespace.sh
```diff
@@ -1 +1,23 @@
...
+
+	# If containerd or runc are in the pstree then we're probably in a namespace
+	if pstree | grep -qE "(containerd|runc)"; then
+		return $IN_A_NS
+	fi
```
ijc: `pgrep` would suffice here I think. You might want to anchor the regex, certainly at the end but perhaps at the beginning and with a full path too (to avoid matching e.g. `docker-containerd`).

Thinking about this though, why does the presence of either of this indicate that we are in a namespace? They are both certainly present in the root too. What this check is really doing is trying to have a heuristic guess at whether we are in the root pid namespace, it needs to be combined with something else to have a guess about being in the root mnt namespace or not (it's the mnt namespace which was confusing people). @justincormack's `readlink` suggestion is probably the way to go here.

deitch:
> I would prefer if the comments mentioned the possibility of false +ve explicitly

Absolutely.

pkg/getty/etc/profile.d/namespace.sh
dave-tucker: I can change that to check that the output from `pgrep` is `1`
Or `/proc/1/cmdline`. I'm easy.

pkg/getty/etc/profile.d/namespace.sh
dave-tucker: @ijc in my experimentation, when you have an `init` getty, you don't see `containerd` or `runc`. You only see these when you are a `service` and sharing a `pid` namespace. Changing the check to look for the `mnt` namespace is probably the better one. I'll adjust appropriately.

ijc: That's very odd -- when you have an `init` getty the getty is running in the host name space and so the `containerd` _should_ be visible.

ijc: Be careful that your `grep` isn't matching itself:
```
(ns: getty) linuxkit-525400123456:~# ps aux| grep -E containerd\|runc
  391 root       0:00 /usr/bin/containerd
  413 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
  465 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
  507 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
  557 root       0:00 grep -E containerd|runc
```

I usually use a single character group to defeat the unwanted matching, but you need it in all the alternatives i.e.

```
(ns: getty) linuxkit-525400123456:~# ps aux| grep -E [c]ontainerd\|[r]unc
  391 root       0:00 /usr/bin/containerd
  413 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
  465 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
  507 root       0:00 containerd-shim --namespace default --address /run/containerd/containerd.sock
```

deitch: @dave-tucker I believe that is because of startup order. I have a getty set as init here. Look at the output:

```
(ns: getty) linuxkit-025000000009:~# ls /etc/init.d
000-issue       001-getty       010-onboot      020-containerd
```

getty will be run before any of the onboot and before containerd. So you don't see it.

At least the init is easily fixable. The `init.d` file for getty is:

```
INSECURE=true /usr/bin/rungetty.sh
```

We could easily set it to 

```
INSECURE=true INITGETTY=true /usr/bin/rungetty.sh
```

and then check for `INITGETTY` in `rungetty.sh`. Again, we could get hit by someone setting the env var on their own, but we can document it.

deitch: Um, hold on a sec. Doesn't that mean that in **all** cases where `[ -z "$INITGETTY" ]` it is in a namespace? Isn't everything _except_ init a namespace? Does that make all of the heuristics unnecessary?

dave-tucker: @ijc note this is `pstree` not `ps`.

Getty as a service:
```
init---containerd---containerd-shim---tini---rungetty.sh-+-rungetty.sh---login-+
                                                         `-rungetty.sh---login-+
```

Getty on init:
```
init---rc.init---001-getty---rungetty.sh-+-rungetty.sh---login---sh
                                         `-rungetty.sh---login---sh---pstree
```

ijc: > when you have an init getty, you don't see containerd or runc

This is because `/etc/init.d/001-getty` is calling `/usr/bin/rungetty.sh` synchronously and blocking the rest of the boot process -- i.e. containerd hasn't been started yet. This is (of course) a bug.

ijc: > @ijc note this is pstree not ps.

Ah, clever!

deitch: > blocking the rest of the boot process

pkg/getty/etc/profile.d/namespace.sh
justincormack: /proc/1/cmdline is much more direct

pkg/getty/etc/profile.d/namespace.sh
ijc: > It only blocks long enough to start all of the background processes.

The final line is a `wait`, so it will block forever since the children it is waiting for are actually the calls to `infinit_loop` which never exits.

justincormack: Backgrounding should be fine. I don't think the order matters much.

deitch: > it will block forever since the children it is waiting for are actually the calls to infinit_loop which never exits

Doh! Of course it is! It has to be to function in the container! Ouch.

ijc: Adding a `&` at the end of the initscript should work for both cases.

deitch: I am putting in a PR that does:

1. Sets `INITGETTY=true` env var for when getty is run as part of `init` (in `/etc/init.d/*getty`
2. `wait` only if `[ -z "$INITGETTY" ]`

deitch: See https://github.com/linuxkit/linuxkit/pull/2241

deitch:
So.. now that running off of `init` will always have `INITGETTY=true` (or will once #2241 gets merged in when CI completes), should we replace all that logic with:

```
if [ -z "$INITGETTY" ]; then
   IN_A_NS=1
else
   IN_A_NS=0
fi
```

deitch:
> should we replace all that logic with:

#2241 merged in.

dave-tucker:
I guess we could add it as another check. I think we'll still need the heuristics for SSH.

ijc:
Currently the `sshd` image only supports running as a container, so it could just be hardcoded for that case.

If we did add a way to run it embedded into the init we should just apply the same methodology now used for `getty` i.e. set a flag in the initscript.

dave-tucker:
Sorted then :) I'll update this PR shortly

dave-tucker:
Updated and ready for review...

deitch:
It doesn't get simpler than that! :-)

LGTM.

deitch:
OK, so interesting idea. Do *not* change this for now. I am just thinking ahead.

Do we want to have `init` set a var `ROOTNAMESPACE=true` (like the `INITGETTY`). That way everything called by root init will have it. Then we can have `runc` and `containerd` explicitly `unset` it (not even sure it is necessary to `unset` but whatever). 

Thoughts?

justincormack:
No, I don't like that, the root namespace should do as little as possible, and getty should look after itself...

justincormack:
This looks fine, but needs the container rebuilt and hashes updated. I think @dave-tucker may be on holiday, so maybe someone else wants to carry this? cc @deitch ?

deitch:
I signed off on this last week. I am hitting approve now. I am happy to do the push to CI and, when clean, merge, but if I am the one doing it, could someone else also "LGTM"?

deitch:
I will update the examples too....

justincormack:
LGTM once it has the package updates+tests pass

deitch:
And... it doesn't work. The problem is that `INITGETTY` is set at the `init` level, so `rungetty` (which is executed directly by it) sees it, but the `sh` itself (which runs `profile.d` does not, since it is launched by `login`, which explicitly destroys the root environment that is passed to the shell (except for `TERM`). You can disable it with `login -p`, but that is a **bad** idea.

Here is an alternate. Since we know the setting of `INITGETTY` when `rungetty.sh` runs, how about instead of creating `namespace.sh` at _image_ creation time, we change `rungetty.sh` (which knows if `INITGETTY` is set) to (pseudo-code):

```
if [ -n "$INITGETTY" -a ! -f namespace.sh]; then
  # create /etc/profile.d/namespace.sh
fi
```

This should work since it will only run if we are in a container and it is safe.

Thoughts?

justincormack:
Sounds reasonable. Remember that if it is not running in a container you won't be able to modify the filesystem at all, so what is in the container needs to just work in that case...

deitch:
> what is in the container needs to just work in that case

Makes sense. I am pushing out the change, you will see in the PR in a minute.

deitch:
Feh. `git`-isms tripping me up. Pulling a branch directly from his repo gets all messed up. Ugh.

justincormack:
You can cherry pick his commit and open a new PR if necessary

deitch:
I think I got it. When you `git pull https://....` it merges in, creating a new commit, etc. I just `git reset --hard <his_last_commit_hash>` and now I can work directly off the branch.

justincormack:
once you add and fetch the remote, you should just be able to do `git checkout ns-detect` and it should make a new branch tracking it for you. You don't ever need to pull...

deitch:
> You don't ever need to pull

Yeah, but you get lazy (or I do) and just copy-paste the command-line instructions from the PR page, and it all is downhill from there...

deitch:
This looks good. I am going to update the hashes in a separate commit. Tell me if this looks good?

justincormack:
yeah looks fine

deitch:
OK, image pushed out, hashes updated.

deitch:
Oh no, conflicts. Sigh.

deitch:
Eh, dumb mistake. Fixing.

deitch:
OK, hashes are correct. Now to fix the conflicts.

deitch:
Phew. Rebased.

deitch:
Thanks. I will merge in once CI is green.

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -41,6 +41,14 @@ start_getty() {
 	infinite_loop setsid.getty -w /sbin/agetty $loginargs $line $speed $tty $term &
 }
 
+
+# check if we are namespaced, and, if so, indicate in the PS1
+if [ -z "$INIGETTY" ]; then
```
ijc: Should this not be `$INITGETTY` (with the first `T`)?

deitch: It sure should, so how the heck did my tests pass??

deitch: Whatever, I don't care. Fixing, pushing. Of course that means CI will be another week until it runs through.

deitch:
OK, pushing out updated image with new hash, then will update the PR.

deitch:
Hashes updated too. I will check on CI in a week or two...

deitch:
OK, all tests pass. Good to merge?

deitch:
I will take that as a "yes".

PR #2235, opened 2017-07-17
Use the upstream dind package to run docker

justincormack:
It is pretty close to our docker package, if we adjust the command
that is run to avoid the actual dind startup script. We can't use
the normal docker image as it does not have mkfs and so on.

cc @djs55 @MagnusS we don't really want to be maintaining Docker packaging in LinuxKit, so can you test this works for you in Docker for Mac (or you can maintain your own packages).

library_cat


blueprints/docker-for-mac/docker-17.06-ce.yml
```diff
@@ -3,7 +3,7 @@ services:
   # Bind mounts /var/run to allow vsudd to connect to docker.sock, /var/vpnkit
   # for vpnkit coordination and /var/config/docker for the configuration file.
   - name: docker-dfm
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
+    image: library/docker:17.06.0-ce-dind
```
riyazdf: nit: you can omit `library` here if you'd like

examples/docker.yml
```diff
@@ -30,7 +30,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:19370f5d9bec84eb91073b7196b732f1301d9c90
   - name: docker
-    image: linuxkit/docker-ce:9b937df179bdbebbc70243779978057df0b54190
+    image: library/docker:17.06.0-ce-dind
```
riyazdf: same as above

justincormack:
Oops, missed some uses of docker, will update

justincormack:
@djs55 any chance you could test this with d4m?

djs55:
I can take a look later today hopefully.

djs55:
I did a little manual testing and it seems to work in D4M.

PR #2244, opened 2017-07-19
tools: multi-arch support for the base alpine image

arm64b:
Alpine is the base docker image for the LinuxKit, but currently
it only support amd64 architecture. This patch try to refactor
the alpine docker image build process in order to support other
architectures, such as arm64, by using the new capability of 'ARG'
and 'FROM' in Dockerfile.
Also this patch splits the standalone packages into 2 parts:
one is common for all archs, another is arch-specific.

**- What I did**
Support muti-arch build for tools/alpine docker image 
**- How I did it**
Take use of the new feature of 'ARG' and 'FROM' supported by newer version of Moby in the Dockerfile, thus we can assign a specific BASE image with '--build-arg BASE=xxx' as the base image used by 'FROM'.
 
**- How to verify it**
Navigate into the tools/alpine folder, 'make ORG=your_org' on amd64 and arm64 platform, then we can get a your_org/alpine:xxx image on different platforms, using this image as the LinuxKit base image for other tools and pkgs. 
$bin/moby build linuxkit.yml
$bin/linuxkit run ... linuxkit 

summit


arm64b:
@justincormack @rn 
please help to review those changes, especially those packages should be installed.

This RP is a key step for us to support ARM64 with LinuxKit mainline code. After that I will clone a brand new LinuxKit repos into different ARM64 platforms to test if we still have some gap to do that.

Thanks!

tools/alpine/.detect-arch-var
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
justincormack: please use `sh` not `bash`. Also having hidden files is kind of annoying, just use a normal one.

arm64b: OK. I will wrap up all the changes and make a re-push

justincormack:
I think there is a much simpler solution with a single `Dockerfile` as since Docker 17.06 you can use build args in the `FROM` line so the following works fine:
```
whale:arg justin$ cat Dockerfile 
ARG BASE=alpine:3.6
FROM $BASE
RUN cat /etc/alpine-release
RUN uname -a
whale:arg justin$ docker build --no-cache .
Sending build context to Docker daemon  2.048kB
Step 1/4 : ARG BASE=alpine:3.6
 ---> 
Step 2/4 : FROM $BASE
 ---> 7328f6f8b418
Step 3/4 : RUN cat /etc/alpine-release
 ---> Running in 288e5cbbe59c
3.6.2
 ---> 938b22c437cc
Removing intermediate container 288e5cbbe59c
Step 4/4 : RUN uname -a
 ---> Running in dda55f6687bd
Linux e1ede117fb1e 4.9.36-moby #1 SMP Wed Jul 12 17:33:58 UTC 2017 x86_64 Linux
 ---> e1b9fa9cb8c1
Removing intermediate container dda55f6687bd
Successfully built e1b9fa9cb8c1
whale:arg justin$ docker build --no-cache --build-arg BASE=arm64v8/alpine:3.6 .
Sending build context to Docker daemon  2.048kB
Step 1/4 : ARG BASE=alpine:3.6
 ---> 
Step 2/4 : FROM $BASE
 ---> 77dabab369a1
Step 3/4 : RUN cat /etc/alpine-release
 ---> Running in e409bdf4f721
3.6.2
 ---> 2c52ad7e3a2e
Removing intermediate container e409bdf4f721
Step 4/4 : RUN uname -a
 ---> Running in 43acd0217e69
Linux bf9f9dc31cee 4.9.36-moby #1 SMP Wed Jul 12 17:33:58 UTC 2017 aarch64 Linux
 ---> 79d388609c80
Removing intermediate container 43acd0217e69
Successfully built 79d388609c80
```

justincormack:
(this would also allow us to override the base image for any architecture, which could be useful for testing alpine `edge` vs release)

arm64b:
Ah, cool! let me try that, we can use only one Dockerfile if that works. My docker version is:
Client:
 Version:      unknown-version
 API version:  1.31
 Go version:   go1.8.3
 Git commit:   5dd3073
 Built:        Sun Jun 18 07:54:10 2017
 OS/Arch:      linux/arm64

Server:
 Version:      17.06.0-dev
 API version:  1.31 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   b931c35
 Built:        Sun Jun 18 08:02:11 2017
 OS/Arch:      linux/arm64
 Experimental: false

justincormack:
That should be new enough. We are fine with requiring developers to have newest docker versions, we dd that before for multi stage builds too.

arm64b:
@justincormack  yes, it works. So that means we can use one Dockerfile. Now have another issue:
https://github.com/linuxkit/linuxkit/blob/master/tools/alpine/Dockerfile#L48
ovmf pkg is missed from arm64v8/alpine, does 'COPY' has some tricks like 'FROM'?

justincormack:
I think the easiest thing to do is something like:

```
RUN mkdir -p /usr/share/ovmf
RUN [ $(uname -m) = "x86_64 ] && apk add -X http://dl-cdn.alpinelinux.org/alpine/edge/community ovmf 

...

COPY --from=mirror /usr/share/ovmf/ /usr/share/ovmf/
```

arm64b:
Good point!
So how about the packages split?  Since some pkgs in the 'packages' file is not available for arm64 such as open-vm-tools, syslinux...

justincormack:
I think move those into `packages.x86_64`, add a (maybe empty) `packages.aarch64` and then always install `packages` and `packages.$(uname -r)`

arm64b:
like this?

COPY packages* /tmp/

# mirror packages
RUN  cat /tmp/packages.$(uname -m) >> /tmp/packages && \
   mkdir -p /mirror/$(uname -m) && \
   apk fetch --recursive -o /mirror/$(uname -m) $(apk info; cat /tmp/packages)

justincormack:
yes thats what I was thinking.

arm64b:
OK, I will wrap up those changes and re-push it again. Thanks for the good suggestion!

tools/alpine/Makefile
```diff
@@ -2,22 +2,24 @@
 
 ORG?=linuxkit
 IMAGE=alpine
-BASE=alpine:3.6
+
+BASE := $(shell bash -c 'source detect-arch-var && echo $${BASE}')
```
rn: Could we not use bash here and in the next?

tools/alpine/Makefile
```diff
@@ -2,22 +2,24 @@
...
 
 default: push
 
 show-tag:
 	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions
 
 hash: Dockerfile Makefile packages
-	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
-	docker build --no-cache -t $(IMAGE):build .
+	DOCKER_CONTENT_TRUST=$(DOCKER_CONT) docker pull $(BASE)
```
rn: It would be better to copy the pattern used in `pkg/package.mk` for setting the content trust.

tools/alpine/detect-arch-var
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
```
rn: Can't this be moved into the Makefile, ie set a makefile variables for `uname -m` and then set other makefile variables based off that?

tools/alpine/Makefile
arm64b: very good suggestion. I will drop the 'detect-arch-var' file and transfer its operation into Makefile, so I promise 'bash' will not show again ;-)

tools/alpine/Makefile
arm64b: Of course. Actually I do think that's will be very elegant, at least looks like - as we'll move all the jobs in 'detect-*" into Makefile.

tools/alpine/detect-arch-var
arm64b: Sure, of course! But interesting, for the 'ARG' and ‘FROM’ feature supported by the newer version of Docker, seems there's a potential bug for the DOCKER_CONTENT_TRUST setting:
$export DOCKER_CONTENT_TRUST=1
$docker build --no-cache --build-arg BASE=alpine:3.6 -t alpine:build .
You will get below message:
Sending build context to Docker daemon 

error during connect: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.31/build?buildargs=%7B%22BASE%22%3A%22alpine%3A3.6%22%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&nocache=1&rm=1&shmsize=0&t=alpine%3Abuild&target=&ulimits=null: invalid reference format: repository name must be lowercase

@justincormack  @rn  please help double check this, if it's done, my PR will be ready quickly to push...

justincormack: hmm, yes I filed an issue https://github.com/moby/moby/issues/34199 will try to get it fixed ASAP.

As a temporary workaround, you could omit setting `DOCKER_CONTENT_TRUST` on the build step if the build arg is set.

arm64b: Great! I will set the DOCKER_CONTENT_TRUST=0 as default in my PR, it will be easy to be tuned once this issue fixed.

arm64b:
@justincormack @rn  The gun is ready to shoot... ;-)

tools/alpine/Makefile
```diff
@@ -2,22 +2,38 @@
...
 default: push
 
 show-tag:
 	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions
 
-hash: Dockerfile Makefile packages
-	DOCKER_CONTENT_TRUST=1 docker pull $(BASE)
-	docker build --no-cache -t $(IMAGE):build .
+hash: Dockerfile Makefile $(PKG_DEPS)
+	docker pull $(BASE)
```
justincormack: you can keep the `DOCKER_CONTENT_TRUST=1` here (unless the arm64v8 packages are not signed, in which case it needs to be conditional; it was only haveing it set on `docker build` that causes issues.

justincormack: Ah, ok, the arm64 images are not signed, so I think you can just set `NOTRUST` for the arm64 case. But you must not disable for the x86_64 case. This Makefile does not set content trust for build as it already does a trusted pull, so should not run into the bug.

justincormack: @riyazdf can probably help work out the best arrangement (I am on a train today)

riyazdf: agreed with @justincormack that we must keep x86_64 trust checking intact.

riyazdf: perhaps we can `export DOCKER_CONTENT_TRUST=1` for this Makefile and set `NOTRUST` in the arm conditional where we override `BASE`?

arm64b: @justincormack @riyazdf  
Ah, thanks for the remind! I totally agree with that we need to keep x86_64 trust checking intact, I am missing that. So now I made below changes:
`diff --git a/tools/alpine/Makefile b/tools/alpine/Makefile
index 9d9b184..4fcad2a 100644
--- a/tools/alpine/Makefile
+++ b/tools/alpine/Makefile
@@ -3,6 +3,7 @@
 ORG?=linuxkit
 IMAGE=alpine
 
+DOCKER_CONTENT_VAR=1
 PKG_DEPS=packages
 ifeq ($(shell uname -m), x86_64)
 BASE=alpine:3.6
@@ -12,12 +13,7 @@ endif
 ifeq ($(shell uname -m), aarch64)
 BASE=arm64v8/alpine:3.6
 PKG_DEPS += packages.aarch64
-endif
-
-ifeq ($(DOCKER_CONTENT_TRUST),)
-ifndef NOTRUST
-export DOCKER_CONTENT_TRUST=0
-endif
+DOCKER_CONTENT_VAR=0
 endif
 
 default: push
@@ -26,14 +22,14 @@ show-tag:
        @sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions
 
 hash: Dockerfile Makefile $(PKG_DEPS)
-       docker pull $(BASE)
+       DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker pull $(BASE)
        docker build --no-cache --build-arg BASE=$(BASE) -t $(IMAGE):build .
        docker run --rm $(IMAGE):build sh -c 'echo Dockerfile /lib/apk/db/installed $$(find /mirror -name '*.apk' -type f) $$(find /go/bin -type f) | xargs cat | sha1sum' | sed 's/ .*//' > $@
 
 push: hash
-       docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
+       DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
                (docker tag $(IMAGE):build $(ORG)/$(IMAGE):$(shell cat hash) && \
-                docker push $(ORG)/$(IMAGE):$(shell cat hash))
+                DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker push $(ORG)/$(IMAGE):$(shell cat hash))`

With this change, we can keep x86_64 has the same behavior as it does before the update.
Let me know if any suggestions. Thanks for the review!

arm64b:
@rn @justincormack @riyazdf 

would you pls help to review the latest update?

tools/alpine/Dockerfile
```diff
@@ -25,7 +27,10 @@ RUN cp /mirror/$(uname -m)/APKINDEX.unsigned.tar.gz /mirror/$(uname -m)/APKINDEX
 RUN abuild-sign /mirror/$(uname -m)/APKINDEX.tar.gz
 
 # fetch OVMF for qemu EFI boot (this is not added as a package)
-RUN apk add -X http://dl-cdn.alpinelinux.org/alpine/edge/community ovmf
+RUN mkdir -p /usr/share/ovmf && \
```
rn: This should probably be added the same way the wireguard tools were added a yesterday. The way this is currently added (and was in the past) does not record the version number in the `versions` file. I know it's not related directly to your PR, so maybe should be addressed separately as it also has implications elsewhere.

tools/alpine/Makefile
```diff
@@ -2,22 +2,34 @@
...
+	docker build --no-cache --build-arg BASE=$(BASE) -t $(IMAGE):build .
 	docker run --rm $(IMAGE):build sh -c 'echo Dockerfile /lib/apk/db/installed $$(find /mirror -name '*.apk' -type f) $$(find /go/bin -type f) | xargs cat | sha1sum' | sed 's/ .*//' > $@
 
 push: hash
-	DOCKER_CONTENT_TRUST=1 docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
+	DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
 		(docker tag $(IMAGE):build $(ORG)/$(IMAGE):$(shell cat hash) && \
-		 DOCKER_CONTENT_TRUST=1 docker push $(ORG)/$(IMAGE):$(shell cat hash))
+		 DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker push $(ORG)/$(IMAGE):$(shell cat hash))
 	echo "# $(ORG)/$(IMAGE):$(shell cat hash)" > versions
```
rn: the `versions` file probably needs to be per arch

arm64b: Not very sure about the exact usage purpose of this file, if it really needs to generate a per arch, then we can do like this:
> versions.$(shell uname -m)
But IMO, if you work on a platform, no matter amd64 or arm64, it's easy for you to know that the version file is the list of packages installed for that platform. 
I am OK for both.

tools/alpine/Dockerfile
arm64b: I am trying to generate a per-arch version file, will double check if the ovmf pkg has been recorded in that file like version.x86_64.

tools/alpine/Makefile
ijc: This file is checked in and is expected to contain the exact versions present in the image hash in the header. Given that this package is built with network access it might be quite hard to ensure that there is no skew due to arches being built at different times (or even atthe same time if there is skew in the upstream alpine repo).

IMHO it would be best to find a way to build a single arch agnostic base mirror image with a multiarch mirrror in it, but for now I think per arch version files are the minimum we need.

arm64b: Ah, I am making the per-arch version file building test, will generate versions.x86_64 on x86_64 platform, versions.aarch64 on AArch64.

arm64b:
@rn @justincormack @ijc @riyazdf 
The latest update to address the 'versions' file, pls help to resolve the conflict if possible. Thanks!

tools/alpine/Makefile
```diff
@@ -2,24 +2,36 @@
...
+ifeq ($(shell uname -m), x86_64)
 BASE=alpine:3.6
+PKG_DEPS += packages.x86_64
+endif
+
+ifeq ($(shell uname -m), aarch64)
+BASE=arm64v8/alpine:3.6
+PKG_DEPS += packages.aarch64
+DOCKER_CONTENT_VAR=0
+endif
```
rn: I lost track of this part, but is there a reason, this is not following the pattern we have elsewhere (`pkg/package.mk`) and simply just sets `NOTRUST` when on arm?

justincormack:
Needs a rebase (versions changed).

tools/alpine/Makefile
arm64b: The reason is, if we export DOCKER_CONTENT_TRUST=1 for x86, then we will encounter the issue filed by https://github.com/moby/moby/issues/34199 no matter we use:
$docker build --no-cache --build-arg BASE=alpine:3.6 -t alpine:build .
or
$docker build --no-cache  -t alpine:build .

That's why we still use the original Makefile content trust mode when building the image.

arm64b:
Will wrap up and edit the commit message in the 1st PR submitted since we have a little big change for now @justincormack

arm64b:
@justincormack @riyazdf @rn 
The latest commit is 516320b

tools/alpine/Makefile
rn: Thanks for reminding me

rn:
@arm64b apologies, but this needs another rebase. I accidentally merged a later PR which introduced conflicts with this one. Otherwise this looks good to me and I'll merge after the rebase

arm64b:
@rn  rebase to the latest code base. it's not a straightforward rebase since we need to take care about the hash value of the versions.* in case installed pkgs changed

Issue #2206, opened 2017-07-14
Add apk to getty and sshd containers


ijc:
These are the login containers so it is reasonable for a user to want to install new tools in them.

I tried to make this happen but ran into a problem which I'm unsure how to approach, hence this issue.

The problem I found was that `/var` is bind mounted from the host in both the `getty` and `sshd` containers which means that various `apk` state directories are missing and `apk` is very unhappy.

Presumably the answer is to do more fine grained binds into those containers but I haven't time to do that now so just synching state here. My WIP patch (based on #2205) is just:
```diff
commit 9c80b3315d20a41d11c71df84caeb8a4cb767fcd
Author: Ian Campbell <ijc@docker.com>
Date:   Fri Jul 14 12:52:56 2017 +0100

    sshd+getty: Add apk to these containers
    
    Since these are the user login containers, having the ability to add packages
    is useful (e.g. I quite often find I want strace).
    
    Signed-off-by: Ian Campbell <ijc@docker.com>

diff --git a/pkg/getty/Dockerfile b/pkg/getty/Dockerfile
index c1f26bf5..81878828 100644
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -1,15 +1,17 @@
-FROM linuxkit/alpine:9bcf61f605ef0ce36cc94d59b8eac307862de6e1 AS mirror
+FROM linuxkit/alpine:3744607156e6b67e3e7d083b15be9e7722215e73 AS mirror
 
 RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out \
     alpine-baselayout \
+    apk-tools \
     busybox \
     ca-certificates \
     musl \
     tini \
     util-linux \
     && true
-RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN mv /out/etc/apk/repositories.upstream /out/etc/apk/repositories
+
 #
 # We require a version of `setsid(1)` which supports the `-w`
 # option, which is not available in all implementations (e.g. the
diff --git a/pkg/sshd/Dockerfile b/pkg/sshd/Dockerfile
index ea724738..0ad93894 100644
--- a/pkg/sshd/Dockerfile
+++ b/pkg/sshd/Dockerfile
@@ -1,8 +1,9 @@
[...]

PR #2249, opened 2017-07-19
Add wireguard to core of LinuxKit

zx2c4:
This contains two commits: the first moves the kernel patch into the main kernel tree. The second moves the userspace tools package into a real package, following the ordinary makefile conventions.

These together drastically reduce complexity and redundant redundancy.


rn:
@zx2c4 thanks for this. I would actually prefer individual patches for the kernel with commit messages and Signed-off-by in the commit. I maintain a branch for every kernel under linuxkit/linux and having smaller patches will allow me to rebase easier if you upstream more bits

zx2c4:
@rn Not parsing your message correctly. Do you mean to say that you like the fact that I've unified the patches into one single patch for all kernels, or that you don't like this?

If you don't like it, the solution would be to just copy it wholesale instead of making symlinks. The code will still be identical. There's only one WireGuard patch, and it works on kernels ≥3.10.

linuxkit.yml
```diff
@@ -6,6 +6,7 @@ init:
   - linuxkit/runc:d5cbeb95bdafedb82ad2cf11cff1a5da7fcae630
   - linuxkit/containerd:b6ffbb669248e3369081a6c4427026aa968a2385
   - linuxkit/ca-certificates:67acf038c44bb191ebb704ec7bb39a1524052cdf
+  - linuxkit/wireguard-tools:26fe3d38455f2d441549e3c54bdec1b26ac819b8
```
justincormack: For now maybe add the tools into the `getty` and `sshd` packages so people can try them out; adding to the top level is not very useful. Obviously in future they may need to be in other packages too (eg Docker) unless they are integrated in instead.

projects/wireguard/README.md
```diff
@@ -15,14 +15,10 @@ A full technical paper from NDSS 2017 is available [here](https://www.wireguard.
 ## Contents
 
 ### Kernel Patches
-This project keeps Linux kernel patches for WireGuard against a 4.9.x kernel.  
-This kernel is built into the `mobylinux/kernel-wireguard` image that is generated by `cd kernel-wireguard && make`.
-
-WireGuard can also be included as a kernel module.
+The `kernel/` sub-directory contains a kernel patch that is included by the main kernel package of LinuxKit.
```
justincormack: I would move the docs to `docs/wireguard.md` - once we merge this we can remove the wireguard project directory, and and a note in projects/README.md saying it was merged.

linuxkit.yml
zx2c4: I assume, though, that many people will use this to make transparent VPN proxies. That is -- boxes that have no ssh or serial or shells at all, but just configure a VPN and enable packet forwarding. Especially on Mac, people have expressed interest in this type of configuration. (And LinuxKit/Moby provides a clear way of doing that.)

So, with that in mind, I'm not sure it makes so much sense to bind this to sshd or getty. It belongs at the same level as `iproute2`. So, wherever that is is where wireguard-tools should go.

projects/wireguard/README.md
zx2c4: Good idea. Will add another commit to this PR for that.

zx2c4:
In light of @justincormack's comment on docs, I'm going to get rid of the symlinking and also get rid of projects/wireguard/*, putting everything in their right place. Will force push changes to this PR.

linuxkit.yml
justincormack: `iproute2` isn't in the base either. The expectation would be that you would have a container that sets up the VPN. This is probably just a short script running in the `wireguard-tools` package.

justincormack: We should have an example of this; does not have to be in this PR though.

rn:
Sorry, I actually commented after only taking a cursory glance at the kernel patch. Now I've taken a closer look and it looks like almost the entire diff is to add `./net/wireguard` with only a tiny patch to `./net/Makefile` at the end.

My workflow is that we have a copy of the linux stable tree at https://github.com/linuxkit/linux. For each release we have a branch off the release tag with our patches applied. For a new release, I just rebase that branch to the new release tag. Then our patches in `kernel/patches-4.Z.x` are generated with `git format-patch` from the new branch.

So, I need to get your patch into https://github.com/linuxkit/linux. For this I need the patch in `git am` format, so that it includes a commit message and crucially the DCO. Alternatively, and even better, if you have a public copy of the linux stable tree with your patches applied, I can just add that as a remote and cherry-pick. The latter has the advantage that if you update your kernel code, I can just re cherry-pick.

For the patch itself, I would slightly prefer if it's broken up into two commits, one which adds `./net/wireguard` and one which modifies `./net/Makefile`, but that's not that important.

does this make it clearer?

zx2c4:
@rn I understand now. Unfortunately what you want isn't going to happen until WireGuard is prepared for upstream inclusion. Right now it's a tree for being an external module, with a script to take that tree and create a patch for a real kernel. The tree is here: https://git.kernel.org/pub/scm/linux/kernel/git/zx2c4/WireGuard.git/tree/src

And the script is here: https://git.kernel.org/pub/scm/linux/kernel/git/zx2c4/WireGuard.git/tree/contrib/kernel-tree/create-patch.sh

So, what I suspect might work somewhat better is adding a git submodule for the WireGuard tree, as one commit, and then adding the net/Makefile and net/Kconfig in another commit.

What would you think of that?

linuxkit.yml
zx2c4: Okay, so it seems then, what you'd like, is for this not to be included in any real `.yml` file, but perhaps just an examples/ file, and then when people actually want to use it, they'll include wireguard-tools themselves?

Or do you not even want this to live in pkgs/ at all, and instead just have people install this from apk from within? This seems like it might be _slightly_ more reasonable, with one small problem being that wireguad-tools is only on Alpine testing.

justincormack: Yes; can add an example later.

We could use the packages from Alpine testing, we currently do a similar thing with one package from community see https://github.com/linuxkit/linuxkit/blob/master/tools/alpine/Dockerfile#L28 - if you add it to here as well it is then easy to add to any package.

rn:
@zx2c4 ah, ok.

I don't really want to mess with submodule but there are two other options:
- we clone the wireguard repo in https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile and build it as a module as part of the kernel build
- we modified `create-patch.sh` slightly to add a commit and DCO in `git am` format and I import that into the https://github.com/linuxkit/linux.

The first option seems easier as we just need to update the git commit hash for `wireguard.git` when you have an update. I can work on the changes tomorrow (I'm UK based so evening here)

zx2c4:
@rn if that first option is appealing to you, then I'm really not certain why the current situation in this PR doesn't work for you. Specifically, doing the repo clone inside the Dockerfile is identical to what we're doing here, but just more of a headache. In both cases, you don't wind up with the patch in the linuxkit/linux repo, but you do wind up with a module (or, better, a built-in, which is what this PR is doing). What's the distinction you were seeing?

rn:
@zx2c4 there is a big difference and that is that currently *all* patches in `./kernel/patches*` come from `https://github.com/linuxkit/linux` and they are uniform in the sense that they have commit messages, DCO, `cherry-pick -x` reference, and a `Origin` field pointing to the linux tree they were cherry picked from  in case they are not from the stable tree (or linus' tree). The current PR breaks this uniformity.

To me, and given how https://git.kernel.org/pub/scm/linux/kernel/git/zx2c4/WireGuard.git is currently organised for external module compile, making the wireguard patch fit with the above (my second option) seems more work than cloning a specific commit and essentially compile it as an external module (which it already is).

Either way, its part of that the standard LinuxKit kernel package, but the update process for the first option is simpler (just update the GIT commit hash in the kernel build Dockerfile).

zx2c4:
@rn Okay that makes sense. I'll fix up this PR to include that logic. It'll clone the WireGuard repository, and then checkout a particular snapshot tag, jerry rig itself into the kernel tree, and then build. Will push a new  commit here in a few hours.

rn:
@zx2c4 I'm not suggesting to jerry rig, I'm suggesting to compiling it as an external module as is and then copy it to the right place that the rest of the kernel build includes it in the right tarball in the kernel package. It won't take me long to do that tomorrow and it might be better/quicker if I add it tomorrow instead of iterating on a PR.

rn:
@zx2c4 there are two other reasons, why I'd like to compile it as an external module as part of the kernel build:
- we need to handle 1-2 kernel updates a week for 4.9.x, 4.11.x, and soon 4.12.x (all in debug on none debug config). If we compile wireguard as classic external module and the compile fails it's easier to debug. If we jerry rig the failure would be as part of the normal kernel compile and harder to debug.
- We may want to add additional external modules in the same way. I'm considering sysdig, for example. So having one pattern for handling this makes thing easier.

zx2c4:
>  If we jerry rig the failure would be as part of the normal kernel compile and harder to debug.

In practice, external modules wind up being a bit harder to debug, actually.

> So having one pattern for handling this makes thing easier.

That makes sense.

I've got the code written and it appears to be working. Validating a few things and then I'll push a commit. Sit tight... a few more minutes.

kernel/Dockerfile
```diff
@@ -98,8 +98,16 @@ RUN make defconfig && \
     cp System.map /out && \
     ([ -n "${DEBUG}" ] && cp vmlinux /out || true)
 
+# WireGuard
+RUN TAG="0.0.$(curl -sSL https://git.zx2c4.com/WireGuard/info/refs/ | sed -n 's/.*refs\/tags\/0\.0\.\([0-9]\+\)$/\1/p' | sort -nr | head -n 1)" && \
```
zx2c4: In this line here, I automatically determine the latest version by examining the tags. I like this, but I would understand if you prefer to instead hard code a particular snapshot version instead. Either way is fine with me; just let me know.

zx2c4:
@rn Okay, all set. Let me know what you think of this.

kernel/Dockerfile
rn: For consistency, I would prefer to use the pattern we use at different places like [here](https://github.com/linuxkit/linuxkit/blob/master/pkg/init-lcow/Dockerfile#L11). Basically define and environment variable for `WIREGUARD_REPO` and `WIREGUARD_COMMIT`.

kernel/Dockerfile
```diff
@@ -98,8 +98,16 @@ RUN make defconfig && \
...
+    git clone https://git.zx2c4.com/WireGuard/ && \
+    git -C WireGuard verify-tag -v "$TAG" && \
+    git -C WireGuard checkout -b docker-build "$TAG" && \
+    make -j "$(getconf _NPROCESSORS_ONLN)" M="$PWD/WireGuard/src" modules
```
rn: I would also suggest to explicitly checkout wireguard in, say `/WireGuard` and then replace the use of `$PWD` with the fixed path

kernel/Dockerfile
```diff
@@ -38,7 +38,7 @@ ENV KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VE
 ENV KERNEL_SHA256_SUMS=https://www.kernel.org/pub/linux/kernel/v4.x/sha256sums.asc
 ENV KERNEL_PGP2_SIGN=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.sign
 
-# PGP keys: 589DA6B1 (greg@kroah.com) & 6092693E (autosigner@kernel.org) & 00411886 (torvalds@linux-foundation.org)
+# PGP keys: 589DA6B1 (greg@kroah.com) & 6092693E (autosigner@kernel.org) & 00411886 (torvalds@linux-foundation.org) & A5DE03AE (jason@zx2c4.com)
```
rn: nice!

kernel/Dockerfile
zx2c4: Will do. Though it'll be `WIREGUARD_TAG` instead of `WIREGUARD_COMMIT`, since I sign tag objects, so then it can be explicitly verified via `git verify-tag`.

kernel/Dockerfile
zx2c4: ACK.

kernel/Dockerfile
```diff
@@ -98,8 +101,14 @@ RUN make defconfig && \
     cp System.map /out && \
     ([ -n "${DEBUG}" ] && cp vmlinux /out || true)
 
+# WireGuard
+RUN git clone --depth 1 -b "$WIREGUARD_TAG" "$WIREGUARD_REPO" /wireguard && \
```
zx2c4: You'll notice here that I'm using doing a shallow clone, since all we really want is the tag. Then in the next line, I verify that the tag has a valid GPG signature. This method works fine.

However, since we're using `WIREGUARD_TAG` now, instead of my fancy auto-latest magic of the last commit, it means we have to update this file for snapshot bumps. That's fine. But it does raise the question: why are we even using git at all? The WireGuard upstream project releases tarballs of these tags along with their checksums in a GPG signed email. It'd probably be more efficient to just download this tarball and check the checksum. Updates would then update not only the `WIREGUARD_TAG` but also the `WIREGUARD_HASH` variable.

Would you prefer this? It makes more sense to me anyhow.

rn: release tarballs are great if they can be verified. We also do that in a couple of places (including in this Dockerfile :). I'd be happy with that

zx2c4:
@rn Moved from using git to using the tarballs and doing the checksum. Should be ready to merge now.

pkg/wireguard-tools/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM linuxkit/alpine:3744607156e6b67e3e7d083b15be9e7722215e73 as tools
+RUN echo http://dl-cdn.alpinelinux.org/alpine/edge/main > /etc/apk/repositories && \
+    echo http://dl-cdn.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories && \
+    apk update && \
```
justincormack: I dont think there is any point having a wireguard-tools package, it doesnt actually do anything - my suggestion was adding these to tools/alpine directly so they can easily be added to another package, and then make packages that do something specific, like setting up a vpn.

zx2c4: I can play with this, but I'm not sure it will work as we want: wireguard-tools is in `edge/testing`, whereas tools/alpine uses a release, and things could get funky mixing and matching. Will play though.

(Maybe we want to merge this PR as is, and I'll submit subsequent ones for doing different things with wireguard-tools?)

justincormack: Yes, it usually works and then it breaks... we should probably build from source. Can we just remove all of `pkg/wireguard-tools` for now, happy to build some real examples next week/tomorrow, lets just merge the kernel bits

rn:
compiled the kernel and verified contents of the package. The kernel side looks all good to me.

@justincormack @zx2c4 I'm not too sure about the `wireguard-tools` package either. How about removing it for now and then in a separate PR work on an example. I definitely would want it in the "interactive" packages `getty` and `ssh` to start playing with it :)

zx2c4:
I'll work on adding wireguard-tools to tools/alpine as part of this PR then. Nearly there.

zx2c4:
That turned out to be very simple. This PR gets shorter every iteration!

ijc:
I've just figured out a solution for https://github.com/linuxkit/linuxkit/issues/2206 (it's based on https://github.com/linuxkit/linuxkit/pull/2258 so waiting for that to merge before posting), with that it should be possible to `apk add -U wireguard-tools` in the `sshd` or `getty` containers (possibly after `echo some-testing-repo >> /etc/apk/repositories`). That probably reduces the need for having it in the default `sshd` or `getty` containers (looks like there are still good reasons to have it in the alpine base though e.g. a `pkg/wireguard-vpn` as suggested by @justincormack).

zx2c4:
@ijc yes indeed it's possible to mix and match like this. But I think given this is something that interacts with our kernel, and most use cases are actually non-interactive, so putting it with sshd and getty doesn't make sense. The latest commit here solves everything. I think we're finished here.

rn:
@zx2c4 sorry, could you rebase. Looks like the other changes to `tools/alpine` required for the containerd update caused this

zx2c4:
@rn Rebased. Please merge before another conflict occurs.

justincormack:
@zx2c4 will do!

riyazdf:
postfacto LGTM, awesome work @zx2c4 🎉

PR #2252, opened 2017-07-19
Use kernel-config project style config for kernel/

tych0:
Hi all,

Here's an attempt at using kernel-config style config management for kernel/. I have built and boot tested the kernel images (except 4.4.x on arm64). However, it's worth noting that the drivers enabled are mostly based on the x86 config. I'm not sure what exactly our criteria for enabling stuff on arm64 is.

Please, have a look at the config files. While they should be equivalent to the ones generated previously (the diffs are close, with some exceptions below) I've separated most of the arch-specific options by hand+trial and error, so it's quite possible I've missed something.

Notable changes between kernel versions:

* 4.4: CONFIG_NET_FOU=y (must match CONFIG_NET_FOU_IP_TUNNELS)
* 4.4 -> 4.9 IPV6_TUNNEL m to y
* 4.4 -> 4.9 BLK_DEV_NVME m to y
* 4.4 -> 4.9 GENEVE y to m
* STRICT_DEVMEM is arch specific in 4.4, but i've listed it in the generic
  files because it moves to generic later on
* 4.9 -> 4.11 drops NETFILTER_XT_MATCH_SOCKET
* 4.9 -> 4.11 PHYLIB m to y
* UPROBE_EVENT was renamed to UPROBE_EVENTS (note the s) in 4.11; we set
  UPROBE_EVENT, but not UPROBE_EVENTS

I think r.e. the last one, we should definitely enable UPROBE_EVENTS. However, I've not done it here, just to keep the diffs semantically as close as possible. I can send follow on commits to fix the above up.

Thoughts welcome!


ijc:
> Notable changes between kernel versions:

Those are the differences between the (new) 4.4, 4.9 and 4.11? Given "the diffs are close, with some exceptions below" I think a more interesting list would be the differences for each kernel version before and after this change i.e. "before 4.4 had CONFIG_FOO=m and now 4.4 has CONFIG_FOO=y because ...".

Ideally with a "build infrastructure" change of this sort the output `.config` file (e.g. what is produced after running `make oldconfig` or some `make fooconfig` in the input config) should be identical before and after this PR.

IMO any actual changes to the desired configuration should come separately with their own justifications.

tych0:
On Thu, Jul 20, 2017 at 09:18:22AM +0000, Ian Campbell wrote:
> > Notable changes between kernel versions:
> 
> Those are the differences between the (new) 4.4, 4.9 and 4.11? Given "the diffs are close, with some exceptions below" I think a more interesting list would be the differences for each kernel version before and after this change i.e. "before 4.4 had CONFIG_FOO=m and now 4.4 has CONFIG_FOO=y because ...".

Sorry, this is poorly written. These are differences between the
kernel configs before this patch (not including any stuff that was
added in later versions). Mostly this is just to point out some
discrepancies that we could/should resolve.

> Ideally with a "build infrastructure" change of this sort the output `.config` file (e.g. what is produced after running `make oldconfig` or some `make fooconfig` in the input config) should be identical before and after this PR.
> 
> IMO any actual changes to the desired configuration should come separately with their own justifications.

Yep, agreed. Of the list I made above, it looks like I kept some and
dropped some. I'll make it so that we keep all the differences, and we
can work out what to do about it in a separate thread.

ijc:
Thanks for the explanation!

tych0:
Ok, I've updated it so the old versions of things are reflected (except for NET_FOU on 4.4.x, which needs to be =y as noted above, other wise it won't build).

rn:
doing a `make build_4.11.x` i get:
```
#
# configuration written to .config
#
scripts/kconfig/conf  --oldconfig Kconfig
.config:3789:warning: override: DEFAULT_SECURITY_DAC changes choice state
*
* Restart config...
```
Is that warning benign?

I also get *lots* of `Error in reading or end of file.` like here:
```
  Perf controller (CGROUP_PERF) [Y/n/?] y
  Support for eBPF programs attached to cgroups (CGROUP_BPF) [N/y/?] (NEW)
Error in reading or end of file.

Error in reading or end of file.

Error in reading or end of file.
```
This is quite distracting at best, but is this an error?

More importantly, the generated `.config` differs from what we had:
```
diff -u master-kernel_config-4.11.x pr-kernel_config-4.11.x | diffstat
 pr-kernel_config-4.11.x |   46 ++++++++++++++++++----------------------------
 1 file changed, 18 insertions(+), 28 deletions(-)
```
I've attached both files but some of the differences are:
- recently added TMP support is not yet in the new config files
- `CONFIG_XEN_BACKEND` is now set and that causes some additional options (not set), like `CONFIG_XEN_BLKDEV_BACKEND` to be added to the config file.
- Some additional devices are set, e.g., `CONFIG_SERIAL_8250_MID`, `CONFIG_R8169`, 
- it also enables `CONFIG_TREE_RCU_TRACE`, disables `CONFIG_CPU_IDLE_GOV_LADDER`, enables `CONFIG_DMA_SHARED_BUFFER`, enables `CONFIG_SYNC_FILE`, enables `CONFIG_RCU_TRACE`, enables `CONFIG_UBSAN_SANITIZE_ALL` and `CONFIG_UBSAN_NULL`

[pr-kernel_config-4.11.x.txt](https://github.com/linuxkit/linuxkit/files/1165417/pr-kernel_config-4.11.x.txt)
[master-kernel_config-4.11.x.txt](https://github.com/linuxkit/linuxkit/files/1165416/master-kernel_config-4.11.x.txt)

tych0:
Hi Rolf,

[sorry for the delay, just got back from PTO]

On Fri, Jul 21, 2017 at 11:59:19AM +0000, Rolf Neugebauer wrote:
> doing a `make build_4.11.x` i get:
> ```
> #
> # configuration written to .config
> #
> scripts/kconfig/conf  --oldconfig Kconfig
> .config:3789:warning: override: DEFAULT_SECURITY_DAC changes choice state
> *
> * Restart config...
> ```
> Is that warning benign?

Definitely not, good catch. I've fixed it.

> I also get *lots* of `Error in reading or end of file.` like here:
> ```
>   Perf controller (CGROUP_PERF) [Y/n/?] y
>   Support for eBPF programs attached to cgroups (CGROUP_BPF) [N/y/?] (NEW)
> Error in reading or end of file.
> 
> Error in reading or end of file.
> 
> Error in reading or end of file.
> ```
> This is quite distracting at best, but is this an error?

This isn't an error as near as I can tell, just kconfig doing its
thing. The output matches what I expect, anyway. We could pipe it to
/dev/null, but then we'd miss the stuff you noted above, so I've just
left it for now.

> More importantly, the generated `.config` differs from what we had:
> ```
> diff -u master-kernel_config-4.11.x pr-kernel_config-4.11.x | diffstat
>  pr-kernel_config-4.11.x |   46 ++++++++++++++++++----------------------------
>  1 file changed, 18 insertions(+), 28 deletions(-)
> ```
> I've attached both files but some of the differences are:
> - recently added TMP support is not yet in the new config files
> - `CONFIG_XEN_BACKEND` is now set and that causes some additional options (not set), like `CONFIG_XEN_BLKDEV_BACKEND` to be added to the config file.
> - Some additional devices are set, e.g., `CONFIG_SERIAL_8250_MID`, `CONFIG_R8169`, 
> - it also enables `CONFIG_TREE_RCU_TRACE`, disables `CONFIG_CPU_IDLE_GOV_LADDER`, enables `CONFIG_DMA_SHARED_BUFFER`, enables `CONFIG_SYNC_FILE`, enables `CONFIG_RCU_TRACE`, enables `CONFIG_UBSAN_SANITIZE_ALL` and `CONFIG_UBSAN_NULL`
> 
> [pr-kernel_config-4.11.x.txt](https://github.com/linuxkit/linuxkit/files/1165417/pr-kernel_config-4.11.x.txt)
> [master-kernel_config-4.11.x.txt](https://github.com/linuxkit/linuxkit/files/1165416/master-kernel_config-4.11.x.txt)

Yes, thanks for these. I'll add the new stuff and figure out what the
deal is/was with the other differences.

rn:
There is a somewhat more general question as to why we are not re-using the debian/ubuntu or android tooling for this? @ijc seems to think [here](https://github.com/linuxkit/linuxkit/issues/635#issuecomment-256362305) that the debian tooling might work

tych0:
As near as I can tell android doesn't have any actual tools for merging things, just different config files (.base, and .recommended, etc.), and then stores the options alphabetically in these files, which is what we're doing here.

I suppose we could re-use parts of the debian stuff from here: https://anonscm.debian.org/cgit/kernel/linux.git/tree/debian/lib/python/debian_linux for makeconfig.sh, but a lot of it we don't need, and the script is only 75 lines so it didn't seem necessary.

The problems you've hit on above are mostly to do with the splitting up process by arch/version, which I think would have been painful no matter what tooling we're using to merge it back together :)

rn:
from the quick look at the debian tooling it looked quite appealing when trying to add new options. for starters it is organised by directory/Kconfig files and from what ian said, you can just throw stuff at it and it tidies it up etc. It also seemed a bit more flexible wrt to different flavours which we may want to support soon (I'm thinking of a special LCOW kernel for example). The debian tooling did not seem to have support for different kernel versions so I'm not sure how easy one can make that fit.

As said, I have never used it, just done a bit of reading around

tych0:
On Mon, Jul 31, 2017 at 1:10 PM, Rolf Neugebauer <notifications@github.com>
wrote:

> from the quick look at the debian tooling it looked quite appealing when
> trying to add new options. for starters it is organised by
> directory/Kconfig files and from what ian said, you can just throw stuff at
> it and it tidies it up etc. It also seemed a bit more flexible wrt to
> different flavours which we may want to support soon (I'm thinking of a
> special LCOW kernel for example).
>

Can you elaborate on this? It seems like people are interested in something
other than the current .{arch}.{version} structure, but I've never really
understood what. Per-Kconfig-file separation would be easy enough to do, if
that's what is desirable.

> The debian tooling did not seem to have support for different kernel
> versions so I'm not sure how easy one can make that fit.
>
Shouldn't be too hard; I think we'd just make our own driver for the python
classes in that directory. They're small, so it's not much code saved, but
it's something.

PR #2256, opened 2017-07-20
metadata: avoid printing large chunks of configuration json to the console

djs55:
**- What I did**

I prevented large chunks of configuration .json (some of which could be secret) from being logged to the console.

**- How I did it**

I suppressed printing the .json value when there is no `perm` or `content` attribute i.e. when the node is a directory.

**- How to verify it**

Using configuration .json like:
```json
{
  "etc": {
    "ssl": {
      "certs": {
        "ca-certificates.crt": {
          "perm": "0644",
          "content": "large amount of certificate text"
        }
      }
    }
  }
}
```
Without this patch, I see on the console:
```
2017/07/20 10:03:04 CDROM: Probe succeeded
2017/07/20 10:03:04 No permission provided ssl:map[certs:map[ca-certificates.crt:map[perm:0644 content:large amount of certificate text]]]
 - 000-metadata
```
With this patch, I see on the console:
```
2017/07/20 09:54:18 CDROM: Probe succeeded
2017/07/20 09:54:18 No permission provided ssl
 - 000-metadata
```

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -10,7 +10,7 @@ init:
 onboot:
   # support metadata for optional config in /var/config
   - name: metadata
-    image: linuxkit/metadata:f122f1b4e873f1d08cd67bd9105385fd923af0cb
+    image: linuxkit/metadata:428093dd1c4178e8ba1952af44b46c0fd16f8e79
   - name: sysctl
     image: linuxkit/sysctl:d1a43c7c91e92374766f962dc8534cf9508756b0
   - name: sysfs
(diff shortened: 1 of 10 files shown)


rn:
looks good to me. Could you add a commit to update the hashs for all uses of the metadata package? The hash will be:
`linuxkit/metadata:428093dd1c4178e8ba1952af44b46c0fd16f8e79` and I've signed/pushed it already

djs55:
I'm glad I discovered that `update-component-sha.sh` script!

ijc:
In general I find metadata to be a bit too chatty on the console (e.g. with the non-json mode it says something) and think it really ought to log to `/var/log/blah` or (`/var/config/blah`) instead of stdio unless it is run in verbose mode.

But that is another story and this change itself LGTM.

PR #2258, opened 2017-07-20
Containerd v1.0.0 alpha1

ijc:
**- What I did**

Primary goal was to update to containerd alpha1, since this involves touching `tools/alpine` I also made some other updates there.

**- How I did it**

* Updated `CONTAINERD_COMMIT` in Alpine base. This is  still tracking https://github.com/containerd/containerd/pull/1141 which has been rebased.
* Updated `service` cli tool for changes to containerd API.
* Updated `pkg/getty`, `pkg/sshd` and a test case to handle merge of `dist` tool into `ctr`.
* Updated Moby commit to latest version compatible with new containerd
* Updated the yml files
* Switched alpine base to use `docker build --iidfile` which enables some simplifications and improvements
* Added xfsprogs-extra to alpine base on request from @dave-tucker 

**- How to verify it**

Run something (e,g. toplevel `linuxkit.yml`).

Updated to containerd v1.0.0.-alpha1

[](https://en.wikipedia.org/wiki/In_Their_Darkened_Shrines)


tools/alpine/Makefile
```diff
@@ -9,25 +9,25 @@ default: push
 show-tag:
 	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions
 
-hash: Dockerfile Makefile packages
+iid: Dockerfile Makefile packages
```
justincormack: I think you should still delete the `iid` file after `tag` and `push` as we always need to rebuild in case the outside world has changed, it doesnt look like it will now.

justincormack: Maybe do the `iid` change in a different PR, as want the containerd stuff merged ASAP as will cause runc version sync issues

ijc: I'd expect to have to run `make clean` if I wanted that.

ijc: You mean rebase this without the first commit? Can do.

(sorry for the ordering, GH dynamic update seems a bit flakey for me at the moment so I only saw your first comment at first)

justincormack: Yeah, its totally unrelated to the rest, best to remove

ijc:
Dropped the iidfile changes.

ijc:
All green.

PR #2259, opened 2017-07-20
Add apk to login (sshd & getty) containers

### LINKED ISSUE #2206: Add apk to getty and sshd containers

ijc:
**- What I did**

Made it possible to run `apk` commands (such as `apk add`) in the `getty` and `sshd` containers

Fixes #2206.

**- How I did it**

Added `apk-tools` to both images, used 00a2f2ac8caca2b10f0b61bfe0116892c5f2687d to provide an appropriate repositories config and stopped nuking the apk state.

I also dropped the bind mount of `/var` from both packages. Quoting the relevant bit of commit message:
>    Doing this requires that we not share `/var` with the login containers since we
>    want the apk database therein. Previously it was thought that the containers
>    might need some parts of `/var` for `ctr` to work (e.g. `/var/lib/containerd`)
>    but this is not the case now (if it ever was) based on my testing.

Although not strictly necessary, bumped the version of alpine base used by both images.

**- How to verify it**

Try `apk -U add strace` from a getty or sshd container.

Login containers (sshd and getty) now have the `apk` installed enabling ad-hoc package installation.

[](https://en.wikipedia.org/wiki/Times_of_Grace_(album))

### DIFF
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -1,15 +1,17 @@
-FROM linuxkit/alpine:9bcf61f605ef0ce36cc94d59b8eac307862de6e1 AS mirror
+FROM linuxkit/alpine:a39a433162a873519910a07beeb3e8db22529956 AS mirror
 
 RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out \
     alpine-baselayout \
+    apk-tools \
     busybox \
     ca-certificates \
     musl \
     tini \
     util-linux \
     && true
-RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN mv /out/etc/apk/repositories.upstream /out/etc/apk/repositories
+
 #
 # We require a version of `setsid(1)` which supports the `-w`
 # option, which is not available in all implementations (e.g. the
@@ -29,4 +31,4 @@ COPY --from=mirror /out/ /
 COPY usr/ /usr/
 COPY etc/ /etc/
 CMD ["/usr/bin/rungetty.sh"]
-LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/tmp:/tmp", "/etc:/hostroot/etc", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/var:/var","/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
+LABEL org.mobyproject.config='{"pid": "host", "net":"host", "binds": ["/run:/run", "/tmp:/tmp", "/etc:/hostroot/etc", "/usr/bin/ctr:/usr/bin/ctr", "/usr/bin/runc:/usr/bin/runc", "/containers:/containers","/dev:/dev","/sys:/sys"], "capabilities": ["all"]}'
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -53,7 +53,7 @@ services:
     image: linuxkit/acpid:1966310cb75e28ffc668863a6577ee991327f918
   # Enable getty for easier debugging
   - name: getty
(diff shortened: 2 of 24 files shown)


pkg/getty/Dockerfile
```diff
@@ -1,15 +1,17 @@
...
     alpine-baselayout \
+    apk-tools \
     busybox \
     ca-certificates \
     musl \
     tini \
     util-linux \
     && true
-RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN mv /out/etc/apk/repositories.upstream /out/etc/apk/repositories
```
deitch: This is what points them at the official apk repos? So `apk add ....` would go over the net, as opposed to the (massive) cache we use for building `pkg` images, right? Makes sense.

deitch:
LGTM.

pkg/getty/Dockerfile
ijc: Correct. The cache is not copied into the output image and full network repo (even larger than what we cache for our build) is what we want anyway.

justincormack:
indeed, eg just now while preparing for a demo...

PR #2260, opened 2017-07-20
alpine: Use `docker build --iidfile`

ijc:
This was added in 17.06 and allows us to avoid using `$(BASE):build` which is
not safe against parallel builds etc.

Having done this restructure the build to not always delete the built container
and to separate out the `hash` and `version` file rules so that they can be
included in both the `tag` and `push` targets.

This carries a change from https://github.com/linuxkit/linuxkit/pull/2258 which was dropped from there due to being off-topic.

**- What I did**

See above.

**- How I did it**

Used `docker build --no-cache --iidfile iid .` to build an `iid` file (which will contain `sha256:[0-9a-f]{64}`) and refactored the `hash` and `versions` generation into dedicated targets depending on that. Updated everything using `$(IMAGE):build` to use `$(shell cat iid)` instead. Updated the dependencies of the existing `tag` and `push` targets. It is no longer necessary to `docker rmi` anything.

**- How to verify it**

Check that `make -C tools/alpine tag` does something sensible.

Internal detail not really worth mentioning

[](https://en.wikipedia.org/wiki/Given_to_the_Rising)

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -9,7 +9,7 @@ COPY Dockerfile /Dockerfile
 COPY packages* /tmp/
 
 # mirror packages
-RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \ 
+RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \
    mkdir -p /mirror/$(uname -m) && \
    apk fetch --recursive -o /mirror/$(uname -m) $(apk info; cat /tmp/packages)
 
@@ -30,7 +30,7 @@ RUN cp /mirror/$(uname -m)/APKINDEX.unsigned.tar.gz /mirror/$(uname -m)/APKINDEX
 RUN abuild-sign /mirror/$(uname -m)/APKINDEX.tar.gz
 
 # fetch OVMF for qemu EFI boot (this is not added as a package)
-RUN mkdir -p /usr/share/ovmf && \ 
+RUN mkdir -p /usr/share/ovmf && \
     if [ $(uname -m) = x86_64 ]; then \
        apk add -X http://dl-cdn.alpinelinux.org/alpine/edge/community ovmf; \
     fi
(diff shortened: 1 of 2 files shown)


ijc:
Running `git grep docker.*rmi` suggests there are plenty of other places under `tools` which could benefit from this (or switching to `pkg/pacakge.mk`, in preference where possible). Possibly `test/pkg/kernel-config/Makefile` too. I think I should look at tackling those here too.

I also see a bunch of stuff under projects which I do not propose to touch.

justincormack:
As discussed on previous issue, I think its totally bizarre to persist a random number on a file in the file system, the `iid` should be deleted after use.

ijc:
I've just spotted that the actual issue causing it to rebuild every time is that `docker build --no-cache --iidfile iid .` includes `--no-cache`, preserving the `iid` was really just working around that. If caching were allowed to do its thing (by default, an override would make sense) then there would be no need to preserve `iid`.

justincormack:
The assumption is though that you only rebuild this when you actually want to, because you know/want to see if upstream has changed.

ijc:
I don't think that assumption holds when doing iterative development (especially on bits later in the `Dockerfile` since the first bits are necessarily the slowest) and in particular having had to rebase various changes to `tools/alpine` over the last few days I'm certainly not rebuilding in the expectation of an upstream change the majority of the time.

justincormack:
Well I don't mind refactoring so there are some more convenient makefile targets for some use cases, but both (a) persisting random numbers on the filesystem and pretending that is meaningful and (b) adding caching by default to something that is designed as our single point of not using caching are both wrong, so not LGTM at all in this form.

ijc:
I added an `rm -f iid`. I'm going to wait for #2244 before I bother rebasing though.

ijc:
And as if by magic it was merged ;-)  Rebasing coming.

ijc:
Rebased. Bonus patch to drop some trailing whitespace from the docker file.

I'm undecided whether `hash` and `iid` should have `.$(ARCH)` suffixes or not.

Also, could `PKG_DEPS` be just inlined into the one usage as `packages packages.$(ARCH)`?

justincormack:
we might build this using qemu emulation for convenience, but I guess you would do a make clean so having suffixes on hash and iid seems unnecessary I think

PR #2261, opened 2017-07-20
containerd: add `exec` subcommand to `service`

ijc:
**- What I did**

This replaces `ctr exec --tty --exec-id mumble <service>`.

Several helpers cribbed from ctr source.

**- How I did it**

Mostly stole code from `ctr`.

**- How to verify it**

Right now:
```
(ns: getty) linuxkit-525400123456:~# nsenter -m/proc/1/ns/mnt service exec --tty nginx ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 nginx: master process nginx -g daemon off;
    5 nginx      0:00 nginx: worker process
   15 root       0:00 ps aux
```

If there is support for this idea I think it makes sense to bind `service` into the `getty` and `sshd` containers.

`service exec` is a convenient way to enter a container namespace.

[](https://en.wikipedia.org/wiki/Scream,_Dracula,_Scream!)

### DIFF
--- a/pkg/containerd/cmd/service/exec.go
+++ b/pkg/containerd/cmd/service/exec.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/containerd/console"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	"golang.org/x/sys/unix"
+)
+
+type resizer interface {
+	Resize(ctx context.Context, w, h uint32) error
+}
+
+type killer interface {
+	Kill(context.Context, syscall.Signal) error
+}
+
+func exec(args []string) {
+	invoked := filepath.Base(os.Args[0])
+	flags := flag.NewFlagSet("exec", flag.ExitOnError)
+	flags.Usage = func() {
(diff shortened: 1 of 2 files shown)


justincormack:
It is a bit weird replacing/enhancing their UI with ours. I would kind of rather `ctr exec` could default to making up an id if not specified, but this also seems fairly harmless

ijc:
Yeah. Feedback from `ctr` folks was that they didn't want to get into the business of making up ids, plus not a stable interfaces etc.

It occurs to me now though that I could just make this a wrapper, i.e. could spawn `ctr` with some appropriate options, would be less faffing with signals etc on our end.

justincormack:
It still feels weird to provide another command. Maybe `getty` and `sshd` should just come with some shell aliases for `ctrexec` instead?

ijc:
My mental model was that `service` would be(come) the primary point of entry for interacting with LinuxKit services for most things including starting, stopping (not implemented), getting logs (not implemented), executing processes within them (this PR), inspecting their config (not implemented) etc.

justincormack:
Hmm, my mental model was that this was an internal command only, with no real user facing parts.

ijc:
That certainly explains the disconnect!

justincormack:
I don't think we should make this user facing at the moment, especially as just running it with no arguments currently runs the whole system init (that could be changed). Also I think it is good for people to be exposed to the underlying commands, and the system is really designed for automation not an interactive experience, and making it usable interactively would require some redesign. SO closing for now.

PR #2262, opened 2017-07-20
tools/alpine: Update alpine base to include wireguard tools

rn:

elephants

### DIFF
--- a/tools/alpine/versions
+++ b/tools/alpine/versions
@@ -1,4 +1,4 @@
-# linuxkit/alpine:a39a433162a873519910a07beeb3e8db22529956
+# linuxkit/alpine:5f82d95d42caa9858117e09d9b77bdf1b0b9c9c9
 # automatically generated list of installed packages
 abuild-3.0.0_rc2-r8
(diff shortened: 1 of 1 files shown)


justincormack:
The `versions` file doesnt have `wireguard-tools` in...

rn:
ah, yes. I think the add is in the wrong place. closing this until fixed

ijc:
The reordering of `versions` here is very odd (and no doubt obscured the fact that wireguard wasn't really added). Perhaps an `LC_COLLATE` issue?

ijc:
Perhaps `s/sort/LANG=C sort/` in the appropriate place in `tools/alpine/Makefile`?

ijc:
BTW It's not that the add is in the wrong place per se, it is that the add should instead be a new line in `versions` so it gets added to the mirror not the image filesystem.

BTW2, it seems likely the ovmf addition is in the wrong place too...

ijc:
I see. Might be better to rework that while fixing the other stuff and have the qemu image include ovmf in its apk add.

PR #2263, opened 2017-07-20
Wireguard tools as package

zx2c4:
As discussed with @rn, wireguard-tools should be a *package available from* tools/alpine. Then, it can be included in sshd and getty, since it's on the same level as iproute2, and wireguard is going to be an essential feature of linuxkit, so it's important for folks to get acquainted with it.

The hashes in getty and sshd need to be updated, so after you've built them and have a tag, please mention it here so I can update the PR.

### DIFF
--- a/docs/wireguard.md
+++ b/docs/wireguard.md
@@ -18,7 +18,7 @@ A full technical paper from NDSS 2017 is available [here](https://www.wireguard.
 The default kernels build WireGuard in as a module.
 
 ### Userspace Tools
-The userspace tools are part of `tools/alpine`.
+The userspace tools are now a package part of `tools/alpine`, which can be installed via `apk add wireguard-tools`.
 
 ## Quickstart
 To give WireGuard a spin, the [official quick start](https://www.wireguard.com/quickstart/) is a good way to get going.  For containers,
--- a/pkg/getty/Dockerfile
+++ b/pkg/getty/Dockerfile
@@ -9,6 +9,7 @@ RUN apk add --no-cache --initdb -p /out \
     musl \
     tini \
     util-linux \
+    wireguard-tools \
     && true
 RUN mv /out/etc/apk/repositories.upstream /out/etc/apk/repositories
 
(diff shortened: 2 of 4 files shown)


deitch:
Do these belong in `getty` and `ssh`? Now that we have full `apk` access, shouldn't someone do `apk add ...` if they need them?

zx2c4:
No. Just as `iproute2` is part of the standard, so is `wg`. At some point they'll be part of the same tool anyway.

deitch:
> No. Just as iproute2 is part of the standard, so is wg. At some point they'll be part of the same tool anyway.

Yeah, but it isn't yet. _Every_ container needs `iproute2` but not every one needs `wg`. I am all for it being so, but until it is, I am not sure it should be installed by default. But open to other input. @ijc you had some comments.

zx2c4:
Did you ignore the other big point? Reproducibility and networklessness.

I'll add another: WireGuard is going to be a big and integral part of LinuxKit. It's important that it's available for people to experiment with right away.

In any case, this was the dual suggestion of @rn and @justincormack, so I'll let them decide.

deitch:
> this was the dual suggestion of @rn and @justincormack, so I'll let them decide

OK. I usually defer to them. :-)

rn:
@deitch `iproute2` is *not* part of every container. The containers which use busybox as the base (which are some but not all) include the busybox version of `ip` which is nowhere near as complete as `iproute2`

deitch:
> iproute2 is not part of every container

Very good point. It is part of `getty` and `sshd` (as `busybox`) because it makes sense there, but, yeah, it is not.

rn:
@zx2c4 thanks for adjusting the alpine base. I've check with:
```
make ORG=rneugeba  tag
docker run --rm -ti rneugeba/alpine:5ca5d83ef72f74bdc2a7b96bc17d58031a42306d /bin/sh
/ # apk search wireguard
wireguard-tools-0.0.20170629-r0
```
that the tools are now part of the base image and show up in `versions`. So that part of the PR is great.

@ijc raised the concern elsewhere if this should be part of `getty`/`sshd` considering that they now can easily add the utilities with `apk add`. I half agree with him. @deitch @justincormack @riyazdf what's your view? Maybe it's best to split the PR and initially just add it to alpine base?

zx2c4:
I split this into two.

PR #2264, opened 2017-07-20
support tap devices in qemu backend

tych0:
The motivation for this is networking out (in particular, testing NFS
support) from the VM.

We could be a lot more user friendly (a la libvirt) by creating the tap
device for users and allowing them to specify a bridge instead, but then
we'd need root to create this tap device. For now, let's make people do
their own tap devices, and just use them. A tap device can be created for a
bridge as follows:

    # ip tuntap add linuxkit0 mode tap user `whoami`
    # ip link set linuxkit0 up
    # ip link set linuxkit0 master $bridge_name

and then used by:

    $ ./bin/linuxkit run qemu -tap-device linuxkit0 linuxkit

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -35,6 +35,7 @@ type QemuConfig struct {
 	QemuBinPath    string
 	QemuImgPath    string
 	PublishedPorts []string
+	TapDevice      string
 }
 
 func haveKVM() bool {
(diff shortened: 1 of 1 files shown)


ijc:
Nice! FWIW I've also been considering supporting system which have the [`qemu-bridge-helper`](http://wiki.qemu.org/Features/HelperNetworking) installed. TL;DR you can make `/usr/lib/qemu/qemu-bridge-helper` suid root and whitelist bridges in `/etc/qemu/bridge.conf`such that users can attach without being root.

Of course involving a suid helper this is not for everyone, and also the whitelisting is system wide and not per user, but in general it is better than running the whole of qemu as root!

I think that can nicely co-exist with this stuff thought.

PR #2265, opened 2017-07-20
wireguard: add tools to alpine as package not into filesystem

zx2c4:
As discussed with @rn, wireguard-tools should be a package available from tools/alpine. Then, it can be included in sshd and getty, since it's on the same level as iproute2, and wireguard is going to be an essential feature of linuxkit, so it's important for folks to get acquainted with it.

This PR takes the first step of making it into a package.

From @rn, it works:

```
make ORG=rneugeba  tag
docker run --rm -ti rneugeba/alpine:5ca5d83ef72f74bdc2a7b96bc17d58031a42306d /bin/sh
/ # apk search wireguard
wireguard-tools-0.0.20170629-r0
```

### DIFF
--- a/docs/wireguard.md
+++ b/docs/wireguard.md
@@ -18,7 +18,7 @@ A full technical paper from NDSS 2017 is available [here](https://www.wireguard.
 The default kernels build WireGuard in as a module.
 
 ### Userspace Tools
-The userspace tools are part of `tools/alpine`.
+The userspace tools are now a package available in `tools/alpine`, which can be installed via `apk add wireguard-tools`.
 
 ## Quickstart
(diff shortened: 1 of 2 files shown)


rn:
thanks for splitting

riyazdf:
just pushed and signed `linuxkit/alpine:4248059c38452217ff63853869df36034a890401`

rn:
@riyazdf thanks, but you also need to to a PR for the updated `versions` file. The alpine base image is tricky in this respect and maybe best just updated by maintainer...

riyazdf:
@rn: ah right.  I'll merge this as-is and open a new PR for `versions` :+1:

PR #2266, opened 2017-07-20
Add wireguard to sshd and getty

zx2c4:
WireGuard is on the same level as iproute2, which is also in sshd and getty. If WireGuard is going to become an important component of LinuxKit, it should be available for people to play with in these containers. Adding the tools adds hardly any increase and size, and will make developing LinuxKit+WireGuard considerably smoother. This is also needed to use these in reproducable/networkless configurations, since /mirrors is not available after boot.

Suggested by @rn and @justincormack.

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -18,7 +18,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:1516d5d70683a5d925fe475eb1b6164a2f67ac3b
   - name: sshd
-    image: linuxkit/sshd:dc98a72c1d1285c30f2db176252f3ce2bf645d5b
+    image: linuxkit/sshd:9e9186dd5989ae9c604eba77e306b0e67500b55c
     binds:
      - /var/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -15,7 +15,7 @@ services:
   - name: dhcpcd
     image: linuxkit/dhcpcd:4b7b8bb024cebb1bbb9c8026d44d7cbc8e202c41
   - name: sshd
-    image: linuxkit/sshd:dc98a72c1d1285c30f2db176252f3ce2bf645d5b
+    image: linuxkit/sshd:9e9186dd5989ae9c604eba77e306b0e67500b55c
 files:
   - path: root/.ssh/authorized_keys
     source: ~/.ssh/id_rsa.pub
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -20,7 +20,7 @@ onboot:
     command: ["/mount.sh", "/var/lib/docker"]
 services:
   - name: getty
-    image: linuxkit/getty:08b704915af0ce90f8f40df5d41d4c1aa14ef83a
+    image: linuxkit/getty:1b651a91f1c17f50357be9873b580ccf81668630
     env:
      - INSECURE=true
   - name: rngd
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -16,13 +16,13 @@ onboot:
     image: linuxkit/metadata:428093dd1c4178e8ba1952af44b46c0fd16f8e79
 services:
   - name: getty
-    image: linuxkit/getty:08b704915af0ce90f8f40df5d41d4c1aa14ef83a
+    image: linuxkit/getty:1b651a91f1c17f50357be9873b580ccf81668630
     env:
      - INSECURE=true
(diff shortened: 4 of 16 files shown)


rn:
this will fail CI and needs an update the `FROM` line to an `linuxkit/alpine` hash which includes the tools, ie #2265

riyazdf:
@rn: I am building/pushing/signing a new alpine base from that PR, will be: `linuxkit/alpine:4248059c38452217ff63853869df36034a890401`

zx2c4:
Just pushed a new commit that has the updated base for this.

riyazdf:
pushed and signed:
- `linuxkit/getty:1b651a91f1c17f50357be9873b580ccf81668630`
- `linuxkit/sshd:9e9186dd5989ae9c604eba77e306b0e67500b55c`

zx2c4:
Alright the new one should be good to go. Example hashes updated. I've tested it works.

First ever WireGuard instance in LinuxKit:

image

ijc:
> WireGuard is on the same level as iproute2, which is also in sshd and getty.

This is not true (also pointed out by @rn [here](https://github.com/linuxkit/linuxkit/pull/2263#issuecomment-316799551)). The `sshd` and `getty` containers have `ip` and `ifconfig` etc as binaries because they are provided by `busybox` which is needed for other reasons, there is no full featured `iproute2` package installed in those environments.

In LinuxKit the network configuration is containerised and occurs in the `dhcpcd` container. I would expect that if I needed to do any non-trivial network configuration I would have to either `ctr exec` my way into the dhcp container (which should have the full feature tools installed, I see that right now it does not, this should be addressed, or maybe we want a "network admin" container) or that I would need to  add the tools to the login package either on the fly (`apk add`). Or if I were e.g. building a dedicated network appliance image want to create a custom/derived login environment (i.e. custom `getty`/`sshd` image(s)) with a selection of tools relevant to that appliance installed.

It is not clear to me why wireguard should be different to that model, i.e. why we shouldn't produce a `pkg/wireguard` which can be included in images which want to use wireguard (at a minimum something under `examples`, but arguments for including in the top-level `linuxkit.yml` too would be of interest). `pkg/wireguard` can both perform necessary start of day configuration and provide a suitable shell environment (perhaps like the `dhcpd` package it would want to be suitable for both `service` and `onboot` use).

I expressed my views on adding tools to the `sshd` and `getty` containers on both https://github.com/linuxkit/linuxkit/pull/2249 (where the last comment on the previous PR seemed to agree https://github.com/linuxkit/linuxkit/pull/2249#issuecomment-316704398) and on the slack channel. See also e.g. https://github.com/linuxkit/linuxkit/pull/2263#issuecomment-316810820.

So, I'm sorry but my feeling is that this PR was merged prematurely and should be reverted while we discuss how this stuff should be properly integrated into LinuxKit.

To be clear, I am not arguing against having wireguard as a core part of LinuxKit, but we should properly discuss how to achieve that within LinuxKit's highly containerised model

zx2c4:
Firstly, I split this PR into two different PRs so that this could be discussed, not for any premature conclusions. Sorry you weren't around yesterday at that time.

I'm actually working on a `pkg/wireguard`. Hopefully I'll have a PR soon for that, though it might be a few weeks until I have that sort of availability to polish and push it.

avsm:
@ijc 
> I would expect that if I needed to do any non-trivial network configuration I would have to either ctr exec my way into the dhcp container (which should have the full feature tools installed, I see that right now it does not, this should be addressed, or maybe we want a "network admin" container) 

I think it's better to have a separate network admin container rather than overload dhcpcd for this purpose -- we would like to swap out the dhcpcd for a very minimal one from the projects/miragesdk work at some point soon, and so depending on it having the full toolchain in a separate container would be helpful to that work.

riyazdf:
hey folks - looking back I think we could use more discussion to determine the best path forward here as @ijc and @avsm raise some good points about how to best make this available.

I've opened a PR to revert this change: https://github.com/linuxkit/linuxkit/pull/2270 (as well as updated docs: https://github.com/linuxkit/linuxkit/pull/2271)

Let's continue the discussion here.

riyazdf:
edit: actually let's use https://github.com/linuxkit/linuxkit/issues/2273 to discuss for better discoverability

deitch:
> I think it's better to have a separate network admin container rather than overload dhcpcd for this purpose

And I think what "network" is and looks like is still in flux. Right now we have just the host (wireguard rapidly becoming an option). There has been talk of more flexible networking options, and per-container network options still need to be worked out. Much to do...

PR #2270, opened 2017-07-21
Revert "wireguard: add to getty and sshd"

riyazdf:
Some discussion on https://github.com/linuxkit/linuxkit/pull/2266 was fragmented across other mediums and it appears that there was not agreement.

Revert in the meantime while we discuss and determine the best path forward.

### DIFF
--- a/docs/wireguard.md
+++ b/docs/wireguard.md
@@ -18,7 +18,9 @@ A full technical paper from NDSS 2017 is available [here](https://www.wireguard.
 The default kernels build WireGuard in as a module.
 
 ### Userspace Tools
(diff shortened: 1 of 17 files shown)


justincormack:
I am ok with leaving this in for now, and removing when there is a better
example of a real setup (thinking about how to do that sensibly, it needs
some changes)

zx2c4:
That makes sense to me. We'd keep things like this now, so people can easily play, and then once there's a `pkg/wireguard` and other things, we remove it.

rn:
I'm also ok with leaving this in for now as it is quite nice to play with it. But we should remove it at some point because it doesn't belong here and network config should be done differently anyway

justincormack:
Closing for now as no appetite for revert and it would need updating should we decide to.

PR #2275, opened 2017-07-23
Add script to update tag for all image matches

deitch:

**- What I did**
Added a script that updates all instances of an image with a given tag. The existing script replaces tag `<OLD>` with `<NEW>`. This changes the hash for any occurrence of `<IMAGE>` to `<NEW>`, no matter what the old one was.

Existing is fine, but often I find the need to just "change them all."

**- How I did it**
Typing.

**- How to verify it**
Run it, see the changes.

Script to support changing tag for all occurrences of an image.

monster

### DIFF
--- a/scripts/update-component-sha.sh
+++ b/scripts/update-component-sha.sh
@@ -1,10 +1,67 @@
 #!/bin/sh
 set -e
-if [ $# -ne 2 ] ; then
-    echo "Need <OLD> and <NEW> as arguments" >&2
+
+##
+#
+# script to replace hashes in config files
+# see usage() for usage and functionality
+#
+
+function usage() {
+    cat >&2 <<EOF
+$0 --<mode> <how-to-find> <new-hash>
+
+Available modes: --hash and --image
+
+Replace by hash:
+	$0 --hash <OLD> <NEW>
+	Example: $0 8675309abcdefg abcdef567899
+    	   Will replace all instances of 8675309abcdefg with abcdef567899
+
+Replace by image: $0 --image <IMAGE> <NEW>
+	$0 --image <IMAGE> <NEW>
+	Example: $0 linuxkit/foo abcdef567899
+	   Will tag all instances of linuxkit/foo with abcdef567899
+
+By default, for convenience, if no mode is given (--image or --hash), the first method (--hash) is assumed. 
+Thus the following are equivalent:
+	$0 <OLD> <NEW>
+	$0 --hash <OLD> <NEW>
+
+EOF
+}
+
+
+# backwards compatibility
+if [ $# -eq 2 ]; then
+  set -- "--hash" "$1" "$2"
+fi
+
+# sufficient arguments
+if [ $# -ne 3 ] ; then
+    usage
     exit 1
 fi
-old=$1
-new=$2
 
-git grep -l "$old" | xargs sed -i -e "s,$old,$new,g"
+
+# which mode?
+case "$1" in
+    --hash)
+        old=$2
+        new=$3
+
+        git grep -w -l "\b$old\b" | xargs sed -i.bak -e "s,$old,$new,g"
+        ;;
+    --image)
+        image=$2
+        hash=$3
+        git grep -E -l "\b$image:" | xargs sed -i.bak -e "s,$image:[[:xdigit:]]"'\{40\}'",$image:$hash,g"
+        ;;
+    *)
+        echo "Unknown mode $1"
+        usage
+        exit 1
+        ;;
+esac
+
+find . -name '*.bak' | xargs rm


justincormack:
Can we just have a single script that takes an `--all` argument?

Also, while you are modifying this, I would prefer if the script removed the leftover original files, as you are using git anyway so you can always undo. Every time I run this script I immediately do `find . -name '*-e' | xargs rm`...

deitch:
Ha! It bothered me too, but I figured it was left for a reason. 

Sure I'll merge them and add the rm.

deitch:
This has been updated as follows:

1. Single script, as requested by @justincormack 
2. Script now has options for `--hash <old> <new>` to replace by hash or `--image <image> <hash>` to replace by image
3. When run with insufficient arguments, it spits out usage of both formats to stderr
4. If no type is given, just `<old> <new>`, it falls back to functioning as if it were `--hash` for backwards-compatibility.

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
-    echo "Need <OLD> and <NEW> as arguments" >&2
+
+##
+#
+# script to replace hashes in config files
+# can be run in 2 modes:
+# 1- $0 --hash $oldhash $newhash              - replace all instances of $oldhash with $newhash
+# 2- $0 --image $image $newhash        - replace all hashes for $image with $newhash
+#
+# for backwards compatibility if no argtype is given, the first method is assumed. Thus the following are equivalent:
```
rn: It doesn't look like this is what the code does. Frankly for a tool this new there is no need to maintain backwards compatibility, so just remove this comment, but also see below

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
 #!/bin/sh
 set -e
-if [ $# -ne 2 ] ; then
-    echo "Need <OLD> and <NEW> as arguments" >&2
+
+##
+#
+# script to replace hashes in config files
```
rn: This is already in the usage, so why repeat it here?

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
+#   $0 $oldhash $newhash 
+#   $0 --hash $oldhash $newhash 
+#
+
+function usage() {
+    echo "Modes of operation:" >&2
+    echo >&2
+    echo "Replace by hash:" >&2
+    echo "	$0 --hash <OLD> <NEW>" >&2
+    echo "	Example: $0 8675309abcdefg abcdef567899" >&2
```
rn: I would prefer if the usage would give a better example, like `$0 linuxkit/foo:8675309abcdefg linuxkit/foo:abcdef567899` and then also reflect that in the next line.

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
+        hash=$3
+        git grep -l "$image:" | xargs sed -i -e "s,$image:[[:xdigit:]][[:xdigit:]]*,$image:$hash,g"
+        ;;
+    *)
+        echo "Unknown mode $1" >&2
+        usage
+        exit 1
+        ;;
+esac
+
```
rn: super minor nit, but one empty line should be enough

ijc:
The creation of `*-e` was unintentional, GNU `sed` doesn't behave that way.

The invocation is `xargs sed -i -e "s,$old,$new,g"`, I suppose on MacOS it is treating the `-e` as the argument to `-i`.

Comparing the MacOS and Linux  man pages I hoped `xargs sed -i '' -e "s,$old,$new,g"` (i.e. an explicitly 0 length extension) would work, but no dice on Linux:
```
$ sh -x ./scripts/update-component-sha.sh linuxkit/runc:d5cbeb95bdafedb82ad2cf11cff1a5da7fcae630 foo
+ set -e
+ [ 2 -ne 2 ]
+ old=linuxkit/runc:d5cbeb95bdafedb82ad2cf11cff1a5da7fcae630
+ new=foo
+ git+ xargs sed -i  -e s,linuxkit/runc:d5cbeb95bdafedb82ad2cf11cff1a5da7fcae630,foo,g
 grep -l linuxkit/runc:d5cbeb95bdafedb82ad2cf11cff1a5da7fcae630
sed: can't read : No such file or directory
```

Reordering the `-i` to the end will just result in the first filename argument being taken as the backup extension. I'm not seeing a way to drive this which works for both OSes. So maybe the current solution of doing a `find` to clean up afterwards is the best way.

Unrelated but last week I ended up writing this quick helper:
```
#!/bin/sh
set -e
if [ $# -lt 1 ] ; then
    echo "Need <PKG...> as argument" >&2
    exit 1
fi

for pkg in $@ ; do
    oldhash=$(make --no-print-directory -C pkg/$pkg HASH_COMMIT=origin/master show-tag)
    newhash=$(make --no-print-directory -C pkg/$pkg show-tag)

    echo "-$oldhash"
    echo "+$newhash"
    ./scripts/update-component-sha.sh "$oldhash" "$newhash"
    echo
done
```
Not sure that it can sanely be integrated with the existing script though.

deitch:
> So maybe the current solution of doing a find to clean up afterwards is the best way.

That was kind of my thought. Whenever I use `sed -i` in a project that can run both inside (Linux) and outside (MacOS) a container, I fall afoul of this. My basic pattern nowadays is to have a `sed` that does it wherever and then some kind of cleanup action afterwards (like the `find ...`).

deitch:
> Unrelated but last week I ended up writing this quick helper

So looks like we both had the same issue. Happy to merge this in or replace that. I originally wrote separate script, but @justincormack asked to make it one with options.

Either way, let me know.

deitch:
>  would prefer if the usage would be:
> update-component-sha.sh <old> <new>
> and
> update-component-sha.sh --image <image> <new>

It does that. It makes the "canonical" way to have `--image` or `--hash`, but if you provide no option, it falls back to the old way of `update <old> <new>`/

scripts/update-component-sha.sh
deitch: Heh, because this was my usage guide before I wrote `usage`! First I doc how it will work, then write it, then realized we needed better `usage` so added it. :-)

scripts/update-component-sha.sh
deitch: I am confused. I thought the two modes were:

1. Existing: replace all instances of hash x with hash y
2. New: change tags for all instances of image q with hash y

Are you asking that we change the first one to: replace all instances of image:tag combo q:x with q:y? I would think the existing functionality of hash is enough?

rn:
> It does that. It makes the "canonical" way to have --image or --hash, but if you provide no option, it falls back to the old way of update <old> <new>/

Hah, I missed the `set --` hack at the top...

scripts/update-component-sha.sh
ijc: I wouldn't call it `backwards compatibility` I'd call it `a (convenient to some) default`.

ijc:
@deitch note that my script in https://github.com/linuxkit/linuxkit/pull/2275#issuecomment-317368775 is a third thing to what is added here, which is to replace all uses of the current `origin/master` of a given `pkg/foo` with the current `HEAD` version (and in fact to operate on a list of `pkg/foo`s). That's pretty niche and specific to whatever it was I was doing last week (I've already forgotten what I used it for!). Probably needs a fair bit of generalisation to warrant being added to this script at this stage.

IOW I think it is fine to just leave that snippet in the comment and not try to integrate into this script at this point. If I (or anyone) finds themselves needing to do something like that on a regular basis then we'll have a better idea what the needed functionality actually is.

justincormack:
@ijc I believe MacOS `sed` is Posix compliant and GNU `sed` has special GNUisms that can do in place update without backup, so it is best to use the Posix form and then delete the backups.

scripts/update-component-sha.sh
deitch: LOL! Do you really want me to change the comment?

ijc: I think "By default the first method is assumed" is a correct description of the behaviour. If you were to remove it like Rolf suggested I think you would also want to remove the code which implements the default (NB: I don't think you should do that).

deitch: > I don't think you should do that
Me neither. I like leaving it with the existing functionality as default.

OK, I am going to update the usage line (and remove the comments while I am at it).

scripts/update-component-sha.sh
ijc: Sounds like we have been using it in different ways. I've (and @rn?) use it as:
```
./scripts/update-component-sha.sh linuxkitprojects/swarmd:1cd4c061cc7327750d2a12c267db6d4d9e26b1d3 linuxkitprojects/swarmd:cce587b5e67b7d3e088a2e527dafc5ad3edb6502
```
Whereas you seem to use it as
```
./scripts/update-component-sha.sh 1cd4c061cc7327750d2a12c267db6d4d9e26b1d3  cce587b5e67b7d3e088a2e527dafc5ad3edb6502
```

If we discount the possibility of hash clashes (super unlikely) then they are equivalent. I just the first form because I can easily pick it out from the `show-tag` output, or just from the build log.

ijc: I suppose in a way your usage matches the `--hash` name you've given it, whereas mine might better be called `--tag`, but given the implementation would be identical (and actually you can omit the argument anyway) then I don't think we need to have both...

scripts/update-component-sha.sh
deitch: Comments and usage updated. Does that work?

ijc:
> @ijc I believe MacOS sed is Posix compliant and GNU sed has special GNUisms that can do in place update without backup, so it is best to use the Posix form and then delete the backups.

Agreed.

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
+##
+#
+# script to replace hashes in config files
+# see usage() for usage and functionality
+#
+
+function usage() {
+    echo "$0 --<mode> <how-to-find> <new-hash>" >&2
+    echo >&2
+    echo "Avaioable modes: --hash and --image" >&2
```
ijc: Typoe here

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
+    echo "	$0 --hash <OLD> <NEW>" >&2
+    echo "	Example: $0 8675309abcdefg abcdef567899" >&2
+    echo "	   Will replace all instances of 8675309abcdefg with abcdef567899" >&2
+    echo >&2
+    echo "Replace by image: $0 --image <IMAGE> <NEW>" >&2
+    echo "	$0 --image <IMAGE> <NEW>" >&2
+    echo "	Example: $0 linuxkit/foo abcdef567899" >&2
+    echo "	   Will tag all instances of linuxkit/foo with abcdef567899" >&2
+    echo >&2
+    echo "By default, for convenience, if no mode is given (--image or --hash), the furst method (--hash) is assumed. Thus the following are equivalent:" >&2
```
ijc: `s/furst/first/g` here. I'd also consider a line break before `Thus` to keep it under common terminal widths.

You could also consider replacing the multitude of `echo "..." >&2` with something like:
```
cat >&2 <<EOF
Blah blah
blah blah
Usage blah:
     ---hash blah
...
EOF
```
Might make it a (little) more readable in source form.

deitch: I thought of that but was too lazy, since it started as one `echo`, then two, then...

I guess I should fix it.

deitch: Fixed.

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,66 @@
...
+    --hash)
+        old=$2
+        new=$3
+
+        git grep -l "$old" | xargs sed -i -e "s,$old,$new,g"
+        ;;
+    --image)
+        image=$2
+        hash=$3
+        git grep -l "$image:" | xargs sed -i -e "s,$image:[[:xdigit:]][[:xdigit:]]*,$image:$hash,g"
```
ijc: Do we have any cases where a given `$image` is a suffix of another? Perhaps use `-E` and `\b` at the start to constrain the match to word boundaries? Oh, or I see `git grep` has `-w` AKA `--word-regexp` which sounds even better (I suppose that applies to my code which became `--hash` mode too).

For the `[[:xdigit:]][[:xdigit:]]` I think POSIX re include counts so `[[:xdigit:]]{40}` should be a safer match (probably not many instances which aren't followed by a character which is not a space or something else not in `[[:xdigit:]]`, but still).

deitch: > I think POSIX re include counts so [[:xdigit:]]{40} should be a safer match
That is a good idea.

deitch: Updated to do precise count

deitch:
Any other changes? Can we merge this in?

deitch:
(after CI passes, of course)

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,67 @@
...
+	$0 --hash <OLD> <NEW>
+	Example: $0 8675309abcdefg abcdef567899
+    	   Will replace all instances of 8675309abcdefg with abcdef567899
+
+Replace by image: $0 --image <IMAGE> <NEW>
+	$0 --image <IMAGE> <NEW>
+	Example: $0 linuxkit/foo abcdef567899
+	   Will tag all instances of linuxkit/foo with abcdef567899
+
+By default, for convenience, if no mode is given (--image or --hash), the furst method (--hash) is assumed. Thus the following are equivalent:
```
ijc: Still a typo here (and a long line)

scripts/update-component-sha.sh
ijc: I think you can nuke this comment now that `usage()` has more details

scripts/update-component-sha.sh
deitch: It has a hidden meaning. The first method actually is the prince or ruler method!

scripts/update-component-sha.sh
ijc: First paragraph here not addressed.

scripts/update-component-sha.sh
deitch: Updated

scripts/update-component-sha.sh
ijc: I agree.

scripts/update-component-sha.sh
deitch: Eh, let's leave it. I like having a short comment at the beginning of a script that says "this is what I do."

scripts/update-component-sha.sh
deitch: OK, cleaned

scripts/update-component-sha.sh
deitch: So using `-w` works for `--hash` mode, since it always is preceded by a `:` and following by whitespace or end of line (or some other non-word char).

It doesn't work for `--image` mode, since we match with `${image}:`, which is followed by a word-char (the beginning of the hash). I was thinking of dropping the `:`, but then it is likely to match in all sorts of docs and maybe Makefiles, so not a good idea.

May need to just use `-E` with `\b`.

deitch: I used that. Updated.

scripts/update-component-sha.sh
```diff
@@ -1,10 +1,67 @@
...
+    --hash)
+        old=$2
+        new=$3
+
+        git grep -w -l "$old" | xargs sed -i -e "s,$old,$new,g"
+        ;;
+    --image)
+        image=$2
+        hash=$3
+        git grep -E -l "\b$image:" | xargs sed -i -e "s,$image:[[:xdigit:]]"'\{40\}'",$image:$hash,g"
```
ijc: I think you may need the `\b` on the sed too, otherwise a file which has references to both `foolinuxkit/foo` and `linuxkit/foo` will potentially get both updated.

WRT `So using -w works for --hash mode, since it always is preceded by a : ` in another comment see https://github.com/linuxkit/linuxkit/pull/2275#discussion_r128997311 -- not everyone is using it that way at the moment.

deitch: > I think you may need the \b on the sed too, otherwise a file which has references to both foolinuxkit/foo and linuxkit/foo will potentially get both updated.

Yep. Worth fixing.

> WRT So using -w works for --hash mode, since it always is preceded by a : in another comment see #2275 (comment) -- not everyone is using it that way at the moment.

Yep, same here too. Besides, it is worth being consistent.

deitch: > I think you may need the \b on the sed too

Aren't `\b` and `\<` and `\>` gnu sed extensions? Any equivalent to use?

deitch: Fixed the second one...

deitch: Apparently Mac supports `[[:<:]]` and `[[:>:]]` but that doesn't appear to be universal. Ideas?

ijc: They aren't mentioned in http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap09.html#tag_09_03 (found from http://pubs.opengroup.org/onlinepubs/009695399/utilities/sed.html), so I suppose they must be. Doesn't look they are covered by `sed -E` either.

It'd have been nice to avoid this particular quirk, but I don't think it is going to be all that critical in practice to jump through too many hoops to avoid, so I guess we can just ignore the problem.

justincormack:
Can we change the `sed` to do `sed -i .bak -e ...` as that is more correct and should work everywhere, then remove all the `.bak` files.

deitch:
> Can we change the sed to do sed -i .bak -e ... 

Does it? Tried it on alpine:3.6 to no success:

```
$ docker run -it --rm alpine:3.6 sh
/ # cat > /tmp/abc <<EOF
> abc
> def
> ghi
> abc
> jkl
> mno
> abc
> EOF
/ # sed -i '.bak' -e 's/abc/LLL/g' /tmp/abc
sed: .bak: No such file or directory
/ #
```

justincormack:
Hmm, busybox `sed` clains to require `sed -i.bak -e ...` with no space, which should also work elsewhere...

deitch:
Phew, yes, that works.

deitch:
Yeah, that worked. Updated.

ijc:
FWIW `-i.bak` works with GNU/Linux `sed` too.

deitch:
@ijc right, but only if there is no whitespace between `-i` and `.bak`. 

OK, pushed out and updated.

Any other comments?

deitch:
Looks like we got everyone's comments in. Going to merge. If we need more improvements, happy to entertain a new PR.

arm64b:
I would like to see you guys can extend the script to replace 'linuxkit/alpine:sha' to 'other_org/alpine:sha' like this:
`#./update-component-sha.sh --image linuxkit/alpine:sha  other_org/alpine:sha path/to/pkgs`
the above will replace all linuxkit/alpine:sha to other_org/alpine:sha in all files under the 'path/to/pkgs'

rn:
@arm64b I personally don't use the script but have a shell alias which does the same:
```
git-replace ()
{
    git grep -l "$1" | xargs sed -i -e "s,$1,$2,g"
}
```

arm64b:
@rn git-replace is good to use 👍

PR #2276, opened 2017-07-23
Refactor OpenStack push support to use Gophercloud

yankcrime:
Signed-off-by: Nick Jones <nick@dischord.org>

**- What I did**

Reactored the OpenStack push support to make use of [Gophercloud](http://gophercloud.io) instead of handling authentication and talking to the image service API directly.

This simplifies the codebase and eases the burden of maintenance.

FWIW Gophercloud is pretty standard, it's well supported and used by numerous other projects.

**- How I did it**

This PR imports the Gophercloud packages as dependancies and then uses their model to define a provider, authenticate, and then configure an image client with appropriate URLs as advertised via the service catalogue (Keystone).

I've also amended some of the options so that they're consistent with other providers (i.e `img-name`) and updated the documentation accordingly.

**- How to verify it**

```shell
linuxkit push openstack -img-name "myimage" \
-authurl https://compute.datacentred.io:5000/v3 \
-username "xxx" -password "xxx" -project "xxx" \
linuxkit.raw
```

Refactor OpenStack push support to take advantage of Gophercloud instead.

Here's a GIF of a Kangaroo who is super stoked for the OpenStack support 🤘

kangarooguitar

### DIFF
--- a/src/cmd/linuxkit/push_openstack.go
+++ b/src/cmd/linuxkit/push_openstack.go
@@ -1,77 +1,19 @@
 package main
 
 import (
-	"bytes"
-	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
-	"net/http"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
-
-// KeyStoneV3 for OpenStack login
-type KeyStoneV3 struct {
-	Auth struct {
-		Identity struct {
-			Methods  []string `json:"methods"`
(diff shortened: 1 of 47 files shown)


src/cmd/linuxkit/push_openstack.go
```diff
@@ -1,99 +1,37 @@
...
-	"bytes"
-	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
-	"net/http"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
```
justincormack: we usually put all the `github.com` URLs at the bottom (where logrus is). Also you should not need the top two?

justincormack:
cc @thebsdbox

yankcrime:
Thanks @justincormack - I've moved those URLs but they're all required it seems.

justincormack:
ok, they can be required but its fairly unusual...

thebsdbox:
Excellent, the library provides more compatibility across releases and in theory could be used for a `run` option.

PR #2278, opened 2017-07-24
tpm tools

deitch:
**- What I did**
Create an image with the tpm 1.2 software stack (tss), as well as the necessary daemon `tcsd` (configured to start by default).

Also adds an example and a doc.

**- How I did it**
1. `FROM linuxkit/alpine`
2. install necessary packages
3. Download specific version of trousers (tss daemon and lib) and tpm-tools
4. compile and save to `/out/`
5. Load into new `FROM scratch` image

**- How to verify it**
Run it on a laptop/server with tpm 1.2 chip and see that `tcsd` starts and you can communicate with it.

Add tpm 1.2 software stack support

gremlin

**Note:** testing on this is a bit slow, so don't approve until `[WIP]` removed.


docs/trusted-computing.md
```diff
@@ -0,0 +1,59 @@
...
+  - name: tss
+    image: linuxkit/tss:3406973699199b47224ad4e819e5728964673772
+```
+
+For a full example, see [tpm.yml](../examples/tpm.yml)
+
+### tss 2.0
+In tss 2.0, the character device `/dev/tpmrm0` can be addressed by as many processes, in parallel, as desired. All of the multiplexing and low-level services are built into the kernel module.
+
+To use a tpm 2.0 device, you to **not** need any special tss container. You just need an container that:
```
justincormack: typoe for "do not"

deitch: I was trying to speak New Jersey-ese, like Tony Soprano: "Youse guys to not!" :-)

Fixing...

deitch: Thanks, fixed.

examples/tpm.yml
```diff
@@ -0,0 +1,30 @@
...
+    image: linuxkit/dhcpcd:4b7b8bb024cebb1bbb9c8026d44d7cbc8e202c41
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+services:
+  - name: getty
+    image: linuxkit/getty:1b651a91f1c17f50357be9873b580ccf81668630
+    env:
+     - INSECURE=true
+  - name: tss
+    image: linuxkit/tss:3406973699199b47224ad4e819e5728964673772
+  - name: rngd
```
justincormack: Don't really like an example which doesn't do anything, eg can't really base a test on it. Is there anything that could talk to it and check its working we could use as a test? An example which just says "put this in your image" is no better than the docs (and from the testing point of view is worse than a simple build test).

pkg/tss/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    openssl-dev \
+    util-linux \
+    && true
+
+RUN mkdir -p /usr/src
+COPY upload/src/glibc_stubs/ /usr/src/glibc_stubs
+WORKDIR /usr/src/glibc_stubs
+RUN make && make install
+
+RUN git clone https://git.code.sf.net/p/trousers/tpm-tools /usr/src/trousers-tpm-tools && cd /usr/src/trousers-tpm-tools && git checkout bdf9f1bc8f63cd6fc370c2deb58d03ac55079e84
```
justincormack: We use env vars to set the commit elsewhere its more readable and cleaner for updates.

pkg/tss/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+RUN git clone https://git.code.sf.net/p/trousers/trousers /usr/src/trousers-trousers && cd /usr/src/trousers-trousers && git checkout de57f069ef2297d6a6b3a0353e217a5a2f66e444
+WORKDIR /usr/src/trousers-trousers
+RUN sh bootstrap.sh && \
+    ./configure --prefix=/ --sysconfdir=/etc LDFLAGS="-L/out/lib/ -lgetpwent_r" && \
```
justincormack: why do you need to set LDFLAGS here?  That seems weird.

examples/tpm.yml
deitch: We could have it up and then run, e.g. `tpm_version` or `tpm_nvinfo`, but they will fail if `tcsd` isn't running, and it will fail unless you test with a chip, or a virtual tpm. `tcsd` only supports real tpms, so you either need the one that loads into the kernel, or one that virtualization supports.

Happy to take some good ideas here.

pkg/tss/upload/out/etc/group
```diff
@@ -0,0 +1,6 @@
+root:x:0:root
+bin:x:1:root,bin,daemon
+daemon:x:2:root,bin,daemon
+sys:x:3:root,bin,adm
+adm:x:4:root,adm,daemon
+tss:x:116:
```
justincormack: generally we do adduser etc in the container rather than adding group and passwd files.

pkg/tss/Dockerfile
deitch: I definite like that better. Just `ENV trousers_commit abcdefg` in the `Dockerfile`?

pkg/tss/upload/out/etc/group
deitch: Yeah, good point. We can do that.

deitch: No, wait. If we have a `scratch` container, does it even _have_ `adduser`/`addgroup`?

pkg/tss/Dockerfile
deitch: Two reasons:

1. We save the libs (and bins) to `/out/...` so it isn't going to find the ones it needs (especially `libtspi` in the usual `/lib/` locations.
2. The software itself assumes that we are using `glibc`, which has `getpwent_r` (musl does not). That is the only `glibc` thing (and isn't really reentrant anyways). We have a small stub lib that provides it. We need to tell the code about that lib.

I would be much happier if trousers just did the right thing here. We have an open issue with them, but I am not holding my breath.

If there are other ways, happy to use them.

examples/tpm.yml
justincormack: Even an example that you can only run with a real tpm is better than one which doesnt relly do anything. But can't we run a test with the qemu emulated tpm relatively easily? Or is it not that easy?

pkg/tss/Dockerfile
justincormack: yes. Usually we use upper case variables for some reason (visibility?)

pkg/tss/upload/out/etc/group
justincormack: you already have `busybox` in it for `chown`, you can do `busybox adduser ...` before you remove it

justincormack: or you can do the adduser first in the parent container and then copy the relevant files from their

examples/tpm.yml
deitch: Not that easy. The tpm support in qemu is just pass-through. It connects to the physical tpm device on the host on which qemu is running. There is a very longstanding patch for qemu to also support other paths (Unix-domain sockets, e.g.). Never been merged in.

pkg/tss/Dockerfile
deitch: OK

pkg/tss/upload/out/etc/group
deitch: OK, sure.

deitch: Hmm, apparently `busybox addgroup` doesn't work on a `scratch` image. It must have `/etc/group` already in place for it to work.

Other idea? I could add it manually... but that is what I already was doing.

pkg/tss/Dockerfile
justincormack: It seems really weird to add a library you are using for compiling to `out/` as that is where you are installing. Just install the libraries it needs to build in `/usr/local/lib` and it will be much easier.

If necessary run `make install` twice, once to /usr/local and once to `/out` for things you need in the final image.

pkg/tss/upload/out/etc/group
justincormack: Just do it in the build container and copy the `passwd` etc files to the scratch container then.

deitch: Cleaner not to. The various and sundry packages create a lot of users and groups:

```
# cat /etc/passwd
root:x:0:0:root:/root:/bin/ash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
news:x:9:13:news:/usr/lib/news:/sbin/nologin
uucp:x:10:14:uucp:/var/spool/uucppublic:/sbin/nologin
operator:x:11:0:operator:/root:/bin/sh
man:x:13:15:man:/usr/man:/sbin/nologin
postmaster:x:14:12:postmaster:/var/spool/mail:/sbin/nologin
cron:x:16:16:cron:/var/spool/cron:/sbin/nologin
ftp:x:21:21::/var/lib/ftp:/sbin/nologin
sshd:x:22:22:sshd:/dev/null:/sbin/nologin
at:x:25:25:at:/var/spool/cron/atjobs:/sbin/nologin
squid:x:31:31:Squid:/var/cache/squid:/sbin/nologin
xfs:x:33:33:X Font Server:/etc/X11/fs:/sbin/nologin
games:x:35:35:games:/usr/games:/sbin/nologin
postgres:x:70:70::/var/lib/postgresql:/bin/sh
cyrus:x:85:12::/usr/cyrus:/sbin/nologin
vpopmail:x:89:89::/var/vpopmail:/sbin/nologin
ntp:x:123:123:NTP:/var/empty:/sbin/nologin
smmsp:x:209:209:smmsp:/var/spool/mqueue:/sbin/nologin
guest:x:405:100:guest:/dev/null:/sbin/nologin
nobody:x:65534:65534:nobody:/:/sbin/nologin
```


```
# cat /etc/group
root:x:0:root
bin:x:1:root,bin,daemon
daemon:x:2:root,bin,daemon
sys:x:3:root,bin,adm
adm:x:4:root,adm,daemon
tty:x:5:
disk:x:6:root,adm
lp:x:7:lp
mem:x:8:
kmem:x:9:
wheel:x:10:root
floppy:x:11:root
mail:x:12:mail
news:x:13:news
uucp:x:14:uucp
man:x:15:man
cron:x:16:cron
console:x:17:
audio:x:18:
cdrom:x:19:
dialout:x:20:root
ftp:x:21:
sshd:x:22:
input:x:23:
at:x:25:at
tape:x:26:root
video:x:27:root
netdev:x:28:
readproc:x:30:
squid:x:31:squid
xfs:x:33:xfs
kvm:x:34:kvm
games:x:35:
shadow:x:42:
postgres:x:70:
cdrw:x:80:
usb:x:85:
vpopmail:x:89:
users:x:100:games
ntp:x:123:
nofiles:x:200:
smmsp:x:209:smmsp
locate:x:245:
abuild:x:300:
utmp:x:406:
ping:x:999:
nogroup:x:65533:
nobody:x:65534:
```

As much as I prefer `adduser` and `addgroup` too, adding _all_ of these users and groups is far messier.

docs/trusted-computing.md
```diff
@@ -0,0 +1,59 @@
...
+
+* `tcsd`
+* the various `tpm_*` tools
+
+To make a `tcsd` available to your LinuxKit image, just include it:
+
+```yml
+services:
+  - name: tss
+    image: linuxkit/tss:3406973699199b47224ad4e819e5728964673772
```
justincormack: Can we omit the exact hash from the docs?

pkg/tss/upload/out/etc/group
justincormack: ok.

Can we not put all these packages in `upload/out` then. These are details of the build process.

justincormack:
WHat is `pkg/tss/upload/out/system.data` binary file in commit?

pkg/tss/upload/out/etc/group
deitch: > Can we not put all these packages in upload/out then. These are details of the build process.

You lost me. You mean to put it in a separate dir and `COPY` directly to the final image, as opposed to the build image's `/out/`?

docs/trusted-computing.md
deitch: Yeah.

deitch: Done.

pkg/tss/upload/out/etc/group
justincormack: I don't like having `upload/out` in the linuxkit tree. Usually we just put things directly under `etc/` say in the build tree and the `Dockerfile` puts things in the right place; if that is really inconvenient, then something like `upload` would be ok, but not with `out` under it as that is only there because that is an internal directory you are using in the build process.

deitch:
> What is pkg/tss/upload/out/system.data binary file in commit?

Something that shouldn't be there. Nice catch. This was a conversion from work I had done elsewhere. 

If you want to know what it really does, ask me offline. It needs to exist, but is created the first time a `tcsd` runs and takes ownership of the tpm (in v1.2 anyways). This was from a project where we bootstrapped the tpm config.

Going away...

pkg/tss/upload/out/etc/group
deitch: That's what I thought you meant.

deitch:
Updated all of those. 

Remember: this is still `[WIP]`. When we feel it is good to go, I want to run a full battery of tests on my local tpm-enabeld laptop.

justincormack:
There is also a zero size `pkg/tss/etc/tcsd.conf`

Apart from that looks fine now.

deitch:
> There is also a zero size pkg/tss/etc/tcsd.conf

There has to be. Even if you stick with defaults, you need to have a `tcsd.conf`. :-(

justincormack:
It seems pointless to check it in though - cant you just `touch` it during build?

deitch:
> cant you just touch it during build?

I thought about it, but decided to leave it as an empty file. It serves as a better reference point if we ever decide to go for something other than default.

justincormack:
Almost always though if people want to change the default they will be doing that in their runtime image by adding a specific config; we have a general rule of leaving the configs as the default unless there is a very good reason, so I think just doing `touch` is fine. It is very confusing having empty files committed...

deitch:
Sigh, yeah. Done. Got to update the hash on the example in a moment.

deitch:
First run with latest build works, `tcsd` starts, can read tpm info, etc.

pkg/tss/Makefile
```diff
@@ -0,0 +1,6 @@
+IMAGE=tss
```
riyazdf: this is great.  Let me know when you'd like me to set up the repo / signing, etc

deitch: Will do. Getting close.

deitch:
Found a bug in my glibc stub function. Will update shortly.

deitch:
This is fully functional. Tested both with and without tpm chip. Removing `[WIP]`

deitch:
@riyazdf @justincormack who already have expressed interest, please take a look.

Anyone else with interest to review among the maintainers?

justincormack:
Can you add a build test in https://github.com/linuxkit/linuxkit/tree/master/test/cases/000_build/100_examples

deitch:
> Can you add a build test

Sure. I will get to it in next day or two.

justincormack:
Given we cant test it running it is the next best thing, at least it checks it is signed and there are no typos in the labels etc.

deitch:
> Can you add a build test in

What do those do? It looks like under `000_build/100_examples/` there is a series of directories, all named for whatever they are testing, with a simple `test.sh` file that does the same thing for all of them: build the linuxkit image using `examples/<something>/yml` and then remove it? Is that it?

justincormack:
Yes thats it. It literally just builds it.

deitch:
Well, if I had known it was _that_ simple...

Done.

deitch:
CI failed because I didn't push the `linuxkit/tss` image out. Wasting until at least one other maintainer says it is good to push out.

deitch:
Thanks @riyazdf . Pushed out signed. Would you mind restarting CI?

riyazdf:
kicked CI!

justincormack:
@talex5 what do we have to do to give @deitch ability to restart CI?

talex5:
@justincormack : add a line at https://github.com/linuxkit/linuxkit-ci/blob/a53e16d898ec86e01478fb617863d0075989ae04/src/ci.ml#L101 with the desired GitHub username.

PR #2281, opened 2017-07-24
Rewrite /init in Go

justincormack:
This uses a more memory efficient copy, and gets us closer to
not having a shell in the base system if not required.

replaces #2098 by using Go not a shell script so we avoid lots of forking. Surprisingly large amount of Go to replace a two line shell script...

walrus


pkg/init/init
```diff
@@ -1,9 +0,0 @@
-#!/bin/sh
-
-/bin/mount -t tmpfs tmpfs /mnt
-
-/bin/cp -a / /mnt 2>/dev/null
-
-/bin/mount -t proc -o noexec,nosuid,nodev proc /proc
```
rn: your go code does not mount `/proc` is that only mounted here for the `switch_root`?

justincormack: Yes, it is not needed, it is done as the first line of the init script in `/bin/rc.init`

justincormack:
Oh, I did update the hash but forgot to push it

justincormack:
Hashes pushed

justincormack:
@rn its green now with updated hashes

PR #2286, opened 2017-07-25
tools/alpine: Resync alpine base image

rn:
Somewhere between the various updates yesterday the hash in
'versions.x86_64' went wrong and there is no image with hash
available on hub.

This commit updates the alpine base to the latest version and
thus rectifies the issue


ijc:
We really need to sort out (pun intended) the stability of the `sort` call on the version list.

Perhaps using `LANG=C sort` is sufficient, but maybe we should also push it into the container so we get a consistent sort binary too?

rn:
@ijc I'm planning to make some changes to the Makefile/Dockerfile this afternoon, will see ig `LANG=C` comes up with a more stable ordering

PR #2287, opened 2017-07-25
Update and extend namespace stress tests

rn:
- Rewrite the stress tests using `runc` instead of just creating a network namespace
- Use the `sock_stress` utility from https://github.com/linuxkit/virtsock instead of `iperf3` as it provides more flexibility over the traffic pattern and socket types to use
- Add tests for unix domain sockets
- Provide more individual tests as well as mixes of tests.
- All tests are disabled by default (ie they are not run by CI) because they take quite a while to run. Some tests can be enabled by specifying the `kernel` label and additional tests can be enabled using the `kernel-extra` label.
- Structure the tests so that it easier to add additional tests.

whale-boat


justincormack:
`runc spec` will output a minimal-ish version of the config file which you can then adapt; might be easier as means less updating and it basically has all the stuff you need for it to work that you have.

rn:
I didn't know about `runc spec` but having taken a look I don't think it will simplify things. I would still have to prod the values into json and having the template checked in makes it easier to test.

rn:
sure, should be a painless update for me

justincormack:
thats merged now

rn:
updated `linuxkit/init` and force pushed

Issue #2288, opened 2017-07-25
fully immutable filesystem


justincormack:
Currently a few parts of the init script cheat and make filesystem modifiable. Fix these

- [x] dhcp /etc/resolv/conf see #1920 
- [ ] /var if not overmounted, should be a tmpfs
- [x] container root filesystem, should add a tmpfs overlay if writeable #2309

At this point we should be able to mount the filesystem `ro` very early, and run off a cdrom filesystem or similar too if needed.

PR #2289, opened 2017-07-25
resolv.conf fixes

### LINKED ISSUE #1920: dhcp cleanup
dhcpcd should not use `/tmp/etc` for storing state. Should be in `/run`, probably using a resolvconf style setup as per Debian/Alpine resolvconf packahes.

justincormack:
Do not try to change /etc/resolv.conf
    
The filesystem is supposed to be immutable, so do not try to make
a symlink; new versions of moby tool should add one anyway. But
try to make the directory a symlink points to, assuming that it
will be on a writeable filesystem.
    
fix #1920
see also #2288
updates moby to include https://github.com/moby/tool/pull/128

cat-in-a-box

### DIFF
--- a/pkg/dhcpcd/Dockerfile
+++ b/pkg/dhcpcd/Dockerfile
@@ -16,4 +16,4 @@ WORKDIR /
 COPY --from=mirror /out/ /
 COPY /dhcpcd.conf /usr/ /
 CMD ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf"]
-LABEL org.mobyproject.config='{"binds": ["/var:/var", "/tmp/etc:/etc"], "capabilities": ["CAP_NET_ADMIN", "CAP_NET_BIND_SERVICE", "CAP_NET_RAW"]}'
+LABEL org.mobyproject.config='{"binds": ["/run/resolvconf:/etc"], "capabilities": ["CAP_NET_ADMIN", "CAP_NET_BIND_SERVICE", "CAP_NET_RAW"]}'
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 PREFIX?=/usr/local/
 
 MOBY_REPO=https://github.com/moby/tool.git
-MOBY_COMMIT=8e720bff08cb9f20488d1ae6f114a5b1e4edf9cd
+MOBY_COMMIT=36217e5145f1e54807490c09cc389f5e3ac99075
 MOBY_VERSION=0.0
 bin/moby: tmp_moby_bin.tar | bin
 	tar xf $<
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ endif
 
 PREFIX?=/usr/local/
 
+MOBY_REPO=https://github.com/moby/tool.git
 MOBY_COMMIT=8e720bff08cb9f20488d1ae6f114a5b1e4edf9cd
 MOBY_VERSION=0.0
 bin/moby: tmp_moby_bin.tar | bin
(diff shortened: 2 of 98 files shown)


justincormack:
The only user facing change should be that any image that wants to have access to `resolv.conf` must add a `binds: - /etc/resolv.conf:/etc/resolv/conf`. They should have been doing that before, but previously they had a default Google DNS if they did not, so some may have been missed.

pkg/dhcpcd/Dockerfile
```diff
@@ -16,4 +16,4 @@ WORKDIR /
 COPY --from=mirror /out/ /
 COPY /dhcpcd.conf /usr/ /
 CMD ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf"]
-LABEL org.mobyproject.config='{"binds": ["/var:/var", "/tmp/etc:/etc"], "capabilities": ["CAP_NET_ADMIN", "CAP_NET_BIND_SERVICE", "CAP_NET_RAW"]}'
+LABEL org.mobyproject.config='{"binds": ["/run/resolvconf:/etc"], "capabilities": ["CAP_NET_ADMIN", "CAP_NET_BIND_SERVICE", "CAP_NET_RAW"]}'
```
ijc: This looks a bit odd, is there nothing else in the image's `/etc/` which is interesting?

Perhaps `/run/resolvconf/resolve.conf:/etc/resolv.conf` would be less intrusive?

justincormack: Yeah it does look "interesting" but it is somewhat hard to change. You cannot bind mount just the file as it writes a new one, so the inode changes. It also writes some other files in there (eg `ntpd.conf`) which are fairly harmless. In future we could customise the scripts to write to more useful places but mapping `etc` has been how we did it so far.

justincormack: But no, it does not use anything else in `/etc`. `dhcp` does not do DNS or anything... If there was we could also bind that over but it seems to run fine.

ijc: Does the dhcp client replace a symlink or the target? If it replaces the target you could do the same symlink to `/run` trick as you do in init.

In any case, I'm ok with this for now.

justincormack: It probably replaces the target, but its really difficult to make `/etc/resolv.conf` a symlink in a container except by changing it at runtime or making them all symlinks in every image, which is also a bit weird. Going to leave for now...

ijc: I was thinking that dhcp clients were a special case worth modifying in this way, but fine to leave it.

justincormack: I dont know how to tell if something is a dhcp client, other than adding a special label...

ijc: I just meant to suggest adding the /run bind to the binds in the label in `pkg/dhcpcd/Dockerfile`, which you know is a client, not to do something clever elsewhere.

justincormack: You can't create a symlink with a bind as `runc` dereferences them so it still needs to have the symlink at `/etc/resolv.conf`. It is all a bit gross.

ijc: I think we are talking at cross purposes, all I was suggesting was that the `pkg/dhcpcd/Dockerfile` have in it:
```
RUN ln -nfs /run/resolvconf/resolve.conf /etc/resolv.conf
LABEL org.mobyproject.config='{"binds": ["/run/resolvconf:/run/resolvconf"], ...
```
Then the dhcpc client writes through it's local `/etc/resolv.conf` into the host/shared `/run/resolvconf/resolve.conf` which the host then picks up via it's own symlink. Other "normal" packages just bind `/etc/resolv.conf:/etc/resolv.conf` as normaly and rely on the symlink deref behaviour of `runc` to get the right thing.

justincormack: you cannot have `RUN ln -nfs /run/resolvconf/resolve.conf /etc/resolv.conf` in any Dockerfile as `/etc/resolv/conf` is an external read only bind mount that you cannot override.

justincormack: (It would help if `--network-none` builds/runs did not bind mount that in)

ijc: > external read only bind mount that you cannot override

Damn, I always forget that :-(

In theory we could probably configure dhcpcd to write to `/etc/resolv.conf.rw` (or just to `/run` directly) but that's probably going too far down this particular rabbit hole.

justincormack: I think not adding the bind mounts for network stuff with network=none is the best long term solution. They are of no real use anyway, and we have network=none on build too.

PR #2290, opened 2017-07-25
tools/alpine: Make sure we use content trust for linuxkit/alpine

rn:
We should always pull and push linuxkit/alpine with content trust
irrespective of architecture, unless explicitly disabled.

Currently, we have to dance around various other issues on arm64,
which are now documented in the Makefile.

tiger

### DIFF
--- a/tools/alpine/Makefile
+++ b/tools/alpine/Makefile
@@ -2,21 +2,38 @@
 
 ORG?=linuxkit
 IMAGE=alpine
+DEPS=packages
 
-DOCKER_CONTENT_VAR=1
-PKG_DEPS=packages
-
-ARCH := $(shell uname -m)
-
-ifeq ($(ARCH), x86_64)
-BASE=alpine:3.6
-PKG_DEPS += packages.x86_64
+# The logic for content trust is a bit convoluted because:
+# - The arm64 base image is currently not signed so we need to pull it
+#   with content trust disabled. This is controlled by
+#   DOCKER_CONTENT_PULL.
+# - 'docker build' with the FROM image supplied as environment
+#   variable *and* with DOCKER_CONTENT_TRUST=1 currently does not work
+#   (https://github.com/moby/moby/issues/34199). We therefor build
+#   with DOCKER_CONTENT_TRUST explicitly set to 0. However, we pull
+#   the base image just before with content trust enabled (if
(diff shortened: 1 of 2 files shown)


tools/alpine/Makefile
```diff
@@ -36,9 +53,9 @@ versions.$(ARCH): Makefile hash iid
 	docker run --rm $(shell cat iid) find /mirror -name '*.apk' -exec basename '{}' .apk \; | sort | (echo '# automatically generated list of installed packages'; cat -) >> versions.$(ARCH)
 
 push: hash iid versions.$(ARCH)
-	DOCKER_CONTENT_TRUST=$(DOCKER_CONTENT_VAR) docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
+	docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
```
justincormack: this is supposed to be a trusted pull if you are going to push with trust, as if there is an unsigned copy it will ignore it and replace it with your signed one.

justincormack: Oh I see you are always setting trust, ok.

rn: yes, but we have `export DOCKER_CONTENT_TRUST=1` set globally

tools/alpine/versions.x86_64
```diff
@@ -1,239 +0,0 @@
-# linuxkit/alpine:34af9cb1990debd17fae6d4198c62ce3910d9908
```
justincormack: why are you removing this file?

rn: Ah, missed that. Shouldn't remove it, but add an updated one once the overall Makefile changes are ok and I have updated the PR

justincormack:
Looks correct to me...

rn:
@arm64b note also that you need a newer version of the docker now (17.06) since the alpine base now uses `--iidfile`. I've build a master version of docker on ubuntu. My steps are here: https://gist.github.com/rn/08d13b7ed30ab5fbde9a6dcaa24831ce

rn:
Update the PR, now with signed ARM alpine base (weirdly, x86 did not change...I thought we'd incorporate the hash of the Makefile...). Removed the RFC bits and added animal picture

PR #2291, opened 2017-07-25
Kubernetes bump to 1.6.7 and improved functionality on Linux platform

ijc:
**- What I did**

Bumped Kubernetes project from 1.6.1 to 1.6.7 and bumped various other components (see commit messages).

Dropped `Drop k8s-dns-* from image cache` from the image cache. They appear to be unused in practice although they are mentioned in https://kubernetes.io/docs/admin/kubeadm/ so I'm unsure. They can always be pulled at runtime.

Fixes and updates to qemu runner and updates to project/kubernetes scripting to make use of those. Various other fixes and updates to the scripting for Linux use.

Use a smaller ssh container image.

**- How I did it**

Emacs

**- How to verify it**

The `README.md` should describe how to build a cluster, which should work out of the box on MacOS and with `KUBE_NETWORK=bridge,«somebridge»` on Linux. With that I've been running the sock shop demo with:
```
      kubectl --namespace=sock-shop get pods
      kubectl create namespace sock-shop
```
Kubernetes project bumped to 1.6.7
Kubernetes project more usable on Linux platform.

[](https://en.wikipedia.org/wiki/The_Elephant_Riders)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -547,7 +615,10 @@ func splitPublish(publish string) (publishedPorts, error) {
 }
 
 func buildQemuForwardings(publishFlags multipleFlag, containerized bool) (string, error) {
-	forwardings := "user"
```
riyazdf: want to confirm my understanding: is this ok because of `qemuNetworkingDefault == user`?

ijc: Correct. The caller now just unconditionally appends the result of this function to the mode, and the caller is responsible for checking that the mode is consistent with doing forwarding, which in practice means `user` (the check is in the `switch netMode[0]` bit).

ijc:
@tych0 sorry, I meant to tag you on this because it changes the option you added in #2264. Hope this is ok...

tych0:
No problem. Works for me, as long as there's some way to accomplish it :)

PR #2294, opened 2017-07-26
pkg: Add forcepush target

ijc:
This is like the `push` target but omits the pulls and depends on forcetag
instead. With the git commit now being embedded into the image this is now a
necessary part of rebasing a PR for which images have already been pushed.

Also adds PHONY targets for existing forcetag and push targets which were
missing.

NB $(error) appends a "." to omit the final one from the error message

### DIFF
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -1,4 +1,4 @@
-.PHONY: image tag show-tag
+.PHONY: image tag forcetag show-tag check-dirty push forcepush
 default: push
 
 ORG?=linuxkit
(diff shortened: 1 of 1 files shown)


ijc:
I'm wondering if this could be simplified further, given that `push` already depends on `tag` which has a `pull` already perhaps the one in `push` is unneeded? If so then we could have a single rule:
```
%push: %tag
ifneq ($(DIRTY),)
	$(error Your repository is not clean. Will not push package image)
endif
	docker push $(TAG)
ifneq ($(RELEASE),)
	docker tag $(TAG) $(ORG)/$(IMAGE):$(RELEASE)
	docker push $(ORG)/$(IMAGE):$(RELEASE)
endif
```

I'm not 100% sure that `%tag` in phony rules is ok though. Alternatively if there is an alternative to `docker pull` which is a nop if there is a local image already it might be possible to just have a single `push`.

Arguably we could have a single `push` which always used `forcetag`, i.e. insist on a fresh build (modulo caching) before pushing.

justincormack:
The `pull` is not equivalent to anything local, as it usually checks content trust.

justincormack:
This looks fine as is - I had to do this yesterday, so was just doing forcetag and then pushing by hand

PR #2296, opened 2017-07-26
Make service start up containerd and services

justincormack:
This moves most of the initialisation of containers to the
service init in the `service` command.

Still leaves remounting root file systems read only but this
will go away shortly. Another step closer to removing shell
scripts in base system.

Will add updating images shortly.

owl

### DIFF
--- a/pkg/containerd/cmd/service/start.go
+++ b/pkg/containerd/cmd/service/start.go
@@ -15,7 +15,7 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
-func start(args []string) {
+func startCmd(args []string) {
 	invoked := filepath.Base(os.Args[0])
 	flags := flag.NewFlagSet("start", flag.ExitOnError)
 	flags.Usage = func() {
@@ -24,7 +24,8 @@ func start(args []string) {
 		flags.PrintDefaults()
 	}
 
-	sock := flags.String("sock", "/run/containerd/containerd.sock", "Path to containerd socket")
+	sock := flags.String("sock", defaultSocket, "Path to containerd socket")
+	path := flags.String("path", defaultPath, "Path to service configs")
 
 	dumpSpec := flags.String("dump-spec", "", "Dump container spec to file before start")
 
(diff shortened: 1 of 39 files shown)


pkg/containerd/cmd/service/start.go
```diff
@@ -40,55 +41,65 @@ func start(args []string) {
...
+	log.Debugf("Started %s pid %d", id, pid)
+}
+
+func start(service, sock, path, dumpSpec string) (string, uint32, string, error) {
+	rootfs := filepath.Join(path, service, "rootfs")
+
+	client, err := containerd.New(sock)
 	if err != nil {
-		log.WithError(err).Fatal("creating containerd client")
+		return "", 0, "creating containerd client", err
```
ijc: We have `github.com/pkg/errors` available via the containerd vendoring so maybe all these should become `errors.Wrap(err, "creating containerd client")` etc?

pkg/containerd/cmd/service/system_init.go
```diff
@@ -71,6 +76,32 @@ func systemInit(args []string) {
 		os.Exit(1)
 	}
 
+	// remove (unlikely) old containerd socket
+	_ = os.Remove(*sock)
+
+	// start up containerd
+	cmd := exec.Command(*binary)
+	cmd.Stderr = os.Stderr
```
ijc: It would be useful to send stdout somewhere, perhaps to `/var/log/system-containerd.log`? Previously all stdio went to the console so doing `cmd.Stdout = os.Stdout` too would retain the status quo.

pkg/containerd/cmd/service/start.go
justincormack: well I was trying to make the minimal change so as not to change the rest of the code that calls this, rather than refactoring everything. Personally I think `logrus` and `errors` and all these other libraries trying to improve the standard library are actually worse for almost all use cases, and wrapping errors is rarely useful...

pkg/containerd/cmd/service/system_init.go
justincormack: I don't believe containerd outputs anything to stdout - I was originally sending stdout but it was silent, and I discovered that it is only logging to stderr (which seems correct), and so I set it up like this.

pkg/containerd/cmd/service/start.go
ijc: ok

justincormack: (but yes these returned arguments are a bit weird)

pkg/containerd/cmd/service/system_init.go
ijc: It might not now but who knows what it will do in the future, this isn't the sort of thing someone is likely to consider updating when bumping the containerd commit.

What's the harm of propagating stdout?

justincormack: well I hope it never does output on stdout, as that would be most incorrect, but I guess it does no harm changing it...

justincormack:
Added hash updates

pkg/containerd/cmd/service/start.go
rn: the return is indeed quite ugly. could we package the values in a struct and return `nil` and `err` on error?

justincormack: well, the string is part of error because of how `logrus` is returning errors as it wants `err` and the message seperately, I might just get rid of `logrus` to clean it up...

rn: IIRC I added logrus to have configurable verbosity which the default `log` package did not offer.

justincormack: service startup doesnt really need configurable verbosity (this isnt the linuxkit tool itself)

ijc: I just cribbed the basic command framework from linuxkit, taking the logging infra wasn't really a conscious decision.

Even with logrus you can do `"creating containerd client: " + err.Error()` if you like, the `WithError` stuff just works nicely with structured logging, but its not mandatory.

PR #2297, opened 2017-07-26
pkg/containerd: Fix the type conflict issue in containerd pkg

arm64b:
Since 'https://github.com/containerd/containerd/pull/1141' has been
merged into containerd git repo, so we don't need the below repo
'https://github.com/ijc/containerd.git' as the containerd git tree in
that pkg. Moreover, with the 'ijc' containerd git tree, we'll have
below error message when build the pkg/containerd:

fatal: reference is not a tree: d42cb88ba2b08d2ca6c8c017d629b394bf1dd08c
The command '/bin/sh -c git checkout $CONTAINERD_COMMIT' returned a
non-zero code: 128.

Except that above issue, we need to update the "github.com/Sirupsen/logrus"
pkg imported, because the latest code has change the 'Sirupsen' to 'sirupsen',
else the build process will complain that that pkg can't be found.

We also need to update the caller of NewTask(.) adapting to the latest
implementation of NewTask() to avoid the type conflict issue.

**- What I did**
Fix the build issue for pkg/containerd with `$make ORG=arm64b NOTRUST=1`
**- How I did it**
1. Update the CONTAINERD_REPO point to the https://github.com/containerd/containerd.git
and CONTAINERD_COMMIT to the latest commit.
2.  Change the "github.com/Sirupsen/logrus" pkg to "github.com/sirupsen/logrus"
3.  Update the parameters when calling NewTask() since this function implementation changed.
**- How to verify it**
Build the pkg/containerd separately.

summit

### DIFF
--- a/pkg/containerd/cmd/service/start.go
+++ b/pkg/containerd/cmd/service/start.go
@@ -9,10 +9,10 @@ import (
 	"os"
 	"path/filepath"
 
-	log "github.com/Sirupsen/logrus"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/namespaces"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+	log "github.com/sirupsen/logrus"
 )
 
 func start(args []string) {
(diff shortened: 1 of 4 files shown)


ijc:
I had been intending to defer this bump to the next containerd alpha release. Is there a pressing need to update to master sooner? You mention `the type conflict issue` in the subject, but that appears to be something which needs to be fixed due to the bump rather than the _reason_ for the bump itself.

I think the issue with d42cb88ba2b08d2ca6c8c017d629b394bf1dd08c not being available has been resolved, it happened because I deleted my branch when my PR was merged, but I've repushed to a `linuxkit` branch to keep it live until we update to the next containerd release.

pkg/containerd/cmd/service/start.go
```diff
@@ -84,21 +84,25 @@ func start(args []string) {
...
+	}
+
+	// io.Writer for both normal and error output upon logfile
+	logOE, err := os.Open(logfile)
+	if err != nil {
+		log.WithError(err).Fatal("failed to open log file")
 	}
 
-	task, err := ctr.NewTask(ctx, io)
+	task, err := ctr.NewTask(ctx, containerd.NewIO(logR, logOE, logOE))
```
ijc: When we do update IMHO we should just update the type of the existing function closure instead of rewriting like this. The new `id string` parameter can be ignored in this context.

arm64b: I'd like to say that no special pressing need to update the containerd git to the master, I just found that the 'fatal: reference is not a tree: d42cb88ba2b08d2ca6c8c017d629b394bf1dd08c' issue and chase down to those stuff. I will double check this if it's resolved when I am back to office tomorrow.

justincormack: I had the same issue yesterday but it is working for me now.

justincormack:
Will close for now as believed resolved; please reopen if not.

PR #2299, opened 2017-07-26
Switch runc and containerd startup to be entirely Go

### LINKED ISSUE #2295: dhcpcd expects resolv.conf to exist
```
/usr/lib/dhcpcd/dhcpcd-run-hooks: /usr/lib/dhcpcd/dhcpcd-hooks/20-resolv.conf: line 204: can't create /etc/resolv.conf: nonexistent directory
chmod: /etc/resolv.conf: No such file or directory
eth0: waiting for carrier
/usr/lib/dhcpcd/dhcpcd-run-hooks: /usr/lib/dhcpcd/dhcpcd-hooks/20-resolv.conf: line 1: can't create /etc/resolv.conf: nonexistent directory
chmod: /etc/resolv.conf: No such file or directory
```

does not seem fatal, but best to create empty file.

justincormack:
At present they use a small shared function called "prepare"
that does the read-write remounts, that I will switch to doing overlay
mounts soon.

also fix #2295 while I was modifying init to fix another issue.

traincat


ijc:
LGTM.

NB `prepare` is duplicated rather than shared.

justincormack:
Yes, I could not work out a reasonable way to share `prepare` between the two, I will add a note at the top saying pleas update in both places.

justincormack:
Pushed hashes (and added aforementioned note)

PR #2301, opened 2017-07-26
Various OVMF related fixes

rn:
- Add the `ovmf` package to the mirror instead of installing it to the `linuxkit/alpine` rootfs. This has the added benefit of the version of the package being recorded.
- Install the `ovmf` package in `tools/qemu`.
- Update the `linuxkit run qemu` code to use the new image
- While working on this, I also noticed that, when run in containerised mode, we would always bind mount the FW into the container. Clearly we'd want to use the FW in the container by default. Now, it only bind mounts the FW if the location specified via `-fw` is not the default.

Note that when run on Docker for Mac in containerised mode qemu currently segfaults when using the efi boot. I doubt this is a regression introduced by theses changes and the EFI image boots fine on Hyper-V.

bluebird

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -18,7 +18,10 @@ import (
 )
 
 // QemuImg is the version of qemu container
-const QemuImg = "linuxkit/qemu:8c07b24790ac5162dfc129791f8afeace159ca20"
+const (
+	QemuImg       = "linuxkit/qemu:8c07b24790ac5162dfc129791f8afeace159ca20"
+	defaultFWPath = "/usr/share/ovmf/bios.bin"
+)
 
 // QemuConfig contains the config for Qemu
 type QemuConfig struct {
@@ -114,7 +117,8 @@ func runQemu(args []string) {
 	data := flags.String("data", "", "Metadata to pass to VM (either a path to a file or a string)")
 
 	// Paths and settings for UEFI firware
-	fw := flags.String("fw", "/usr/share/ovmf/bios.bin", "Path to OVMF firmware for UEFI boot")
+	// Note, we do not use defaultFWPath here as we have a special case for containerised execution
+	fw := flags.String("fw", "", "Path to OVMF firmware for UEFI boot")
 
 	// VM configuration
 	enableKVM := flags.Bool("kvm", haveKVM(), "Enable KVM acceleration")
@@ -326,6 +330,9 @@ func runQemuLocal(config QemuConfig) error {
 
 	// Check for OVMF firmware before running
 	if config.UEFI {
+		if config.FWPath == "" {
+			config.FWPath = defaultFWPath
+		}
 		if _, err := os.Stat(config.FWPath); err != nil {
 			if os.IsNotExist(err) {
 				return fmt.Errorf("File [%s] does not exist, please ensure OVMF is installed", config.FWPath)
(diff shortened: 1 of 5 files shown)


tools/alpine/Dockerfile
```diff
@@ -16,6 +16,10 @@ RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \
 # add the tools for WireGuard, since the kernel module is now included, but from edge/testing
 RUN apk fetch --recursive -o /mirror/$(uname -m) -X http://dl-cdn.alpinelinux.org/alpine/edge/testing -U wireguard-tools
 
+# add OVMF for EFI boot (x86 only)
+RUN [ $(uname -m) = x86_64 ] && \
+    apk fetch --recursive -o /mirror/$(uname -m) -X http://dl-cdn.alpinelinux.org/alpine/edge/community -U ovmf
```
ijc: https://pkgs.alpinelinux.org/packages?name=ovmf&branch=&repo=&arch=&maintainer= suggests that the alpine 3.6 version would suffice, I suspect the use of edge is from pre-3.6 days?

src/cmd/linuxkit/run_qemu.go
```diff
@@ -375,10 +378,12 @@ func runQemuContainer(config QemuConfig) error {
...
 
-	// if user specify the "-fw" parameter, this should override the default in container context,
-	// with "-v" option, we will have the chance to assign an external FW binary to the containerized qemu
-	// instead of the fixed FW bin instealled by the build process of the image.
-	if config.UEFI {
+	// If we are running in a container and if the the user
+	// specifies the "-fw" parameter to a non-default, then we
+	// assume that the FW binary is on the local filesystem and
+	// bind mount it into the container. Otherwise, the FW image
+	// in the container is used.
```
ijc: This means I'm out of luck if I want to use my host's `/usr/share/ovmf/bios.bin`?

Might be better to default to `""` and use that to detect if the user tried to override vs wanting the default?

tools/alpine/Dockerfile
justincormack: Yes I think thats true. It is still in community, hence why we need to fetch it.

src/cmd/linuxkit/run_qemu.go
rn: I don't really want to default to `""` as this is a bit tedious for users. I would think that using the systems default `/usr/share/ovmf/bios.bin` is far less common. Or would you object?

tools/alpine/Dockerfile
rn: `docker run --rm alpine:3.6 apk add --no-cache ovmf` works, so I should be able to just add it to `packages.x86_64`

src/cmd/linuxkit/run_qemu.go
ijc: I meant that if the setting is `""` then use `/usr/share/ovmf/bios.bin` from within the container, so the user just doesn't give the `-fw` option if that is what they want, i.e. the common case.

If they give `-fw /usr/share/ovmf/bios.bin` then that means the _host_ version of that file.

tools/alpine/Dockerfile
justincormack: I thought we explictly didnt have community enabled (although it probably makes no sense as we have our own package cache now)

rn: looks like community is enabled by default:
```
docker run --rm -ti alpine:3.6 cat /etc/apk/repositories
http://dl-cdn.alpinelinux.org/alpine/v3.6/main
http://dl-cdn.alpinelinux.org/alpine/v3.6/community
```
and we don't explicitly disable it.

Should we?

src/cmd/linuxkit/run_qemu.go
rn: IC, that makes sense. Just need to bang in a couple of conditionals in the right places.

tools/alpine/Dockerfile
justincormack: I think the reason we used to was just to make sure everything we were using was actually supported, being updated, patched etc. We should probably just ask @ncopa to move `ovmf` to `main`.

rn: `go` and `tini` also come from community so we have to use it

ijc: Perhaps we should do
```
apk fetch --recursive -o /mirror/$(uname -m) -X http://dl-cdn.alpinelinux.org/alpine/edge/community -U $(cat packages.community)
```
and list them there instead of in the main `packages`, so at least we are explicit about what we are willing to take from community?

justincormack: hmm, we should definitely try to get `go` in `main`.

rn:
force pushed an update. I have now changed to use `ovmf` via `packages.x86_64` as we already use other bits from community. Also changed the logic for FW selection as @ijc suggested (I think)

ijc:
LGTM.

PR #2302, opened 2017-07-26
linuxkit: Better GCP disk handling

dave-tucker:
This commit allows the GCP backend to use the familiar `-disk` behaviour
that the local hypervisors use. The `file` attribute is used as the disk
name in GCP. The size is converted to GB and is always > 1GB.
This has the benefit of allowing multiple disks to be used with GCP
instances.

### DIFF
--- a/src/cmd/linuxkit/gcp.go
+++ b/src/cmd/linuxkit/gcp.go
@@ -178,7 +178,7 @@ func (g GCPClient) DeleteImage(name string) error {
 }
 
 // CreateInstance creates and starts an instance on GCP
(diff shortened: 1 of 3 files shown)


justincormack:
It seems weird to use `path` as `name` it might be better to add a different (optional) `name` field? I would expect `path` to actually create a disk with the contents of the file. But maybe if this was just betetr documented it might be ok...

dave-tucker:
> it might be better to add a different (optional) name field? 

I did consider that... but then again, what would `name` mean in the qemu case?
If you supplied `name` and not `path` would it create `${name}.{extension}` in the current directory?

Issue #2307, opened 2017-07-27
Initialise persistent disk post formatting


ijc:
It's not that uncommon to want to initialise some data (typically directories) on the persistent disk after formatting, for example in `projects/kubernetes` there is a [fairly crude hack](https://github.com/linuxkit/linuxkit/blob/master/projects/kubernetes/kube-master.yml#L26...L30) to ensure `/var/lib/kubeadm` exists after putting the persistent disk on `/var/lib` (solving this properly is a bullet in #2236)  and the `docker-for-mac` blueprint has [something similar](https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac/base.yml#L26..L31).

A while back I wrote a [quick patch](https://gist.github.com/ijc/c985ab61ea8d113385156fb64bdd1a03), which I subsequently discarded, which simply added a list of directories to be created to the arguments of the mount container.

That approach seemed a bit inflexible and I'm not really sure mount (rather than format) is the right place to do this, since mount happens on every boot while format is a one shot operation (for a given disk/filesystem) which better suites initialisation. The problem with format is that writing to the filesystem requires the fs to be mounted, so doing it on mount is certainly more convenient, but format could also do a temporary mount. Or maybe there should be a third onboot image to do this stuff.

I also wonder if it might be more useful to be able to do something more sophisticated, like unpacking a tarball (or image?) into the newly created filesystem.

/cc @djs55 @MagnusS for the d4m case. /cc @dave-tucker who has been involved with format+mount recently.

justincormack:
I am thinking about adding a "runtime config" section to the config, which would allow `mkdir -p` (on writeable paths) before a service runs. This config is needed for stuff I am planning on pinning network namespaces and moving interfaces into containers, which has to be done at runtime, which is another reason why I am rewriting the startup in Go right now...

It would look something like
```
name: kubernetes
image: blah:
runtime:
  - mkdir
      - /var/log/kubernetes
      - /var/lib/kubeadm
  - interface: eth0
```

ijc:
Assuming that happens after the bind mounts are configured then that would work.

justincormack:
Hmm, well I was thinking of the use case where the bind mounts will fail if directories are not created too, so I was thinking of before, and that these would be absolute root namespace paths. Would that not work?

ijc:
Ah, yes, because `/var/lib/kubeadm` is also a root path, I'd gotten myself confused there.

justincormack:
[...]

PR #2309, opened 2017-07-27
Use overlay for writeable containers

justincormack:
Previously we would sneakily remount as `rw` but of course you can't
really do that on a truly immutable filesystem.

See https://github.com/moby/tool/pull/129 for the `moby` side.

See #2288

dog-in-bed

### DIFF
--- a/pkg/containerd/cmd/service/prepare.go
+++ b/pkg/containerd/cmd/service/prepare.go
@@ -4,18 +4,84 @@ package main
 // Update it in both places if you make changes
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"syscall"
 )
 
 func prepare(path string) error {
+	// see if we are dealing with a read only or read write container
+	if _, err := os.Stat(filepath.Join(path, "lower")); err != nil {
+		if os.IsNotExist(err) {
+			return prepareRO(path)
+		}
+		return err
+	}
+	return prepareRW(path)
+}
+
+func prepareRO(path string) error {
+	// make rootfs a mount point, as runc doesn't like it much otherwise
 	rootfs := filepath.Join(path, "rootfs")
 	if err := syscall.Mount(rootfs, rootfs, "", syscall.MS_BIND, ""); err != nil {
 		return err
 	}
-	// remount rw
-	if err := syscall.Mount("", rootfs, "", syscall.MS_REMOUNT, ""); err != nil {
+	return nil
+}
+
+func prepareRW(path string) error {
+	// mount a tmpfs on tmp for upper and workdirs
+	// make it private as nothing else should be using this
+	tmp := filepath.Join(path, "tmp")
+	if err := syscall.Mount("tmpfs", tmp, "tmpfs", 0, "size=10%"); err != nil {
+		return err
+	}
+	// make it private as nothing else should be using this
+	if err := syscall.Mount("", tmp, "", syscall.MS_REMOUNT|syscall.MS_PRIVATE, ""); err != nil {
+		return err
+	}
+	upper := filepath.Join(tmp, "upper")
+	// make the mount points
+	if err := os.Mkdir(upper, 0744); err != nil {
+		return err
+	}
+	work := filepath.Join(tmp, "work")
+	if err := os.Mkdir(work, 0744); err != nil {
+		return err
+	}
+	lower := filepath.Join(path, "lower")
+	rootfs := filepath.Join(path, "rootfs")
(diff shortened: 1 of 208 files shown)


justincormack:
Needs `moby` update and hashes.

pkg/containerd/cmd/service/prepare.go
```diff
@@ -4,18 +4,79 @@ package main
...
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"syscall"
 )
 
 func prepare(path string) error {
+	// see if we are dealing with a read only or read write container
+	if _, err := os.Stat(filepath.Join(path, "lower")); err != nil {
```
ijc: Strictly this ought to check for `ENOTEXIST` and report any other error, but meh.

justincormack: yeah, but if root can't stat it, then it may as well not exist...

pkg/containerd/cmd/service/prepare.go
```diff
@@ -4,18 +4,79 @@ package main
...
-	// remount rw
-	if err := syscall.Mount("", rootfs, "", syscall.MS_REMOUNT, ""); err != nil {
+	return nil
+}
+
+func prepareRW(path string) error {
+	// mount a tmpfs on tmp for upper and workdirs
+	// make it private as nothing else should be using this
+	tmp := filepath.Join(path, "tmp")
+	if err := syscall.Mount("tmpfs", tmp, "tmpfs", 0, "size=10%"); err != nil {
```
ijc: The hardcoding of 10% is a bit of a shame.

What about this: have moby tool write a list of mounts (as a JSON array) into some file (lets say `rootfs.json`) and this tool reads it and passes it (perhaps after API conversion) to `WithRootFS` when calling `container.NewTask`? Then the tool has some control over all this and we can more easily expose via the yml files etc?

pkg/containerd/cmd/service/prepare.go
```diff
@@ -4,18 +4,79 @@ package main
...
+		return err
+	}
+	// make it private as nothing else should be using this
+	if err := syscall.Mount("", tmp, "", syscall.MS_REMOUNT|syscall.MS_PRIVATE, ""); err != nil {
```
ijc: What (if any) effect will this have on `rootfsPropagation` settings on the services?

pkg/containerd/cmd/service/prepare.go
ijc: It'd be nice to know about this when diagnosing why ones system isn't booting right though

pkg/containerd/cmd/service/prepare.go
justincormack: Well, maybe when we have something like the runtime config discussed in https://github.com/linuxkit/linuxkit/issues/2307#issuecomment-318340226 then we can make the overlay config configurable as part of that, but I don't really want to (a) make this contingent on that design and (b) have two kinds of runtime config.

10% is huge, should not limit anything. Containers should absolutely not store data in their images.

pkg/containerd/cmd/service/prepare.go
justincormack: This is not the rootfs mount point, it is the tmpfs mount for the writeable overlay, so I dont believe it can have any effect.

pkg/containerd/cmd/service/prepare.go
justincormack: fixed

justincormack:
Updated with `moby` update, and hashes.

There is one effect of this, in the Kubernetes build at present we delete a bunch of packages after use, so free up memory. This will no longer work, as they will just be deleted in the overlay layer not in the root filesystem. However I have different plans around this, and it should not have been relied on.

This also means that in theory we can now run multiple containers from the same image as they do not write on each other, although there is no config for this yet.

justincormack:
CI seems to have hung, possibly when it was updated?

justincormack:
Ah, CI is actually hanging, I need to update the `mkimage` stuff in tool I think...

justincormack:
Updated `moby` with fixes for read only containers defined in labels https://github.com/moby/tool/pull/132

justincormack:
Some of the tests were making incorrect assumptions about read-only, so added fixes.

justincormack:
Ok, hoping it might pass tests now.

I had to disable `containerd` test temporarily as part was failing after I moved it to an external disk, will invetogate later cc @AkihiroSuda

justincormack:
Ok this is green now!

PR #2312, opened 2017-07-27
Fixed commands and resolv.conf in docker example

mbentley:
Signed-off-by: Matt Bentley <matt.bentley@docker.com>

**- What I did**
Changed command for docker to run and added a bind mount for resolv.conf because DNS is great - remembering all of those IPs is really hard.  Fixed for both the docker example and the static & dynamic compose project examples

**- How I did it**
vim :)

**- How to verify it**
moby build foo.yml
linuxkit run foo

Updated commands as the path for docker changed and added resolv.conf to bind mounts

image

### DIFF
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -38,6 +38,7 @@ services:
      - type: cgroup
        options: ["rw","nosuid","noexec","nodev","relatime"]
     binds:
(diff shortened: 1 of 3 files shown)


justincormack:
Thanks!

PR #2313, opened 2017-07-28
package.mk: Adjust the targe sequence to avoid early abort

arm64b:
Current package.mk will operate on the first dependence object
'check-dirty', which result in the make process abort earlier
with below message:
pkg/package.mk:60: *** Your repository is not clean. Will not push package image.  Stop.

This is not expected behavior since we have no chance to build
a docker image locally in case of the user has no intention to
push that image.

This patch adjust the dependece order for the those targets, thus
we can build the docker image locally but can't push that image
if we're in a dirty git repository.

**- What I did**
make the docker image build can succeed locally but the image can't be pushed for reason of dirty git repository 
**- How I did it**
change the dependence order for the target: build first, then check the git repo is dirty or not
**- How to verify it**
make a individual pkg like this `make ORG=mainline NOTRUST=1`

summit

### DIFF
--- a/pkg/package.mk
+++ b/pkg/package.mk
@@ -60,14 +60,14 @@ ifneq ($(DIRTY),)
 	$(error Your repository is not clean. Will not push package image)
 endif
 
-push: check-dirty tag
+push: tag check-dirty
 	docker pull $(TAG) || docker push $(TAG)
 ifneq ($(RELEASE),)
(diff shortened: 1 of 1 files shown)


ijc:
If you want top build locally but not try and push why don't you `make tag` or `make forcetag` rather than relying on the dependencies of the `push` targets to do the build.

The reason for having the `push` do the dirty check early is so that if you want to actually do a push it will tell you before you waste time building everything only to find it is dirty and can't be pushed.

PR #2317, opened 2017-07-28
Remove kernel.random.write_wakeup_threshold from sysctl

justincormack:
The value we set is the default anyway.

racoons


rn:
Hmm, failed in trusted pull of redis. hit rebuild

justincormack:
Hmm, this doesnt seem to be default, going to revert in #2331

PR #2320, opened 2017-07-28
allow passing arguments directly to qemu

zacheryph:
**- What I did**

Added a `-args` and overriding `LINUXKIT_QEMU_ARGS` environment variable to the qemu runner

**- How I did it**

mimicked current flag/environment handling to add this, and append the flags within `buildQemuCmdline`

**- How to verify it**

see addition of `-args` in `linuxkit run qemu -help`. simplest validation being to run linuxkit image doing `-args "-badflag"` or with `LINUXKIT_QEMU_ARGS=-badflag` and qemu will fail nothing the bad flag being added as the arguments being sent to it.

* allow passing custom additional arguments directly to qemu

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -40,6 +40,7 @@ type QemuConfig struct {
 	Containerized  bool
 	QemuBinPath    string
 	QemuImgPath    string
+	QemuArgs       []string
 	PublishedPorts []string
 	NetdevConfig   string
 	UUID           uuid.UUID
@@ -71,6 +72,15 @@ func envOverrideBool(env string, b *bool) {
 	}
 }
 
+func envOverrideString(env string, s *string) {
+	val := os.Getenv(env)
+	if val == "" {
+		return
+	}
+
+	*s = val
+}
+
 func generateMAC() net.HardwareAddr {
 	mac := make([]byte, 6)
 	n, err := rand.Read(mac)
@@ -99,6 +109,9 @@ func runQemu(args []string) {
 		fmt.Printf("setuid network helper and appropriate host configuration, see\n")
 		fmt.Printf("http://wiki.qemu.org/Features/HelperNetworking.\n")
 	}
+
+	// Arguments to pass to qemu
+	qemuDirectArgs := flags.String("args", "", "Arguments to pass directly to qemu")
 
 	// Display flags
 	enableGUI := flags.Bool("gui", false, "Set qemu to use video output instead of stdio")
@@ -145,6 +158,7 @@ func runQemu(args []string) {
 
 	// These envvars override the corresponding command line
 	// options. So this must remain after the `flags.Parse` above.
+	envOverrideString("LINUXKIT_QEMU_ARGS", qemuDirectArgs)
 	envOverrideBool("LINUXKIT_QEMU_KVM", enableKVM)
 	envOverrideBool("LINUXKIT_QEMU_CONTAINERIZED", qemuContainerized)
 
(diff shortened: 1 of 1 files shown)


zacheryph:
I used `-args`.. wasnt totally sure what to call it. `-qemuFlags` maybe? can be changed easily enough though.

justincormack:
I think I prefer something with `qemu` in the name, so it is clearly non portable.

justincormack:
I think the CI failures may well be related to your change.

zacheryph:
I updated option to be `-qemuOptions` and `LINUXKIT_QEMU_OPTIONS`.  I also rebased master for the sake of purity.

As for the failed tests I am not totally sure what is going on here? All `linuxkit.platform.qemu.*` tests pass locally (asside from SKIP/CANCEL) and everything before as well.  I get package failings locally though with no clear reason why if i do -v=3.

justincormack:
It is not immediately obvious why it should fail. Can you rebase as there have been a bunch of upstream changes.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -289,6 +303,7 @@ func runQemu(args []string) {
 		Memory:         *mem,
 		KVM:            *enableKVM,
 		Containerized:  *qemuContainerized,
+		QemuOptions:    strings.Split(*qemuDirectOptions, " "),
```
ijc: This won't allow any QEMU options which happen to contain a space. Maybe that is not an issue in practice? An alternative would be to allow the option to be given multiple times to populate a slice.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -147,6 +160,7 @@ func runQemu(args []string) {
 	// options. So this must remain after the `flags.Parse` above.
 	envOverrideBool("LINUXKIT_QEMU_KVM", enableKVM)
 	envOverrideBool("LINUXKIT_QEMU_CONTAINERIZED", qemuContainerized)
+	envOverrideString("LINUXKIT_QEMU_OPTIONS", qemuDirectOptions)
```
ijc: The original two were added to allow for quick/dirty performance testing of this stuff when linuxkit is used (internally) by the moby tool (we were seeing slowness on CI, where KVM is not an option).

What is the rationale for allowing arbitrary command line overrides via that mechanism?

ijc:
What is your actual use case, i.e. what options do you actually want to add. We should consider whether or not they should be added as first class options.

One downside of this change would be that in the future people would just use this without exposing their requirements to us which makes it harder for us to know what first class options we should be adding. I suppose that it unavoidable.

zacheryph:
@ijc right now the use case is purely to allow for `-pidfile {filename}`.  I'm am playing with the `projects/kubernetes` and writing a script that involves automating the the setup of the cluster. Although after opening this I thought about writing a more full featured tool that would continue to run in the background.

So the current only use case is for -pidfile to be able to run linuxkit / qemu in the background. Though maybe I should just be concerned with the linuxkit pid and it will handle killing qemu if its given the proper signal?

justincormack:
We could definitely add a global `pidfile` option to `linuxkit`, would seem to make sense.

ijc:
IIRC the hyperkit backend will put a pid file into the state dir, it might not be a bad idea for the qemu backend to do the same (overideable by a global flag if we like)

rn:
I have plans to have a `~/.linuxkit` directory to stash per VM state and the offer backgrounding, list, kill etc subcommands for `linuxkit`. very tentative at the moment and haven't started work on it so it might be a while. Also don't want to re-invent `libvirt` and friends...

zacheryph:
I can look at submitting a PR for storing the pidfile into the state directory.  Does it make sense to store the Qemu pid or linuxkit itself.  I suppose I should look into how signal handling works.  I'll get a better idea for a solution on what my intentions are and create an issue for discussion.

Thanks :)

ijc:
> Does it make sense to store the Qemu pid or linuxkit itself

The hyperkit backend stores the PID of hyperkit, I think that makes sense and the QEMU backend should follow suite and store the QEMU pid.

PR #2326, opened 2017-07-31
pkg/getty: Add ARM 'ttyAMA0' console into securetty

arm64b:
We need to add 'ttyAMA0' console used on ARM64 platform into
securetty file to make it's possible to log into the system
as root. Also it will dismiss the below warning message before
login:
"getty: cmdline has console=ttyAMA0 but does not exist in
/etc/securetty; will not be able to log in as root on this tty ttyAMA0."

**- What I did**
 Dismiss the console warning message before login.
**- How I did it**
Add the 'ttyAMA0' used on ARM64 into /etc/securetty
**- How to verify it**
The warning message is disappeared after this patch applied. 

summit

### DIFF
--- a/pkg/getty/etc/securetty
+++ b/pkg/getty/etc/securetty
@@ -28,6 +28,8 @@ hvsi1
 # ARM "AMBA" serial ports
 ttyAM0
 ttyAM1
+ttyAMA0
+ttyAMA1
 
 # s390 and s390x ports in LPAR mode
 ttysclp0


justincormack:
Does it make sense to add  `ttyAMA1` as well?

arm64b:
@justincormack  yes, make sense. it doesn't hurt the expected behavior and provide more choices. it will be better if it support some kinds of wildcard like ttyAMA*, adding ttyAMA1 will mitigate this condition in some degree. Will re-push a revision one.

PR #2331, opened 2017-07-31
Replace rngd with a Go version

### LINKED ISSUE #1247: initialise rng properly
We need a one shot rng initialisation before the daemon scripts start, as well as running rngd later, in case the `onboot` scripts need entropy.

justincormack:
Only supports the use cases we currently need, currently support
for using Intel hardware rng to initialise and add entropy.

Supports oneshot and service mode. Call as `rngd -1` for one shot
mode.

fix #1247

panda1


justincormack:
The other functionality in the rng-tools package is to use other external rngs, in particular the tpm one, but these can be easily added if necessary.

rn:
how do we know this is working as expected?

justincormack:
@rn we don't have any way of testing with `qemu` but I have tested it quite extensively in one shot and service modes on machines I have with `rdrand` support - I will also try to dig out a machine with `rdseed` support.

justincormack:
Tested with `rdseed`, disabling `rdrand` and that is working too. Going to revery #2317 though, as that value was not the default it seems.

justincormack:
It is difficult to test on linuxkit but have tested the binary on various platforms, in d4m etc. You can drain entropy and `strace` the binary to see it is adding it back.

deitch:
@justincormack given the loading and compilation, does it make sense to turn this into a standard package in alpine linux, and then this `pkg/rngd` could be nothing more than `apk add rngd`?

I have been thinking the same for the tpm tools (but haven't gotten around to it, and am not familiar with the process of making an alpine package part of the standard distro).

justincormack:
@deitch I don't think it would be accepted upstream until (a) it is not written in Go, as Go still has limited portability and still generates rather large binaries; we have a plan to rewrite in Rust which should fix both of these, and I think it is easier to do that form simple Go code (b) it does still have reduced functionality, it is mainly that it adds specific functionality that we want (c) could do with better tests, `rngd` has a fips randomness test, that is kind of useless in my experience as they randomly fail, but some platform tests would be more useful it is just that we dont have test coverage on platforms this is useful yet. But eventually, yes.

deitch:
> it is not written in Go... But eventually, yes.

OK.

Issue #2131, opened 2017-06-29
projects/kubernetes: kubelet container mounts not visible to dockerd container


btheado:
**Description**

Commit 0b70e7e3a79585d5b24bae6c9c731ba21d40c4c2

When I start the kubernetes master from projects/kubernetes and then run kubeadm-init.sh, neither of the kube-proxy or weave-net pods come up. In both the pod's logs there are errors complaining about files missing in /var/run/secrets/kubernetes.io/serviceaccount.

It looks to me like the kubelet is creating those files in the kubelet container, but the dockerd container is not able to see those same files.

If I go back to an earlier commit (e0bcac3a) from over a month ago, both kube-proxy and weave-net can come up fine and when I look in the dockerd container, it can see the same kubernetes secrets mounts as in the kublet container.

**Steps to reproduce the issue:**

1. `cd projects/kubernetes`
1. `make build-vm-images`
1. `./boot.sh`
1. `ctr exec --tty kubelet ash`
1. `kubeadm-init.sh`
1. `kubectl --namespace=kube-system get pods`
1. Repeat step 6 until the kube-proxy and weave-net pods to go to Error or CrashLoopBackOff state
1. `kubectl --namespace=kube-system logs kube-proxy-7wdtn`  ;# Replace with your actual value from step 6
1. Look for a complaint in the output about `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt: no such file or directory`
1. `kubectl --namespace=kube-system logs weave-net-wqt0m weave`  ;# Replace with your actual value from step 6
1. Look for a complaint in the output about `/var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory`
1. `df -h | grep secret` and notice this kubelet container has two tmpfs mounts for each of weave-net and kube-proxy
1. `exit` ;# to leave the kubelet container
1. `ctr exec docker df -h | grep secret` and notice there is no output
1. `ctr exec docker docker ps -a` and look for the container id for a kube-proxy instance
1. `ctr exec docker docker inspect <containerid from previous step> | grep secret` and notice the source of the mount is the same path as the tmpfs from step 12. But step 13 shows those tmpfs mounts are missing from the dockerd container. This is why steps 9 and 11 are complaining about no such file or directory.

**Describe the results you received:**

Neither the weave-net nor the kube-proxy pods came up due to the missing secrets mount

**Describe the results you expected:**

The weave-net and the kube-proxy pods should have come up

[...]

PR #2334, opened 2017-08-01
projects/kubernetes: fixup and bump to kube 1.7.2

ijc:
**- What I did**

- Updated `getty` and `sshd` which were lagging.
- Fixes up some fallout from the switch to a `tmpfs` on `/var` and in particular the change to `/var/run` being a symlink to `/run`.
- Updated to Kube 1.7.2
- Reintroduced k8s-dns images in the cache, since they are now used.

**- How I did it**

Emacs

**- How to verify it**

I like the sock shop demo for this.

Kubernetes project is now based on Kube 1.7.2.

[](https://en.wikipedia.org/wiki/The_Congregation_(Leprous_album))

### DIFF
--- a/projects/kubernetes/kube-master.yml
+++ b/projects/kubernetes/kube-master.yml
@@ -59,11 +59,11 @@ services:
     rootfsPropagation: shared
     command: ["/usr/local/bin/docker-init", "/usr/local/bin/dockerd"]
   - name: kubernetes-image-cache-common
-    image: linuxkitprojects/kubernetes-image-cache-common:6fccda74ea301f9a62cdcfc2fe4952cff2c8c97b
+    image: linuxkitprojects/kubernetes-image-cache-common:ba16b1f8cfe4f415a5946d521e59f67eaeecd9ce
   - name: kubernetes-image-cache-control-plane
-    image: linuxkitprojects/kubernetes-image-cache-control-plane:6fccda74ea301f9a62cdcfc2fe4952cff2c8c97b
+    image: linuxkitprojects/kubernetes-image-cache-control-plane:ba16b1f8cfe4f415a5946d521e59f67eaeecd9ce
   - name: kubelet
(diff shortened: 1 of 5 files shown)


projects/kubernetes/kube-master.yml
```diff
@@ -51,18 +51,19 @@ services:
      - /dev:/dev
      - /etc/resolv.conf:/etc/resolv.conf
      - /lib/modules:/lib/modules
+     - /run:/run:rshared,rbind
```
justincormack: there is no real need for `/run` to be `rshared,rbind` as no one ever mounts anything in `/run` AFAIK

ijc: The problematic paths in #2131 were `/var/run/secrets/blahblah` which due to the symlink are now actually `/run/secrets/blahblah`. This path needs to exist in both the kubelet and docker containers.

I didn't actually test it but due to that I think having `/run` would break things. I could try it if you want though.

justincormack: Hmm, is `/var/run/secrets` actually a mount point? If it is not you shouldnt need the `rshared,rbind`, if it is then I guess so. Weird place to put a mountpoint if so...

ijc: Re-reviewing #2131 it might be there is a symlink to `/var/lib` in there somewhere. I'll try this out (probably tomorrow at this stage).

ijc: Ah, I got it, on the control plane side things are mounted/populated under `/var/lib/kubernetes/...` they are then bound into various runtime containers as `/var/run/secrets` which is why I was seeing that. But the runtime mount is a kube managed mount not something we need to be concerned with and I think there is indeed no need for `/run` to be shared by the control plane and the host.

I'll double check by actually making the change rather than just inspecting things, but expect an update.

Issue #2292, opened 2017-07-25
multiarch support for packages


rn:
We now have both x86_64 and arm64 `linuxkit/alpine` images as well as initial support for building arm64 kernels. So we are getting close to full arm64 support.

We want this support to be transparent, ie ideally we can use the same YAML file without any changes to build `x86_64` and `arm64` LinuxKit images (and other archs in the future). The best way to achieve this is using multiarch images/manifests, currently using the [`manifest-tool`](https://github.com/estesp/manifest-tool). The tool pushes a manifest to hub which contains references to to the different images (e.g. for different architectures). I suggest we add the `manifest-tool` the the top-level `Makefile` in `linuxkit`.

## Packages and kernel

Assuming a multiarch `linuxkit/alpine` (see below), for normal packages (and with some minor mods, for the kernel packages) we can use the command line options for `manifest-tool` as the hash is based on the git content hash, so it should be the same on `x86_64` and `arm64`. We need to modify the package build to first push/sign images with the architecture appended, so instead of, e.g.,
```
linuxkit/binfmt:0bde4ebd422099f45c5ee03217413523ad2223e5
```
we would push/sign:
```
linuxkit/binfmt:0bde4ebd422099f45c5ee03217413523ad2223e5-x86_64
linuxkit/binfmt:0bde4ebd422099f45c5ee03217413523ad2223e5-arm64
```
Then in a final step, we'd invoke:
```
DOCKER_CONTENT_TRUST=1 manifest-tool push from-args \
    --ignore-missing \
    --platforms linux/x86_64,linux/arm64 \
    --template linuxkit/binfmt:0bde4ebd422099f45c5ee03217413523ad2223e5-ARCH \
    --target linuxkit/binfmt:0bde4ebd422099f45c5ee03217413523ad2223e5
```
This should create a multiarch image on hub which can be used as before, except it should work both on `x86_64` and `arm64` provided that both images exist. The `--ignore-missing` should allow us to first run `make push` on one platform (which pushes a multiarch image with just one platform) and then, afterwards on the other (which then should repush the image with both architectures. The advantage of this approach is that it requires very little coordination between the architectures and that it can be applied pretty much to all our packages with just changes to `package.mk`.

## Base image

[...]

PR #2335, opened 2017-08-01
Create a multi-arch alpine base image

rn:
- Add a modified version of the `manifest-tool` to the top-level Makefile
- Append `-amd64` and `-arm64` to the architecture specific base images
- After pushing the image use `push-manifest.sh`, which:
  - Extracts the current hashes from the `version.*` files
  - Creates a `manifest-tool` YAML file
  - Push the manifest list to hub using the hash of the `x86_64` images (without the `-amd64`)
  - Sign the manifest using `notary`
- Note, the build requires a modified version of `manifest-tool` and `jq` to be installed. On macOS it also requires the `docker-credential-osxkeychain.bin` which is installed with Docker for Mac, and on Linux it requires `base64` (though that is currently not used).

There now is signed a multi-arch image on hub as: `linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630`

Next step is to add support for multi-arch images to `package.mk`, then rebuild the various packages and tools as well as the kernel package. I think that has to be done in one PR and can't be done incrementally.

wip: #2292

/cc @ijc @arm64b @justincormack @riyazdf I've tried to implement the `mirror` apprach suggested by @ijc [here](https://github.com/linuxkit/linuxkit/issues/2292#issuecomment-317997094) but one can't `apk fetch --arch $arch` at the moment.

sleep

### DIFF
--- a/tools/alpine/push-manifest.sh
+++ b/tools/alpine/push-manifest.sh
@@ -1,0 +1,71 @@
+#! /bin/sh
+
+# This script creates a multiarch manifest for the 'linuxkit/alpine'
+# image, pushes and signs it. The manifest is pushed with the tag of
+# the amd64 images (which is the suffix removed). On macOS we use the
+# credentials helper to extract the Hub credentials. We need to
+# manually sign the manifest using 'notary'.
+#
+# This script is specific to 'linuxkit/alpine'. For normal packages we
+# use a different scheme.
+#
+# This should all be replaced with 'docker manifest' once it lands.
+
+ORG=$1
+IMAGE=$2
+
+IMG_X86_64=$(head -1 versions.x86_64 | sed 's,[#| ]*,,')
+IMG_ARM64=$(head -1 versions.aarch64 | sed 's,[#| ]*,,')
+IMG_MANIFEST=$(echo "$IMG_X86_64" | sed 's,\-.*$,,')
+IMG_TAG=$(echo "$IMG_MANIFEST" | sed 's,.*:,,')
+
+YAML=$(mktemp)
+cat <<EOF > "$YAML"
+image: $IMG_MANIFEST
+manifests:
+  - image: $IMG_ARM64
+    platform:
+      architecture: arm64
+      os: linux
+  - image: $IMG_X86_64
+    platform:
+      architecture: amd64
+      os: linux
+EOF
+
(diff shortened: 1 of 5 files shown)


tools/alpine/versions.aarch64
```diff
@@ -87,14 +87,14 @@ libattr-2.4.47-r6
 libblkid-2.28.2-r2
 libburn-1.4.6-r0
 libbz2-1.0.6-r5
+libc-dev-0.7.1-r0
+libc-utils-0.7.1-r0
```
ijc: Hrm, looks like adding `LANG=C` to the sort hasn't resolved the instabilities here :-/

tools/alpine/push-manifest.sh
```diff
@@ -0,0 +1,71 @@
...
+esac
+
+# Push manifest list
+OUT=$(manifest-tool $MT_ARGS push from-spec "$YAML")
+rm "$YAML"
+echo "$OUT"
+SHA256=$(echo "$OUT" | cut -d' ' -f2 | cut -d':' -f2)
+LEN=$(echo "$OUT" | cut -d' ' -f3)
+
+# Sign manifest (TODO: Use $USERPASS and pass them into notary)
```
riyazdf: we could pass in as input, that's how we've been scripting notary today as there is no keychain/envvar support (see: https://github.com/docker/notary/blob/master/buildscripts/dockertest.py#L386-L389)

riyazdf: We do have an env var for delegation key passphrases: `NOTARY_DELEGATION_PASSPHRASE`

rn:
I'm merging this now, as I got two LGTM and I'd like to move forward to move the packages to multi-arch.

tools/alpine/push-manifest.sh
rn: I tried the input method but it didn't work. May have been my mistake. I'll retry

As for `NOTARY_DELEGATION_PASSPHRASE` I'll will set this based on the docker variable in a subsequent PR so that eventually CI can just set the related docker variable

tools/alpine/versions.aarch64
rn: yes, i tried that a while back as well and it didn't fix it. forgot to mention it...

PR #2337, opened 2017-08-01
save qemu pid in state directory

zacheryph:
**- What I did**

Made it so qemu stores pidfile within state directory the same way hyperkit does

**- How I did it**

* Unconditionally create state directory as its required no matter what for our pidfile
* pass `-pidfile $STATE_DIR/qemu.pid` to qemu to create pid file

**- How to verify it**

`./bin/linuxkit run linuxkit` (or any image) and `cat linuxkit-state/qemu.pid` (replace linuxkit-state if you specified `-state` to linuxkit)

Save qemu pid to file within state directory

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -32,6 +32,7 @@ type QemuConfig struct {
 	GUI            bool
 	Disks          Disks
 	MetadataPath   string
+	StatePath      string
 	FWPath         string
 	Arch           string
 	CPUs           string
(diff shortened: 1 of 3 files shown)


ijc:
Failure is:
```
[+ 34m 11s] [FAIL    ] linuxkit.packages.getty-containerd 68.56s
```
[`linuxkit.packages.getty-containerd.log`](https://github.com/linuxkit/linuxkit-logs/blob/e9ccae9621c75e75b1509db99c1d71fad14c8a84/value/1afee414-935c-4426-9813-7007fa789b74/linuxkit.packages.getty-containerd.log) says:
```
[STDERR  ] 2017-08-01T17:05:06.175295555Z: + find . -iname test-ctr* -not -iname *.yml -exec rm -rf {} ;
[STDERR ] 2017-08-01T17:05:06.176910606Z: find: `./test-ctr-state': No such file or directory
```
Which I think is a latent bug in that test which is exposed by the fact that `test-ctr-state` now exists (as a directory) where it did not before.

The patter used in most tests is:
```
find . -depth -iname "test-foo*" -not -iname "*.yml" -exec rm -rf {} \;
```
(with the `-depth` option). Looks like `test/cases/040_packages/003_containerd/test.sh` and `test/cases/040_packages/007_getty-containerd/test.sh` are the two exceptions, could you fix them up here please?

Not sure why `test/cases/040_packages/003_containerd/test.sh` didn't fail too, I guess it is just racey...

PR #2338, opened 2017-08-02
Initial support for launching instances on OpenStack

yankcrime:
Signed-off-by: Nick Jones <nick@dischord.org>

**- What I did**

Added basic support for launching instances on an OpenStack platform.

**- How I did it**

Made use of the Gophercloud library in order to handle creating an instance with a set of options such as flavor and which network to attach the virtual machine to.

**- How to verify it**

As an example:

```shell
linuxkit run openstack \
  -authurl https://compute.datacentred.io:5000/v3 \
  -username "username" \
  -password "xxx" \
  -project "default" \
  -network c5d02c5f-c625-4539-8aed-1dab3aa85a0a \
  -name LinuxKitTest
```

Or it'll inherit a standard set of [OpenStack client environment variables](https://docs.openstack.org/mitaka/install-guide-obs/keystone-openrc.html) for these options, simplifying the command to something such as:

```shell
linuxkit run openstack \                 
  -img-ID 394d9a54-fb08-4e59-8a39-b3ffa80dff97 \ 
  -network c5d02c5f-c625-4539-8aed-1dab3aa85a0a \
  -name LinuxKit
```

Provide basic support for launching instances on an OpenStack platform.

"Hey Nelly, I hear OpenStack is dying!"

calmowl

### DIFF
--- a/src/cmd/linuxkit/run.go
+++ b/src/cmd/linuxkit/run.go
@@ -22,6 +22,7 @@ func runUsage() {
 	fmt.Printf("  gcp\n")
 	fmt.Printf("  hyperkit [macOS]\n")
 	fmt.Printf("  hyperv [Windows]\n")
+	fmt.Printf("  openstack\n")
 	fmt.Printf("  packet\n")
 	fmt.Printf("  qemu [linux]\n")
 	fmt.Printf("  vcenter\n")
@@ -54,6 +55,8 @@ func run(args []string) {
 		runHyperKit(args[1:])
 	case "hyperv":
 		runHyperV(args[1:])
+	case "openstack":
+		runOpenStack(args[1:])
 	case "packet":
 		runPacket(args[1:])
 	case "qemu":
--- a/src/cmd/linuxkit/run_openstack.go
+++ b/src/cmd/linuxkit/run_openstack.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+)
+
+const (
+	defaultOSFlavor = "m1.tiny"
+	authurlVar      = "OS_AUTH_URL"
+	usernameVar     = "OS_USERNAME"
+	passwordVar     = "OS_PASSWORD"
+	projectNameVar  = "OS_PROJECT_NAME"
+	userDomainVar   = "OS_USER_DOMAIN_NAME"
+)
+
+func runOpenStack(args []string) {
+	flags := flag.NewFlagSet("openstack", flag.ExitOnError)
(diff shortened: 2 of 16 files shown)


justincormack:
Hmm, the test log indicated that the cross build failed, but without showing any error. Not sure why that would be either, there is no cgo code or anything as far as I can see...

justincormack:
Hmm, can you try `make test-cross` and see what the error is?

yankcrime:
The only error from the output of `make test-cross` is:

```
make[1]: *** [tmp_linuxkit_bin.tar] Error 1
make[1]: *** Waiting for unfinished jobs....
```
`make bin/linuxkit` fails with a similarly nondescript message:

```$ make bin/linuxkit
tar cf - -C src/cmd/linuxkit . | docker run --rm --net=none --log-driver=none -i -e GOOS=darwin linuxkit/go-compile:a1e3cd3a54b1ad683b555655282272d0eb27ee40 --package github.com/linuxkit/linuxkit --ldflags "-X main.GitCommit=66c9afedfac45f41944f47a9f6e085558c64675c -X main.Version="0.0" " -o bin/linuxkit > tmp_linuxkit_bin.tar
gofmt...
run_openstack.go
make: *** [tmp_linuxkit_bin.tar] Error 1
```

Doing `go build -v` within `src/cmd/linuxkit` compiles the binary quite happily, however, so I'm at a bit of a loss without doing more digging around.

justincormack:
Oh, it may just need you to run `go fmt` on your new code...

yankcrime:
If it only it were that simple - I run `go fmt` on every save so it's not that unfortunately...

```
$ go fmt src/cmd/linuxkit/run_openstack.go
$ echo $?                                 
0
$ make bin/linuxkit                       
tar cf - -C src/cmd/linuxkit . | docker run --rm --net=none --log-driver=none -i -e GOOS=darwin linuxkit/go-compile:a1e3cd3a54b1ad683b555655282272d0eb27ee40 --package github.com/linuxkit/linuxkit --ldflags "-X main.GitCommit=bf5f05b2acdbdc0b6905e453fe3631ff41e17d07 -X main.Version="0.0" " -o bin/linuxkit > tmp_linuxkit_bin.tar
gofmt...
run_openstack.go
make: *** [tmp_linuxkit_bin.tar] Error 1
```

justincormack:
hmm, I will lcone your branch in a bit, but I was a bit suspicious as it prints
```
gofmt...
run_openstack.go
```
which is the normal output from `go fmt` changing something...

The script also checks linting etc see https://github.com/linuxkit/linuxkit/blob/master/tools/alpine/go-compile.sh#L16

yankcrime:
It looks like it's the `-s` option to `gofmt`.  It wants to "simplify" line 72 of `run_openstack.go`:

```diff
72c72
< 		Networks:      []servers.Network{servers.Network{UUID: *networkID}},
---
> 		Networks:      []servers.Network{{UUID: *networkID}},
```

Which is actually invalid and so throws an error.

justincormack:
Ah I see you found a workaround... Probably worth filing a bug?

yankcrime:
@justincormack Yup will follow up, thanks for the help 👍

PR #2339, opened 2017-08-02
Unpack filesystem for BIOS ISO

justincormack:
Now that LinuxKit works on a truly immutable filesystem, for an
ISO it makes sense to use the ISO filesystem rather than using
an initrd. Only major difference is you may need to specify the
device for the root filesystem in the kernel command line, but
we set a sane default of `/dev/sr0` if unspecified.

cat-mummy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 PREFIX?=/usr/local/
 
 MOBY_REPO=https://github.com/moby/tool.git
-MOBY_COMMIT=14a4d923aef7cf5a8a6f162b128315d9e0f7884e
+MOBY_COMMIT=b9de5ff4c64e01d8fbaa495776890a82e9647f20
 MOBY_VERSION=0.0
 bin/moby: tmp_moby_bin.tar | bin
 	tar xf $<
--- a/tools/mkimage-iso-bios/make-iso
+++ b/tools/mkimage-iso-bios/make-iso
@@ -5,22 +5,17 @@ set -e
 mkdir -p /tmp/iso
 cd /tmp/iso
 
-# input is a tarball of kernel and initrd.img on stdin
+# input is a tarball of filesystem on stdin with kernel in /boot
 # output is an iso on stdout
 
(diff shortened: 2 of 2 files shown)


justincormack:
I will bump the `moby` version when https://github.com/moby/tool/pull/135 is merged before merging this.

justincormack:
I haven't changed the EFI ISO yet as there are some issues with testing it...

deitch:
Is there a concern with the structure being different in an ISO that elsewhere? I don't mean technically as much as maintenance and runtime debugging? I don't see it as a major issue, but I also am unsure what the benefit is of having the files laid out in the ISO vs a single file `initrd.img`?

rn:
@deitch we don't need to uncompress the initrd during boot so should be faster. Overall the system might be slower, though give that now all rootfs access go to the cdrom instead of the ram disk. @justincormack have you seen any issues with this?

justincormack:
The benefits are
(a) actually immutable, while with `initrd` an attacker can remount writeable.
(b) faster boot time as there is no unpack stage for unpacking the initramfs
(c) (the main reason I have been working on this) much lower memory usage for huge images, in particular Kubernetes will not boot in under 4G of RAM, large amounts of which are permanently going to be used to store the boot images, especially since as we now have properly immutable container images you cannot free up space. We are roughly throwing away 1.5GB RAM permanently booting Kubernetes at present, which is kind of expensive and unworkable. (Yes we can trim it down a bit, but it is a gigantic piece of software right now).
(d) there are other filesystems that it makes sense to unpack, for example `squashfs` for embedded devices is the most sensible option, and would have the same benefits, will not throw away precious RAM, and can be used in place.

deitch:
> much lower memory usage for huge images

That makes a lot of sense. I can see that.

justincormack:
So far I have only run it with `qemu` so it is a little hard to judge speed, but it seems ok. Going to test hyperkit next (we dont actually have an ISO option in `linuxkit run hyperkit` right now)

PR #2340, opened 2017-08-02
[WIP] Grand multi-arch conversion

rn:
This is WIP to get some early feedback before embarking on converting everything to multi-arch images.

This adds a script to push and sign manifest (similar to the one used for `linuxkit/alpine`), adds it to the generic package building makefile.

It then converts the the `linuxkit/go-compile` package to a multi-arch images and uses it in the top-level Makefile file to build the tools.

This works! and bootstraps the mechanism to convert all the other packages. It would be good to get some  early feedback on the integration before embarking on the grand conversion.

If the approach is ok, I'll close this and finish the work before submitting a proper PR

@justincormack PTAL

### DIFF
--- a/scripts/push-manifest.sh
+++ b/scripts/push-manifest.sh
@@ -1,40 +1,20 @@
 #! /bin/sh
 
-# This script creates a multiarch manifest for the 'linuxkit/alpine'
-# image, pushes and signs it. The manifest is pushed with the tag of
-# the amd64 images (which is the suffix removed). On macOS we use the
-# credentials helper to extract the Hub credentials. We need to
-# manually sign the manifest using 'notary'.
+# This script pushes a multiarch manifest for packages and signs it.
+# 
+# The TARGET must be of the form <org>/<image>:<tag> and this is what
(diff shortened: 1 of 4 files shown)


scripts/push-manifest.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Push manifest list
+OUT=$(manifest-tool $MT_ARGS push from-args \
+                    --ignore-missing \
+                    --platforms linux/amd64,linux/arm64 \
+                    --template $TARGET-ARCH \
+                    --target $TARGET)
+
+echo "$OUT"
+if [ -z ${TRUST+x} ]; then
```
justincormack: you can use `[ -z "$TRUST" ]` I think.

justincormack:
some minor comments but LGTM

scripts/push-manifest.sh
rn: yes, in this case that works.

rn:
Thanks for the review. Closing this as otherwise our CI would go crazy from all the force pushing back and force will building stuff on different machines

PR #2345, opened 2017-08-03
Azure metadata

bketelsen:
**- What I did**
I added Azure metadata support.

**- How I did it**
I copied the AWS Metadata implementation and modified it to use Azure endpoints, removing pieces that aren't supported by azure.

**- How to verify it**
I ran this in two different Azure vm's and they each provided correct output, manually verified.

Metadata: Provide Azure support.

dgql-igxuaa24yq

### DIFF
--- a/pkg/metadata/provider_azure.go
+++ b/pkg/metadata/provider_azure.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"time"
+)
+
+const (
+	azMetaDataURL     = "http://169.254.169.254/metadata/instance/"
+	azMetadataVersion = "2017-04-02"
+	azUserDataLoc     = "/var/lib/waagent/ovf-env.xml"
+)
+
+// ProviderAzure is the type implementing the Provider interface for Azure
+type ProviderAzure struct {
+}
+
+// NewAzure returns a new ProviderAzure
+func NewAzure() *ProviderAzure {
+	return &ProviderAzure{}
+}
+
+func (p *ProviderAzure) String() string {
+	return "Azure"
+}
+
+// Probe checks if we are running on Azure
+func (p *ProviderAzure) Probe() bool {
+	// Getting the hostname should always work...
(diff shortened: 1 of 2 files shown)


pkg/metadata/provider_azure.go
```diff
@@ -0,0 +1,125 @@
...
+		log.Printf("Azure: Failed to get user-data: %s", err)
+		// This is not an error
+		return nil, nil
+	}
+	return userData, nil
+}
+
+// getUserData returns the userdata stored on disk
+func getUserData() ([]byte, error) {
+	return ioutil.ReadFile(azUserDataLoc)
```
justincormack: I don't understand how this gets into `/var/lib/waagent/ovf-env.xml`? We don't really want the `metadata` package to need to run after `waagent` (ideally, I wouldnt want to run that at all, it is overly complicated). Can we just fetch this directly from a URL?

bketelsen: sadly there is no URL for fetching userdata.  We can either remove it or let it fail silently if it's not there as I wrote it.

justincormack: According to the docs the ovf file is actually on a DVD https://docs.microsoft.com/en-us/azure/virtual-machines/linux/agent-user-guide?toc=%2fazure%2fvirtual-machines%2flinux%2ftoc.json

(I really dislike the waagent, as it is huge and tries to do everything at once which is very complicated; we do currently run it on Docker for Azure, but I would like to replace its functions with standalone containers really, and so I don't think that eg this metadata package should require it to be running, indeed it will be quite hard as it needs to run as a service so there is no way you can get this file mounted as `metadata` needs to run as an `onboot` container.)

justincormack:
Needs a rebase

pkg/metadata/provider_azure.go
bketelsen: What's the best course of action on userdata?

justincormack: well if you can see the DVD and find the file on it, that should be relatively simple. Or we could just remove it for now and address userdata in another PR.

rn:
Closing due to inactivity. Please re-submit after rebase and addressing the comments. Thanks

PR #2346, opened 2017-08-03
Support multiple CDROMS in metadata

justincormack:
This checks if any CDROM in the system has metadata. Useful if
you boot off CDROM and also have a metadata CDROM.

Also switch the metadata container to read only, as it is only
writing to bind mounted in directories, and /tmp.

dog-with-disc


justincormack:
Added hash updates

PR #2353, opened 2017-08-03
static mac address for qemu 

### LINKED ISSUE #2351: static mac address in qemu networking
#the new qemu networking capabilities generate [new mac addresses](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_qemu.go#L528) with every vm start which makes it hard to assign a static ip/hostname for the vm via dhcp.

Are there any future plans for this as the existing networking parameter or would an additional optional setting like -networking device,devicename[,xx:xx:xx:xx:xx:xx] be suitable?

w9n:
resolves  #2351
 

**- What I did**
Check if there is a mac address appended to the networking parameter, if not keep generating a random one, if there is one provided also delete it from the parameter.
**- How I did it**
Regex.
**- How to verify it**
Run a vm with and without additional Mac.
Add some regex to parse a mac address from the qemu networking parameter

</img>

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -134,7 +135,7 @@ func runQemu(args []string) {
 	vmUUID := uuid.NewV4()
 
 	// Networking
-	networking := flags.String("networking", qemuNetworkingDefault, "Networking mode. Valid options are 'default', 'user', 'bridge[,name]', tap[,name] and 'none'. 'user' uses QEMUs userspace networking. 'bridge' connects to a preexisting bridge. 'tap' uses a prexisting tap device. 'none' disables networking.`")
+	networking := flags.String("networking", qemuNetworkingDefault, "Networking mode. Valid options are 'default', 'user', 'bridge[,name]', tap[,name] and 'none'. 'user' uses QEMUs userspace networking. 'bridge' connects to a preexisting bridge. 'tap' uses a prexisting tap device. 'none' disables networking. A custom MAC address can be appended with ,xx:xx:xx:xx:xx:xx`")
 
 	publishFlags := multipleFlag{}
 	flags.Var(&publishFlags, "publish", "Publish a vm's port(s) to the host (default [])")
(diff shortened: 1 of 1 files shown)


w9n:
there seems to happened a bug in CI.

rn:
I've hit rebuild

ijc:
Not too sure about this.

My understanding was that the linuxkit tool was intended for dev environments, rather than production. The rationale for this change given in #2351 is for static IP via DHCP which in particular seems to be straying towards the "production" side of things.

Also, I believe this would not be the same as on Mac, which gets a new IP on every `linuxkit run` too. I don't know about other backends.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -525,13 +526,19 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
 	if config.NetdevConfig == "" {
 		qemuArgs = append(qemuArgs, "-net", "none")
 	} else {
-		mac := generateMAC()
-		qemuArgs = append(qemuArgs, "-net", "nic,model=virtio,macaddr="+mac.String())
+		var macRemoveRegExp = regexp.MustCompile(",((?:[0-9A-Fa-f]{2}[:-]){5}[0-9A-Fa-f]{2})|((?:[0-9A-Fa-f]{2}[:-]){5}[0-9A-Fa-f]{2}),")
+		var macRegExp = regexp.MustCompile("([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})")
+		var mac = macRegExp.FindString(config.NetdevConfig)
```
rn: There seems to be a function in the go library to parse a MAC address: https://golang.org/pkg/net/#ParseMAC

Any reason you are not using it?

w9n: I did not know this function but i dont think its very helpful. It seems to take a string which must be a valid mac and returns a type HardwareAddr. In the actual case we get something like -networking tap,lin0,22:11:11:11:11:11 and need to parse it to the variables mac=22:11:11:11:11:11 and netdevConfig=tap,lin0. the remove Regex also correctly parses tap,22:11:11:11:11:11,lin0 or 22:11:11:11:11:11,lin0,tap format by Regex duplication

rn: I'm somewhat allergic to regex's for command line parsing and would much rather prefer parsing this explicit. I would also prefer if the order of arguments where fixed. You could still use `ParseMAC` to verify that the MAC address is valid.

w9n: this would end in a string split by ',', checking each element of the resulting array if it is a valid mac and then rejoin the array without the mac with a ',' delimeter. i would be fine with this solution.

justincormack:
We merged #2368 which (a) conflicts and (b) may resolve this issue sufficiently...

w9n:
if needed the mac can be specified in the mac-addr file of the state path, thanks to #2368

PR #2355, opened 2017-08-04
Support UEFI ISO boot on hyperkit

justincormack:
Code in line with QEMU, other than hyperkit only boots off UEFI ISOs not BIOS ISOs at present.

belka-strelka

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -92,7 +92,9 @@ func runHyperKit(args []string) {
 			*isoBoot = true
 			prefix = strings.TrimSuffix(path, ".iso")
 			// hyperkit only supports UEFI ISO boot at present
-			*uefiBoot = true
+			if !*uefiBoot {
+				log.Fatalf("Hyperkit requires --uefi to be set to boot an ISO")
+			}
 		}
 	}
 
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -46,6 +46,15 @@ func runHyperKit(args []string) {
 	vsockports := flags.String("vsock-ports", "", "List of vsock ports to forward from the guest on startup (comma separated). A unix domain socket for each port will be created in the state directory")
 	networking := flags.String("networking", hyperkitNetworkingDefault, "Networking mode. Valid options are 'default', 'docker-for-mac', 'vpnkit[,socket-path]', 'vmnet' and 'none'. 'docker-for-mac' connects to the network used by Docker for Mac. 'vpnkit' connects to the VPNKit socket specified. If socket-path is omitted a new VPNKit instance will be started and 'vpnkit_eth.sock' will be created in the state directory. 'vmnet' uses the Apple vmnet framework, requires root/sudo. 'none' disables networking.`")
 
+	// Boot type; we try to determine automatically
+	uefiBoot := flags.Bool("uefi", false, "Use UEFI boot")
+	isoBoot := flags.Bool("iso", false, "Boot image is an ISO")
+	kernelBoot := flags.Bool("kernel", false, "Boot image is kernel+initrd+cmdline 'path'-kernel/-initrd/-cmdline")
+
+	// Paths and settings for UEFI firware
+	// Note, the default uses the firmware shipped with Docker for Mac
+	fw := flags.String("fw", "/Applications/Docker.app/Contents/Resources/uefi/UEFI.fd", "Path to OVMF firmware for UEFI boot")
+
 	if err := flags.Parse(args); err != nil {
 		log.Fatal("Unable to parse args")
 	}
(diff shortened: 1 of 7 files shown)


deitch:
So this will support booting the ISO (as an iso) on UEFI-only firmwares.

1. Does it support isohybrid, to boot as a drive as well (thinking flash drive, but others as well)
2. I didn't look deeply at the code, but are we using some boot loader, or actually placing the kernel as a `.EFI` file (efistub)?

rn:
I would prefer having the vendoring in a separate commit (as it would help triaging in case something else broke with the update...). But it's ok for this change.

Could you update `docs/platform-hyperkit.md`?

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -55,7 +64,48 @@ func runHyperKit(args []string) {
...
+		if !statInitrd {
+			log.Fatalf("Cannot find initrd file (%s): %v", path+"-initrd.img", err)
+		}
+	} else {
+		// if path ends in .iso they meant an ISO
+		if strings.HasSuffix(path, ".iso") {
+			*isoBoot = true
+			prefix = strings.TrimSuffix(path, ".iso")
+			// hyperkit only supports UEFI ISO boot at present
+			*uefiBoot = true
```
rn: I would prefer if this was in line with `qemu` so I would suggest to error here if `uefi` is not set.

justincormack:
@deitch this does not change the code that generates the EFI ISO at all. I am currently looking at that. The current code is in https://github.com/linuxkit/linuxkit/blob/master/tools/mkimage-iso-efi/make-efi

Hyperkit uses a generic EFI firmware, so can boot anything, so yes it can also boot a drive.

src/cmd/linuxkit/run_hyperkit.go
justincormack: you don't generally have to set this on qemu, as it will set it if the path has `-efi.iso` in, although I am wondering about making all the ISOs hybrid which would change this... The most consistent would probably be to set it if the path has `-efi.iso` in for now.

rn: Ugh, that's kinda ugly, deriving the boot mode from a convention used in a different tool. Hybrid ISOs would work, but when I last tried to build one I found many misleading information on how to go about it

justincormack: (although maybe in general requiring it to be set makes more sense)

justincormack: ok, removed the autodetect in qemu as well.

deitch:
> this does not change the code that generates the EFI ISO at all. I am currently looking at that. 

So a script that wraps xorriso. I have banged my head on the wall with it a few times. I used to have code that generates a hybrid just right, but I cannot find it now. If I can, I happily will post.

> Hyperkit uses a generic EFI firmware, so can boot anything, so yes it can also boot a drive

I got that, I meant if it was hybrid and thus bootable off a drive.

So it still generates a kernel, filesystem and bootloader? I have not done efistub, wish I had time in the next week.

justincormack:
@deitch there is no bootloader in that script, it is doing the efistub thing see https://github.com/linuxkit/linuxkit/blob/master/tools/mkimage-iso-efi/make-efi#L30

If we had a script that generated a hybrid, you could boot it with hyperkit, but you would need to tweak this script to accept a disk image as well as an ISO

deitch:
> it is doing the efistub thing see

Aha! I glossed over that line. Very nice indeed.

> If we had a script that generated a hybrid, you could boot it with hyperkit, but you would need to tweak this script to accept a disk image as well as an ISO

What do you mean by "accept a disk image"? Hybrids usually are just iso files that can be booted as iso, burned to optical media, of `dd` copied to drive.

justincormack:
I just meant if you wanted to test booting it as a drive vs as an ISO in hyperkit you would have to tweak the script (or use hyperkit directly).

deitch:
Oh! Got it, thanks.

In any case, we still don't have option for booting EFI off of (flash or other) drive, correct? Raw image boots BIOS, and EFI ISO is not yet hybrid. 

Will try and find some time to help there....

justincormack:
I think if you had a hybrid image you could boot it on qemu with `linuxkit run qemu --uefi blah.raw` already.

deitch:
> if you had a hybrid image

But we don't right? Our `raw` image is BIOS only, and our EFI-ISO is not isohybrid?

justincormack:
correct.

deitch:
Sorry @justincormack . I looked for the exact xorriso commands (and sizing, etc.) where I had built it before, could not find it. It was one repo, I did `git co <hash>` for dozens of commits going back to dawn of project, and nothing. I must be confused as to where it is. I guess build from scratch.

deitch:
Aha! It was on an old Dell that I multi-boot Windows/Ubuntu (my main is a MacBook) and use as a boot tester, under a directory in Ubuntu. I will try to get it working here next week. :-)

justincormack:
Actually the BIOS ISO is hybrid, just not the EFI one, see https://github.com/linuxkit/linuxkit/blob/master/tools/mkimage-iso-bios/make-iso#L39

deitch:
Looks like it. We probably should unify to a single hybrid EFI/BIOS iso+disk bootable. I will try.

PR #2356, opened 2017-08-04
Enable sysctl kernel.dmesg_restrict

justincormack:
This requires that users have `CAP_SYSLOG` in order to access `dmesg`.
This means that containers by default have no access to `dmesg` (which
can leak information about the host or other containers) unless they
have this capability added.

message in a bottle


justincormack:
Will update hashes shortly.

justincormack:
cc @riyazdf

justincormack:
Added hashes, did arm64 build...

justincormack:
cc @tych0

tych0:
+1, thanks!

PR #2358, opened 2017-08-04
cmd/qemu: Better handling of arch and console for aarch64

rn:
- When executing on aarch64, use it as the defaul arch
- When selecting aarch64 on a non aarch64 system set the
  CPU flag to a default value (not 'host').
- On aarch64, rewrite ttyS0 to ttyAMA0. This is a bit skanky
  but allows us to re-use execiting YAML files.

Can now boot a arm64 kernel on my mac with:
```
linuxkit run qemu -arch aarch64 linuxkit
```

turtle-beach

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -53,6 +54,19 @@ const (
 	qemuNetworkingBridge         = "bridge"
 	qemuNetworkingDefault        = qemuNetworkingUser
 )
+
+var (
+	defaultArch string
+)
+
+func init() {
+	switch runtime.GOARCH {
+	case "arm64":
+		defaultArch = "aarch64"
+	case "amd64":
+		defaultArch = "x86_64"
+	}
+}
 
 func haveKVM() bool {
 	_, err := os.Stat("/dev/kvm")
@@ -123,7 +137,7 @@ func runQemu(args []string) {
 
 	// VM configuration
 	enableKVM := flags.Bool("kvm", haveKVM(), "Enable KVM acceleration")
-	arch := flags.String("arch", "x86_64", "Type of architecture to use, e.g. x86_64, aarch64")
+	arch := flags.String("arch", defaultArch, "Type of architecture to use, e.g. x86_64, aarch64")
 	cpus := flags.String("cpus", "1", "Number of CPUs")
 	mem := flags.String("mem", "1024", "Amount of memory in MB")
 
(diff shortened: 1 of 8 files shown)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -458,8 +472,10 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
 	qemuArgs = append(qemuArgs, "-pidfile", filepath.Join(config.StatePath, "qemu.pid"))
 	// Need to specify the vcpu type when running qemu on arm64 platform, for security reason,
 	// the vcpu should be "host" instead of other names such as "cortex-a53"...
-	if config.Arch == "aarch64" {
+	if config.Arch == "aarch64" && runtime.GOARCH == "arm64" {
 		qemuArgs = append(qemuArgs, "-cpu", "host")
+	} else {
+		qemuArgs = append(qemuArgs, "-cpu", "cortex-a57")
```
ijc: This will try and use cortex-a57 even if you are running on (and _want_) an x86 won't it?

src/cmd/linuxkit/run_qemu.go
```diff
@@ -54,6 +55,19 @@ const (
...
+var (
+	defaultArch string
+)
+
+func init() {
+	switch runtime.GOARCH {
+	case "arm64":
+		defaultArch = "aarch64"
+	default:
+		defaultArch = "x86_64"
```
ijc: maybe explicit case x86 here and a "huh" in the default case? Or maybe leave as `""` so that if you run linuxkiut on e.g. powerpc you have to pick if you want to emulate arm or x86 (until there is a ppc port!)

src/cmd/linuxkit/run_qemu.go
```diff
@@ -509,11 +525,15 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
...
+		if cmdline, err := ioutil.ReadFile(config.Path + "-cmdline"); err != nil {
 			log.Errorf("Cannot open cmdline file: %v", err)
 		} else {
-			qemuArgs = append(qemuArgs, "-append", string(cmdlineString))
+			cmdlineString := string(cmdline)
+			if config.Arch == "aarch64" && strings.Contains(cmdlineString, "ttyS0") {
+				// On aarch64 the serial is ttyAMA0, rewrite command line
+				cmdlineString = strings.Replace(cmdlineString, "ttyS0", "ttyAMA0", 1)
+			}
+			qemuArgs = append(qemuArgs, "-append", cmdlineString)
```
ijc: Not super keen on this, but not much alternative.

Only thing I can think of is removing `console=` from the examples altogether and have linuxkit inject the options based on `-console=serial` vs `-console=tty` (for all backends where it makes sense I guess)

justincormack: Seems to work fine if you change the default to `"console=tty0 console=ttyS0 console=ttyAMA0"` as unknown ones seem to be ignored just fine.

src/cmd/linuxkit/run_qemu.go
rn: ah, yes. damn boolean logic

src/cmd/linuxkit/run_qemu.go
rn: @ijc yeah it's skanky
@justincormack I'm not sure I like adding `console=ttyAMA0` to all YAML files but I also don't like what I did above...what do you think about `console=serial` in the YAML and then we replace it in `linuxkit`?

justincormack: You cant just replace in `linuxkit` you need to also replace in every piece of code that makes images as it gets hard coded in the bootloader. But seems reasonable.

PR #2361, opened 2017-08-04
Update to tip of manifest-tool and opengcs/LCOW

rn:
- PR to manifest-tool to get the length for notary has been merged. Use official version now and update docs that `go get` now also works
- Update `opengcs` to latest and then update the LCOW blueprint
dog-rabbit


justincormack:
You could also mention that manifest-tool is now available from our homebrew repo

rn:
I would have if I'd known

rn:
added a note on brew install as well

justincormack:
LGTM

PR #2363, opened 2017-08-04
Put full filesystem on EFI ISO

justincormack:
In line with BIOS ISO changes previously.

Remove boot options, user can add if required, they made boot very slow.

stairs

### DIFF
--- a/tools/mkimage-iso-efi/make-efi
+++ b/tools/mkimage-iso-efi/make-efi
@@ -5,49 +5,47 @@ set -e
 mkdir -p /tmp/efi
 cd /tmp/efi
 
-# input is a tarball of kernel and initrd.img on stdin
+# input is a tarball on stdin with kernel and cmdline in /boot
 # output is an iso on stdout
 
 # extract. BSD tar auto recognises compression, unlike GNU tar
 # only if stdin is a tty, if so need files volume mounted...
 [ -t 0 ] || bsdtar xzf -
 
-INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name kernel -or -name bzImage)"
-CMDLINE="$*"
+CMDLINE="$(cat boot/cmdline)"
+# if no root= root device specified, assume /dev/sr0 ie first CD drive
+echo "${CMDLINE}" | grep -q 'root=' || CMDLINE="${CMDLINE} root=/dev/sr0"
+rm boot/cmdline
 
-[ "$KERNEL" = "./kernel" ] || mv "$KERNEL" kernel
-[ "$INITRD" = "./initrd.img" ] || mv "$INITRD" initrd.img
-
-# clean up subdirectories
-find . -mindepth 1 -maxdepth 1 -type d | xargs rm -rf
-
-# Create a EFI boot file with kernel and initrd. From:
+# Create a EFI boot file with kernel. From:
 # https://github.com/haraldh/mkrescue-uefi/blob/master/mkrescue-uefi.sh
 cp /usr/lib/gummiboot/linuxx64.efi.stub .
-echo "${CMDLINE} rootdelay=300 noautodetect" > cmdline.txt
+echo "${CMDLINE}" > cmdline.txt
 
 objcopy \
 	--add-section .osrel=/etc/os-release --change-section-vma .osrel=0x20000 \
 	--add-section .cmdline=./cmdline.txt --change-section-vma .cmdline=0x30000 \
-	--add-section .linux=./kernel --change-section-vma .linux=0x40000 \
-	--add-section .initrd=./initrd.img --change-section-vma .initrd=0x3000000 \
+	--add-section .linux=./boot/kernel --change-section-vma .linux=0x40000 \
 	./linuxx64.efi.stub \
 	linuxkit.efi
 
+rm cmdline.txt
+
 # create a ISO with a EFI boot partition
-mkdir -p iso
 # Stuff it into a FAT filesystem, making it as small as possible.  511KiB
 # headroom seems to be enough; (x+31)/32*32 rounds up to multiple of 32.
-mkfs.vfat -v -C iso/efi.raw \
+mkfs.vfat -v -C boot/efi.raw \
         $(( ($(stat -c %s "linuxkit.efi") / 1024 + 511) \
 		/ 32 * 32 )) > /dev/null
 echo "mtools_skip_check=1" >> /etc/mtools.conf && \
-mmd -i iso/efi.raw ::/EFI
-mmd -i iso/efi.raw ::/EFI/BOOT
-mcopy -i iso/efi.raw linuxkit.efi ::/EFI/BOOT/BOOTX64.EFI
+mmd -i boot/efi.raw ::/EFI
+mmd -i boot/efi.raw ::/EFI/BOOT
+mcopy -i boot/efi.raw linuxkit.efi ::/EFI/BOOT/BOOTX64.EFI
+
+rm linuxkit.efi linuxx64.efi.stub
+
 xorriso -as mkisofs \
-	-R -f -e efi.raw -no-emul-boot -o linuxkit-efi.iso iso
+	-R -e boot/efi.raw -no-emul-boot -o linuxkit-efi.iso .
 
 cat linuxkit-efi.iso
 
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 PREFIX?=/usr/local/
 
 MOBY_REPO=https://github.com/moby/tool.git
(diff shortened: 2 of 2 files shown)


tools/mkimage-iso-efi/make-efi
```diff
@@ -5,49 +5,47 @@ set -e
...
         $(( ($(stat -c %s "linuxkit.efi") / 1024 + 511) \
 		/ 32 * 32 )) > /dev/null
 echo "mtools_skip_check=1" >> /etc/mtools.conf && \
-mmd -i iso/efi.raw ::/EFI
-mmd -i iso/efi.raw ::/EFI/BOOT
-mcopy -i iso/efi.raw linuxkit.efi ::/EFI/BOOT/BOOTX64.EFI
+mmd -i boot/efi.raw ::/EFI
+mmd -i boot/efi.raw ::/EFI/BOOT
+mcopy -i boot/efi.raw linuxkit.efi ::/EFI/BOOT/BOOTX64.EFI
+
```
rn: any idea what these changes are?Looks the same to me

justincormack: changed the directory....

justincormack:
No, will test on Windows.

tools/mkimage-iso-efi/make-efi
rn: doh! Blind

deitch:
I rather like this. So instead of packing everything (initrd and kernel) into the vfat filesystem image, you pack just the kernel and the cmdline, and when it boots it uses the `.cmdline` `objcopy` option to tell it where root is, i.e. whatever was on the command line (or `/dev/sr0` by default)? 

One question: does this make it more brittle? Most systems can boot off of any cd/dvd they find (cd0, cd1, qq999, whatever it is called by the system). With the entire thing packed into the ESP, we didn't care all that much, since one entrypoint was enough. 

Now, we need to tell it _at build time_ what the boot device is going to be (`sr0` e.g.). Does that make it less flexible? Conversely, is it possibly for the kernel (or a tiny initrd) to deduce "I was booted from device `/dev/sr2` so I will use _that_ as my root"?

justincormack:
@deitch thats why it has the `echo "${CMDLINE}" | grep -q 'root=' || CMDLINE="${CMDLINE} root=/dev/sr0"` line in the patch - if you don't specify a root at build time it will choose the most likely one. I made sure that at least in the `linuxkit run` case it adds the boot CD before any additional CDs, which seems a reasonable constraint.

I couldn't find a way to detect what you were booted from, although I have some vague memory that this was possible. Certainly it does not try by default in Bios, I can double check on efi.

deitch:
> if you don't specify a root at build time it will choose the most likely one

Being `sr0`, got it. But you either have to specify which one it is, or you are bound to `sr0`. Would be nice to auto-detect.

There definitely is some way, but I cannot recall it. If I do, I will post it here.

deitch:
You know, I wonder. Could we use `efibootmgr` to do it? It queries the firmware (EFI) to determine what it is set for and what it did `CurrentBoot`. It reports exactly which device it booted and where it found the `.EFI` file it is using. Its format is more hardware-isa than Linux-ish, but it should be parsable.

deitch:
You can, although I would prefer to get it from `/sys/firmware/efi`. If you run `efibootmgr -v`, it will give you all of the potentials and the current one, e.g. 

```
BootCurrent: 0008
BootOrder: 0008,0005,....
Boot0000 Diskette Drive BBS(Floppy,,,0x0)
..
Boot0008 Ubuntu HD(2,GPT,<uuid_of_partition>,0xe1800,0x32000)/File(\EFI\ubuntu\shimx64.efi)
```

Above taken from an Ubuntu machine I booted for purpose.

I verified that the entry for `0008` is a HD, partition 2, GPT, and the UUID is the UUID of the partition (not the filesystem) for that partition (in this case `sda2`).

If we could get it out of `/proc/firmware/efi` it would be cleaner that trying to parse that. I still haven't figured out how though.

justincormack:
The problem is at build time you arent running a system. It is possible the kernel can read from there in efi case (I only tested on bios where it definitely didnt but if it is available on efi it is more plausible)

deitch:
Yeah, EFI gives standard tools/api for interacting with the firmware, which includes the kernel knowing about which partition it was booted from.

deitch:
Oh, yes, it is doable. Start with the `EFI_GLOBAL_GUID`, which is preset. Then you can retrieve the current boot device from `/sys/firmware/efi/efivars/BootCurrent-<GLOBAL_GUID>`, which is all of 6 bytes. 

In my current system, it is 

```
06 00 00 00 08 00
```

Given that I *know* I booted device `0008`, it obviously is the second to last byte, but not sure what the rules are.

Then I can take that and get the info from `/sys/firmware/efi/efivars/Boot<number>-<GUID>` so, e.g. `Boot0008-<GUID>`

That in turn has to be parsed in just the right way, but is doable.

justincormack:
@deitch we would have to do that in a tiny `initramfs` though I guess and then pivot to the real root?

It also looks like we might be able to set the root to a UUID if we can set a UUID on the ISO filesystem

deitch:
> we would have to do that in a tiny initramfs though I guess and then pivot to the real root?

I think so. Even if it is a single compiled binary.

> It also looks like we might be able to set the root to a UUID if we can set a UUID on the ISO filesystem

Oh, I didn't think of that. I _like_ it. Generate a UUID at `mkisofs` time and put it on the ISO, and then make that the root. Much much cleaner. You probably could do it for BIOS as well. 

Nice!

justincormack:
Hmm, so Grub uses the ISO modification date as the UUID, only question is if Linux does https://lists.gnu.org/archive/html/bug-xorriso/2015-06/msg00019.html

deitch:
Does it matter? As long as we are consistent?

justincormack:
Well I dont even know yet if Linux recognises ISO uuids itself at all.

deitch:
> Well I dont even know yet if Linux recognises ISO uuids itself at all.

It does, but not like we would want. The ISO uuid is a _filesystem_ UUID:

```
(ns: getty) linuxkit-025000000005:~# lsblk -f -o NAME,UUID,PARTTYPE,PARTUUID,FST
YPE
NAME UUID                                 PARTTYPE PARTUUID FSTYPE
sr0  2017-08-06-19-38-18-00                                 iso9660
```

Notice that the `PARTUUID` is empty, but the date is `UUID` (i.e. filesystem). Unfortunately, in the cmdline, the kernel can take only _partition_ UUID. From the GitHub mirror [here](https://github.com/torvalds/linux/blob/master/init/do_mounts.c#L183)

```C
/*
 *	Convert a name into device number.  We accept the following variants:
 *
 *	1) <hex_major><hex_minor> device number in hexadecimal represents itself
 *         no leading 0x, for example b302.
 *	2) /dev/nfs represents Root_NFS (0xff)
 *	3) /dev/<disk_name> represents the device number of disk
 *	4) /dev/<disk_name><decimal> represents the device number
 *         of partition - device number of disk plus the partition number
 *	5) /dev/<disk_name>p<decimal> - same as the above, that form is
 *	   used when disk name of partitioned disk ends on a digit.
 *	6) PARTUUID=00112233-4455-6677-8899-AABBCCDDEEFF representing the
 *	   unique id of a partition if the partition table provides it.
 *	   The UUID may be either an EFI/GPT UUID, or refer to an MSDOS
 *	   partition using the format SSSSSSSS-PP, where SSSSSSSS is a zero-
 *	   filled hex representation of the 32-bit "NT disk signature", and PP
 *	   is a zero-filled hex representation of the 1-based partition number.
 *	7) PARTUUID=<UUID>/PARTNROFF=<int> to select a partition in relation to
 *	   a partition with a known unique id.
 *	8) <major>:<minor> major and minor number of the device separated by
 *	   a colon.
 *
 *	If name doesn't have fall into the categories above, we return (0,0).
 *	block_class is used to check if something is a disk name. If the disk
 *	name contains slashes, the device name has them replaced with
 *	bangs.
 */
```

So... kernel needs `PARTUUID`, and iso9660 _only_ has `UUID`. There also is a "relative to a known UUID". Perhaps we can put a PARTUUID on the ESP we create to put in the iso9660, and then relative from there?

justincormack:
As far as I can tell this does work on Windows, it is certainly mounting the rootfs and starting containerd.

justincormack:
@deitch I have opened #2375 to continue discussing improvements for boot selection, but it looks like it still needs some more experimentation.

justincormack:
image built, tool pull is https://github.com/moby/tool/pull/137 will add new `moby` hash here after that is merged.

justincormack:
Added the update to the `moby` tool that uses this image.

PR #2368, opened 2017-08-07
save mac address within state for qemu

zacheryph:
Signed-off-by: Zachery Hostens <zacheryph@gmail.com>

**- What I did**
Have qemu runner save the mac address that is generated within state to be reused on restart. Without this each run (even with an existing state) generates a new mac address. In turn the hostname changes since it uses the mac address.

This is useful for [at least] `projects/kubernetes` if you want the master/nodes to stick around after a reboot without having to drain/delete the nodes.

**- How I did it**
Check state directory for a `mac-addr` file and use this for the mac address. If not found generate a random one and save for future runs.

**- How to verify it**
run an image and check state directory for `mac-addr` file.  Kill / run again using same state directory and validate same mac address is used.

* save mac address for restarts on qemu runner

**- Notes**
Relates to #2351 and will probably clash with #2353 

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -86,6 +86,25 @@ func envOverrideBool(env string, b *bool) {
 	}
 }
 
+func retrieveMAC(statePath string) net.HardwareAddr {
+	var mac net.HardwareAddr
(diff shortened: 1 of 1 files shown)


justincormack:
This seems like a good solution to me; not sure if it can easily be extended to other platforms but could in principle anyway.

PR #2371, opened 2017-08-07
Add script to clean up older images

deitch:

**- What I did**
Added a script to `./scripts/` that will clean up all images that start with a particular name (namespace, part of namespace, namespace+repo, namespace+part of repo all valid) and how many recent versions to keep.

Per discussion on slack channel with @rn might be useful here.

E.g.

```
./scripts/cleanup-older-images.sh linuxkit/ 3
```

Will remove all images starting with `linuxkit/` except for the 3 most recent of each full repo name.

It sort of is `system prune -a` except that it adds:

* ability to limit to a particular set of images by namespace, namespace+repo, or part
* ability to keep a certain number of most recent

**- How I did it**
shell script wrapping docker, awk, etc. And, yes, no bash :-) .

**- How to verify it**
Run it and see. Includes an option `-n` where it will tell you what it will do but not delete any.

Tool to prune images more selectively.

wombat

### DIFF
--- a/scripts/cleanup-older-images.sh
+++ b/scripts/cleanup-older-images.sh
@@ -1,0 +1,88 @@
+#!/bin/sh
+set -e
+
+##
+#
+# script to remove all older linuxkit images
+# see usage() for usage and functionality
+#
+
+usage() {
+    cat >&2 <<EOF
+$0 [-n] <base> <versions>
+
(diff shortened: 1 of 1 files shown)


deitch:
CI is clean. Do we need to await another maintainer to approve, or can we just merge?

PR #2372, opened 2017-08-07
Do an extra chmod after chown

justincormack:
Chown clears suid bits even for root on Linux.

Also move a few functions to x/sys/unix from syscall, to be
more arm64 friendly.

Fixes issue with suid binaries losing their suid flags.

sudo make me a sandwich

### DIFF
--- a/pkg/init/init.go
+++ b/pkg/init/init.go
@@ -22,12 +22,18 @@ import (
 func copyMetadata(info os.FileInfo, path string) error {
 	// would rather use fd than path but Go makes this very difficult at present
 	stat := info.Sys().(*syscall.Stat_t)
-	if err := syscall.Lchown(path, int(stat.Uid), int(stat.Gid)); err != nil {
+	if err := unix.Lchown(path, int(stat.Uid), int(stat.Gid)); err != nil {
 		return err
 	}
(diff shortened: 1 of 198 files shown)


pkg/init/init.go
```diff
@@ -22,13 +22,19 @@ import (
...
-	if err := syscall.Lchown(path, int(stat.Uid), int(stat.Gid)); err != nil {
+	if err := unix.Lchown(path, int(stat.Uid), int(stat.Gid)); err != nil {
 		return err
 	}
 	timespec := []unix.Timespec{unix.Timespec(stat.Atim), unix.Timespec(stat.Mtim)}
 	if err := unix.UtimesNanoAt(unix.AT_FDCWD, path, timespec, unix.AT_SYMLINK_NOFOLLOW); err != nil {
 		return err
 	}
+	// after chown suid bits may be dropped; also setting them with os.OpenFile is not well defined
+	if info.Mode().IsRegular() {
```
ijc: Why only for regular files?

justincormack: amended to be not symlinks (you cant set mode on a symlink without following in Linux)

ijc:
LGTM

justincormack:
Updated hashes

justincormack:
cc @kencochrane this should fix your suid issue.

kencochrane:
@justincormack thank you!

Issue #2064, opened 2017-06-16
[tracking] Packet provider support open issues and missing features


vielmetti:
**Description**

"it would be good to open an issue that lists all the missing features as a checklist."

Reference 

**Steps to reproduce the issue:**

Boot Linuxkit on Packet, notice that it's not feature complete. Examine [run_packet.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_packet.go) and notice some TODOs.

**Feature checklist**

Somewhat ordered by ease of implementation, work in progress, or need

- [ ] Improved documentation
- [ ] Persistent PXE boot support
- [ ] "User data" to support boot status reporting
- [ ] TODO: poll events api for bringup (requires extpacknogo)
- [ ] TODO: connect to serial console (requires API extension to get SSH URI)
- [ ] TODO: add ssh keys via API registered keys
- [ ]  Storage support via on-board SSD and/or HDD
- [ ] Boot on Type 2A (ARMv8)
- [ ] Boot on Type 2 (Intel, bonded Mellanox NICs)
- [ ] Bonding of network cards on devices that support that
- [ ] Boot on Type S (multiple HDD)

rn:
thank you for putting this together

vielmetti:
#1245 has some of this discussion and some additional detail

vielmetti:
#1377 has some specifics and tracks the "type 2a" ARM issue

vielmetti:
Updated to reflect #2376 changes for always-on PXE, serial console.

vielmetti:
Add #2378 for RAID support on Type 2's.

vielmetti:
Updated to reflect that #2376 has been merged.

justincormack:
#2438 fixes metadata, userdata, ssh keys, bonding.

vielmetti:
Added #2434 as an open issue, Networking on Cavium Thunder is not quite working (4.9.x kernel).

rn:
@vielmetti with https://github.com/linuxkit/linuxkit/pull/2852 we changed iPXE support and now load the iPXE script via an URL instead of user-data. This frees up the userdata for metadata and other stuff.

PR #2376, opened 2017-08-07
Always PXE boot and add serial console to packet driver

### LINKED ISSUE #2065: run_packet: persisting iPXE

rn:
- vendor a fork of the packet.net Go bindings which supports always PXE
- enable always PXE by default resolves #2065
- without this, only the first boot of a machine would use PXE
- ssh into the Serial over SSH (SOS) host once the machine is booted so get console access
- update the documentation, in particular with some notes on how to serve the images with http and `ngrok`. I'll soon add better support for this.

this is wip for #2064


docs/platform-packet.md
```diff
@@ -18,64 +18,78 @@ is a work in progress.
 [Type 1]:https://www.packet.net/bare-metal/servers/type-1/
 [Type 2A]:https://www.packet.net/bare-metal/servers/type-2a/
 
+The `linuxjit run packet` command can mostly either be configured via
```
vielmetti: Shouldn't this be `linuxkit run packet`?

docs/platform-packet.md
```diff
@@ -18,64 +18,78 @@ is a work in progress.
...
+command line options or with environment variables. see `linuxkit run
+packet --help` for the options and environment variables.
+
 ## Boot
 
-Build an image with `moby build`. The [packet.yml](https://github.com/vielmetti/linuxkit/blob/master/examples/packet.yml)
-example file provides a suitable template to start from.
+LinuxKit on
+Packet
+[boots via iPXE](https://github.com/vielmetti/linuxkit/blob/master/examples/packet.yml). It
```
vielmetti: It would be better, I think, for this example not to live in https://github.com/vielmetti/linuxkit but instead in https://github.com/linuxkit/linuxkit .

rn:
@vielmetti thanks for the comments. I've updated the docs. I rephrased it a bit more

vielmetti:
Looks good. I'll pursue the `packngo` Go API changes so you don't have to vendor a fork for very long.

docs/platform-packet.md
```diff
@@ -18,64 +18,81 @@ is a work in progress.
...
+
+func main() {
+	// Simple static webserver:
+	log.Fatal(http.ListenAndServe(":8080", http.FileServer(http.Dir("."))))
+}
+```
+
+and then `go run` this in the directory with the `kernel+initrd`.
+
+The web server must me accessible from Packet. You can either run the
```
justincormack: "must me"->"must be"

justincormack:
small typo but otherwise LGTM

rn:
pushed fix. will merge when green (which it should be)

Issue #2311, opened 2017-07-27
Issues with `linuxkit.packages.extend.{ext4,btrfs,xfs}`


ijc:
These are failing in CI on master (see [here](https://linuxkit.datakit.ci/log/live/linuxkit-build-of-2c58fbc01ab3a52f8e875f4e6fa4ebb2be34279a-default-ci) and [here](https://linuxkit.datakit.ci/log/saved/linuxkit-build-of-16674a1557ca85876dcd8d536341fe2602a7a7f6-default-ci/baa4b01f1abab399c5394d5e29977dd655a8d4bc) for the two most recent at time of writing).

Strangely in the PR which merged them (#2152) and on PRs since they seem to pass. They pass for me locally

The tests look to have multiple issues (which may or may not be related to the failures, and some of which make other moot):
* Misnamed (in `cases/040_packages/` the number is the index of the first char of the test name).
* They rely on docker.
* All use the same image name tag (`test-extend`).
* For some reason they bind mount in the script instead of `ADD` in the `Dockerfile`.
* They all leak `test-create*` into the filesystem.

@justincormack suggests that the use of docker and much of the related complexity could be replaced by simply using `format=raw` when creating the disk and then using `dd` to append some (sparse) empty space.

That would just leave the leak of `test-create*` which should be a simple case of adjusting the cleanup runes.

justincormack:
weirdly one of the tests just failed on https://github.com/linuxkit/linuxkit/pull/2309 so probably racy

dave-tucker:
> Misnamed (in cases/040_packages/ the number is the index of the first char of the test name).

That seems quite an odd convention and should probably be documented somewhere.
IMHO (unless there are strong objections to the contrary) I'd recommend just picking the next available number. If we have no constraints for order, I *think* that `rtf` is smart enough to order alphabetically which would be much simpler IMO.

> All use the same image name tag (test-extend).

Laziness on my part 😴. Sorry.

> -They rely on docker.
> - For some reason they bind mount in the script instead of ADD in the Dockerfile.

This is mainly a hangover from many failed attempts as, IMO, these tests were a pain in the a**e to write. `raw` formatted disks and `dd` would make a lot more sense. I did try this route but ran out of patience as I hit silly issues with the disks not being recognized and complaints that the disks were too small for journaling filesystems etc...

> They all leak test-create* into the filesystem.

These should be called `create-*` as it's using `moby build -name create test-create.yml`.
[...]

PR #2380, opened 2017-08-08
Cleanup the extend test cases

dave-tucker:
This PR fixes the ordering of the test cases and removes the dependency on Docker.

Fixes: #2311


test/cases/040_packages/005_extend/002_xfs/check.sh
```diff
@@ -9,5 +9,5 @@ function failed {
 
 [ -f /var/lib/docker/bar ] || failed
 touch /var/lib/docker/foo || failed
-df -h | grep -q "507.7M" || failed
+df -h | grep -q "504.8M" || failed
```
justincormack: I would be reluctant to check that exactly - check the first digit or something, a different version may have slightly different space usage.

dave-tucker: ack. replaced with a regex and re-running locally before a force push

PR #2378, opened 2017-08-07
Installed the MEGARAID_SAS driver as a kernel module

TheOriginalAlex:
Added support for RAID controllers on Packet.net Type 2 servers
Solves part of issue #1245

**- What I did**
Added `CONFIG_MEGARAID_SAS=m` to kernel config.
**- How I did it**
Updated kernel_config-4.9.x and kernel_config-4.4.x
**- How to verify it**
Build kernel and run on Packet's Type 2 server through iPXE.  Run `fdisk -l`.  The drives show up now.

**- Description for the changelog**

Added MEGARAID_SAS drivers a kernel module

**- A picture of a cute animal (not mandatory but encouraged)**
Cute Bunny

PR #2381, opened 2017-08-08
Update kernels to 4.9.41/4.4.80

rn:
x86 kernels already build/pushed/signed so that the tests will be run and the arm64 builds are in progress. This also picks up the megaraid driver from #2378

/cc @TheOriginalAlex


rn:
Had a merge conflict and I force pushed an update. I didn't rebuild the kernels which may mean that the git commits are off.

Weird that the CI didn't get restarted...

PR #2386, opened 2017-08-09
arm64: kernel 4.4.x bringup enablement

arm64b:
Enable DEVPTS_MULTIPLE_INSTANCES in kernel configuration file
to avoid the devpts mounting hang issue during bootup when 
running LinuxKit.

**- What I did**
enable the kernel 4.4.x on LinuxKit
**- How I did it**
update the kernel config file to enable DEVPTS multiple instance in the kernel
**- How to verify it**
build the individual kernel with `make NOTURST=1 build_4.4.x`, then replace the kernel container image with the new built one in linuxkit.yml. `bin/linuxkit run` it and boot to the shell.

summit


justincormack:
Not sure anyone should be using 4.4 at this point as it only has 6 months of support left, but LGTM.

PR #2393, opened 2017-08-09
Use ISO for kubernetes

justincormack:
Because Kubernetes is 1.5GB, ISO makes sense as the files do not take up memory, so you can boot a 1GB machine rather than a 4GB one.

Also cherry pick a patch of Ian's, and de-bashify the script.

img_20170809_003915

### DIFF
--- a/projects/kubernetes/boot.sh
+++ b/projects/kubernetes/boot.sh
@@ -1,10 +1,20 @@
-#!/bin/bash -eu
+#!/bin/sh
+
+set -e
+
 : ${KUBE_PORT_BASE:=2222}
 : ${KUBE_VCPUS:=2}
-: ${KUBE_MEM:=4096}
+: ${KUBE_MEM:=1024}
 : ${KUBE_DISK:=4G}
 : ${KUBE_NETWORKING:=default}
 : ${KUBE_RUN_ARGS:=}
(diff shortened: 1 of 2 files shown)


projects/kubernetes/boot.sh
```diff
@@ -1,10 +1,20 @@
...
+: ${KUBE_MEM:=1024}
 : ${KUBE_DISK:=4G}
 : ${KUBE_NETWORKING:=default}
 : ${KUBE_RUN_ARGS:=}
+: ${KUBE_EFI:=}
+
+[ "$(uname -s)" = "Darwin" ] && KUBE_EFI=1
+
+suffix=".iso"
+[ -n "${KUBE_EFI}" ] && suffix="-efi.iso" && uefi="--uefi"
```
ijc: The QEMU backend supports this (supposedly, I've not tried it), so maybe this could all just be unconditionally UEFI?

justincormack: It could, but not everyone has firmware, and we had some reliability issues with it in CI, so decided to not default to uefi for qemu... @deitch is looking into making the ISOs dual efi/bios which would make it all simpler though.

ijc: SGTM.

PR #2395, opened 2017-08-10
Fix root directory permissions with overlay

justincormack:
At least when running under Docker this seems to cause some issues,
possibly also explains some other oddnesses.

dog-in-cage


justincormack:
building new images now

PR #2399, opened 2017-08-11
package: Do not add an arch suffix if an explicit HASH is given

ijc:
During development I tend to do:
	make -C pkg/foo tag HASH=dev ORG=ijc
and I expect to get an image `ijc/foo:dev` and not `ijc/foo:dev-amd64`. In
general I think if the HASH has been explicitly provided the build should
just honour that.

To compensate and avoid mistaken pushes adjust the check for dirtiness to also
insist on a non-empty suffix.


ijc:
Somewhat related (although maybe it should be a separate issue) the workflow for testing exactly what one intends to push (to both hub and a PR) without actually pushing anything isn't too smooth with the arch suffix.

I usually write  `«repo»/«package»:«hash»` (from `show-tag`) into the `yml` files for the final test, but running `make -C pkg/«package» tag` now produces `«repo»/«package»:«hash»-«arch»` so the `yml` file isn't usable until a `make -C pkg/<package> push` also creates the manifest. I could write `«hash»-«arch»` in the `yml` for testing or (with the change in this PR) do `make -C pkg HASH=«hash» tag` but both are error prone and not actually testing what I will push.

Note sure what the answer is, perhaps a `local-manifest` target?

PR #2400, opened 2017-08-11
[WIP] kubernetes: use cri-containerd to use containerd as runtime

ijc:
**- What I did**

Switched `project/kubernetes` to using [cri-containerd](https://github.com/kubernetes-incubator/cri-containerd/) as the runtime. This cri implementation drivers containerd v1.0.0 directly.

**- How I did it**

Mainly it is just a case of packaging cri-containerd, configuring kubelet and making a adjustments to the images.

This PR is a WIP for a variety of reasons:
* It is based on cri-containerd targeting v1.0.0-alpha3, current cri-containerd targets a version between alpha2 and alpha3. I have a local [update to alpha3](https://github.com/Ijc/cri-containerd/tree/containerd-v1.0.0-alpha3) which is pretty trivial but alpha3 lacks some fixes (such as https://github.com/containerd/containerd/pull/1279) from the cri-containerd maintainers. I'm not sure how critical these are in practice. The fixes are in alpha4 (if noone beats me to it I'll make a PR on cri-containerd after I've bumped linuxkit to alpha4 in a separate PR here).
* The system isn't quite functional! Currently it seems to fail to pull the weave image. I haven't fully investigated. The system, is trying to pull `weaveworks/weave-kube:2.0.1` (which comes from the `weave.yaml`), whereas containerd requires one to ask for `docker.io/weaveworks/weave-kube:2.0.1` I've not yet investigated who (if anyone) is responsible for the imposition of a default registry in the new configuration.
* The image cache images are not ported and are commented out for now. Images will be downloaded at runtime instead (not the end of the world, but not ideal).
* I'm unsure if we should just make this transition or if we should continue to support both configurations (and if so how I would do that)

**- How to verify it**

You should be able to boot and run an image and run `kubeadm-init.sh` and it should bring up most pods except networking. It's not yet possible to run a workload (or rather I've not tried because networking isn't up).

Kubernetes project now uses cri-containerd by default.

[](https://www.metal-archives.com/albums/Vae_Solis/Throw_the_Horns/36469)

### DIFF
--- a/projects/kubernetes/cri-containerd/Dockerfile
+++ b/projects/kubernetes/cri-containerd/Dockerfile
@@ -12,8 +12,8 @@ RUN \
 ENV GOPATH=/go PATH=$PATH:/go/bin
 
 ENV CRI_CONTAINERD_URL https://github.com/kubernetes-incubator/cri-containerd.git
-#ENV CRI_CONTAINERD_BRANCH ...
-ENV CRI_CONTAINERD_COMMIT 86a0f6a59b0d61381d60473680f1b6283fbfd8f9
+ENV CRI_CONTAINERD_BRANCH pull/113/head
+ENV CRI_CONTAINERD_COMMIT 32e03134186077f0dc4ef48767542a7165d181b0
 RUN mkdir -p $GOPATH/src/github.com/kubernetes-incubator && \
     cd $GOPATH/src/github.com/kubernetes-incubator && \
     git clone $CRI_CONTAINERD_URL cri-containerd
--- a/projects/kubernetes/cri-containerd/Dockerfile
+++ b/projects/kubernetes/cri-containerd/Dockerfile
@@ -1,0 +1,44 @@
+FROM linuxkit/alpine:4a9e5f80a774bbea494d49324428a22c6f018865 AS build
+
+RUN \
+  apk add \
+  bash \
+  gcc \
+  git \
+  go \
+  libc-dev \
+  make \
+  && true
+ENV GOPATH=/go PATH=$PATH:/go/bin
+
+ENV CRI_CONTAINERD_URL https://github.com/kubernetes-incubator/cri-containerd.git
+#ENV CRI_CONTAINERD_BRANCH ...
+ENV CRI_CONTAINERD_COMMIT 86a0f6a59b0d61381d60473680f1b6283fbfd8f9
+RUN mkdir -p $GOPATH/src/github.com/kubernetes-incubator && \
+    cd $GOPATH/src/github.com/kubernetes-incubator && \
+    git clone $CRI_CONTAINERD_URL cri-containerd
+WORKDIR $GOPATH/src/github.com/kubernetes-incubator/cri-containerd
+RUN set -e; \
+    if [ -n "$CRI_CONTAINERD_BRANCH" ] ; then \
+        git fetch origin "$CRI_CONTAINERD_BRANCH"; \
+    fi; \
+    git checkout $CRI_CONTAINERD_COMMIT
+RUN make static-binaries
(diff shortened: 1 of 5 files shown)


justincormack:
I think it might make sense to provide examples with both docker (possibly an older certified one not the recent one we have?) and cri-containerd. Will have a think about how we want to merge the kubernetes project into the main repo, as it is pretty mature now in many ways.

justincormack:
Maybe we should merge the standard kubernetes as a graduated project and add a new cri-containerd project while that is still in alpha?

ijc:
Not a bad plan. I think if we did that I would want to avoid maintaining two forks of the actual kubernetes package (`pkg/kubernetes` and `projects/kubernetes/kubernetes`) so I'd likely want to find a sensible way to make https://github.com/linuxkit/linuxkit/pull/2400/files#diff-2e2dba33f31c4cacc705fdd4b976e815 be possible at image build time rather than at package build time. Shouldn't be too hard I think.

Same might be true of the `projects/kubernetes/image-cache` stuff, need to have a think about how that will work in a containerd based system.

justincormack:
When I was experimenting with using image cache for images, I found it wasnt too hard to create at build time, so we could have another section in the yaml config that lists images to preload.

ijc:
Sounds like a great idea.

ijc:
Is there a particular mechanism for promoting things from projects to mainline? Just move in a PR?

How about:
* Check if anything here is applicable to the existing Docker based projects/kubernetes and if so apply it.
* Modify `kubelet.sh` to get `--container-runtime=remote` and `--container-runtime-endpoint=unix:///var/run/cri-containerd.sock` from a file (which can be overridden by the `files` stanza in a yml file).
* Move `projects/kubernetes/kubernetes` to `pkg/kubernetes`
* Move `projects/kubernetes/image-cache` to `pkg/docker-image-cache` (renamed as it goes to clarify the scope)
* Move the following from `projects/kubernetes/` to `examples/kubernetes/`:
  * `README.md`
  * `kube-{master,node}.yaml`
  * `Makefile`
  * `boot.sh`
  * perhaps `ssh_into_kubelet.sh` (I don't use this myself, so I'm not sure how much utility it has, maybe it could be removed and the docs updated)
* Introduce `projects/kubernetes/cri-containerd` (the new pkg added here) and reintroduce a bunch of the above things which were moved above as derived versions targetting cri-containerd. Or perhaps consider naming it `projects/kubernetes-containerd` or something to avoid confusion?

The goal would be that `pkg/kubernetes` would be usable by both `examples/kubernetes` and `projects/kubernetes`.

Separately (and somewhat independently) implement the suggestion in https://github.com/linuxkit/linuxkit/pull/2400#issuecomment-322196782 to allow the system containerd's content store to be prepopulated with images via the `yml`.

ijc:
Rebased onto alpha4 and now using https://github.com/kubernetes-incubator/cri-containerd/pull/113 instead of my fork. I think we expect that to be merged quite soon too.

Couple of cleanups/fixes to the build. I also had to bind `/tmp` into the `cri-containerd` container due to changes between alpha3 and alpha4 (I think).

This is just the rebase and update, nothing suggested in https://github.com/linuxkit/linuxkit/pull/2400#issuecomment-322896671 has been done.

justincormack:
That sounds about right. I think `projects/kubernetes-containerd` is better. or cri-containerd as its shorter....

ijc:
https://github.com/kubernetes-incubator/cri-containerd/pull/113 is merged so we can now use `cri-containerd#master`.

I'll go with `projects/kubernetes-containerd`.

ijc:
Replacing with #2463 and followups.

PR #2402, opened 2017-08-11
Bump to containerd v1.0.0-alpha4 and runc v1.0.0-rc4

### LINKED ISSUE #2328: re-enable containerd test
disabled in https://github.com/linuxkit/linuxkit/pull/2309#issuecomment-319061184 as it seems to fail on an external disk. The failure looked like it might be genuine...

ijc:
Updates to the latest releases of runc and containerd.

Also a few fixes found during #2328 investigations so the containerd test is now actually testing the thing we wanted.

fix #2328


ijc:
It's worth noting that a bunch of `ctr` top level commands are now subcommands of either `c`/`containers` or `t`/`tasks`. Specifically `ctr exec ...` is now `ctr t exec ...`

ijc:
Now also with fixed `linuxkit.packages.containerd` test. Fixes #2328.

justincormack:
yay, all green!

ijc:
Another change to be aware of is the `ctr c ls` now lists all containers matching the filter "ls", which is likely to be none... To list all containers just `ctr c`.

PR #2403, opened 2017-08-12
Add sgdisk package to linuxkit/alpine

deitch:

**- What I did**
Added `sgdisk` package to `tools/alpine` so we can use it in builds of packages

**- How I did it**
Just added it to the `packages` file

**- How to verify it**
Create a Dockerfile `FROM` the version, and `add sgdisk`. If it works, it was installed.

Add sgdisk package to tools/alpine

Canadian Elk

I know I am supposed to do something as well with `packages.versions`, but not sure what?

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -72,6 +72,7 @@ qemu-system-arm
 qemu-system-x86_64
 sed
 sfdisk
+sgdisk
 slang-dev
 squashfs-tools
 strace


rn:
What do you need this for? I don't mind adding it, but an explanation would be good.

Regarding the `versions` files, you need to:
- on arm64: `make push`. This will update `versions.aarch64` and create a multi-arch image with the arm64 image uptodate and the x86_64 on the previous version.
- `git commit --amend` + `git push --force`
- on x86: `git fetch` to update to latest and then `make push`. This will update `versions.x86_64` and create a multi-arch image with both arm64 and x86_64 on latest
- `git commit --amend` + `git push --force` and everything should be ready

deitch:
> What do you need this for? I don't mind adding it, but an explanation would be good.

I'm working on getting the raw image to be EFI bootable. Actually trying for bootable on both. Almost there. But to do so we need to partition the disk GPT, so you need the gdisk tools.

deitch:
@rn wrote:

> Regarding the versions files, you need to:

Doing now. Is that documented somewhere? I looked and don't think I missed it.

Either way, this is a bit of a headache. We looking at automating all of this through some form of CI?

deitch:
I got half way through. All the `arm64` went fine, then back to `x86`:

```
$ make push
DOCKER_CONTENT_TRUST=1 docker pull alpine:3.6
WARN[0012] Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely
ERRO[0012] Metadata for timestamp expired
you are not authorized to perform this operation: server returned 401.
make: *** [iid] Error 1
```

Yes, I did set the `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` env var correctly.

@rn? Or should I be asking @riyazdf about this?

justincormack:
@deitch sounds like you dont have network connectivity?

deitch:
> sounds like you dont have network connectivity?

I am on a slow connection, but same laptop as I am writing this, and it works. Maybe it just was too slow and timed out? Trying again now...

... and this time it worked. OK, not complaining!

rn:
can you share the hashes from the `versions` files? I see a new `arm64` image but no `amd64`

deitch:
>  I see a new arm64 image but no amd64

Impossibly slow internet connection at this aibnb. It will finish pushing eventually.

deitch:
Or not. I will try again in the am. Sigh.

deitch:
Running for 1.5 hrs and 20% done. I will just leave it all day and post here when upload complete.

rn:
Let me merge and I'll do a followup PR this morning with an update of the packages

deitch:
Thanks @rn, much appreciated.

PR #2408, opened 2017-08-14
Add a HTTP server to packet.net to serve iPXE files

rn:
This adds a `-serve` option to the packet.net backend to serve the iPXE files locally. In particualr with `ngrok` this makes booting locally build images via iPXE much simpler. The server only serves the files needed by iPXE. See updated docs for an example

image


src/cmd/linuxkit/run_packet.go
```diff
@@ -267,3 +302,28 @@ func sshHostKey(host string) (ssh.PublicKey, error) {
...
 	return hostKey, nil
 }
+
+// This implements a http.FileSystem which only responds to specific files.
+type serveFiles struct {
+	files []string
+}
+
+func (fs serveFiles) Open(name string) (http.File, error) {
+	name = strings.TrimLeft(name, "/")
```
justincormack: shouldnt this account for Windows file seperator?

rn: not this part, but some other part, I think. URL paths are always `/` but the filename passed in might be windows so the `name == fn` may fail

justincormack: Are you going to make changes then?

rn: yes, but then started adding other bits while at it...

rn:
Update pushed. This now uses the official packet go bindings as they incorporated the iPXE changes and, by default removes the machine when done

PR #2410, opened 2017-08-15
tools/alpine: Update base image

rn:
Completes #2403


deitch:
Thanks again for taking care of this, @rn.

Do we need to update the `FROM linuxkit/alpine:<hash>` for the various `tools/` and `pkg/`?

rn:
we don't normally do this because it's a PITA to rebuild all packages. So we resync once in a while

deitch:
Works for me.

Issue #2151, opened 2017-07-04
Proposal: syntax for loading data into containers


deitch:
As discussed here #2138 (and specific comment by @justincormack [here](https://github.com/linuxkit/linuxkit/pull/2138#issuecomment-312449909) we need to explore the best way to get files into containers at build time.

The use case at hand is when someone wants to use a fairly standard container (docker, sshd, getty, etc.), but wants to load some data or config files in. 

There are 2 ways to do it:

1. Extend the standard image via a Dockerfile (`FROM linux kit/docker:...`)
2. Use `files:` in the moby `.yml`

The problem with the first (extending) is that it requires significant extra work just to load some config files or data in, and makes it harder to keep up to date (rebuilding container image and _then_ linuxkit image, as opposed to just changing a hash and rebuilding linuxkit image). 

The problems with the second (`files`) are:

* Pollutes the host filesystem, which we are trying to make cleaner and simpler as we go on
* Can lead to conflicts between files for the host fs and the desired container, or two or more different `onboot` or `services` containers
* Requires using a `binds` syntax on the container, which causes the container to lose all of its defaults defined in the `LABEL` on the container image, which can lead to messiness, boilerplate, etc.

Some possible ways to address:

1. Leave as is and rely on user to use paths intelligently - this works for now, but as linuxkit adoption grows, I suspect we will see more and more issues - or build images
2. Leave as is but create special key for `services` and `onboot` called, e.g. `bind_adds:` that _appends_ bind mounts to the container's defaults, potentially overriding
3. Create a "data-volume" like structure for loading data into containers

The structure for "data-volume" might look something like this:

```yml
data:
   myvol:
   - path: /foo/bar
      source: /some/path
   - path: /foo/why
      content: "parce que je veux"
  othervol:
   - path: /profile_foo
      source: config_path.sh
```

It then would be consumed by a variant on `binds` as follows:

```yml
services:
  - name: mycontainer
     data:
       - myvol:/foo/bar:/my/path/bar
       - myvol:/foo/why:/french/truth
  - name: mergec
     data:
       - myvol:/foo/bar:/my/path/bar
       - othervol:/profile_foo:/etc/profile.d/foo
```

The differences between `data` and `binds` are:

* Consumes from named volumes (which really just are special paths)
* Complements `binds`
* Has a "source" component, indicating what named "volume" in `data` it is from

Lots of other ways to do this, but this should get the discussion started.

justincormack:
[...]

PR #2413, opened 2017-08-15
Add a runtime config

justincormack:
This adds support for a runtime configuration file that can do:
- `mkdir` to make a directory at runtime, eg in `/var` or `/tmp`, to avoid workarounds that we currently have such as a busybox container doing `mkdir`
- `interface` that can create network interfaces and move them from the root to a container
- `bindNS` that can bind mount a namespace of an `onboot` container to a file so a service can be started in that namespace.

It merges the `service` and `onboot` tools (in `init`) to avoid duplication. This also saves some size for eg LCOW which did not use the `onboot` code in `runc`.

This is currently a work in progress, for discussion, some parts are missing/TODO still, just want to discuss the overall design as it is a big feature change. Currently looking at other networking use cases (eg bonding on packet) to see if it is useful/needed.

It can move a host interface, while keeping the config, eg to move a management interface into a management container, or moving the host interface to an isolated container.

It can now bring up `wireguard` eg with config like:
```
kernel:
  image: linuxkit/kernel:4.9.41
init:
  - linuxkit/init:d6cde616b6bd48746194b4f6e42fec80fb0acf1d-dirty-amd64
  - linuxkit/runc:838259153885c0c40460379d6cdb7baebaf3fa36-dirty-amd64
  - linuxkit/containerd:09a21d6606a4011efebacff62a88b956ecac01be-dirty-amd64
  - linuxkit/ca-certificates:e44b0a66df5a102c0e220f0066b0d904710dcb10
onboot:
  - name: dhcpcd
    image: linuxkit/dhcpcd:f3f5413abb78fae9020e35bd4788fa93df4530b7
  - name: wg
    image: linuxkit/ip:XXXXX
    net: new
    binds:
    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1 peer 192.168.2.2; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
    runtime:
      interfaces:
        - name: wg0
          add: wireguard
          createInRoot: true
      bindNS:
          net: /run/netns/wg
services:
  - name: getty
    image: linuxkit/getty:797cb79e0a229fcd16ebf44a0da74bcec03968ec
    net: /run/netns/wg
    env:
     - INSECURE=true
  - name: nginx
    image: nginx:alpine
    net: /run/netns/wg
    capabilities:
     - CAP_NET_BIND_SERVICE
     - CAP_CHOWN
     - CAP_SETUID
     - CAP_SETGID
     - CAP_DAC_OVERRIDE
    runtime:
      mkdir: ["/tmp/created-directory"]
files:
  - path: etc/wireguard/wg0.conf
    contents: |
      [Interface] 
      PrivateKey = yAnz5TF+lXXJte14tji3zlMNq+hd2rYUIgJBgB3fBmk= 
      ListenPort = 51820
      [Peer] 
      PublicKey = xTIBA5rboUvnH4htodjb6e697QjLERt1NAB4mZqp8Dg= 
      AllowedIPs = 0.0.0.0/0
```

This creates a wireguard interface in the root namespace, moves it to the `wg` container and then keeps that network namespace so it can be used for the `getty` and `nginx` services.

orcas


justincormack:
Yes, thats why I said ignore them... I just opened this for discussion about the interface overall.

rn:
Another though which occurred to me while cycling home...Maybe some of this would be easier if we'd expose an interface to the [`runc hooks`](https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks) and then we could probably implement at least the networking setup there without having to add them to the YAML.

justincormack:
I made it like the `containerd` model, where you do things between create and start, rather than the `runc` hook model. `runc` hook scripts would need to be in the host namespaces, which is a bit annoying.

rn:
I think the mechanism (hook vs explicitly calling between create and start) does not matter as much. I was more thinking about instead of trying to cover some cases (like network config) in the YAML, like:
```
    runtime:
      interfaces:
        - name: wg0
          type: wireguard
          add: move
          up: true
```
we could do something like this:
```
    runtime:
      pre-hooks:
        - command: ["/sbin/prep-netns", "-name", "wg0", ...]
```
The `prep-netns` would need to be added via an init section (or comes by default in the `init` package) and they would be run in the host namespace, similar to `prepareNetNS`.

This way we don't need to hard code different cases in the YAML/service command.

justincormack:
That would basically be equivalent to committing to having `ip` in `init`, although I think we would still need to do the `bindNetNS` part for it, as that would be the functionality we would need to provide (I cant see a way with `ip` for creating a bind from a particular pid, and you would have to template the pid). Although `ip` doesnt provide a sane way of moving interfaces while keeping config, while it does provide things that are not so useful, like setting addresses, that is not really useful as you generally don't statically know what they are.

If we provide a command that is like `ip` it will just end up being `ip`. I was trying to provide the bare minimum that means you can run all the commands in containers without privileges, so we don't need to expand the scope of `init`.

We could remove interface create and up support, as that is easy to do in an `ip` container, and just leave `move` and `bindns`. We could also simplify it by removing the keep settings on move, which are not really needed (you can run dhcp in the namespace after move for the common case).

In which case we would leave just:
```
interfaces: wg0
bindNetNS: /run/wg
```
and remove most of the code, add an `ip` container (maybe just rename the `wg` one which is just `ip`+`wireguard-tools`).

justincormack:
Ok as an interim measure, I got rid of the preserve addresses on move, the `up` option, so it is much more minimal.

justincormack:
Ok I simplified the `interface` config further it now has:
- `name` - name of the interface
- `add` - if non blank, specifies type of interface to create
- `createInRoot` - special wireguard case, create in root namespace before moving.

So you can either take an existing interface and move it into a container - just specify the `name`, or create an interface in the container, specify `name` and `add` which has the interface type, plus the special case for wireguard which creates in root then moves. You can't create an interface and leave it in the root namespace (unless your container is running in root namespace).

justincormack:
Ok, I think this is the minimal sane API.

We could add general scripts. At the moment there are two phases, one of which gets the path, the other gets the pid, so there could be two corresponding script phases. Given there is no use case yet for scripts, I think I might leave them for now...

rn:
I think this looks cleaner (and is less code, which is good)

A couple of comments/questions:
- If I wanted to use this to set up a bridge and `veth` pairs to network some system containers, this wouldn't work with this approach because we can only create simple network interfaces, right?
- It would be good (and would make reviewing easier) if the move of the vendored code would be a separate commit, Having one large commit is quite tedious.
- This would also need an example and some documentation before the WIP is removed

justincormack:
Yes, I was just thinking about `veth`. I think it will almost work, will take a look.

Will try to clean up the commits a bit...

Yes, will add examples, docs and the other things once its finalized

justincormack:
Supporting `veth` creation requires adding a `Peer` field to name the other end, and then choosing which end to move, so fairly simple. Creating a bridge and adding veths to it probably is best done in a container as it needs to sort out addresses etc.

rn:
Yeah, I was mostly concerned about the `veth` pair as one of the peers needs to be moved into the namespace. The bridge and other stuff should indeed be configured within another container.

justincormack:
Ok added veth pair support. Will clean up commits.

justincormack:
cleaned up the commits. Still to add update of moby tool, `wg` container.

pkg/init/cmd/service/prepare.go
```diff
@@ -0,0 +1,220 @@
...
+	nsenterPath = "/usr/bin/nsenter-net"
+)
+
+// Note these definitions are from moby/tool/src/moby/config.go and should be kept in sync
+
+// Runtime is the type of config processed at runtime, not used to build the OCI spec
+type Runtime struct {
+	Mkdir      []string    `yaml:"mkdir" json:"mkdir,omitempty"`
+	Interfaces []Interface `yaml:"interfaces" json:"interfaces,omitempty"`
+	BindNetNS  string      `yaml:"bindNetNS" json:"bindNetNS,omitempty"`
```
ijc: Might there be cause to bind other types of ns in the future (pid, etc?)? Perhaps make it like:
```
  bindns:
     - net
     - pid
```
Which I think is just `BindNS []string` here.

pkg/init/cmd/service/prepare.go
```diff
@@ -0,0 +1,220 @@
...
+const (
+	wgPath      = "/usr/bin/wg"
+	nsenterPath = "/usr/bin/nsenter-net"
+)
+
+// Note these definitions are from moby/tool/src/moby/config.go and should be kept in sync
+
+// Runtime is the type of config processed at runtime, not used to build the OCI spec
+type Runtime struct {
+	Mkdir      []string    `yaml:"mkdir" json:"mkdir,omitempty"`
```
ijc: This is a _host_ path, I think? I only know this because we discussed it once IRL, a less well connected user might find this surprising though so it might need mentioning somewhere, probably docs in preference to e.g. a comment here (although both would be better).

pkg/init/cmd/service/prepare.go
justincormack: you need to specify the path too, so it would be 
```
bindns:
  - net: /run/net
  - pid: /run/pid
```
which is a bit more verbose. It is trivial to implement the rest so maybe it makes sense.

justincormack:
Changed to allow binding any type of namespace, not just network namespaces.

justincormack:
Added an `ip` container, with wireguard, (not yet pushed, waiting for signing setup).

justincormack:
`moby` PR https://github.com/moby/tool/pull/144

pkg/init/cmd/service/prepare.go
justincormack: what do you mean "host"? it is a path in the running LinuxKit machine...

justincormack:
cc @zx2c4 I think this is pretty nice for setting up wireguard now, as well as working for other types of networking setup. As I mentioned in Slack, the only reason I can't use `wg-quick` is that it errors if the interface already exists, which is a bit annoying as I need to create it myself in this setup as I want to control the namespace setup, and move it myself as that is much simpler.

pkg/init/cmd/service/prepare.go
justincormack: Ah I see what you mean. Added a comment to that effect in the `moby` docs.

justincormack:
Added hash updates for the init, runc, containerd - should run the tests with the changes, although there are no runtime configs yet, as I need to add the moby update first.

pkg/init/cmd/service/prepare.go
ijc: It occurred to me overnight that perhaps naming the field `HostDirectory` might be more self-documenting (plus I think it is better to name it for the "resource" rather than the action needed to create the resource IYSWIM).

justincormack: I think that `mkdir` is a bit clearer as it really is doing that, `HostDirectory` sounds like a configuration option...

ijc:
What do you think about (perhaps in a subsequent PR) adding support for a per-service `files` section similar to the global one but operating on paths within the container (perhaps supported only for rw containers for now)? Currently to create a file in a container you need to either hardcode knowledge about the layout (`containers/services/getty/lower/«...»`) or do a two step to create a file in the host and then bind it into the container (which may also require duplicating the binds from the image label).

justincormack:
@ijc thats definitely possible. There is some overlap with the discussion in https://github.com/linuxkit/linuxkit/issues/2151 about having a `data` bind which is easily additive to `binds` and is like a named volume.

justincormack:
If everyone is reasonably ok with this as is please let me know and I will sort out the remaining things that need doing on this PR...

zx2c4:
Hey guys -- I'm a bit out of connectivity on holiday in the mountains but
will be back online in a few weeks and will dig into this PR. Please wait
for my review before merging.

pkg/init/cmd/service/main.go
```diff
@@ -66,8 +70,17 @@ func main() {
...
-		systemInitCmd(args)
-		os.Exit(0)
+		// check if called form startup scripts
+		command := os.Args[0]
+		switch {
+		case strings.Contains(command, "onboot"):
+			os.Exit(runcInit(onbootPath))
+		case strings.Contains(command, "onshutdown"):
+			os.Exit(runcInit(shutdownPath))
+		case strings.Contains(command, "containerd"):
```
ijc: Having all of these be `strings.HasSuffix` might have less potential for confusion?

pkg/init/Dockerfile
```diff
@@ -1,17 +1,28 @@
...
 ENV GOPATH=/go PATH=$PATH:/go/bin
 
-COPY init.go /go/src/init/
-COPY vendor /go/src/init/vendor/
-RUN go-compile.sh /go/src/init/
+COPY cmd /go/src/cmd
+RUN go-compile.sh /go/src/cmd/init
 
-FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS mirror
+# checkout containerd for vendoring
```
ijc: Not completely related to this PR, but I wonder if we should just do the clone in the alpine base andthen consumers can just use it (or copy it, or symlink it)

justincormack:
@rn it just needs some examples, updating `moby` etc.

pkg/init/Dockerfile
justincormack: Yes, I was wondering about that too, but thought I might do in a follow up.

ijc:
The init/service changes LGTM as does the config format. Don't have much of an opinion on the network container stuff (but seems fine)

pkg/init/cmd/service/prepare.go
ijc: OK

ijc:
Another possible usecase for this sort of thing I just tripped over is that the `project/kubernetes` really wants to have writeable `/opt` and `/home/kubernetes`, so perhaps having a `tmpfs` type thing here might be interesting. Haven't quite got to the bottom of it (I suspect switching to .iso/really-ro might have broken this).

justincormack:
@ijc shouldnt those just be bind mounts to somewhere writeable or tmpfs mounts?

ijc:
@justincormack possibly, I'm still investigating. Right now it does look like it really does mean the _host_ paths, i.e. it expects to be able to escape from the `kubelet` container and use the real host versions of these things. There's a `--containerized` option to `kubelet` which we aren't using but there seems to be some other magic related to the use of `/rootfs` in the `kubelet` container which I'm still figuring out.

justincormack:
Added wireguard example and updated to include the moby tool with runtime support.

pkg/ip/Makefile
```diff
@@ -0,0 +1,3 @@
+IMAGE=ip
```
rn: Maybe it's better to call this `netcfg` or similar? Don't have a particularly strong view on this, though.

pkg/init/cmd/service/main.go
justincormack: This is how they were previously in `onshutdown`; they could be changed in a followup PR.

ijc:
I think all my comment have either been address, explained or explicitly punted to a future PR.

justincormack:
@zx2c4 going to merge this now to unblock some other updates, but we can always make changes later, so please review...

PR #2422, opened 2017-08-16
Add documentation for how to run a single node Kube cluster

justincormack:
As I keep forgetting where this is documented...

cub


ijc:
Thanks! If I had seen this in time it would have gotten an enthusiastic LGTM from me...

PR #2427, opened 2017-08-17
package.mk: Add docker build option '--build-arg'

arm64b:
We need a method to add the docker build options '--build-arg'
in top 'package.mk' file, thus we will have chance to provide
some '--build-arg' arguments in the local Makefile like this:

BUILD_ARG=--build-arg BASE=$BASE

**- What I did**
Add '--build-arg' option based on package.mk
**- How I did it**
Add a new variable 'BUILD_ARG' into package.mk
**- How to verify it**
Add below line into a local Makefile:
`BUILD_ARG=--build-arg BASE=aarch64/ubuntu:16.04`
then `make --just-print` to print the make step message, we'll find the `docker build` will become
```
docker build --build-arg BASE=aarch64/ubuntu:16.04 --label org.opencontainers.image.source=https://github.com/linuxkit/linuxkit  --network=none...
```

summit


rn:
Could you elaborate where you need this?

arm64b:
@rn I am preparing the `mkimage-iso-efi` tool patch, where I will use below structure in the local Dockerfile:
```
ARG BASE=ubuntu:16.04
FROM $BASE AS grub-build
...
FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS make-img
...
COPY --from=grub-build /grub*.efi  /usr/local/grub
```
In the Makefile, I need to construct BUILD_ARG like this:
```
ARCH := $(shell uname -m)
ifeq ($(ARCH), x86_64)
BASE=ubuntu:16.04
endif
ifeq ($(ARCH), aarch64)
BASE=aarch64/ubuntu:16.04
endif

BUILD_ARG=--build-arg BASE=$(BASE)
```

ijc:
Is the pattern always that if `BASE` is `foo` on x86 then it is `aarch64/foo` on arm64? IOW could we get away with (in `package.mk`):
```
ifneq ($(BASE),)
ifeq ($(ARCH),arm64)
BASE=aarch64/$(BASE)
endif
endif
```
IOW if the package specifies `BASE` and we are running on `arm64` then unconditionally prefix.

ijc:
Also, IMHO whenever we are playing tricks like this to support multiple architectures there should always be a corresponding code comment with a reference to the issue which has been filed against the relevant hub/library/image/whatever requesting that it be made properly multiarch.

arm64b:
> s the pattern always that if BASE is foo on x86 then it is aarch64/foo on arm64?
a quick answer is NO. for example, we have aarch64/ubuntu:16.04, but for alpine:3.6 we have arm64v8/alpine:3.6, a little bit confusion but it's reality :(

arm64b:
We're now trying to seek the help from linaro org as the official release agency for Docker Images hosted in the hub...

justincormack:
@arm64b also @tianon can help with official images issues

arm64b:
@justincormack Now the LinuxKit already support the multi-arch docker image, it's time to promote this method so we will not be bothered by the prefix like `aarch64` or `arm64v8` 👍

tianon:
If you look at the image description for `aarch64/ubuntu`, it clearly states that all images under that organization are deprecated.  Please switch to `arm64v8/ubuntu` if you would like up-to-date images.

arm64b:
close it since a new PR is replacing this one

PR #2429, opened 2017-08-17
Allow specifying a set of metadata providers

justincormack:
run with `cmd: ["metadata", "aws"]` to just check for AWS metadata for example.

elephant


justincormack:
Added updated hashes

PR #2436, opened 2017-08-18
Update Kernel to 4.9.44/4.4.83

rn:
Kernels already compiled/pushed/signed for ARM and x86

It's a rainy kinda day
cheetah


rn:
Turns out I missed the update to the Makefile. Force pushed and update and re-compiling kernels now. Will then hit rebuild on CI

PR #2437, opened 2017-08-18
Bump to containerd v1.0.0 alpha5

ijc:


ijc:
I missed that init acually needs real code changes. Investigating.

ijc:
The IO to a log file stuff in https://github.com/linuxkit/linuxkit/blob/master/pkg/init/cmd/service/start.go#L105...L120 has changed a lot, trying to figure out if there is a good way to rework it or if I need to expose some new primitive from the containerd library.

ijc:
Minor rework of container IO handling added. I booted the toplevel `linuxkit.yml` and confirmed that `/var/log/rngd.log` contained the usual complaint about lack of an entropy source which I think validates it is basically ok.

ijc:
I think the linuxkit.packages.getty-containerd failure is a real containerd issue which was resolved post alpha5 by https://github.com/containerd/containerd/pull/1391 or https://github.com/containerd/containerd/pull/1396 (I've run the test manually with a containerd 8095244c26fa2daaef850be862e5b1b56d7cec66 outside of linuxkit to check). I guess our choices are to defer the update until alpha6, pull in containerd master or to mark that test as skip for now.

I'm going off on PTO for 1 week in ~30 minutes. I'll push a patch adding the skip for now in case you want to merge that, but I think it might be better to either bump to 8095244c26fa2daaef850be862e5b1b56d7cec66 or just wait for alpha6.

justincormack:
I guess alpha6 will probably be quite soon given this, so I think will wait.

ijc:
:+1: feel free to carry while I'm away, or else I will tackle it as soon as I am back.

justincormack:
Closing as still no alpha6, and needs a rebase.

PR #2438, opened 2017-08-18
Add Packet metadata and bonding support

### LINKED ISSUE #2379: blacklist drivers for packet
As discussed with  @TheOriginalAlex and `manny.packet` on the [packet.net community slack](https://packetcommunity.slack.com/archives/C4FEV702E/p1502139597080318) on the type 2 servers there are onboard ixgbe NICs which are not used and Mellanox cards which are used in a bonded config. When booting LinuxKit on on a Type 2 it seems that `mdev` will assign `eth0` to the non-connected, onboard  `ixgbe` NIC. One option is to blacklist the `ixgbe` driver.

Maybe the best option is to have a `packet-type-2` package in LinuxKit which can be added to the `init` section which adds the relevant rules.

justincormack:
This adds metadata support for Packet.net, with support for using the bonding information.

~~~I am going to test this a bit further as I seem to be getting some packet loss on Type 2 machines; it may be an issue with the setup, or it could be a different issue.~~~

cc @vielmetti 

fix #2379
see also #2064

![bonding](https://user-images.githubusercontent.com/482364/29476924-1234a160-845e-11e7-8930-23c189b86ec2.jpg)

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -10,7 +10,7 @@ init:
 onboot:
   # support metadata for optional config in /var/config
   - name: metadata
-    image: linuxkit/metadata:0c29c3ac1e5d2ee591d9710eca987f7df3fb9ca6
+    image: linuxkit/metadata:da3138079c168e0c5608d8f3853366c113ed91d2
   - name: sysctl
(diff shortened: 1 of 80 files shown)


justincormack:
Ah I think I know what the issue is...

justincormack:
Ok, fixed the issue, this is now setting up bonding correctly with no errors on type 2. 20Gb networking is quick!

PR #2440, opened 2017-08-20
Improvements to the kernel build

rn:
- For consistency always use a architecture suffix for the kernel config file
- Verify that the kernel config did not change after `make defconfig/oldconfig`. This should catch when a config options got added/removed/changed during kernel updates.
- Add a `fetch` target which downloads the kernel source tarballs. These local copies are then used for the build if present. This speeds up testing/experimenting with kernel builds/configuration as it saves the repeated download. The signature of the source tarball are still verified. 

@tych0 I think you requested this last feature a while back :)

dog-flowers


justincormack:
I did notice that kernel downloads only take half a second on the Packet machines with 10Gb ethernet when testing yesterday :)

rn:
not always on packet :)

I was also considering to skip verification of the content if the kernel is locally supplied as this takes some time too, but then decided against. I'll likely add a `Dockerfile.kconfig` to make it easier to experiment with kernel configurations and will skip it there.

PR #2442, opened 2017-08-21
Add some more tools to base image and only clone containerd once

justincormack:
Add the extra tools into base image that Kubernetes is using; currently it clones from upstream Alpine instead of our base image.

As discussed in comments in #2413 it is better to just clone `containerd` once in the base, not the three times it was being cloned.


justincormack:
Just doing the package updates

justincormack:
Package updates added

PR #2445, opened 2017-08-21
Kernel config improvements

rn:
- Add a `Dockerfile.kconfig` which builds an image with all patched kernels and kernel config files. This can be used to experiment/modify the kernel configurations
- Align both the arm64 and the x86_64 4.4 kernel configs with the 4.9 equivalent, reducing the difference between them.
- Add a utility to split kernel configurations in to common and config specific files. This is mostly To aid with aligning the configs
- Update kernel documentation throughout.

swans


rn:
I've tested the 4.4 kernel on arm64 and it works. I have't pushed any new kernel images. Will do that with the inevitable kernel update later this week

scripts/kconfig-split.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+
+# This is a slightly modified version of ChromiumOS' splitconfig
+# https://chromium.googlesource.com/chromiumos/third_party/kernel/+/stabilize-5899.B-chromeos-3.14/chromeos/scripts/splitconfig
```
riyazdf: cc @tych0 this might be of interest to you

rn: @tych0 and others. I will update #635 either tonight or tomorrow. I have some more thoughts which I would like us to discuss

PR #2314, opened 2017-07-28
Use a tmpfs for /var

justincormack:
Previously we were cheating and remounting /var `rw` but this does not
work if the filesystem is really read only. Nount a tmpfs, which may
be overmounted later by a persistent filesystem.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

fix #2288 

racoon

PR #2449, opened 2017-08-22
docker-for-mac: supply the daemon.json via metadata

djs55:
**- What I did**

Fixed a bug in `blueprints/docker-for-mac` which prevented the Docker daemon from starting if you follow the instructions in the `README.md`

**- How I did it**

The Docker daemon needs a `daemon.json` file. Previously a default trivial file was supplied by a `files` declaration in the `.yml` which wrote it to `/var/config/docker/daemon.json` -- a path which is really owned by the `metadata` package. This trick used to work but doesn't work any more -- the `/var/config` directory is empty and the daemon doesn't start. We may have been relying on undefined behaviour?

This patch tweaks the `README.md` to supply the `daemon.json` via the `metadata` package which is the mechanism a real deployment is expected to use. It also adds the `ca-certificates` package to make the `metadata` package start properly, as described in #2443 

**- How to verify it**

Follow the instructions in the `README.md`
```
cd blueprints/docker-for-mac
moby build -name docker-for-mac base.yml docker-17.06-ce.yml
linuxkit run hyperkit -networking=vpnkit -vsock-ports=2376 -disk size=500M -data ./metadata.json docker-for-mac
```

and observe that the `docker-dfm` task is running:
```
ctr t ls
TASK                    PID     STATUS    
vsudd                   1003    RUNNING
acpid                   697     RUNNING
docker-dfm              738     RUNNING
getty                   789     RUNNING
host-timesync-daemon    837     RUNNING
ntpd                    878     RUNNING
trim-after-delete       918     RUNNING
vpnkit-forwarder        959     RUNNING
```

### DIFF
--- a/blueprints/README.md
+++ b/blueprints/README.md
@@ -39,7 +39,7 @@ linuxkit run hyperkit -networking=vpnkit -vsock-ports=2376 -disk size=500M -data
 In another terminal you should now be able to access docker via the socket `guest.00000947` in the state directory (`docker-for-mac-state/` by default):
 
 ```
-$ docker -H unix://docker-for-mac-state/guest.00000947 ps
+$ docker -H unix://docker-for-mac-state/guest.00000948 ps
 CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
 ```
 
--- a/blueprints/README.md
+++ b/blueprints/README.md
@@ -33,7 +33,7 @@ $ moby build -name docker-for-mac base.yml docker-ce.yml
 To run the VM with a 500M disk:
 
 ```
-linuxkit run hyperkit -networking=vpnkit -vsock-ports=2376 -disk size=500M docker-for-mac
+linuxkit run hyperkit -networking=vpnkit -vsock-ports=2376 -disk size=500M -data ./metadata.json docker-for-mac
 ```
 
 In another terminal you should now be able to access docker via the socket `guest.00000947` in the state directory (`docker-for-mac-state/` by default):
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -24,12 +24,6 @@ onboot:
   - name: mount
     image: linuxkit/mount:4fe245efb01384e42622c36302e13e386bbaeb08
     command: ["/usr/bin/mountie", "/var/lib"]
-  # create docker dir on mounted drive if it doesn't exist
-  - name: mkdir-docker
-    image: alpine:3.6
-    binds:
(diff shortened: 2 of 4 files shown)


justincormack:
I can't see any obvious reason why this should break - there is nothing obvious wrong with the default file, would rather understand why it broke...

justincormack:
Also while you are changing things, the bit
```
  # create docker dir on mounted drive if it doesn't exist
  - name: mkdir-docker
    image: alpine:3.6
    binds:
        - /var/lib:/host_var_lib
    command: ["sh", "-c", "mkdir -p /host_var_lib/docker"]
```
is not needed any more, you can add
```
  runtime:
     mkdir: ["/var/lib/docker"]
```
to the docker container instead to create the directory if it is missing.

djs55:
I did a bit of searching and narrowed the `/var/config` problem down to between the following merge commits:
- f436e66a77ce8fb3d412c33c8ac9e8f2ab822dc3 -- this works
- 9f66f3392200a106eadbf4c03813c628d98e2a1a -- this fails

I think the problem is that we switched from keeping `/var/` and remounting read/write to using a `tmpfs` in https://github.com/linuxkit/linuxkit/pull/2314/commits/2c20ee2f8a2122989d13a5477ac5cc83eeb186f3 -- the `tmpfs` mount will shadow files already in `/var/config/docker`.

justincormack:
Ah I see yes, you can't really specify `files` in `/var` any more (like `/tmp` and `/run`). This probably ought to be documented...

This needs a rebase.

ijc:
> Ah I see yes, you can't really specify files in /var any more (like /tmp and /run). This probably ought to be documented...

Might be another use case for a files stanza in the `runtime` section?

ijc:
or could be done with a global `files` to create `/local-overlay` and then use the new `runtime.mounts` stuff to do it -- bit round the houses though and getting a bit OT for this PR.

MagnusS:
Could this be merged as is, with additional `files:` changes later? It LGTM and would be good to get it in to fix the README example and avoid conflicts.

ijc:
Yeah, my comments were certainly around possible future directions and not something needed for this PR, sorry for not making that at all clear.

This PR looks good to me, and I think @justincormack  was OK (I don't think he meant to hold up this PR for docs). I'll merge.

ijc:
> Ah I see yes, you can't really specify files in /var any more (like /tmp and /run). This probably ought to be documented...

@justincormack would it be sane to make /var an overlay with a tmpfs somewhere else?

justincormack:
I don't know about overlay, but maybe copying up the files makes sense.

PR #2451, opened 2017-08-22
Add support for mount in runtime config

justincormack:
This could be used in LinuxKit now, as there are some examples, eg
https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac/base.yml#L33
which are creating containers to do a mount.

Also removed the ad hoc code to generate overlay mounts for writeable containers
with a runtime config which does the same thing.

See https://github.com/moby/tool/pull/145

cheetah

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 PREFIX?=/usr/local/
 
 MOBY_REPO=https://github.com/moby/tool.git
-MOBY_COMMIT=0d58d332be0afc27be4402301f7c7950bd3ae189
+MOBY_COMMIT=69596e17ddb01fc3d3e75774bffeb760812b2f97
 MOBY_VERSION=0.0
 bin/moby: tmp_moby_bin.tar | bin
 	tar xf $<
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -4,7 +4,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/vpnkit-expose-port:fa4ab4ac78b83fe392e39b861b4114c3bb02d170 # install vpnkit-expose-port and vpnkit-iptables-wrapper on host
-  - linuxkit/init:2122f8b7202b383c1be0a91a02122b0c078ca6ac
+  - linuxkit/init:438d377802498afb3a6ed2a86175ab4beb07a31a
   - linuxkit/runc:a1b564248a0d0b118c11e61db9f84ecf41dd2d2a
   - linuxkit/containerd:8e4aa6c09e9bceee8300a315c23e0333e187f5fa
 onboot:
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.9.44
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:2122f8b7202b383c1be0a91a02122b0c078ca6ac
+  - linuxkit/init:438d377802498afb3a6ed2a86175ab4beb07a31a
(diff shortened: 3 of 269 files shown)


justincormack:
Updated to remove all the code that worked out how to do overlay mounts - this is now specified at build time by the `moby` tool and we just use the specified mounts.

justincormack:
The `moby` code will now share mounts from identical images, so using the same container twice in a system has no space penalty.

justincormack:
Ok this is ready to go other than updating `moby` and the images hashes. So it
- allows `mount` in runtime config
- `moby` uses that to set up the initial mounts needed to run a container (eg overlay)
- the code that needed to know about how `moby` wanted to set that up has gone
- `moby` now uses this to de-duplicate mounts of identical images, telling the runtime to only mount the base image once.

justincormack:
Ok, added updated hashes and a test of the updated tool (not merged yet) to make sure CI passes.

This should decouple the `moby` tool from the LinuxKIt `init` much better now as the actions to do are explicit, so updating should need less synchronisation in future.

Makefile
```diff
@@ -20,8 +20,8 @@ endif
 
 PREFIX?=/usr/local/
 
-MOBY_REPO=https://github.com/moby/tool.git
-MOBY_COMMIT=0d58d332be0afc27be4402301f7c7950bd3ae189
+MOBY_REPO=https://github.com/justin/tool.git
+MOBY_COMMIT=cfa5d273b7d5d45345d4312dfdc5ccf6247cd01e
```
rn: you removed the `WIP` from the title, but this still pulls your version

justincormack: yeah, I hadnt decided what to do. But yes, the last commit is WIP

justincormack:
Ok this is passing with unmerged `moby`

justincormack:
@rn any chance of getting a review? I cant finalize this without merging the `moby` PR which will break anyone using `moby` master without this being merged (which is the decoupling that this PR should address as in future can change them in a less sync fashion)

rn:
I think this looks OK, though I have to admit I don't fully understand the setup

justincormack:
Basically, rather than the service having a fixed idea of how to mount the container rootfs (which is currently either a simple bind mount for a read only container, or an overlay filesystem of a tmpfs for a read write one), the `moby` tool just passes a runtime config saying what to mount before running the container, so it has all the logic; rather than the init code guessing based on the filesystem layout it is explicit in the json.

So a read-only container will get a runtime.json saying:
```
cat /containers/onboot/001-sysctl2/runtime.json 
{
    "mounts": [
        {
            "destination": "/containers/onboot/001-sysctl2/rootfs",
            "source": "/containers/onboot/000-sysctl/rootfs",
            "options": [
                "bind"
            ]
        }
    ],
    "bindNS": {}
}
```

while a read write one will get
```
{
    "mounts": [
        {
            "destination": "/containers/onboot/003-dhcpcd/tmp",
            "type": "tmpfs",
            "source": "tmpfs"
        },
        {
            "destination": "/containers/onboot/003-dhcpcd/tmp",
            "options": [
                "remount",
                "private"
            ]
        },
        {
            "destination": "/containers/onboot/003-dhcpcd/rootfs",
            "type": "overlay",
            "source": "overlay",
            "options": [
                "lowerdir=/containers/onboot/003-dhcpcd/lower",
                "upperdir=/containers/onboot/003-dhcpcd/tmp/upper",
                "workdir=/containers/onboot/003-dhcpcd/tmp/work"
            ]
        }
    ],
    "bindNS": {}
}
```

which is the same sequence of mounts it would have guessed it had to do before.

rn:
OK, that makes sense and this is kinda what I understood from the commits.

justincormack:
ok, rebased and updated to final.

Issue #2322, opened 2017-07-29
4.12 kernel support


rn:
This issue is to track progress on tracking progress on supporting 4.12 now that 4.11 has been EOLed.

First step is get the patches applied. There is a new https://github.com/dcui/linux/commits/decui/msft-4.12.y branch for hyper-v sockets but it does not seem complete, so here is the status of the 4.11 patch queue applied to 4.12:

- 0001-tools-build-Add-test-for-sched_getcpu.patch
  Did not apply cleanly, maybe not needed. Fixed perf compilation

- 0002-vmbus-vmbus_open-reset-onchannel_callback-on-error.patch
  mew version: b20ffd850ad97 at git@github.com:dcui/linux.git

- 0003-vmbus-add-the-matching-tasklet_enable-in-vmbus_close.patch
  Already in as 5116f5e2e05cf("vmbus: re-enable channel tasklet")

- 0004-vmbus-remove-goto-error_clean_msglist-in-vmbus_open
  new version: 981ae0c39f4a0 at git@github.com:dcui/linux.git

- 0005-vmbus-dynamically-enqueue-dequeue-a-channel-on-vmbus.patch
  not in 4.12 and not in
  https://github.com/dcui/linux/commits/decui/msft-4.12.y
  Applies cleanly.

- 0006-hv_sock-implements-Hyper-V-transport-for-Virtual-Soc.patch
  new version: ca99361977429 at git@github.com:dcui/linux.git

- 0007-VMCI-only-try-to-load-on-VMware-hypervisor
  new version: 52d06e92fca7e at git@github.com:dcui/linux.git

- 0008-hv_sock-add-the-support-of-auto-loading.patch
  new version: 6c9b2ebd7fc79 at git@github.com:dcui/linux.git

- 0009-tools-hv_sock-2-simple-test-cases.patch
  drop. This is just test code

- 0010-vmbus-introduce-in-place-packet-iterator.patch
  Already in as f3dd3f4797652("vmbus: introduce in-place packet iterator")

- 0011-hvsock-fix-a-race-in-hvs_stream_dequeue.patch
  new version: f06df7f9be11c9 at git@github.com:dcui/linux.git

- 0012-hvsock-fix-vsock_dequeue-enqueue_accept-race.patch
  new version: 719999b4908874 at git@github.com:dcui/linux.git

- 0013-Drivers-hv-vmbus-Fix-rescind-handling.patch
  not in 4.12 and not in
[...]

PR #2452, opened 2017-08-22
Add initial support for 4.12.x kernels

rn:
Add support for 4.12 kernels.

- The Hyper-V/VMBus patches are taken from 4.11 and from https://github.com/dcui/linux/commits/decui/msft-4.12.y. 
- The kernel config for x86_64 was derived from 4.11 and then updated with latest changes. The arm64 config was done using the 4.9 config and then updated

I have only very lightly tested this (`linuxkit.yaml` on `hyperkit` and `qemu-aarch64`). I specifically have not yet tested the Hyper-V socket implementation and I would suspect potential issue. But nevertheless having initial support for 4.12 should be good enough to include.

wip #2322 

fox


rn:
Forgot to mention, the x86_64 kernels have already been pushed (as needed for CI) and the arm64 kernels are currently being build/pushed

rn:
debug build for 4.12 on arm64 failed. possibly on x86 as well. Will look into it tomorrow

rn:
closing this for now. The 4.4 kernels also don't build on x86. Need to investigate a bit more

PR #2455, opened 2017-08-23
Initial support for 4.12 kernels (and assorted fixes)

rn:
- Fix kernel compilation for debug kernels. Unsurprisingly the sanity check for the kernel config failed
- Fix compilation of 4.4 kernels. The VHOST_VSOCK code does not compile so disable it. It's not needed by the VSOCK code inside a VM anyway.
- Add support for 4.12 kernels.
  - The Hyper-V/VMBus patches are taken from 4.11 and from https://github.com/dcui/linux/commits/decui/msft-4.12.y.
  - The kernel config for x86_64 was derived from 4.11 and then updated with latest changes. The arm64 config was done using the 4.9 config and then updated

I have only very lightly tested this (linuxkit.yaml on hyperkit and qemu-aarch64). I specifically have not yet tested the Hyper-V socket implementation and I would suspect potential issue. But nevertheless having initial support for 4.12 should be good enough to include.

wip #2322

image


rn:
All kernels for x86 and arm64 have been compiled/pushed/signed

justincormack:
The new yaml file additions will conflict with `init` update in #2451

rn:
#2451 has not been merged. I'm easy either way re merge order

justincormack:
Yeah I was just noting it as there wont be a merge conflict but the hashes will be inconsistent

rn:
tests pass. @justincormack how do you want to handle the YAML files?

PR #2457, opened 2017-08-23
Use a custom script for wireguard

justincormack:
the `wg-quick` script did not quite work for us, so ship a slightly modified one that @zx2c4 made.

This makes the config much cleaner.

This does not show the nameserver options as we don't have DNS but it will update `/etc/resolv.conf` is writeable so this can also be bind mounted.

squirrel-on-wire

### DIFF
--- a/examples/wireguard.yml
+++ b/examples/wireguard.yml
@@ -13,11 +13,11 @@ onboot:
     image: linuxkit/dhcpcd:f3f5413abb78fae9020e35bd4788fa93df4530b7
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: wg0
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:722329e06dccb26348876a00ba5e2d247cad206b
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1/24; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
+    command: ["/usr/bin/wg-config", "configure", "/etc/wireguard/wg0.conf"]
     runtime:
       interfaces:
         - name: wg0
@@ -26,11 +26,11 @@ onboot:
       bindNS:
           net: /run/netns/wg0
   - name: wg1
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:722329e06dccb26348876a00ba5e2d247cad206b
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg1 up; ip address add dev wg1 192.168.2.2/24; wg setconf wg1 /etc/wireguard/wg1.conf; wg show wg1"]
+    command: ["/usr/bin/wg-config", "configure", "/etc/wireguard/wg1.conf"]
     runtime:
       interfaces:
         - name: wg1
@@ -58,7 +58,8 @@ services:
 files:
   - path: etc/wireguard/wg0.conf
     contents: |
-      [Interface] 
+      [Interface]
+      Address = 192.168.2.1/24
       PrivateKey = KG7EKkHDkp7THfW5mOXcZzo3RbGAjq+ARMfJGFm5G1s=
       ListenPort = 51820
       [Peer] 
@@ -68,6 +69,7 @@ files:
   - path: etc/wireguard/wg1.conf
     contents: |
       [Interface]
+      Address = 192.168.2.2/24
       PrivateKey = SCGCKDuTm4PMOw+LXdK/2s8mxnv145QHOohKRq3vc2A=
       ListenPort = 51821
       [Peer]
--- a/pkg/ip/Dockerfile
+++ b/pkg/ip/Dockerfile
@@ -16,4 +16,5 @@ ENTRYPOINT []
 CMD []
 WORKDIR /
 COPY --from=mirror /out/ /
+COPY linuxkit-configure-wireguard-iface.sh /usr/bin/wg-config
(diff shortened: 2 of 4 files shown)


pkg/ip/Dockerfile
```diff
@@ -16,4 +16,5 @@ ENTRYPOINT []
 CMD []
 WORKDIR /
 COPY --from=mirror /out/ /
+COPY linuxkit-configure-wireguard-iface.sh /usr/bin/wg-config
```
zx2c4: Please don't name this binary `wg-config`, as that's a name previously used and possibly used in the future by upstream. Something linuxkit-specific might be better, such as `linuxkit-configure-wireguard-iface`. Alternatively, maybe you prefer a short linuxkit prefix -- `lk-wg-config`?

test/cases/040_packages/023_wireguard/test-wireguard.yml
```diff
@@ -11,11 +11,11 @@ onboot:
     image: linuxkit/dhcpcd:f3f5413abb78fae9020e35bd4788fa93df4530b7
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: wg0
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:722329e06dccb26348876a00ba5e2d247cad206b
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1/24; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
+    command: ["/usr/bin/wg-config", "configure", "/etc/wireguard/wg0.conf"]
```
zx2c4: Same comment re:naming.

test/cases/040_packages/023_wireguard/test-wireguard.yml
```diff
@@ -24,11 +24,11 @@ onboot:
       bindNS:
           net: /run/netns/wg0
   - name: wg1
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:722329e06dccb26348876a00ba5e2d247cad206b
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg1 up; ip address add dev wg1 192.168.2.2/24; wg setconf wg1 /etc/wireguard/wg1.conf; wg show wg1"]
+    command: ["/usr/bin/wg-config", "configure", "/etc/wireguard/wg1.conf"]
```
zx2c4: Same comment re:naming.

test/cases/040_packages/023_wireguard/test-wireguard.yml
```diff
@@ -66,7 +66,8 @@ services:
 files:
   - path: etc/wireguard/wg0.conf
     contents: |
-      [Interface] 
+      [Interface]
+      Address = 192.168.2.1/24
```
zx2c4: Could add an example `DNS = ` entry in there?

pkg/ip/Dockerfile
justincormack: ah ok, I picked it at random!

test/cases/040_packages/023_wireguard/test-wireguard.yml
justincormack: Yes, that probably makes sense even though there is no DNS server actually available as it shows how to plumb it in...

examples/wireguard.yml
```diff
@@ -68,6 +69,7 @@ files:
   - path: etc/wireguard/wg1.conf
     contents: |
       [Interface]
+      Address = 192.168.2.2/24
       PrivateKey = SCGCKDuTm4PMOw+LXdK/2s8mxnv145QHOohKRq3vc2A=
       ListenPort = 51821
```
zx2c4: By the way, if you don't specify ListenPort, the kernel will choose one for you at random, which could be desirable for some.

test/cases/040_packages/023_wireguard/test-wireguard.yml
zx2c4: Actually, probably makes more sense to do that in `examples/wireguard.yml`.

zx2c4:
In case you'd like to also have a dynamically configured example, the [wireguard demo server](https://www.wireguard.com/quickstart/#demo-server) can be used to make something on the fly.

justincormack:
I am going on PTO for a week so won't have time to revise this until after, if anyone does have time feel free.

pkg/ip/linuxkit-configure-wireguard-iface.sh
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash
+#
+# Copyright (C) 2016-2017 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
+#
+# This file is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2
```
justincormack: We really do not want this to be GPL, we are not planning to ship GPL code in this repo. @zx2c4 can this be relicensed?

justincormack: (cant request changes on my own review!)

zx2c4: As this script is a derivative of something from the WireGuard upstream project, it will remain GPLv2. Fortunately, it's easy enough that you could cleanroom re-implement it in a different language if you really want.

But I think for this repo, having a bash script that's GPLv2 shouldn't wind up being a problem for you. It's not as though this is a critical .go file or .c file that's linked into some final executable object. Rather it's just a utility bash script that goes inside images.

Alternatively, you could host it elsewhere -- such as a gist? or something more robust? -- and just wget it in via the package script.

justincormack:
Not going to ship a GPL bash script, it is too much license compliance for users.

zx2c4:
That's too bad. Alright, I'll try to rework this pull request so that the script doesn't have to explicitly live in this repo.

PR #2458, opened 2017-08-24
Further improvements to the kernel build process

rn:
The main part of this PR allows customisation of the build process with out modifying the the local files. Specifically it allows specifying a `make EXTRA=-foo build_4.9.x-foo`. This will take `kernel_config-foo` and append it to the kernel config.

Other changes are:
- Add `NOTRUST=1` as an option to disable image signing (just like in `package.mk`
- Add the content of `EXTRA` to `LOCALVERSION` so that `uname -r` picks it up
- Rename `_dbg` to `-dbg`
- Fix of docker warning about empty continuations.

sealion


linuxkit.yml
```diff
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:4.9.44
```
riyazdf: should this still be `linuxkit`?

rn:
@riyazdf thanks for spotting. Should be fixed now

PR #2460, opened 2017-08-29
Bump to containerd v1.0.0 alpha6

ijc:
This carries all the changes from #2437 (bump to alpha5) and rolls in the bump to alpha6 (with no additional changes needed compared with #2437). This should fix the test regression which alpha5 had.

### DIFF
--- a/scripts/update-component-sha.sh
+++ b/scripts/update-component-sha.sh
@@ -15,12 +15,12 @@ Available modes: --hash and --image
 
 Replace by hash:
 	$0 --hash <OLD> <NEW>
-	Example: $0 8675309abcdefg abcdef567899
+	Example: $0 --hash 8675309abcdefg abcdef567899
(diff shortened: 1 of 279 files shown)


scripts/update-component-sha.sh
```diff
@@ -15,12 +15,12 @@ Available modes: --hash and --image
...
 Replace by hash:
 	$0 --hash <OLD> <NEW>
-	Example: $0 8675309abcdefg abcdef567899
+	Example: $0 --hash 8675309abcdefg abcdef567899
     	   Will replace all instances of 8675309abcdefg with abcdef567899
 
 Replace by image: $0 --image <IMAGE> <NEW>
 	$0 --image <IMAGE> <NEW>
-	Example: $0 linuxkit/foo abcdef567899
+	Example: $0 --image linuxkit/foo abcdef567899
```
riyazdf: thanks for fixing these too 👍

PR #2461, opened 2017-08-29
qemu: Support ISO boot at the same time as metadata

ijc:
Tested with:
 - boot from kernel + initrd with metadata (appears as sr0)
 - boot from iso with no metadata (root is sr0)
 - boot from iso with metadata (root is sr0, metadata is sr1)
 - boot from iso with metadata using containerized qemu

In all cases where it was present the metadata was correctly expanded to
/var/config.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -32,7 +32,7 @@ type QemuConfig struct {
 	Kernel         bool
 	GUI            bool
 	Disks          Disks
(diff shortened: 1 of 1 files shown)


ijc:
Looks like we don't have any tests for metadata at all (I can't find any tests using `-data` anywhere). I'll open an issue to track that and merge this as is.

ijc:
#2466 is the issue

PR #2464, opened 2017-08-30
Update DinD from 17.06.1-ce to 17.07.0-ce

rn:

owl 1


blueprints/docker-for-mac/docker-17.06-ce.yml
```diff
@@ -3,7 +3,7 @@ services:
   # Bind mounts /var/run to allow vsudd to connect to docker.sock, /var/vpnkit
   # for vpnkit coordination and /var/config/docker for the configuration file.
   - name: docker-dfm
-    image: docker:17.06.1-ce-dind
+    image: docker:17.07.0-ce-dind
```
MagnusS: This doesn't match the filename anymore... Perhaps it would be better to rename this to `docker-ce.yml` and update the docs so it can track latest release (or add a new `docker-17.07-ce.yml`)

rn: ah, good point, I just did a `sed` of all 17.06

rn: I renamed `docker-17.06-ce.yml` to `docker-ce.yml`. Is that better?

MagnusS: LGTM with the readme/test update. Probably better to not track a specific version here

PR #2465, opened 2017-08-30
Updates to okernel project build files

t-koulouris:
A (long-overdue) update to the okernel project build files:
 - enabled multi-stage building
 - cleaned-up example yaml file
 - updated README

No changes outside of okernel directory.

Signed-off-by: Theo Koulouris <theo.koulouris@hpe.com>


projects/okernel/Dockerfile.okernel
```diff
@@ -1,4 +1,4 @@
-FROM linuxkit/alpine-build-kernel:cfdd576c36a52ed2dd62f237f79eeedc2dd3697b@sha256:3fe08db373a9373ba1616a485858f01ebd2d7a3cb364a099d0ed8b45fa419da2
+FROM linuxkit/kernel-compile:1b396c221af673757703258159ddc8539843b02b@sha256:6b32d205bfc6407568324337b707d195d027328dbfec554428ea93e7b0a8299b AS kernel-build
```
rn: `linuxkit/kernel-compile` is no longer a thing. It would be better to replicate what's here https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile#L1

or probably https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile.kconfig#L2 might be sufficient

t-koulouris:
Hi Rolf, thanks for the suggestion.

I removed the dependency to linuxkit/kernel-compile. Although arguably this is not as important in a LinuxKit environment, we build the kernel with the CONFIG_MODULE_SIG option set, so we need openssl which is pulled from alpine edge. Please let me know if you think this will become an issue.

I took the opportunity to squeeze in a couple more fixes - again all local to the okernel directory.

Theo

projects/okernel/Dockerfile.okernel
```diff
@@ -1,13 +1,33 @@
...
+    kmod \
+    libelf-dev \
+    linux-headers \
+    sed \
+    tar \
+    xz \
+    zlib-dev \
+    openssl-dev
+
+RUN apk --no-cache add --repository http://dl-cdn.alpinelinux.org/alpine/edge/main openssl
```
rn: This makes the build non-reproducible as it pulls an unknown version over the internet, but we can leave this for now

t-koulouris: Agreed, I am not happy with this either.

Somehow openssl is not found if attempting to pull it alongside the rest of the packages in the first RUN statement. No such issue with openssl-dev. Which repos is linuxkit/alpine using internally - aren't they those for v3.6?

The other alternative is to unset CONFIG_MODULE_SIG for now in the kernel configuration and remove the dependency.

rn: I'm adding it to `linuxkit/alpine` now, but might take a day or so before it get's merged

t-koulouris: Great, I'll fix the Dockerfile once the merge goes through. Thanks

PR #2467, opened 2017-08-30
Update kernels to 4.12.9/4.9.45/4.9.46/4.4.84/4.4.85

rn:
This also includes a workaround for https://github.com/moby/moby/issues/34199 for the perf build which got accidentally broken in https://github.com/linuxkit/linuxkit/pull/2458/commits/8b84baf226b21583622edb7a759f85b29dbb0689

Also for 4.12.9 a new config option (`HARDLOCKUP_CHECK_TIMESTAMP`) was introduced which was added to the config file with the default being on.

baby-deer


rn:
x86 kernels all pushed/signed. arm64 kernels being compiled/pushed as I write this, so should be avail well before CI finishes

kernel/patches-4.4.x/0016-drivers-hv-Define-the-channel-type-for-Hyper-V-PCI-E.patch
```diff
@@ -40,10 +40,11 @@ diff --git a/include/linux/hyperv.h b/include/linux/hyperv.h
 index ae6a711dcd1d..10dda1e3b560 100644
 --- a/include/linux/hyperv.h
 +++ b/include/linux/hyperv.h
-@@ -1156,6 +1156,17 @@ int vmbus_allocate_mmio(struct resource **new, struct hv_device *device_obj,
+@@ -1155,6 +1155,17 @@ int vmbus_allocate_mmio(struct resource **new, struct hv_device *device_obj,
+ 			0xab, 0x99, 0xbd, 0x1f, 0x1c, 0x86, 0xb5, 0x01 \
```
riyazdf: is this intended?  Looks odd.

rn: I noticed this too, but I used the same commands I always use. Maybe some subtle difference in how git handles whitespace in patch files...The kernel compiled fine, though.

riyazdf: ok sounds good then 👍

PR #2472, opened 2017-08-31
tools: Add openssl, attr-dev, and libtirpc-dev to the alpine base image

rn:
- openssl is for the okernel project (see #2465)
- attr-dev and libtirpc-dev are require to configure ZFS to build the kernel modules

/cc @t-koulouris

frogs

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -3,6 +3,7 @@ alpine-baselayout
 alpine-keys
 apk-tools
 argp-standalone
+attr-dev
 autoconf
 automake
 bash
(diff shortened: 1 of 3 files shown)


rn:
I'll add a few more packages shortly

rn:
packages added

rn:
I've already build/pushed the packages. The hash now is the hash from the `x86_64` versions file (sans the `-amd64`)

rn:
I take the LGTM from @riyazdf as approval and merge

Issue #2333, opened 2017-08-01
ZFS support


matjohn2:
Trying to build kernel modules for ZFS using the moby/LinuxKit image `linuxkit/kernel-compile`

Have tried using the default image `linuxkit/kernel-compile:1b396c221af673757703258159ddc8539843b02b` and also a modified version where the image has been upgraded to Alpine edge.

Having issues with alpines musl vs glibc libraries and unable to compile & run tests for building / testing the module.

Could you please provide the Dockerfile for the linuxkit/kernel-compile image? I cant find it anywhere and i'm hoping looking through it may allow me to build a linuxkit compatible module via another userspace env.

Thanks,
Matt

justincormack:
We havent used the `kernel-compile` package for ages, you should be able to find the source in the repo history though.

The Alpine zfs patches are here https://git.alpinelinux.org/cgit/aports/tree/main/zfs?h=master which may help.

rn:
what is the error? Looks like we need to update the docs (and the kmod test) but basically the `./kernel/Dockerfile` now installs the relevant tools from `linuxkit/alpine`

matjohn2:
Thanks for the superfast response @justincormack !

Just to clarify, if `kernel-compile` hasn't been used for ages, is that now the wrong process for building modules?
(documented here: https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#compiling-external-kernel-modules ) 

If so, any pointers to the correct process would be helpful and i'll PR the docs when i figure it out :)

matjohn2:
*bump* looking for pointers to the "correct" tools/process to be using to build linuxkit compatible and packaged kernel modules?

Thanks.

rn:
@matjohn2 The ZFS kernel modules compile perfectly fine with the exisiting kernel headers. The main issue was finding all the alpine packages required for `configure` to pass and create the `Makefile`. Here is the Dockerfile which I uses:
```
FROM linuxkit/kernel:4.9.46 AS ksrc
FROM linuxkit/alpine:21f2d52a0e20a78a0a81bbf5bebc652a3ec21b01 AS build
RUN apk add \
    attr-dev \
    autoconf \
    automake \
    build-base \
    git \
    libtirpc-dev \
    libtool \
    util-linux-dev \
    zlib-dev

COPY --from=ksrc /kernel-dev.tar /
RUN tar xf kernel-dev.tar

# Note: ZFS and SPL commit must match and we pick the current
[...]

PR #2475, opened 2017-08-31
Add support for building ZFS kernel modules

rn:
This adds support for building the ZFS kernel modules. The modules can be build as a separate package with tags matching the kernel build version. The modules can be added to a linuxkit image by adding the `zfs-kmod` package to the `init` section of a YAML file.

The `zfs-kmod` package also contains the standard kernel modules with `depmod` run over all of them so that `modprobe` should work.

The kernel modules do not compile for debug kernels due to CDDL incompatibility, apparently (the ZFS `./configure` script complains about `CONFIG_DEBUG_LOCK_ALLOC`)

Note, the `zfs-kmod` package is not build as the standard kernel build and is *not* hosted under the `linuxkit` Hub org due licensing uncertainty. Potential users will have to build the package themselves. `./docs/kernels.md` contains some notes.

/cc @matjohn2 (re #2333)

piglet

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -307,3 +307,39 @@ nsenter -m/proc/1/ns/mnt ash
 
 Alternatively, you can add the `kernel-perf` package as stage in a
 multi-stage build to add it to a custom package.
+
+
+## ZFS
+
+The kernel build Makefile has support for building the ZFS kernel
+modules. Note, the modules are currently not distributed as standard
+LinuxKit packages and if you wish to use them you have to compile them
+yourself:
+
+```sh
+cd kernel
+make ORG=<foo> NOTRUST=1 push_zfs_4.9.x # or different kernel version
+```
+
+will build and push a `zfs-kmod-4.9.<version>` image to Docker Hub
+under the `ORG` specified. This package contains the all the standard
+kernel modules from the kernel specified plus the `spl` and `zfs`
+kernel modules, with `depmod` run over them, so they can be
+`modprobe`ed. To use the modules do something like this in your YAML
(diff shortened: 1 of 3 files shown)


justincormack:
We could make a "runtime `depmod`" setup - it could mount an overlay for example. Not sure if this is the best option. 

Couldnt you run `depmod` after building the zfs modules again? It would need to unpack the built modules and run `depmod` again, then copy the output of that to the final output which would then override the default. That doesnt work so well if you want to add several sets of kernel module addons though.

rn:
@justincormack I can certainly include ZFS in the standard kernel package and then depmod would be run. This PR create a seperate `zfs-kmod` package. I'm fine with including the ZFS kmods in the standard kernel package as well, if we decided to do so.

Runnning `depmod` takes some time so I don't really want to run it on boot with a suitable overlay by default

justincormack:
@rn no, I meant that this package could get all the modules from the build and then run depmod on the existing modules plus the zfs modules, then include the resulting files in its build, which will override the default ones.

I don't think including zfs in the default build is a good idea, in case people have legal concerns and want to totally ignore zfs.

rn:
Ah, I hadn't thought about this approach, but sounds doable. Will try next week when back.

rn:
OK, I've updated the commits:
- Don't build/push ZFS packages as part of the standard kernel build. Users have to build themselves.
- Add standard kernel modules to the package and run `depmod` so that `modprobe` should work
- Added some notes to the documentation

I've removed the `WIP/RFC` as I think this is ready for review/merge if folks are happy

PR #2477, opened 2017-09-01
okernel: build dockerfile now gets openssl package from alpine base image

t-koulouris:
Updated okernel build dockerfile to pull openssl package from alpine base image (added in #2472).


justincormack:
Do you specifically need `openssl`? Alpine defaults to `libressl` now and would encourage using that if possible.

t-koulouris:
The Linux kernel build process seems to explicitly use openssl to generate a key pair and sign kernel modules. I did try to use libressl instead but haven't been successful. If you are aware of a fix I'd be very happy to make changes.

justincormack:
Ah ok that makes sense, yes.

PR #2479, opened 2017-09-02
Setup the bond interfaces with proper default routes

### LINKED ISSUE #2446: packet: only add default gateway once
The code at https://github.com/linuxkit/linuxkit/blob/master/pkg/metadata/provider_packet.go#L156 is not quite right - only the first ipv4 and ipv6 addresses should set the default gateway, later public addresses are additional IP assignments and should not set the default gateway.

TheOriginalAlex:
We need to add the addresses first and then set the routes, that way if the private ip gets added last, we're not setting the private ip as the default route
And using netlink.Bond to create the bondded interface is more readable, I think.

Signed-off-by: Alex Johnson <hello@alex-johnson.net>

Fixes #2446 

**- What I did**
Cleaned up the packet metadata.

**- How I did it**
Created the bond0 interface using the netlink.Bond struct.  
Add addresses and then setup routes in two discreet loops, that way the private IP doesn't end up being the last route that's added.

**- How to verify it**
Run on a Packet Type 0 server.

Cleaned up and fixed Packet metadata/bonding

Cute Dolphin

### DIFF
--- a/pkg/metadata/provider_packet.go
+++ b/pkg/metadata/provider_packet.go
@@ -6,8 +6,8 @@ import (
 	"net"
 	"os"
 	"path"
-	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/packethost/packngo/metadata"
 	"github.com/vishvananda/netlink"
@@ -51,6 +51,10 @@ func (p *ProviderPacket) Extract() ([]byte, error) {
 		return nil, fmt.Errorf("Packet: Failed to write hostname: %s", err)
 	}
 
+	if err := syscall.Sethostname([]byte(p.metadata.Hostname)); err != nil {
+		return nil, fmt.Errorf("Packet: Failed to set hostname: %s", err)
+	}
+
 	if err := os.MkdirAll(path.Join(ConfigPath, SSH), 0755); err != nil {
 		return nil, fmt.Errorf("Failed to create %s: %s", SSH, err)
 	}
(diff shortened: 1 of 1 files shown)


justincormack:
re "- How to verify it - Run on a Packet Type 0 server." you need to run on type 2 as well.

pkg/metadata/provider_packet.go
```diff
@@ -84,6 +88,28 @@ func (p *ProviderPacket) Extract() ([]byte, error) {
...
+	// set up bonding
+	bond, err := netlink.LinkByName("bond0")
+	if err != nil {
+		if strings.Contains(err.Error(), "Link not found") {
+			newBond := new(netlink.Bond)
+			newBond.LinkAttrs.Name = "bond0"
+			newBond.Mode = netlink.BondMode(ni.BondingMode())
+
+			if err := netlink.LinkAdd(newBond); err != nil {
+				if !strings.Contains(err.Error(), "file exists") {
```
justincormack: you should be able to check if the error is EEXIST without using strings.Contains.

I still dont understand why netlink gives this error...

pkg/metadata/provider_packet.go
```diff
@@ -84,6 +88,28 @@ func (p *ProviderPacket) Extract() ([]byte, error) {
...
+				}
+			}
+			bond, err = netlink.LinkByName("bond0")
+			if err != nil {
+				return fmt.Errorf("Could not get link by name for bond0: %s", err)
+			}
+		} else {
+			return fmt.Errorf("Could not get link by name for bond0: %s", err)
+		}
+	}
```
justincormack: it would make sense to print a message saying "using existing bond0 interface" in this case.

justincormack:
@TheOriginalAlex hi, have you had a chance to test this on type 2? I have been a bit busy so I haven't yet, will try to get to it in a while...

justincormack:
I am working on an updated version of this.

TheOriginalAlex:
Sure!  Sorry, I haven't responded much.  Been busy.  Also, I got stuck on getting the default gateway to work for some reason.  Running the binary directly on the host worked, but wouldn't work inside the metadata container even though the metadata container has `CAP_NET_ADMIN` and `/dev` is mounted in it.

justincormack:
@TheOriginalAlex what problems were you seeing with the default gateway?

TheOriginalAlex:
It just wasn't getting set.  If I printed the output of nsenter.RouteList, I could see it, but it wasn't showing up if I ran `ip routes show` on the shell.

justincormack:
Routes seemed to be working for me ok, but I need to rebase against the new metadata package again.

PR #2483, opened 2017-09-04
wireguard: use custom wg-quick

zx2c4:
This makes things in general much cleaner. This is a reworking of #2457 from @justincormack , which avoids putting GPLd code directly in this repo, while also fixing up a few other small things.

Currently untested, but I wanted to get your feedback on the idea.

Note that downloading custom scripts and tarballs and whatnot via curl is already quite common throughout LinuxKit:

```
test/pkg/ltp/Dockerfile:RUN curl -fsSL -o ltp-full-${LTP_VERSION}.tar.xz ${LTP_SOURCE}
projects/clear-containers/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/selinux/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/landlock/kernel-landlock/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/shiftfs/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/ima-namespace/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/kernel-config/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/memorizer/kernel-memorizer/Dockerfile:RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
projects/ebpf/ebpf.build/Dockerfile:RUN curl -sSL -O https://fedorahosted.org/releases/e/l/elfutils/0.165/elfutils-$ELFUTILS_VERSION.tar.bz2
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /tmp/cni.tgz https://github.com/containernetworking/cni/releases/download/v0.5.2/cni-amd64-${cni_version}.tgz && \
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/etc/weave.yaml https://cloud.weave.works/k8s/v1.7/net?v=${weave_version} 
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubelet https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubelet && chmod 0755 /out/usr/bin/kubelet
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubeadm https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubeadm && chmod 0755 /out/usr/bin/kubeadm
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubectl https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubectl && chmod 0755 /out/usr/bin/kubectl
kernel/Dockerfile:RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
kernel/Dockerfile:RUN curl -sSL -o /wireguard.tar.xz "${WIREGUARD_URL}" && \
```

### DIFF
--- a/examples/wireguard.yml
+++ b/examples/wireguard.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:4.9.43
+  image: linuxkit/kernel:4.9.49
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
   - linuxkit/init:6d11a1f9d299d3425e78cce80dfba8b236d20412
@@ -13,11 +13,11 @@ onboot:
     image: linuxkit/dhcpcd:f3f5413abb78fae9020e35bd4788fa93df4530b7
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: wg0
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:044287be54a62d1590dcd4d22543c1a1031479e6
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1/24; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
+    command: ["/usr/bin/lk-wg-config", "configure", "/etc/wireguard/wg0.conf"]
     runtime:
       interfaces:
         - name: wg0
@@ -26,11 +26,11 @@ onboot:
       bindNS:
           net: /run/netns/wg0
   - name: wg1
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:044287be54a62d1590dcd4d22543c1a1031479e6
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg1 up; ip address add dev wg1 192.168.2.2/24; wg setconf wg1 /etc/wireguard/wg1.conf; wg show wg1"]
+    command: ["/usr/bin/lk-wg-config", "configure", "/etc/wireguard/wg1.conf"]
     runtime:
       interfaces:
         - name: wg1
@@ -40,7 +40,7 @@ onboot:
           net: /run/netns/wg1
 services:
   - name: getty
-    image: linuxkit/getty:797cb79e0a229fcd16ebf44a0da74bcec03968ec
+    image: linuxkit/getty:48870d0f92c65fecd6ddb18004d68d85f9b9cde3
     env:
      - INSECURE=true
     net: /run/netns/wg1
@@ -58,7 +58,9 @@ services:
 files:
   - path: etc/wireguard/wg0.conf
     contents: |
-      [Interface] 
+      [Interface]
+      Address = 192.168.2.1/24
+      DNS = 192.168.2.2
       PrivateKey = KG7EKkHDkp7THfW5mOXcZzo3RbGAjq+ARMfJGFm5G1s=
       ListenPort = 51820
       [Peer] 
@@ -68,6 +70,7 @@ files:
   - path: etc/wireguard/wg1.conf
     contents: |
       [Interface]
+      Address = 192.168.2.2/24
       PrivateKey = SCGCKDuTm4PMOw+LXdK/2s8mxnv145QHOohKRq3vc2A=
       ListenPort = 51821
       [Peer]
--- a/pkg/ip/Dockerfile
+++ b/pkg/ip/Dockerfile
@@ -1,5 +1,6 @@
-FROM linuxkit/alpine:a120ad6aead3fe583eaa20e9b75a05ac1b3487da AS mirror
+FROM linuxkit/alpine:1e67b8fdba849ed2f7e8537c13b7f280639623d6 AS mirror
 RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add curl
 RUN apk add --no-cache --initdb -p /out \
     alpine-baselayout \
     bash \
@@ -7,6 +8,9 @@ RUN apk add --no-cache --initdb -p /out \
     iproute2 \
     musl \
     wireguard-tools
+
+# We grab our version of wg-quick, called lk-wg-config.sh
+RUN curl -sSL -o /out/usr/bin/lk-wg-config https://gist.githubusercontent.com/zx2c4/3624de869ab7eaef3de5ea8f2b867be9/raw/de72b018f4f4548858ce6aae2898b34db0474221/lk-wg-config.sh && chmod 755 /out/usr/bin/lk-wg-config
 
 # Remove apk residuals
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
--- a/test/cases/040_packages/023_wireguard/test-wireguard.yml
+++ b/test/cases/040_packages/023_wireguard/test-wireguard.yml
@@ -11,11 +11,11 @@ onboot:
     image: linuxkit/dhcpcd:f3f5413abb78fae9020e35bd4788fa93df4530b7
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: wg0
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:044287be54a62d1590dcd4d22543c1a1031479e6
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1/24; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
+    command: ["/usr/bin/lk-wg-config", "configure", "/etc/wireguard/wg0.conf"]
     runtime:
       interfaces:
         - name: wg0
@@ -24,11 +24,11 @@ onboot:
       bindNS:
           net: /run/netns/wg0
   - name: wg1
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:044287be54a62d1590dcd4d22543c1a1031479e6
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg1 up; ip address add dev wg1 192.168.2.2/24; wg setconf wg1 /etc/wireguard/wg1.conf; wg show wg1"]
+    command: ["/usr/bin/lk-wg-config", "configure", "/etc/wireguard/wg1.conf"]
     runtime:
       interfaces:
         - name: wg1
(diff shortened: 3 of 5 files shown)


pkg/ip/Dockerfile
```diff
@@ -8,6 +8,9 @@ RUN apk add --no-cache --initdb -p /out \
     musl \
     wireguard-tools
 
+# We grab our version of wg-quick, called lk-wg-config.sh
+RUN wget -O /usr/bin/lk-wg-config https://gist.githubusercontent.com/zx2c4/3624de869ab7eaef3de5ea8f2b867be9/raw/de72b018f4f4548858ce6aae2898b34db0474221/lk-wg-config.sh
```
ijc: I don't think moving the file behind a webserver to be fetched at build time changes anything wrt the GPL obligations upon the resulting image which is being built (which AIUI is the objection to the script, although perhaps I was mistaken).

zx2c4: In that case, you're not allowed to use any GPL software in any images? Which obviously isn't the case, considering the use of bash, Linux, etc in some resultant images.

justincormack: No, we need it to be easy for people to get the source for the GPL code they are using. The way we have been doing this is to make sure it is all in Alpine packages or kernel source, where we can easily trace back from the source and tags to which source is being shipped.

ijc: > No, we need it to be easy for people to get the source for the GPL code they are using. 

That's trivially achieved for a shell script like this though, isn't it? It's basically a nop...

zx2c4:
Isn't Github's gist considered to be pretty easy to use?

If not, any suggestions for other easy places to host this file? I can
amend this patch set and host elsewhere.

pkg/ip/Dockerfile
zx2c4: > No, we need it to be easy for people to get the source

Isn't Github's gist considered to be pretty easy to use?

If not, any suggestions for other easy places to host this file? I can amend this patch set and host elsewhere.

zx2c4: Actually, it looks like you're already doing this tons of places using `curl`, so I think this method will be okay. I'll fix up this commit to use `curl` instead of `wget`.

```
test/pkg/ltp/Dockerfile:RUN curl -fsSL -o ltp-full-${LTP_VERSION}.tar.xz ${LTP_SOURCE}
projects/clear-containers/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/selinux/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/landlock/kernel-landlock/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/shiftfs/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/ima-namespace/kernel/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/kernel-config/Dockerfile:RUN curl -fsSL -o linux-${KERNEL_VERSION}.tar.xz ${KERNEL_SOURCE}
projects/memorizer/kernel-memorizer/Dockerfile:RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
projects/ebpf/ebpf.build/Dockerfile:RUN curl -sSL -O https://fedorahosted.org/releases/e/l/elfutils/0.165/elfutils-$ELFUTILS_VERSION.tar.bz2
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /tmp/cni.tgz https://github.com/containernetworking/cni/releases/download/v0.5.2/cni-amd64-${cni_version}.tgz && \
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/etc/weave.yaml https://cloud.weave.works/k8s/v1.7/net?v=${weave_version} 
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubelet https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubelet && chmod 0755 /out/usr/bin/kubelet
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubeadm https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubeadm && chmod 0755 /out/usr/bin/kubeadm
projects/kubernetes/kubernetes/Dockerfile:RUN curl -fSL -o /out/usr/bin/kubectl https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64/kubectl && chmod 0755 /out/usr/bin/kubectl
kernel/Dockerfile:RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
kernel/Dockerfile:RUN curl -sSL -o /wireguard.tar.xz "${WIREGUARD_URL}" && \
```

test/cases/040_packages/023_wireguard/test-wireguard.yml
```diff
@@ -11,11 +11,11 @@ onboot:
...
-    image: linuxkit/ip:b084fd2ada446015e00e43a441bb0ae41ff8b227
+    image: linuxkit/ip:722329e06dccb26348876a00ba5e2d247cad206b
     net: new
     binds:
       - /etc/wireguard:/etc/wireguard
-    command: ["sh", "-c", "ip link set dev wg0 up; ip address add dev wg0 192.168.2.1/24; wg setconf wg0 /etc/wireguard/wg0.conf; wg show wg0"]
+    command: ["/usr/bin/lk-wg-config", "configure", "/etc/wireguard/wg0.conf"]
     runtime:
       interfaces:
         - name: wg0
```
zx2c4: Another way of going about this would to simply reimplement lk-wg-config in Go, and then add ` - config: /etc/wireguard/wg1.conf` to this line, and have Go handle it from the yaml. However, it's probably easiest to just do it this way with the script.

pkg/ip/Dockerfile
ijc: My point was not that there is any reason to avoid curl/wget per se, just that if the only reason for doing so is to avoid licensing issues the it is not actually making any difference to the situation, the license problems are the same whether the file is directly in this repo or fetched at build time.

So either the licensing situation for distributing this script in an image is fine (from a license compliance PoV) and we can just put this file in this repo and move on or the licensing is not fine and fetching it at runtime does not somehow make it ok.

If there are other places in the code which are also using curl/wget (or any other build time fetch) solely in a (IMHO misguided) attempt to address some licensing issue or other then we should raise issues here so we can confirm whether or not remote fetching makes any difference to the license in question (it seems unlikely to me that it would for any normal OSS license).

Note that there are other legitimate reasons for fetching things at build time.

To be clear, my opinion is that the GPL is pretty trivially satisfiable for a shell script, about the only sticking point would be the last bit of clause 1:
> and give any other recipients of the Program a copy of this License along with the Program.
which we may or may not be doing depending on what is in the base image.

It looks like @justincormack might have a different opinion on that. It also seems he wants to address licensing more generally via https://github.com/moby/tool/issues/147 (and presumably that will also necessitate changes to various LinuxKit package too).

zx2c4: > just that if the only reason for doing so is to avoid licensing issues the it is not actually making any difference to the situation, the license problems are the same whether the file is directly in this repo or fetched at build time. So either the licensing situation for distributing this script

I think your premises might be flawed. The objection, in the first place, was that @justincormack didn't want to put GPL files actually inside this repo. I find that a bit silly, but it's not up to me. The solution was therefore just to move the file outside of the repo. Putting GPL code inside the final images has always been okay, and has never been what this discussion is about.

justincormack:
rebuilt CI to see what up as it timed out.

test/cases/040_packages/023_wireguard/test-wireguard.yml
justincormack: cant do that easily as this code is run outside the container namespace, so config gets lost.

zx2c4:
Thanks. Will tail along the output and see.

zx2c4:
```
[+ 62m 10s] [PASS    ] linuxkit.packages.mkimage 157.04s
[+ 62m 52s] [PASS    ] linuxkit.packages.sysctl 42.28s[2017-09-12 18:27.20] Switch was turned off, so killing process "ssh" "-i" "/run/secrets/builder-ssh" "-o" "UserKnownHostsFile=/dev/null" "-o" "StrictHostKeyChecking=no" "root@35.195.32.250" "/usr/local/bin/test.sh"
[2017-09-12 18:27.20] Removing test VM...
[2017-09-12 18:27.20] Running "/gcloud/destroy.py" "linuxkit-ci-2"
[2017-09-12 18:27.20] Fetching results
[2017-09-12 18:27.20] Running "scp" "-r" "-i" "/run/secrets/builder-ssh" "-o" "StrictHostKeyChecking=no" "root@35.195.32.250:/tmp/build/test/_results" "."
[2017-09-12 18:27.20] Switch was turned off, so killing process "scp" "-r" "-i" "/run/secrets/builder-ssh" "-o" "StrictHostKeyChecking=no" "root@35.195.32.250:/tmp/build/test/_results" "."
[2017-09-12 18:27.20] Error fetching results: (Failure Cancelled)
[2017-09-12 18:27.20] Results directory "/tmp/linuxkit-242919d4/_results" not found, so not saving
[2017-09-12 18:27.20] Failed: Exceeded timeout of 120.00 minutes
```

Hm.

zx2c4:
@justincormack This new commit should satisfy CI now.

zx2c4:
@justincormack I fixed things up and this works very well now in my testing. However, this commit uses builds/tags of linuxkit/getty and linuxkit/ip that you'll need to build and push before you can safely commit this.

riyazdf:
I've pushed & signed the new linuxkit/ip and linuxkit/getty images

zx2c4:
CI builds now. Merge time?

examples/wireguard.yml
```diff
@@ -1,6 +1,6 @@
 kernel:
-  image: linuxkit/kernel:4.9.43
-  cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
+  image: linuxkit/kernel:4.9.49
+  cmdline: "console=ttyS0"
```
justincormack: can we not change the `console=` line - although it can be annoying for debug, this is generally what we use so it works everywhere...

zx2c4: Yes, no problem, it doesn't make a difference to me. I'll submit a change now. However, things are not as consistent as you say:

```
linuxkit/examples $ grep -hr 'cmdline:.*console=' | sort | uniq -c | sort -nr
      7   cmdline: "console=ttyS0"
      6   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
      2   cmdline: "console=tty0 console=ttyS0"
      1   cmdline: "console=ttyS1"
      1   cmdline: "console=tty0"
```

zx2c4:
Latest round of review fixed up, CI green. @justincormack merge time?

PR #2485, opened 2017-09-05
tweak dockerfiles to not clone the whole kit and caboodle

friism:
* make curl silent
 * optimize to just clone what's required to build


ijc:
Seems to have been made obsolete by the split of lcow into a separate repo. Please rebase onto that if this change is still relevant.

PR #2491, opened 2017-09-07
Use Linux kernel 4.11.1 for LCOW for now

rn:
Newer 4.11.x kernels seem to trigger a bug in either Linux or the Hyper-V NVDIMM backend which causes the Linux container VM to fail silently. These commits resurrect the 4.11.x kernel build (again!) and then revert to 4.11.1 before removing the changes and update the LCOW YAML file accordingly.

Tested with `docker run --rm -ti busybox ls -l /`

This is obviously a temporary measure until the real issue has been root caused. Details are in the individual commit messages.

southeast-africa-monkey


rn:
forgot to mention the kernel is already pushed and signed

Issue #2488, opened 2017-09-07
-networking "tap, tap0" doesn't work on ARM64


arm64b:
**Description**
Run below command to run the linuxkit with '-networking' specified: 
```
root@entos-huawei-d05:/linuxkit#bin/linuxkit run qemu -uefi -fw /home/dennis/qemu/images/bios.bin -networking tap,tap0 -iso linuxkit-efi.iso
```
The system will not enter into shell, it stops here:
```
Welcome to LinuxKit

                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
          {                       /  ===-
           \______ O           __/
             \    \         __/
              \____\_______/

[    1.415704] 8021q: adding VLAN 0 to HW filter on device eth0
eth0: waiting for carrier
eth0: carrier acquired
DUID 00:01:00:01:21:43:b1:89:ca:ec:41:85:4a:85
eth0: IAID 41:85:4a:85
eth0: adding address fe80::4386:5262:2a40:c103
eth0: soliciting an IPv6 router
eth0: soliciting a DHCP lease
[    2.505711] IPv6: eth0: IPv6 duplicate address fe80::4386:5262:2a40:c103 detected!
[   44.033985] nf_conntrack: default automatic helper assignment has been turned off for security reasons and CT-based  firewall rule not found. Use the iptables CT target to attach helpers instead.
```
But the same command (ovmf path need to be changed) can work on x86.

**Steps to reproduce the issue:**
In Host (Ubuntu 16.04 on ARM64): create a br0 and tap0, the `brctl show` output:
```
root@entos-huawei-d05:/home# brctl show
bridge name     bridge id               STP enabled     interfaces
br0             8000.521995f091f8       no                   eth0
                                                                            tap0
```
br0 has a valid IP address which can communicate with outside world (eg. ping from another physical machine) while the ip address of eth0 has been flushed by `ip addr flush dev eth0`.

**Describe the results you received:**
The LinuxKit can not run into shell.

**Describe the results you expected:**
The LinuxKit can run into shell and have an valid IP which can communicate with external world, so I can access the `nginx` service externally. OR, if there's another solution can get that target is also OK for the time being. 

**Additional information you deem important (e.g. issue happens only occasionally):**

rn:
do you have a DHCP server on the network? If you remove the `dhcpd` container can you then up the the interface manually from the shell and have connectivity?

[...]

PR #2497, opened 2017-09-11
qemu: Fix the networking issue in 'tap' mode on arm64

arm64b:
This PR is used to fix the issue #2488.

Currently we use '-net' the old way to initialize a host nic
interface, this method will not work on arm64 platform(#2488 issue),
so we use the '-netdev' method which will work on both arm64 and amd64.

**- What I did**
Fix the networking issue in 'tap' mode on arm64 platform
**- How I did it**
1. Add a new networking device 2. Add a networking interface in QEMU backed by the device created in step 1.
**- How to verify it**
Test it on both arm64 and amd64 platform after applied this patch, both system can enter into shell with valid network connection with outside world, eg, we can access the nginx service remotely from another machine.

summit


arm64b:
Will check the CI failure before push a new PR.

PR #2500, opened 2017-09-12
Convert rc.init and rc.shutdown to Go

justincormack:
This removes more shell scripts to improve maintainability.

This now also works correctly in userspace, so it can be used for
running LinuxKit images in Docker and other such use cases.

It is a literal conversion of the shell scripts with a few small
tweaks.

zebroid


pkg/init/Dockerfile
```diff
@@ -28,8 +31,9 @@ ENTRYPOINT []
 CMD []
 WORKDIR /
 COPY --from=build /go/bin/init /
+COPY --from=build /tmp/bin /bin/
 COPY --from=build /go/bin/service /usr/bin/
 COPY --from=build usermode-helper /sbin/
 COPY --from=mirror /out/ /
 COPY etc etc/
-COPY bin bin/
+#COPY bin bin/
```
ijc: You can nuke this I think

justincormack: Ah yes I think that is left over from testing

deitch:
> This now also works correctly in userspace, so it can be used for
running LinuxKit images in Docker 

What does that mean?

> This removes more shell scripts to improve maintainability

What is the big advantage of a compiled go binary over a shell script in this instance?

Either way, the actual commit looks good here too (although I also didn't do line-by-line).

justincormack:
@deitch you can run LinuxKit images as docker containers, there is a small example here https://github.com/moby/tool/tree/master/examples - doing more work on other userspace runtimes.

deitch:
> you can run LinuxKit images as docker containers

Interesting. I guess `-format docker` leaves out the unnecessary kernel. Would be cool if the same format could run in all formats.

Since linuxkit images contain containerd and runc and run all of the "internal" images that way, any issues?

justincormack:
With this PR, you just need to add kernel, dhcp and anything else systemy (eg rngd), but otherwise you can run exactly the same yaml. I will update the example shortly.

justincormack:
Updated hashes

justincormack:
https://github.com/moby/tool/pull/149 updates the `moby` examples to use this, making them simpler.

justincormack:
CI seems to have hung on the wireguard test; going to (a) restart and (b) test it myself

justincormack:
Ok, I updated this to do exactly the same shutdown code as the previous version, so should hopefully pass CI now.

justincormack:
Ok this is passing now, so fixed the `qemu` shutdown issue.

PR #2506, opened 2017-09-13
new pkg/modprobe from linuxkit/alpine base onto scratch

eyz:
**- What I did**
Created a minimal pkg/modprobe package which is based on scratch, with only the linuxkit/alpine /bin/busybox, and /bin/busybox -> /sbin/modprobe symlinked. This also includes the required capabilities and bind mounts for loading modules.

**- How I did it**
I followed the established pattern for the other included LinuxKit packages

**- How to verify it**

1. Build Linuxkit packages, and use as follows -
```
  - name: modprobe
    image: linuxkit/modprobe:HASH
    command: ["modprobe", "-a", "iscsi_tcp", "dm_multipath"]
```
(substituting the module you need to load in the above)

2. Create and boot your LinuxKit image
3. Run `lsmod` to verify the specified module has been installed

added minimal pkg/modprobe

maddie

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -38,6 +38,15 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Loading kernel modules
+
+Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a module manually you can use the `modprobe` package in the `onboot` section like this:
+```
+  - name: modprobe
+    image: linuxkit/modprobe:<hash>
+    command: ["modprobe", "-a", "iscsi_tcp", "dm_multipath"]
+```
+
 ## Compiling external kernel modules
 
 This section describes how to build external (out-of-tree) kernel
--- a/pkg/modprobe/Dockerfile
+++ b/pkg/modprobe/Dockerfile
@@ -1,0 +1,13 @@
+FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS mirror
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    busybox
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+LABEL org.mobyproject.config='{"binds": ["/lib/modules:/lib/modules", "/sys:/sys"], "capabilities": ["CAP_SYS_MODULE"]}'
--- a/pkg/modprobe/Makefile
+++ b/pkg/modprobe/Makefile
@@ -1,0 +1,3 @@
+IMAGE=modprobe
+
+include ../package.mk


rn:
haven't looked at this in detail, but can you squash your commits into one and also needs a signed-off-by.

eyz:
Had some issues trying to squash with a subsequent commit. Thanks, @rn

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    && true
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN mkdir /out/bin && mkdir /out/sbin && cp /bin/busybox /out/bin/busybox
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+RUN ["/bin/busybox", "ln", "-s", "/bin/busybox", "/sbin/modprobe"]
```
eyz: This is simply to keep the symlink to /bin/busybox in place, rather than letting the `COPY` place the /bin/busybox binary in /sbin as /sbin/modprobe

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+LABEL org.mobyproject.config='{"binds": ["/lib/modules:/lib/modules", "/sys:/sys"], "capabilities": ["CAP_SYS_MODULE"]}'
```
eyz: I did not need the bind of `/sys:/sys` in my testing, but I found that the following tests included it in their `binds`, so I included `/sys:/sys` in the `LABEL` for `pkg/modprobe` -

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/005_extend/001_btrfs/test.yml#L13

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/005_extend/001_btrfs/test-create.yml#L13

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/006_format_mount/003_btrfs/test.yml#L13

pkg/modprobe/Dockerfile
rn: Is there a reason why you use a different format of the `RUN` command here compare to above? I prefer consistency

ijc: It avoids an unnecessary shell (the string version is mutated into `sh -c 'foo'`). Not sure why that is an issue here though.

ijc:
To load multiple modules you would need multiple such containers, I think, since `modprobe` can only deal with one at a time. I suppose: `Command: sh -c 'modprobe foo && modprobe bar'` works.

I can' think of a way to avoid that without adding a shell (or a custom utility) to the image. Might be worth mentioning in the example of the `modprobe && modprobe` pattern in the docs @rn has asked for though?

ijc:
Actually, by having `busybox` in the image there is already a shell. So we _could_ have a default command/entry point which pointed to a script which iterated over the lines in `/etc/modules` which you could bind in. But lets leave that for another day.

justincormack:
Busybox `modprobe` does support `/etc/modprobe.d` which could make sense for multiple modules. Could be the default to use this.

ijc:
`/etc/modprobe.d` lets you configure the options to use when a module is loaded, but AFAIK doesn't enumerate a set of modules to actually load. On Debian at least that latter functionality is `/etc/modules` and `/etc/modules-load.d` which are processed by `/etc/init.d/kmod` (or, I presume, its systemd equivalent).

ijc:
Seems like it is possible to load multiple at once with `-a`:
```
/ # modprobe --help
BusyBox v1.26.2 (2017-06-11 06:38:32 GMT) multi-call binary.

Usage: modprobe [-alrqvsDb] MODULE [SYMBOL=VALUE]...

	-a	Load multiple MODULEs
```

The help doc doesn't really make it clear what the syntax is (neither does the manpage for full-fat `modprobe`). Perhaps `modprobe moda modaopta=foo modb` is treated as ``modprobe moda modaopta=foo && modprobe modb`?

riyazdf:
@rn: just set up signing for `linuxkit/modprobe`

pkg/modprobe/Dockerfile
eyz: Thanks. I am now copying the `/sbin/modprobe` symlink properly with `cp -P`, rather than creating it in the final stage.

eyz:
@rn I am now correctly copying the `/sbin/modprobe` symlink with `cp -P`. I have squashed the commit in. Now there is no need to symlink in the final stage.

docs/kernels.md
```diff
@@ -38,6 +38,15 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Loading kernel modules
+
+Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a module manually you can use the `modprobe` package in the `onboot` section like this:
+```
+  - name: modprobe
+    image: linuxkit/modprobe:HASH
+    command: ["modprobe", "-a", iscsi_tcp", "dm_multipath"]
```
eyz: @ijc I tested busybox's `modprobe`. This works fine: `modprobe -a modulea moduleb`. This form has been included in the documented example.

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 as mirror
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    musl \
+    && true
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+RUN mkdir /out/bin && mkdir /out/sbin && cp /bin/busybox /out/bin/busybox && cp -P /sbin/modprobe /out/sbin/modprobe
```
eyz: @rn I am now correctly copying the `/sbin/modprobe` symlink with `cp -P` in the first stage in lieu of creating the symlink in the second stage. The `CMD ["/bin/busybox", ...]` is no longer needed.

docs/kernels.md
```diff
@@ -38,6 +38,15 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Loading kernel modules
+
+Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a module manually you can use the `modprobe` package in the `onboot` section like this:
+```
+  - name: modprobe
+    image: linuxkit/modprobe:HASH
+    command: ["modprobe", "-a", "iscsi_tcp", "dm_multipath"]
```
eyz: @justincormack @ijc I tested busybox's `modprobe`. This works fine: `modprobe -a moduleA moduleB`. This form has been included in the documented example

docs/kernels.md
eyz: There was a typo (missing double quote), which I have since fixed.

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=mirror /out/ /
+LABEL org.mobyproject.config='{"binds": ["/lib/modules:/lib/modules", "/sys:/sys"], "capabilities": ["CAP_SYS_MODULE"]}'
```
eyz: I did not need the bind of `/sys:/sys` in my testing, but I found that the following tests included it in their `binds`, so I included the `/sys:/sys` bind in the `LABEL` for `pkg/modprobe` -

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/005_extend/001_btrfs/test.yml#L13

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/005_extend/001_btrfs/test-create.yml#L13

https://github.com/linuxkit/linuxkit/blob/3b097e6056b7b0b9396c7908c82d72212813a847/test/cases/040_packages/006_format_mount/003_btrfs/test.yml#L13

eyz:
Pretty sure this build fail is not related: `[2017-09-14 19:18.45] Failed: Failed to build init tarball from linuxkit/containerd:06876ceef325e49e9ba119659357768d5df89075: Could not pull image linuxkit/containerd:06876ceef325e49e9ba119659357768d5df89075: Error response from daemon: manifest for linuxkit/containerd@sha256:c02c1e1fbbf7efeeba437423a4d13064fb7c9082d5e8d2f3d648740464c31637 not found`

docs/kernels.md
```diff
@@ -38,6 +38,15 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Loading kernel modules
+
+Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a module manually you can use the `modprobe` package in the `onboot` section like this:
+```
+  - name: modprobe
+    image: linuxkit/modprobe:HASH
```
rn: super minor nit, but elsewhere we use `<hash>` instead of `HASH`

pkg/modprobe/Dockerfile
rn: another super minor nit but I prefer a command per line so using a `&& \` and then move to the next line is preferred

eyz:
ok @rn, I've cleaned up the Dockerfile per your preferred style (which is my preferred style as well); I was simply trying to keep the existing pattern from the other included `pkg`s

docs/kernels.md
eyz: @rn fixed

docs/kernels.md
```diff
@@ -38,6 +38,15 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
 
+## Loading kernel modules
+
+Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a module manually you can use the `modprobe` package in the `onboot` section like this:
+```
+  - name: modprobe
+    image: linuxkit/modprobe:<hash>
```
eyz: @rn updated to <hash> here

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 as mirror
+
+RUN mkdir -p /out/etc/apk && \
+    cp -r /etc/apk/* /out/etc/apk/ && \
+    apk add --no-cache --initdb -p /out musl && \
+    rm -rf /out/etc/apk /out/lib/apk /out/var/cache && \
+    mkdir /out/bin && \
+    mkdir /out/sbin && \
+    cp /bin/busybox /out/bin/busybox && \
+    cp -P /sbin/modprobe /out/sbin/modprobe
```
eyz: @rn, I've cleaned up the `pkg/modprobe/Dockerfile` here per your preferred style (which is my actually my preferred style as well). I was simply trying to keep the existing pattern from the other included `pkg`; I think you'll find the single line `&&` style is common in a lot of the existing `pkg/*/Dockerfile`

rn: Argh, I think we misunderstood...the previous version had one line with multiple commands in it and I was proposing to just split that line up. something like:
```
RUN apk add --no-cache --initdb -p /out \
    alpine-baselayout \
    busybox \
    musl

RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
```

You don't actually need to copy `cp /bin/busybox /out/bin/busybox` and `cp -P /sbin/modprobe /out/sbin/modprobe` as the `-p /out` argument to `apk` installs the packages directly in `/out`. So what you are doing is copy the `busybox` binary from `linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630` to `/out` insteda of the one `apk` installs. (they should be same).

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    musl
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+# keep only /bin/busybox and /sbin/modprobe symlink from busybox package
+RUN mkdir /out/bin && \
+    mkdir /out/sbin && \
+    cp /bin/busybox /out/bin/busybox && \
+    cp -P /sbin/modprobe /out/sbin/modprobe
```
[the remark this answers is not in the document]
rn: I think the space savings are minimal so in this case I prefer to just keep the base `busybox` intact instead of trying to save a couple of symlinks and text files.

This `modprobe` package is mostly for convenience so the main part is the `org.mobyproject.config` below.

pkg/modprobe/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS mirror
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    busybox
```
eyz: @rn just keeping busybox; musl installed already as a dep. Please review and merge if all looks good

rn:
I've build/pushed/signed the package and it should be avail as: `linuxkit/modprobe:69494c187eddc12b3f24e4e0cbd5d3360bda3504`

PR #2507, opened 2017-09-14
service: log stdout and stderr to separate files

djs55:
**- What I did**

Fixed a bug in `service` which fails to read data from the `stdout` of the container (in my case leading to the process blocking when the pipe filled up)

**- How I did it**

Previously we would pass the path `/var/log/service.log` for both stdout and stderr to containerd. containerd would construct a dict with the [paths as keys](https://github.com/containerd/containerd/blob/49437711c31c5f8ca8e1a5931be7284198497350/linux/shim/io.go#L51) and, due to the duplicate key, would only open one of the files and start one `io.Copy` instance. Writes to the other stream would be buffered by the pipe connected to containerd-shim and would eventually block.

If we modified containerd to open the file twice and start two `io.Copy` instances, we would end up with the two streams interleaved together. It seems cleaner to keep the streams separate; therefore this patch logs stdout to `/var/log/service.out.log` and stderr to `/var/log/service.err.log`.

**- How to verify it**

Create a service which writes to `stdout`: before this change, the write will not be logged. After this patch it will be logged to `/var/log/service.out.log`

Change the log paths: write stdout to `/var/log/service.out.log` and stderr to `/var/log/service.err.log`

beaver sitting on logs

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -4,7 +4,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/vpnkit-expose-port:fa4ab4ac78b83fe392e39b861b4114c3bb02d170 # install vpnkit-expose-port and vpnkit-iptables-wrapper on host
-  - linuxkit/init:79973d34faf7d4654462f3408c5eb00befaf03b8
+  - linuxkit/init:851e9c3ad0574d640b733b92fdb26c368d2f7f8f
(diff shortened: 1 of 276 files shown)


rn:
This also needs rebuilding of the init package and update of hashes...

I've build and pushed x86 and arrm64 packages and the new hash is `linuxkit/init:851e9c3ad0574d640b733b92fdb26c368d2f7f8f`

can you add a commit which updates the YAML files

Otherwise the changes look good to me

PR #2514, opened 2017-09-15
Ensure depmod is run on ZFS module builds (fixes ZFS modprobe)

matjohn2:
Depmod in the zfs makefiles will never run as `/boot/` and relevant map files dont exist in our build environments.

Signed-off-by: Matt Johnson <matjohn2@cisco.com>

**- What I did**
Depmod in the zfs makefiles will never run as the`/boot/` dir, system maps and relevant files don't exist in our build environments.

Running depmod on whatever is in the `/lib/modules` directory solves this issue.
Tested with local builds successfully.

**- How to verify it**
If you can modprobe zfs once adding the image to `init:` it's working.

Depmod now runs correctly on the ZFS kernel module build.

image

### DIFF
--- a/kernel/Dockerfile.zfs
+++ b/kernel/Dockerfile.zfs
@@ -8,7 +8,7 @@ RUN apk add \
     build-base \
     file \
     git \
-    libtirpc-dev \ 
+    libtirpc-dev \
     libtool \
     util-linux-dev \
     zlib-dev
@@ -51,6 +51,10 @@ RUN ./autogen.sh && \
     cd module && \
     make -j "$(getconf _NPROCESSORS_ONLN)" && \
     make install
+
+# Run depmod against the new module directory.
+RUN cd /lib/modules && \
+    depmod -ae * 
 
 FROM scratch
 ENTRYPOINT []


kernel/Dockerfile.zfs
```diff
@@ -52,6 +52,9 @@ RUN ./autogen.sh && \
     make -j "$(getconf _NPROCESSORS_ONLN)" && \
     make install
 
+# Run depmod against the new module directory.
+RUN depmod -ae `ls /lib/modules`
```
justincormack: some shell script style issues:
1. can you use `$(...)` not backticks
2. can you use `find` not `ls`
Possibly you can just do `depmod -ae /lib/modules/*` though instead...

matjohn2: depmod won't take a path, tried a few variants; thought `ls` was the most readable, happy to change.

```
Removing intermediate container 47e6e498c868
Step 20/25 : RUN depmod -ae /lib/modules/*
 ---> Running in 10e5ae3abf3a
depmod: can't change directory to 'lib/modules/4.9.41-moby': No such file or directory
The command '/bin/sh -c depmod -ae /lib/modules/*' returned a non-zero code: 1
```

kernel/Dockerfile.zfs
```diff
@@ -53,7 +53,7 @@ RUN ./autogen.sh && \
     make install
 
 # Run depmod against the new module directory.
-RUN depmod -ae `ls /lib/modules`
+RUN cd /lib/modules ; depmod -ae *
```
rn: I would make that
```
RUN cd /lib/modules && \
    depmod -ae *
```
It's unlikely to fail, but is more in line with the general style we use.

rn:
LGTM but can you squash your commits into one?

PR #2515, opened 2017-09-16
added support for guest shut down in pkg/open-vm-tools

eyz:
**- What I did**
I enabled hypervisor-initiated guest shut down in `pkg/open-vm-tools`

**- How I did it**
I added `poweroff-vm-default.d/poweroff.sh` to `/etc/vmware-tools/scripts`, which sends a `SIGTERM` to PID 1 (the LinuxKit init system) when a guest shut down event is received through `vmtoolsd`. The `poweroff.sh` script then returns exit code 0 through `vmtoolsd` to send success back to the hypervisor UX.

**- How to verify it**
With a running LinuxKit OS in VMware vSphere, choose `Power` -> `Shut Down Guest` on the applicable VM. LinuxKit will shut down via the init system.

added support for guest shut down in pkg/open-vm-tools

56269_1559857750243_5587211_o

### DIFF
--- a/pkg/open-vm-tools/Dockerfile
+++ b/pkg/open-vm-tools/Dockerfile
@@ -13,4 +13,7 @@ ENTRYPOINT []
 CMD []
 WORKDIR /
 COPY --from=mirror /out/ /
+COPY scripts /etc/vmware-tools/scripts
 CMD ["/usr/bin/vmtoolsd"]
+
+LABEL org.mobyproject.config='{"pid": "host", "capabilities": ["CAP_SYS_BOOT"]}'
--- a/pkg/open-vm-tools/scripts/poweroff-vm-default.d/poweroff.sh
+++ b/pkg/open-vm-tools/scripts/poweroff-vm-default.d/poweroff.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+# send SIGTERM to the init system (PID 1), which causes a clean VM host-initiated shutdown
+kill -s SIGTERM 1
+exit 0


eyz:
Please let me know if anyone would prefer explicit checking that PID 1 is `init`, or any condition to be allowed to make guest shutdown configurable via an environment variable, or similar.

pkg/open-vm-tools/Dockerfile
```diff
@@ -13,4 +13,7 @@ ENTRYPOINT []
 CMD []
 WORKDIR /
 COPY --from=mirror /out/ /
+COPY scripts /etc/vmware-tools/scripts
 CMD ["/usr/bin/vmtoolsd"]
+
+LABEL org.mobyproject.config='{"pid": "host", "capabilities": ["CAP_SYS_BOOT"]}'
```
justincormack: pid=host is actually the default, although it probably still makes sense to have this explicitly.

eyz: I thought that when I tested it was required, but I'll try without to verify.

eyz: @justincormack yep, it looks like I don't need it. Should I leave `"pid": "host"` or leave it in?

rn: I prefer if it is explicit esp if it is really needed to work.

eyz: @rn This requires access to the host pid space to kill `init`, so I'm quite certain this is the case. Thanks again for the merge.

PR #2517, opened 2017-09-18
Update/Unify tests

rn:
Originally, the plan was to reduce the number of YAML files which need updating on kernel/init/runc updates by using `moby build` for the kernel tests with multiple YAML files. Then, some minor OCDness kicked in and I ended up cleaning up most of the test scripts to unify them.

- Update `rtf` to the latest version
- Share kernel/init/runc section for kernel namespace tests. This significantly reduces the number of YAML files which need updating on frequent kernel/init/runc hash updates.
- Unify on using `test.yml` for all tests
- Unify cleanup code and use of env variables in `test.sh` scripts
- Fix removal for temporary files for some of the tests
- Add a test for building the LCOW blueprint

baby-donkey


rn:
yeah, a lot of mechanical changes with `sed` and friends. I spot checked a few tests and CI seems to be green, so merging

PR #2519, opened 2017-09-18
Specify punctual install targets for "make install"

fntlnz:
I just noted that I had the `rtf` command in my path, I don't think that the intended behavior of the install target is to copy everything in bin to the installation directory but I may be wrong, so please lemme knoww!

Signed-off-by: Lorenzo Fontana <lo@linux.com>

**- What I did**
I just specified intended binaries to be installed in while calling the `install` target of the Makefile and added set install as phony target 🦄 

**- How I did it**
You can just look at it!

**- How to verify it**
Do a make install and see that `rtf` is not installed

Installing only moby and linuxkit as result of `make install`

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,8 +79,9 @@ local: $(LINUXKIT_DEPS) | bin
 bin:
 	mkdir -p $@
 
+.PHONY: install
 install:
-	cp -R ./bin/* $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


justincormack:
I think it was the intention, install all the things we build. `rtf` and `manifest-tool` are only needed for some use cases but it is hard to guess what use cases people want. Ideally we build packages instead.

fntlnz:
Yea thanks for the clarification

PR #2522, opened 2017-09-19
Bump kuberrnetes versions

ijc:
Updates various Kube components:
* Kube itself from 1.7.4 to 1.7.6
* Weave from 2.0.1 to 2.0.6
* CNI from 0.5.2 to 0.6.0

I also added `force{push,tag}` targets to the image cache `Makefile` and scripted up the mapping from images to digests to be pulled in since it was a bit tedious to do before.


projects/kubernetes/image-cache/mkversions
```diff
@@ -0,0 +1,45 @@
...
+	kube-scheduler-amd64:$kube_version
+	etcd-amd64:$etcd_version"
+
+for i in $common $control ; do
+	docker image pull "$repo/$i" 1>&2
+done
+
+oi() {
+	local i="$1"
+	digest=$(docker image inspect "$repo/$i" | jq -r '.[].RepoDigests[0]')
```
rn: `docker inspect --format '{{index .RepoDigests 0}}' "$repo/$i"` might be simpler

ijc: It would indeed, thanks.

Since this code is added before the "Bump yml" commit and it's a pain to rebuild I'm going to fix this at the head of the branch rather than fold it into the right place.

PR #2523, opened 2017-09-20
Updates including kernel update to 4.12.14/4.9.51

### LINKED ISSUE #2511: add HP devices
add
```
CONFIG_SCSI_HPSA=m
```

Also review other storage drivers.

rn:
- Update alpine base (picking up new wireguard tools)
- Update packages which need new wireguard tools
- Enable HPSA/SMARTPQI in kernel config (resolves #2511)
- Update kernels to 4.9.51/4.12.14
- Might as well update AUTHORS while at it

seahorse-qtip


rn:
x86/arm64 kernels pushed (x86 was pushed before otherwise CI wouldn't work...)

PR #2531, opened 2017-09-22
pkg/init onboot /var/log/NAME.{out,err}.log

eyz:
**- What I did**
Enabled logging to `/var/log/NAME.out.log` (stdout) and `/var/log/NAME.err.log` (stderr) for `onboot`

**- How I did it**

1. attempt to open `/var/log/NAME.out.log` and `/var/log/NAME.err.log` for append/create/write
2. sending `onboot` `cmd` stdout and stderr to the logs (if able to open), as well as to the standard `os` stdout and stderr

**- How to verify it**
Start LinuxKit, and verify that `onboot` services have written to /var/log as `##-NAME.out.log` and `##-NAME.err.log`, as well as to the console (as applicable)

pkg/init onboot /var/log/NAME.{out,err}.log logging

16174540_10209809662012121_3957783309968380981_n

### DIFF
--- a/pkg/init/cmd/service/runc.go
+++ b/pkg/init/cmd/service/runc.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io"
 	"io/ioutil"
 	"log"
 	"os"
@@ -50,22 +51,78 @@ func runcInit(rootPath string) int {
 			status = 1
 			continue
 		}
+
 		pidfile := filepath.Join(tmpdir, name)
 		cmd := exec.Command(runcBinary, "create", "--bundle", path, "--pid-file", pidfile, name)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		if err := cmd.Run(); err != nil {
-			log.Printf("Error creating %s: %v", name, err)
+
+		stdoutFile := filepath.Join("/var/log", "onboot-"+name+".out.log")
+
+		stdoutFd, stdoutFdErr := os.OpenFile(stdoutFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+		defer stdoutFd.Close()
+
+		var stdoutCopyErr error
+
+		if stdoutFdErr == nil {
+			stdoutIn, _ := cmd.StdoutPipe()
+			stdoutWriter := io.MultiWriter(os.Stdout, stdoutFd)
+			go func() {
+				_, stdoutCopyErr = io.Copy(stdoutWriter, stdoutIn)
+			}()
+		} else {
+			log.Printf("Could not open %s for writing: %v", stdoutFile, stdoutFdErr)
+			cmd.Stdout = os.Stdout
+		}
+
+		stderrFile := filepath.Join("/var/log", "onboot-"+name+".err.log")
+
+		stderrFd, stderrFdErr := os.OpenFile(stderrFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+		defer stderrFd.Close()
+
+		var stderrCopyErr error
+
+		if stderrFdErr == nil {
+			stderrIn, _ := cmd.StderrPipe()
+			stderrWriter := io.MultiWriter(os.Stderr, stderrFd)
+			go func() {
+				_, stderrCopyErr = io.Copy(stderrWriter, stderrIn)
+			}()
+		} else {
+			log.Printf("Could not open %s for writing: %v", stderrFile, stderrFdErr)
+			cmd.Stderr = os.Stderr
+		}
+
+		if err := cmd.Start(); err != nil {
+			log.Printf("Error starting %s: %v", name, err)
 			status = 1
 			// skip cleanup on error for debug
 			continue
 		}
+
+		if err := cmd.Wait(); err != nil {
+			log.Printf("Error waiting for completion of %s: %v", name, err)
+			status = 1
+			continue
+		}
+
+		if stdoutFdErr == nil && stdoutCopyErr != nil {
+			log.Printf("Could not stream stdout for %s: %v", name, stdoutCopyErr)
+			status = 1
+			continue
+		}
+
+		if stderrFdErr == nil && stderrCopyErr != nil {
+			log.Printf("Could not stream stderr for %s: %v", name, stderrCopyErr)
+			status = 1
+			continue
+		}
+
 		pf, err := ioutil.ReadFile(pidfile)
 		if err != nil {
 			log.Printf("Cannot read pidfile: %v", err)
 			status = 1
 			continue
 		}
+
 		pid, err := strconv.Atoi(string(pf))
 		if err != nil {
 			log.Printf("Cannot parse pid from pidfile: %v", err)
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -4,7 +4,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/vpnkit-expose-port:fa4ab4ac78b83fe392e39b861b4114c3bb02d170 # install vpnkit-expose-port and vpnkit-iptables-wrapper on host
-  - linuxkit/init:7804129bd06218b72c298139a25698a748d253c6
+  - linuxkit/init:2a82fddd29836e4b1b422a3d8b0124fbd185277c
   - linuxkit/runc:a1b564248a0d0b118c11e61db9f84ecf41dd2d2a
   - linuxkit/containerd:417f83f7b8dc1fa36acf90effe44f99c7397480a
   - linuxkit/ca-certificates:e44b0a66df5a102c0e220f0066b0d904710dcb10
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.9.51
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:7804129bd06218b72c298139a25698a748d253c6
+  - linuxkit/init:2a82fddd29836e4b1b422a3d8b0124fbd185277c
   - linuxkit/runc:a1b564248a0d0b118c11e61db9f84ecf41dd2d2a
   - linuxkit/containerd:417f83f7b8dc1fa36acf90effe44f99c7397480a
   - linuxkit/ca-certificates:e44b0a66df5a102c0e220f0066b0d904710dcb10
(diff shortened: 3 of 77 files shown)


justincormack:
I think we should write to both the log and to stdout/err. As the output is short you can probably just use a buffer, or maybe `io.TeeReader`.

eyz:
@justincormack, OK. I'll check out the tee writer.

eyz:
I tried the io.Multiwriter, but I get no output, but no compile errors.

justincormack:
Hmm, are you sure you were getting things on the console before? It helps if you remove all the `console=` parts except the one you are using, normally `console=ttyS0`.

The reason I suggested a buffer is that it should always work, ie read the output into a buffer and then copy that to both stdout and the file, but that is ugly and a multiwriter is nicer....

eyz:
I'll check. The console is blank and hangs with the above code. Init never finishes.

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
-			cmd.Stdout = stdoutFd
+
+		var stdoutCopyErr error
+
+		if stdoutFdErr == nil {
+			stdoutIn, _ := cmd.StdoutPipe()
+			stdoutWriter := io.MultiWriter(os.Stdout, stdoutFd)
+			go func() {
+				_, stdoutCopyErr = io.Copy(stdoutWriter, stdoutIn)
+			}()
```
eyz: if a stdout log file can be opened for write/append, then io.Copy the cmd.StdoutPipe to a io.MultiWriter (of both the os.Stdout and the stdout log file)

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
-			cmd.Stderr = stderrFd
+
+		var stderrCopyErr error
+
+		if stderrFdErr == nil {
+			stderrIn, _ := cmd.StderrPipe()
+			stderrWriter := io.MultiWriter(os.Stderr, stderrFd)
+			go func() {
+				_, stderrCopyErr = io.Copy(stderrWriter, stderrIn)
+			}()
```
eyz: if a stderr log file can be opened for write/append, then io.Copy the cmd.StderrPipe to a io.MultiWriter (of both the os.Stderr and the stderr log file)

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
+			log.Printf("Error waiting for completion of %s: %v", name, err)
+			status = 1
+			continue
+		}
+
+		if stdoutFdErr == nil && stdoutCopyErr != nil {
+			log.Printf("Could not stream stdout for %s: %v", name, stdoutCopyErr)
+			status = 1
+			continue
+		}
```
eyz: if io.Copy fails then a fatal error is produced

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
+
+		if stderrFdErr == nil && stderrCopyErr != nil {
+			log.Printf("Could not stream stderr for %s: %v", name, stderrCopyErr)
+			status = 1
+			continue
+		}
+
 		pf, err := ioutil.ReadFile(pidfile)
```
eyz: if io.Copy fails during stdout streaming then a fatal error is produced

eyz: if io.Copy fails during stderr streaming then a fatal error is produced

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
+			status = 1
+			continue
+		}
+
+		if stderrFdErr == nil && stderrCopyErr != nil {
+			log.Printf("Could not stream stderr for %s: %v", name, stderrCopyErr)
+			status = 1
+			continue
+		}
+
```
eyz: if io.Copy fails during stderr streaming then a fatal error is produced

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
 		pf, err := ioutil.ReadFile(pidfile)
 		if err != nil {
 			log.Printf("Cannot read pidfile: %v", err)
 			status = 1
 			continue
 		}
+
```
eyz: added newline for consistency with the next sections

pkg/init/cmd/service/runc.go
```diff
@@ -50,42 +51,78 @@ func runcInit(rootPath string) int {
...
```
eyz: added newline for consistency with the following sections

pkg/init/cmd/service/runc.go
```diff
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io"
```
eyz: `"io"` required for `io.Copy` and `io.MultiWriter`

eyz:
@justincormack OK. With the above changes, SGTM now. I tested again by changing the /var/log/ target to a non-existent folder, and output still is sent to the console through the prior method. If the logs can be written, then output will be sent to both the logs and the console with these changes.

eyz:
I also attempted to factor the stdout and stderr handling, but I found there is a large amount of boilerplate that only seems to complicate the code further. I would suggest going with this, unless anyone else can think of a clean way to factor it further.

ijc:
I've often wished for logs of onboot stuff, thanks!
We should either prefix all the log files with `onboot-` and `service-` as appropriate or enforce that the names are unique across both namespaces (as it stands I think only services have a uniqueness check). Or else services can clobber similarly named onboot services. The most obvious case where this might happen would be `dhcpcd` where running onboot and as a service is a realistic proposition.

Alternatively we could log all `onboot` stuff to a single log file (and insert our own progress markers/delimiters).

Also this new code seems like it is duplicating (or at least shares some similarities with) a bunch of stuff which is also in the service stuff. Perhaps something could be refactored to be used by both?

justincormack:
Hmm, I think the uniqueness check should be global really (I thought it was).

eyz:
I added in prefixes of "onboot-" and "service-" in b48394b per @ijc's comment

@ijc It looks to me like the `start.go` (service launch) and `runc.go` (onboot launch) code is not similar, other than the filename concatenation -- which I pulled from `start.go` to use in `runc.go`

Here's using the io.MultiWriter and goroutine for io.Copy in my code -

justincormack:
I would rather leave the `service` log names unchanged. Then later can remove the `onboot-` prefix when names are required to be globally unique

eyz:
@justincormack @ijc I removed the "service-" prefix and left the "onboot-" prefix

justincormack:
This looks good, I will build and push the images so you can do a hash update sometime this afternoon.

eyz:
@justincormack I am squashing now.

justincormack:
Ok I built and pushed `linuxkit/init:489133fb1af16b3161af2f03a1780eca5e506b37` so you can add a commit that updates to that

eyz:
```

**Comment by justincormack** [comment]
Argh, no you cant squash it after I built the init, as it has the commit ID in the metadata, I need to rebuild it now its squashed...

**Comment by justincormack** [comment]
Oh its ok, I think you squashed it first

**Comment by eyz** [comment]
Justin is building again post-squash

**Comment by justincormack** [comment]
`linuxkit/init:489133fb1af16b3161af2f03a1780eca5e506b37`

**Comment by eyz** [comment]
I have commited the new tag for `pkg/linuxkit` (linuxkit/init:489133fb1af16b3161af2f03a1780eca5e506b37) -

**Comment by eyz** [comment]
I have done `git fetch upstream; git rebase upstream/master` and I'm awaiting another build before I commit the tag changes in a new commit

**Comment by eyz** [comment]
Updated `linuxkit/init` tag in d130d9a3 -
```
$ grep -PRo 'linuxkit/init:\w+' | grep -Pv '^\.git/' | cut -d ':' -f 2,3 | sort -u
linuxkit/init:7804129bd06218b72c298139a25698a748d253c6

$ scripts/update-component-sha.sh linuxkit/init:7804129bd06218b72c298139a25698a748d253c6 linuxkit/init:2a82fddd29836e4b1b422a3d8b0124fbd185277c

$ grep -PRo 'linuxkit/init:\w+' | grep -Pv '^\.git/' | cut -d ':' -f 2,3 | sort -u
linuxkit/init:2a82fddd29836e4b1b422a3d8b0124fbd185277c
```

eyz:
@justincormack Yeah, pretty sure something is either different output, or hanging or something with that init version. I checked out a commit from master earlier in the day and pointed the test.yml at my build of init (locally) and it failed too.
So, I'd say that's pretty conclusive. But, like I said, I saw odd behavior with using pipes for stdout and stderr.
It worked fine in my testing (with my .yml), but it seems that it doesn't work across the board for some reason.
So yeah- go ahead and close the PR if you folks can't figure out what's up, please. And also keep in mind that there is that odd behavior with os exec write/wait per https://go-review.googlesource.com/c/go/+/65490 that I found

eyz:
@justincormack, if you get stuck looking at the fail root cause then I'll try migrating the `pkg/init` `onboot` launcher to the containerd API (https://github.com/containerd/containerd/blob/master/README.md#tasks)  instead of calling it via `exec`. The `pkg/init` service launcher is already using that, and since `onboot` is not shell (any longer?) it probably makes sense to use that pattern in `pkg/init` `onboot` as well.

eyz:
It sounds like reading from cmd via pipes is prone to race conditions per https://github.com/go-cmd/cmd. This PR likely is not viable as-is. It may be best to close this for now.

ijc:
It looks like you were pretty close, just needs a waitgroup to synchronise between the goroutines finishing and the call to Wait. Or just use `go-cmd/cmd` since it claims to have fixed this?

eyz:
Thanks, @ijc :) I actually spent about 2 to 3 days trying many variations, including using waitgroups, channels, and pouring through examples. Next steps on this for me are to change this code over to the containerd API (such as used in `pkg/init/cmd/service/start.go`) and troubleshoot from there, if needed.

It seems that `exec.Command` with pipes is simply a very race-heavy area, and use of that pattern is highly discouraged, as https://github.com/go-cmd/cmd points out by needing to exist. The common `cmd.Stdout = os.Stdout` pattern is apparently not prone to race issues at all, from what I've seen and read. Another thought would also be to use some sort of logging library that specializes in this area, and defer to it to do the right thing.

I'll probably pick this back up again, but I need to move onto other things in the meantime. I have the branch available still, so if anyone wants to fork my branch and pick it up they are certainly welcome to. Thanks for the comment!

PR #2532, opened 2017-09-22
format: blkid content check, format argument, and verbose argument

eyz:
**- What I did**
1) Added new argument -force to the format utility. This will force formatting of the specified single device if it exists and is a block device
2) By default, the format package will no longer format the specified single device if content exists on that device
3) Added new blkid-based check for content, in addition to the existing check for partitions on autoformat candidate devices.
4) Cleanup of old unused code.
5) Refactoring of block device verification.
6) Added test/cases/040_packages/006_format_mount/005_by_device_force
7) Added document of new arguments to /docs/external-disk.md. Also sorted the arguments in the doc alphabetically and added them as bullets so they do not run together on the page.

**- How I did it**
- By default, the format package will no longer format the specified single device if content exists on that device
- Added new blkid-based check for content, in addition to the existing check for partitions on autoformat candidate devices.

**- How to verify it**
I have verified that the following will ignore auto-formatting devices other than `/dev/sda` -
```
  - name: format
    image: MYREGISTRY/format:<hash>
    command: ["/usr/bin/format", "-force", "/dev/sda"]
```

19575408_10211240830670443_6190493428953175106_o

### DIFF
--- a/pkg/format/Dockerfile
+++ b/pkg/format/Dockerfile
@@ -1,4 +1,4 @@
-FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS mirror
+FROM linuxkit/alpine:d1778ee29f11475548f0e861f57005a84e82ba4e AS mirror
 
 RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out \
@@ -11,10 +11,11 @@ RUN apk add --no-cache --initdb -p /out \
     musl \
     sfdisk \
     util-linux \
+    blkid \
     && true
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
 
-FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630 AS build
+FROM linuxkit/alpine:d1778ee29f11475548f0e861f57005a84e82ba4e AS build
 
 RUN apk add --no-cache go musl-dev
 ENV GOPATH=/go PATH=$PATH:/go/bin
--- a/pkg/format/format.go
+++ b/pkg/format/format.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"sort"
 	"strings"
 	"syscall"
 	"time"
@@ -21,36 +20,127 @@ const (
 )
 
 var (
-	labelVar  string
-	fsTypeVar string
-	drives    map[string]bool
-	driveKeys []string
+	labelVar   string
+	fsTypeVar  string
+	forceVar   bool
+	verboseVar bool
+	drives     map[string]bool
+	driveKeys  []string
 )
+
+func hasPartitions(d string) bool {
+	err := exec.Command("sfdisk", "-d", d).Run()
+	return err == nil
+}
+
+func isEmptyDevice(d string) (bool, error) {
+	// default result
+	isEmpty := false
+
+	if verboseVar {
+		log.Printf("Checking if %s is empty", d)
+	}
+
+	out, err := exec.Command("blkid", d).Output()
+	if err == nil {
+		log.Printf("%s has content. blkid returned: %s", d, out)
+		// there is content, so exit early
+		return false, nil
+	}
+
+	if exiterr, ok := err.(*exec.ExitError); ok {
+		if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+			// blkid exitcode 2 (from the non-busybox version) signifies the block device has no detectable content signatures
+			if status.ExitStatus() == 2 {
+				if verboseVar {
+					log.Printf("blkid did not find any existing content on %s.", d)
+				}
+				// no content detected, but continue through all checks
+				isEmpty = true
+			} else {
+				return isEmpty, fmt.Errorf("Could not determine if %s was empty. blkid %s returned: %s (exitcode %d)", d, d, out, status.ExitStatus())
+			}
+		}
+	}
+
+	if hasPartitions(d) {
+		log.Printf("Partition table found on device %s. Skipping.", d)
+		return false, nil
+	}
+
+	// return final result
+	return isEmpty, nil
+}
 
 func autoformat(label, fsType string) error {
 	var first string
 	for _, d := range driveKeys {
-		err := exec.Command("sfdisk", "-d", d).Run()
-		if err == nil {
-			log.Printf("Partition table found on device %s. Skipping.", d)
-			continue
+		if verboseVar {
+			log.Printf("Considering auto format for device %s", d)
 		}
-		first = d
-		break
+		// break the loop with the first empty device we find
+		isEmpty, err := isEmptyDevice(d)
+		if err != nil {
+			return err
+		}
+		if isEmpty == true {
+			first = d
+			break
+		}
 	}
 
 	if first == "" {
 		return fmt.Errorf("No eligible disks found")
 	}
 
-	if err := format(first, label, fsType); err != nil {
-		return err
+	return format(first, label, fsType, false)
+}
+
+func refreshDevicesAndWaitFor(awaitedDevice string) error {
+	exec.Command("mdev", "-s").Run()
+
+	// wait for device
+	var done bool
+	for i := 0; i < timeout; i++ {
+		stat, err := os.Stat(awaitedDevice)
+		if err != nil {
+			return err
+		}
+		if isBlockDevice(&stat) {
+			done = true
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+		exec.Command("mdev", "-s").Run()
 	}
+	if !done {
+		return fmt.Errorf("Error waiting for device %s", awaitedDevice)
+	}
+	// even after the device appears we still have a race
+	time.Sleep(1 * time.Second)
 
 	return nil
 }
 
-func format(d, label, fsType string) error {
+func format(d, label, fsType string, forced bool) error {
+	if forced {
+		// clear partitions on device if forced format and they exist
+		if hasPartitions(d) {
+			if verboseVar {
+				log.Printf("Clearing partitions on %s because forced format was requested", d)
+			}
+			partCmd := exec.Command("sfdisk", "--quiet", "--delete", d)
+			partCmd.Stdin = strings.NewReader(";")
+			if out, err := partCmd.CombinedOutput(); err != nil {
+				return fmt.Errorf("Error deleting partitions with sfdisk: %v\n%s", err, out)
+			}
+		} else {
+			if verboseVar {
+				log.Printf("No need to clear partitions.")
+			}
+		}
+	}
+
 	log.Printf("Creating partition on %s", d)
 	/* new disks do not have an DOS signature in sector 0
 	this makes sfdisk complain. We can workaround this by letting
@@ -75,28 +165,10 @@ func format(d, label, fsType string) error {
 		return fmt.Errorf("Error running blockdev: %v", err)
 	}
 
-	exec.Command("mdev", "-s").Run()
-
 	partition := fmt.Sprintf("%s1", d)
-	// wait for device
-	var done bool
-	for i := 0; i < timeout; i++ {
-		stat, err := os.Stat(partition)
-		if err == nil {
-			mode := stat.Sys().(*syscall.Stat_t).Mode
-			if (mode & syscall.S_IFMT) == syscall.S_IFBLK {
-				done = true
-				break
-			}
-		}
-		time.Sleep(100 * time.Millisecond)
-		exec.Command("mdev", "-s").Run()
+	if err := refreshDevicesAndWaitFor(partition); err != nil {
+		return err
 	}
-	if !done {
-		return fmt.Errorf("Error waiting for device %s", partition)
-	}
-	// even after the device appears we still have a race
-	time.Sleep(1 * time.Second)
 
 	// mkfs
 	mkfsArgs := []string{"-t", fsType}
@@ -133,6 +205,13 @@ func format(d, label, fsType string) error {
 	return nil
 }
 
+func isBlockDevice(d *os.FileInfo) bool {
+	// this probably shouldn't be so hard
+	// but d.Mode()&os.ModeDevice == 0 doesn't work as expected
+	mode := (*d).Sys().(*syscall.Stat_t).Mode
+	return (mode & syscall.S_IFMT) == syscall.S_IFBLK
+}
+
 // return a list of all available drives
 func findDrives() {
 	drives = make(map[string]bool)
@@ -140,45 +219,76 @@ func findDrives() {
 	ignoreExp := regexp.MustCompile(`^loop.*$|^nbd.*$|^[a-z]+[0-9]+$`)
 	devs, _ := ioutil.ReadDir("/dev")
 	for _, d := range devs {
-		// this probably shouldn't be so hard
-		// but d.Mode()&os.ModeDevice == 0 doesn't work as expected
-		mode := d.Sys().(*syscall.Stat_t).Mode
-		if (mode & syscall.S_IFMT) != syscall.S_IFBLK {
+		if isBlockDevice(&d) {
+			if verboseVar {
+				log.Printf("/dev/%s is a block device", d.Name())
+			}
+		} else {
+			if verboseVar {
+				log.Printf("/dev/%s is not a block device", d.Name())
+			}
 			continue
 		}
 		// ignore if it matches regexp
 		if ignoreExp.MatchString(d.Name()) {
+			if verboseVar {
+				log.Printf("ignored device /dev/%s during drive autodetection", d.Name())
+			}
 			continue
 		}
 		driveKeys = append(driveKeys, filepath.Join("/dev", d.Name()))
 	}
-	sort.Strings(driveKeys)
-	for _, d := range driveKeys {
-		drives[d] = true
-	}
 }
 
 func init() {
+	flag.BoolVar(&forceVar, "force", false, "Force format of specified single device (default false)")
 	flag.StringVar(&labelVar, "label", "", "Disk label to apply")
 	flag.StringVar(&fsTypeVar, "type", "ext4", "Type of filesystem to create")
+	flag.BoolVar(&verboseVar, "verbose", false, "Enable verbose output (default false)")
+}
+
+func verifyBlockDevice(device string) error {
+	d, err := os.Stat(device)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("%s does not exist", device)
+	}
+	if !isBlockDevice(&d) {
+		return fmt.Errorf("%s is not a block device", device)
+	}
+	// passed checks
+	return nil
 }
 
 func main() {
 	flag.Parse()
-
-	findDrives()
 
 	if flag.NArg() > 1 {
 		log.Fatalf("Too many arguments provided")
 	}
 
 	if flag.NArg() == 0 {
+		// auto-detect drives if a device to format is not explicitly specified
+		findDrives()
 		if err := autoformat(labelVar, fsTypeVar); err != nil {
 			log.Fatalf("%v", err)
 		}
 	} else {
-		if err := format(flag.Args()[0], labelVar, fsTypeVar); err != nil {
+		candidateDevice := flag.Args()[0]
+
+		if err := verifyBlockDevice(candidateDevice); err != nil {
 			log.Fatalf("%v", err)
+		}
+
+		if forceVar == true {
+			if err := format(candidateDevice, labelVar, fsTypeVar, forceVar); err != nil {
+				log.Fatalf("%v", err)
+			}
+		} else {
+			// add the deviceVar to the array of devices to consider autoformatting
+			driveKeys = []string{candidateDevice}
+			if err := autoformat(labelVar, fsTypeVar); err != nil {
+				log.Fatalf("%v", err)
+			}
 		}
 	}
 }
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -20,7 +20,7 @@ onboot:
     image: linuxkit/binfmt:472eeba777d056c5f98fe074aa0f581c67ccc7ff
   # Format and mount the disk image in /var/lib/docker
   - name: format
-    image: linuxkit/format:158d992b7bf7ab984100c697d7e72161ea7d7382
+    image: linuxkit/format:f180a74d878c8c0c86f6208f9311474c91452a79
   - name: mount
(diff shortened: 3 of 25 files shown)


justincormack:
Hmm, wouldn't this be more useful if you could specify multiple devices here? eg format sda or sdb but definitely dont mess with sdc?

eyz:
@justincormack, sure. I can do that

Something like `--devices sda,sdb,sdc`?

justincormack:
yeah that would be fine.

deitch:
> specify multiple devices here

I thought our philosophy here was to have single device, and just run the (lightweight) format container multiple times, once for each device, if wanted? 

I know we had discussed this before - cannot recall where at the moment - but that was the thrust. 

Personally, I have no issues with multiple, but I do recall that as the direction, at least then.

eyz:
@deitch, that was how I modeled my PR. Listing every device to format is nearly a full fdtab mapping

deitch:
> that was how I modeled my PR

Meaning one container of `format` for each device?

eyz:
@deitch one device per format

eyz:
Other thoughts? I can make another argument which specifies all of the devices which should be considered for auto formatting if not partitioned

deitch:
@justincormack whichever you think.

justincormack:
Hmm, well I was thinking of the multiple devices as a whitelist not a list to format, but I don't think that will work. I think this is fine as is.

pkg/format/format.go
```diff
@@ -154,13 +155,17 @@ func findDrives() {
 	}
 	sort.Strings(driveKeys)
 	for _, d := range driveKeys {
-		drives[d] = true
+		if deviceVar == "" || deviceVar == d {
+			log.Printf("Considering auto format for device %s", d)
```
eyz: Logging to make the auto format candidate list visible to the end user

eyz: I figured this would be helpful, as well.

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
 var (
 	labelVar  string
 	fsTypeVar string
+	deviceVar string
```
eyz: new var for --device argument

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
 	drives    map[string]bool
 	driveKeys []string
 )
 
 func autoformat(label, fsType string) error {
 	var first string
 	for _, d := range driveKeys {
-		err := exec.Command("sfdisk", "-d", d).Run()
```
eyz: `sfdisk` check has been moved after new `blkid` check

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
+		log.Printf("Considering auto format for device %s", d)
```
eyz: explicitly logging which devices we consider for auto-formatting to the console

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
+
+		out, err := exec.Command("blkid", d).Output()
+		if err == nil {
+			log.Printf("Skipping autoformat of %s. blkid returned: %s", d, out)
+			continue
```
eyz: if `blkid` returns with exitcode 0, then there was content detected on the device; skip the autoformat, and also include the uuid / label / type which was found, as applicable

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
+		} else {
+			if exiterr, ok := err.(*exec.ExitError); ok {
+				if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+					// blkid exitcode 2 (from the non-busybox version) signifies the block device has no detectable content signatures
+					if status.ExitStatus() == 2 {
+						log.Printf("blkid did not find any existing content on %s.", d)
```
eyz: do not skip, as `blkid` exitcode 2 is returned when no content is detected (NOTE: only the `blkid` package from Alpine; the busybox `blkid` does not have this feature, so I install the alpine `blkid` package for this reason)

https://linux.die.net/man/8/blkid mentions this exitcode, should you like to see that man page

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
+					} else {
+						log.Printf("Skipping autoformat of %s. blkid %s returned: %s (exitcode %d)", d, out, status.ExitStatus())
+						continue
```
eyz: if we received an exitcode other than 0 or 2, then we assume there was a fatal error with the check, so skip formatting, and output the result from the `blkid` command; we should not proceed if `blkid` can't read the block device to be safe

pkg/format/format.go
```diff
@@ -23,18 +23,40 @@ const (
...
+					}
+				}
+			}
+		}
+
+		err = exec.Command("sfdisk", "-d", d).Run()
```
eyz: the existing `sfdisk` check is performed if the `blkid` test encounters no issues, and empty content

pkg/format/format.go
```diff
@@ -143,24 +165,31 @@ func findDrives() {
...
 		if ignoreExp.MatchString(d.Name()) {
 			continue
 		}
-		driveKeys = append(driveKeys, filepath.Join("/dev", d.Name()))
-	}
-	sort.Strings(driveKeys)
-	for _, d := range driveKeys {
-		drives[d] = true
+
+		driveKey := filepath.Join("/dev", d.Name())
```
eyz: factored the `driveKey` out of the prior `append` so I can compare it with the value of the `--device` argument (if applicable)

pkg/format/format.go
```diff
@@ -143,24 +165,31 @@ func findDrives() {
...
+
+		// ignore if the deviceVar argument is defined and the driveKey doesn't match the deviceVar
+		if deviceVar != "" && driveKey != deviceVar {
+			continue
+		}
```
eyz: if the `--device` argument exists (non-empty) and the `driveKey` does not match the specified device from `--device`, then skip this device; this causes us to follow the above pattern of ignoring devices, but as applied to the new `--device` argument in this case

pkg/format/format.go
```diff
@@ -143,24 +165,31 @@ func findDrives() {
...
+
+		driveKeys = append(driveKeys, driveKey)
```
eyz: append the factored `driveKey` into `driveKeys` as it was before

pkg/format/format.go
```diff
@@ -143,24 +165,31 @@ func findDrives() {
...
 	}
 }
 
 func init() {
-	flag.StringVar(&labelVar, "label", "", "Disk label to apply")
 	flag.StringVar(&fsTypeVar, "type", "ext4", "Type of filesystem to create")
+	flag.StringVar(&labelVar, "label", "", "Disk label to apply")
+	flag.StringVar(&deviceVar, "device", "", "Device to format")
```
eyz: new `--device` argument, which is the explicit single device to allow to be formatted

eyz:
@dave-tucker you mentioned you'd like to take a look

pkg/format/format.go
dave-tucker: I assume that `--device` should be in the format `/dev/sda` then?

pkg/format/format.go
```diff
@@ -23,18 +22,40 @@ const (
...
+				}
+			}
+		}
+
+		err = exec.Command("sfdisk", "-d", d).Run()
 		if err == nil {
 			log.Printf("Partition table found on device %s. Skipping.", d)
 			continue
 		}
+
```
dave-tucker: Why the additional whitespace?

pkg/format/format.go
```diff
@@ -143,24 +164,31 @@ func findDrives() {
 		// this probably shouldn't be so hard
 		// but d.Mode()&os.ModeDevice == 0 doesn't work as expected
 		mode := d.Sys().(*syscall.Stat_t).Mode
+
```
dave-tucker: and here...

pkg/format/format.go
```diff
@@ -143,24 +164,31 @@ func findDrives() {
...
 		if (mode & syscall.S_IFMT) != syscall.S_IFBLK {
 			continue
 		}
+
```
dave-tucker: ... and here ...

pkg/format/format.go
dave-tucker: nit: these were in alphabetical order before, it would be nice to keep them that way.
`device`, `label`, `type`.

dave-tucker:
As another minor nit, could you please include a description of the changes in the git commit message? It's easier to see in the history. Also, the title might be better phrased as `format: Add device argument and blkid content check` - you can include more detail in the commit message.

eyz:
@dave-tucker is there preference about the device format? `/dev/sdX` or just `sdX`?

I was trying not to make large changes to the concepts of auto-discovery, but I'd be glad to have this flag skip findDrives.

So far: I had added spacing to split some of the checks into logical areas, to my eye when in the same function, but I removed the ones you mentioned. I also updated the commit log and the PR title.

eyz:
Dave Tucker and I agreed to change the arguments around a little. `--force` will be used to explicitly format the specified device, otherwise it will be checked for content and will be skipped if content is found.

pkg/format/format.go
```diff
@@ -21,36 +20,106 @@ const (
...
+	forceVar   bool
+	verboseVar bool
+	drives     map[string]bool
+	driveKeys  []string
 )
 
+func hasPartitions(d string) bool {
+	err := exec.Command("sfdisk", "-d", d).Run()
+	return err == nil
+}
```
eyz: factored out `sfdisk` partition check, as it is now used in multiple areas

pkg/format/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apk add --no-cache --initdb -p /out \
     musl \
     sfdisk \
     util-linux \
+    blkid \
```
eyz: the alpine package `blkid` is required
- the `busybox` version returns exitcode 0 when content is found
- the `blkid` package version returns exitcode 2 when content is found on the block device (distinct result)

pkg/format/format.go
```diff
@@ -21,36 +20,106 @@ const (
...
 	}
 
 	if first == "" {
 		return fmt.Errorf("No eligible disks found")
 	}
 
-	if err := format(first, label, fsType); err != nil {
+	if err := format(first, label, fsType, false); err != nil {
 		return err
 	}
```
eyz: - false = do not force re-creation of partitions nor re-format
- true = clear partitions, re-partition, and format (if applicable)

pkg/format/format.go
```diff
@@ -21,36 +20,106 @@ const (
...
+			if out, err := partCmd.CombinedOutput(); err != nil {
+				return fmt.Errorf("Error deleting partitions with sfdisk: %v\n%s", err, out)
+			}
+		} else {
+			if verboseVar {
+				log.Printf("No need to clear partitions, as partitions already exist")
+			}
+		}
+	}
+
```
eyz: support for `format` to optionally clear the partitions and re-partition via `forced` bool

pkg/format/format.go
```diff
@@ -133,52 +203,91 @@ func format(d, label, fsType string) error {
 	return nil
 }
 
+func isBlockDevice(d *os.FileInfo) bool {
+	// this probably shouldn't be so hard
+	// but d.Mode()&os.ModeDevice == 0 doesn't work as expected
+	mode := (*d).Sys().(*syscall.Stat_t).Mode
+	return (mode & syscall.S_IFMT) == syscall.S_IFBLK
+}
+
```
eyz: factored out the block device check, as it is used in multiple places now

pkg/format/format.go
```diff
@@ -83,11 +152,12 @@ func format(d, label, fsType string) error {
 	for i := 0; i < timeout; i++ {
 		stat, err := os.Stat(partition)
 		if err == nil {
-			mode := stat.Sys().(*syscall.Stat_t).Mode
-			if (mode & syscall.S_IFMT) == syscall.S_IFBLK {
+			if isBlockDevice(&stat) {
 				done = true
 				break
 			}
```
eyz: use new factored `isBlockDevice`

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
+		if isBlockDevice(&d) {
+			if verboseVar {
+				log.Printf("/dev/%s is a block device", d.Name())
+			}
+		} else {
+			if verboseVar {
+				log.Printf("/dev/%s is not a block device", d.Name())
+			}
 			continue
 		}
```
eyz: - skip any devices which are not block devices
- using the new `isBlockDevice`

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
+				log.Printf("ignored device /dev/%s during drive autodetection", d.Name())
+			}
 			continue
 		}
 		driveKeys = append(driveKeys, filepath.Join("/dev", d.Name()))
 	}
-	sort.Strings(driveKeys)
-	for _, d := range driveKeys {
-		drives[d] = true
-	}
```
eyz: not needed any longer

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
 }
 
 func init() {
+	flag.BoolVar(&forceVar, "force", false, "Force format of specified single device (default false)")
 	flag.StringVar(&labelVar, "label", "", "Disk label to apply")
 	flag.StringVar(&fsTypeVar, "type", "ext4", "Type of filesystem to create")
+	flag.BoolVar(&verboseVar, "verbose", false, "Enable verbose output (default false)")
```
eyz: new `-force` and `-verbose` flags

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
+	d, err := os.Stat(device)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("%s does not exist", device)
+	}
+	if !isBlockDevice(&d) {
+		return fmt.Errorf("%s is not a block device", device)
+	}
+	// passed checks
+	return nil
 }
```
eyz: factored out device existence and block device checking

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
 
-	findDrives()
-
 	if flag.NArg() > 1 {
 		log.Fatalf("Too many arguments provided")
 	}
 
 	if flag.NArg() == 0 {
+		// auto-detect drives if a device to format is not explicitly specified
+		findDrives()
```
eyz: auto-detecting drives only if there are no explicit devices specified to consider singly

pkg/format/format.go
```diff
@@ -133,52 +203,90 @@ func format(d, label, fsType string) error {
...
+			if err := format(candidateDevice, labelVar, fsTypeVar, forceVar); err != nil {
+				log.Fatalf("%v", err)
+			}
+		} else {
+			// add the deviceVar to the array of devices to consider autoformatting
+			driveKeys = []string{candidateDevice}
+			if err := autoformat(labelVar, fsTypeVar); err != nil {
+				log.Fatalf("%v", err)
+			}
+		}
```
eyz: if an explicit device was specified:
- ensure it is a block device, or error fatally
- if we are forcing (with `-force`) then format, or error fatally
- set `driveKeys` to an array with the specified device as the only element, which is still used by `autoformat`

pkg/format/format.go
eyz: Personal style. I reverted since.

pkg/format/format.go
```diff
@@ -21,36 +20,106 @@ const (
...
+		// there is content, so exit early
+		return false, nil
+	}
+
+	if exiterr, ok := err.(*exec.ExitError); ok {
+		if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+			// blkid exitcode 2 (from the non-busybox version) signifies the block device has no detectable content signatures
+			if status.ExitStatus() == 2 {
+				if verboseVar {
+					log.Printf("blkid did not find any existing content on %s.", d)
```
rn: why is this conditional on `verboseVar` but other output, like above, is not?

eyz: @rn, which other line are you referring to?

pkg/format/format.go
```diff
@@ -21,36 +20,106 @@ const (
...
+				}
+				// no content detected, but continue through all checks
+				isEmpty = true
+			} else {
+				return isEmpty, fmt.Errorf("Could not determine if %s was empty. blkid %s returned: %s (exitcode %d)", d, d, out, status.ExitStatus())
```
eyz: @rn was this the other line you meant? (see above)

eyz:
@rn I still owe @dave-tucker a new test case named `040_packages/006_format_mount/005_by_device_force` or such. I expect to work on that today, but I would also be curious to see if official CI passes. I have already confirmed that `rtf -x run` passes locally in qemu on a local build of this pkg.

eyz:
@dave-tucker @rn the docs have been updated, a new test has been written (and passes locally), and so I think it's up to @dave-tucker to give this a review, and then perhaps we can get this signed and pushed so it can be tested on LinuxKit CI.

pkg/format/format.go
eyz: @rn This line is the verbose output, without a failure condition. The next one is a failure condition.

`blkid` exitcodes:
- `0` = content exists
- `2` = content does not exist
- other = error

eyz: See below -

pkg/format/format.go
```diff
@@ -21,36 +20,133 @@ const (
...
+		exec.Command("mdev", "-s").Run()
+	}
+	if !done {
+		return fmt.Errorf("Error waiting for device %s", awaitedDevice)
+	}
+	// even after the device appears we still have a race
+	time.Sleep(1 * time.Second)
+
+	return nil
+}
```
eyz: @rn just wanted to let you know that I factored this function out of `format` since you looked at this PR. It still passes tests locally.

pkg/format/format.go
rn: NM, just realised that you unconditionally print a message if, for whatever reason, the disk is not formatted and this case is not one of them, so all good

eyz: The above isn't an error case. The one below is an error case. I only log verbose/debugging if `-verbose` is enabled, other than if it's an error (such as below)

rn:
@eyz could you squash the commits? I will then build/push the updated package and it would be great if you could then update the hashs in the YAML (in a separate commit) so that the CI has a chance to test with the update.

eyz:
@rn I just squashed that final commit.

rn:
Thanks! I've build and pushed the updated package for x86 and arm64. The new hash for the format package is `linuxkit/format:440a08cde1778aec60d9fb5de02d27f391cdfe37`. Could you create a *separate* commit which updates the YAMLs files. To do this, in the top-level directory run something like:
```
./scripts/update-component-sha.sh linuxkit/format:158d992b7bf7ab984100c697d7e72161ea7d7382 linuxkit/format:440a08cde1778aec60d9fb5de02d27f391cdfe37
```
and then commit the change.

eyz:
Updated `linuxkit/format' tag, and confirmed they match across the project -

```
$ grep -RF 'linuxkit/format' * | grep -Po 'linuxkit/format:\w+' | sort -u
linuxkit/format:440a08cde1778aec60d9fb5de02d27f391cdfe37

**Comment by eyz** [comment]
@rn no problem. I dropped the hash update again just now. Let me know and I'll make the change ASAP

**Comment by rn** [comment]
Images are pushed for arm64 and x86. Could you update the Dockerfile to use `linuxkit/alpine:d1778ee29f11475548f0e861f57005a84e82ba4e`

**Comment by eyz** [comment]
linuxkit/alpine:d1778ee29f11475548f0e861f57005a84e82ba4e has been updated in pkg/format/Dockerfile, and re-squashed

File: pkg/format/format.go [hunk 31]
```diff
@@ -21,36 +20,133 @@ const (
...
-		first = d
-		break
 	}
 
 	if first == "" {
 		return fmt.Errorf("No eligible disks found")
 	}
 
-	if err := format(first, label, fsType); err != nil {
+	if err := format(first, label, fsType, false); err != nil {
```
rn: `golint` complains about this:
```
./format.go:97:2: redundant if ...; err != nil check, just return error instead.
sh: redundant: unknown operand
```

pkg/format/format.go
```diff
@@ -21,36 +20,133 @@ const (
...
+
 func autoformat(label, fsType string) error {
 	var first string
 	for _, d := range driveKeys {
-		err := exec.Command("sfdisk", "-d", d).Run()
+		if verboseVar {
+			log.Printf("Considering auto format for device %s", d)
+		}
+		// break the loop with the first empty device we find
+		isEmpty, err := isEmptyDevice(d)
```
rn: Since the code below needs changing, might as well refactor this as well to
```
if isEmpty, err := isEmptyDevice(d); err != nil {
  return err
} else {
  if isEmpty == true {
    first = d
    break
}
```
This limits the scope of `isEmpty`

pkg/format/format.go
```diff
@@ -21,36 +20,133 @@ const (
...
 }
 
-func format(d, label, fsType string) error {
+func refreshDevicesAndWaitFor(awaitedDevice string) error {
+	exec.Command("mdev", "-s").Run()
+
+	// wait for device
+	var done bool
+	for i := 0; i < timeout; i++ {
+		stat, err := os.Stat(awaitedDevice)
```
rn: same as my comment here above.

pkg/format/format.go
eyz: Doing so now. Good point.

pkg/format/format.go
eyz: Trying change to just `
	return format(first, label, fsType, false)

pkg/format/format.go
eyz: Trying change to just
```
	return format(first, label, fsType, false)
```

eyz:
@rn can you run CI style checks on 91bcd4d3045c38df22e050ecaa6d5fc729e234ee ? I inverted and collapsed the err like you requested.

eyz:
@rn made your recommended changes. Please re-test golint

rn:
build and pushed x86/arm64 package. Please update YAML files with the new pacakge: `linuxkit/format:f180a74d878c8c0c86f6208f9311474c91452a79`

eyz:
Pushed tag update to linuxkit/format:f180a74d878c8c0c86f6208f9311474c91452a79

eyz:
@rn Sorry about that. I hadn't reverted my testing before I pushed. Confirmed no testing tags. Also, can you please redact the tag with my testing registry name in your comment? I am not supposed to refer to my organization per my agreement to contribute to open source.

PR #2534, opened 2017-09-24
Better error message when running qemu on macOS when --fw not supplied

### LINKED ISSUE #2510: linuxkit run qemu uefi fails to find firmware on macOS
**Description**
`linuxkit run qemu --uefi` fails because it cannot find ovmf:

```
$ linuxkit run qemu --uefi lk.iso
FATA[0000] File [/usr/share/ovmf/bios.bin] does not exist, please ensure OVMF is installed
```

This can be fixed manually by downloading Tianocore explicitly and then passing it the path:

```
$ linuxkit run qemu --uefi --fw ./OVMF.fd lk.iso
```

The default path is set [here](https://github.com/linuxkit/linuxkit/blob/95a1c51d0da4f5774e1ca4c6ab6d3a4a7e49d430/src/cmd/linuxkit/run_qemu.go#L24) to `/usr/share/ovmf/bios.bin`.

While that might work on Linux, it does not work on macOS. You cannot install anything into `/usr/share` because it is a protected path.

We probably need some sane default, perhaps shipping it with `linuxkit`? That would be an issue as a separate binary.

```
linuxkit run qemu --uefi someimg
```

As above.

It should work.

deitch:

**- What I did**
Added a more descriptive error message when running qemu on macOS

**- How I did it**
Added an additional check if `runtime.GOOS == "darwin"`

**- How to verify it**
Run it on macOS with `linuxkit run qemu ` both with and without `--fw <fw>`. It should give the more descriptive error message when `--fw` is not supplied.

More descriptive error message when firmware not supplied on macOS for qemu.

Armadillo

Fixes #2510

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -359,6 +359,10 @@ func runQemuLocal(config QemuConfig) error {
 	// Check for OVMF firmware before running
 	if config.UEFI {
 		if config.FWPath == "" {
+			// there is no default on mac
+			if runtime.GOOS == "darwin" {
+				return fmt.Errorf("To run qemu with UEFI firmware on macOS, you must specify the path to locally installed OVMF firmware as `--fw <path>`. You can download OVMF from https://sourceforge.net/projects/edk2/files/OVMF/ ")
+			}
 			config.FWPath = defaultFWPath
 		}
 		if _, err := os.Stat(config.FWPath); err != nil {


deitch:
> LGTM on green

Will merge on green.

Any reason we don't have GitHub block merges until checks have completed?

deitch:
Wow, CI running for an almost an hour and still going? Anything we can do to boost it? Worth getting a bare metal or two from packet?

Issue #2530, opened 2017-09-22
add metadata to linuxkit.yml example and add to some blueprints


rn:
This is more a reminder to self that we should add an example and test for the `metadata` file section flag introduced by https://github.com/moby/tool/pull/122. We should also add this to the blueprints.

rn:
the yml files have the metadata, just need a test

rn:
closing. toplevel `linuxkit.yml` has an example

PR #2536, opened 2017-09-25
Update kernel and LCOW

rn:
- Update the 4.12.x kernel with latest VMBus/Hyper-V socket patches
  This *seems* to fix the `vmbus_on_event()` crash we have been seeing.
  based on: https://github.com/Microsoft/opengcs/pull/147
- Update OpenGCS to latest (only minor changes to logging)
- Add the YAML file used to generate the image to the initrd (wip #2530)

tapir


blueprints/lcow.yml
```diff
@@ -1,10 +1,14 @@
 kernel:
-  image: linuxkit/kernel:4.12.14
+  # XXX Use the content hash as there are several 4.12.14 images on hub
+  image: linuxkit/kernel:4.12.14-c4d19a4de21def5a5134a6f70ef7212c31104bf1
```
ijc: OK, but does a force push not overwrite?

Is it really possible to have multiple tags with the same name? (I suspect you meant something else)

rn: It's to prevent the case where you may have had a `linuxkit/kernel:4.12.14` in your local docker store so I'm explicitly pulling the version with the hash. I'll drop the hash again on 4.12.15 update, if there is one

ijc: OIC, the unusual case here is that we are doing a rev of the packaging (and patches) but not the underlying kernel version so there is a greater than usual chance of picking up a stale image.

This seems like a shortcoming in our packaging strategy, but not one which should be addressed in this PR. So my LGTM stands once things go green.

rn: we do have `moby build -pull` which will always pull. Any downstream builds should use this. But me as a dev, I never do because I typically have most/all the images locally, so this is juts to protect me (and others).

PR #2537, opened 2017-09-25
kubernetes: build kubelet from source & add openssl

ijc:
This builds the binaries for the kubernetes image from scratch rather than pulling unverified images of the Internet.

I also added openssl to the container since I found I needed it to sign my csr for RBAC purposes.

I had hoped to also bump cri-containerd here, but need to sort out a new build-time dependency on `libapparmor-dev` (which is only in Alpine edge/testing at the moment) so that will come later.

### DIFF
--- a/projects/kubernetes/kubernetes/Dockerfile
+++ b/projects/kubernetes/kubernetes/Dockerfile
@@ -50,6 +50,7 @@ RUN apk add --no-cache --initdb -p /out \
     iptables \
     libc6-compat \
     musl \
+    openssl \
     socat \
     util-linux \
     && true
--- a/projects/kubernetes/kubernetes/Dockerfile
+++ b/projects/kubernetes/kubernetes/Dockerfile
@@ -1,13 +1,38 @@
-FROM linuxkit/alpine:79987c65c66700171c073151c1d3f0372597bec2 AS build
+FROM linuxkit/alpine:28254e4530703db4caa6b0199a025c30a987dfa1 AS build
 
 ENV kubernetes_version v1.7.6
 ENV cni_version        v0.6.0
 
-ENV kube_release_artefacts "https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64"
-
 RUN apk add -U --no-cache \
+  bash \
+  coreutils \
   curl \
+  findutils \
+  git \
+  go \
+  grep \
+  libc-dev \
+  linux-headers \
+  make \
+  rsync \
   && true
+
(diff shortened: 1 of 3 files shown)


projects/kubernetes/kubernetes/Dockerfile
```diff
@@ -28,15 +56,19 @@ RUN apk add --no-cache --initdb -p /out \
     socat \
     util-linux \
     && true
+
+RUN cp _output/bin/kubelet /out/usr/bin/kubelet
+RUN cp _output/bin/kubeadm /out/usr/bin/kubeadm
+RUN cp _output/bin/kubectl /out/usr/bin/kubectl
+
+RUN strip /out/usr/bin/kubelet /out/usr/bin/kubeadm /out/usr/bin/kubectl
```
rn: You also strip above, once should be enough

ijc: Dang, I _knew_ I did it on Friday but missed it when I checked so I added it again this morning, Doh!

justincormack: I don't believe it is safe to use `strip` on Go binaries. You can use the Go build options that we use in our build script that does a partial `strip`.

projects/kubernetes/kubernetes/Dockerfile
```diff
@@ -1,14 +1,42 @@
-FROM linuxkit/alpine:79987c65c66700171c073151c1d3f0372597bec2 AS build
+FROM linuxkit/alpine:28254e4530703db4caa6b0199a025c30a987dfa1 AS build
 
 ENV kubernetes_version v1.7.6
 ENV cni_version        v0.6.0
 
 ENV kube_release_artefacts "https://dl.k8s.io/${kubernetes_version}/bin/linux/amd64"
```
justincormack: this is not being used?

projects/kubernetes/kubernetes/Dockerfile
ijc: What are the symptoms? These ones worked well enough to run the sock shop demo, but perhaps the issues are subtle?

ijc:
Force pushed with:
* Dropped both strips for now until the reasons for avoid are understood (the kube build system needs a bit more investigation if we are to add `-ldflags "-s"` which I think is what @justincormack was referring to).
* Dropped the now unused env
* New patch to disable cAdvisor on public port (the default in kubelet).

The apparmor issues with cri-containerd are resolved but that is still not updated here due to a new issue which is a requirement for `/dev/disk/by-uuid` which I am investigating.

PR #2540, opened 2017-09-26
Add support to make raw efi images

deitch:

UPDATE: removed `raw-bios` from this PR, separated to [2558](https://github.com/linuxkit/linuxkit/pull/2558)

**- What I did**
Created a new `tools/` entry and therefore images:

* `tools/mkimage-raw-efi/`: create raw images that are bootable EFI

**- How I did it**
1. Created new directory
2. Created entrypoint for EFI that creates correct image given the `tar` input on stdin from moby

The intent, once this is merged, is to update moby build to add the output format `raw-efi`.

**Note:** There is some overlap between this raw format builder `raw-efi` and `iso-efi`. I would _like_ to extract common code to all, perhaps simplifying to just 1 docker image, but it was enough work to get this here and working fully. Let's merge it in first and get it working, then remove duplication.

**- How to verify it**
1. Create a `linuxkit.tar` as the usual input format for builders. This is *not* the output of `moby build -format tar ...`, since that puts everything in the tar as if on an iso. Instead, recreate the output of `tarToInitrd()` by either getting kernel+cmdline+initrd and tarring up, or doing `moby build -format raw ...` and taking the 3 files off the iso image and tarring them up.
2. Build the images using `make tag`
3. Create a raw image the way `moby build` would: `cat linuxkit.tar | docker run -i --rm <image_name> > linuxkit-bios.efi`
4. Run the image

To run the image, you can load it into qemu, vbox, a bare metal server, or hyperkit.

For qemu, my command-lines (shamelessly hijacked and modified from `linuxkit run qemu ...`  are:

```
$ /usr/local/bin/qemu-system-x86_64 -device virtio-rng-pci -smp 1 -m 1024 -uuid 940c5cec-0640-47cf-aa48-feea2dd23037 -pidfile linuxkit.raw-state/qemu.pid -drive file=./linuxkit-efi.raw,index=0,media=disk -drive if=pflash,format=raw,file=./OVMF.fd -device virtio-net-pci,netdev=t0,mac=12:7b:f1:13:29:2a -netdev user,id=t0 --nographic
```

You can run the EFI image in hyperkit with:

```
$ /usr/local/bin/hyperkit -A -u -F linuxkit-state/hyperkit.pid -c 1 -m 1024M -s 0:0,hostbridge -s 31,lpc -s 1:0,virtio-vpnkit,path=$HOME/Library/Containers/com.docker.docker/Data/s50,uuid=1edc6091-1bb8-4f9e-9550-d5a4844589a9 -U 841601a7-189b-4a13-8dea-6e65df9875f2 -s 2:0,virtio-blk,./linuxkit-efi.raw  -s 3,virtio-sock,guest_cid=3,path=./linuxkit-state -s 4,virtio-rnd -l com1,stdio -f bootrom,/Applications/Docker.app/Contents/Resources/uefi/UEFI.fd,,
```

Add support for raw EFI images

Marmot

Cannot recall if I used a marmot before, but saw them in France and in Western Canada :-)

### DIFF
--- a/tools/mkimage-raw-efi/make-efi
+++ b/tools/mkimage-raw-efi/make-efi
@@ -1,0 +1,124 @@
+#!/bin/sh
+
+set -e
+# for debugging
+[ -n "$DEBUG" ] && set -x
+
+IMGFILE=$PWD/disk.img
+
+
+# we want everything except the final result to stderr
+( exec 1>&2;
+
+ESP_FILE=$PWD/boot.img
+
+
+
+# get the GRUB2 boot file name
+ARCH=`uname -m`
+case $ARCH in
+x86_64)
+  BOOTFILE=BOOTX64.EFI
+  LINUX_ENTRY=linuxefi
+  INITRD_ENTRY=initrdefi
+  ;;
+aarch64)
+  BOOTFILE=BOOTAA64.EFI
+  LINUX_ENTRY=linux
+  INITRD_ENTRY=initrd
+  ;;
+esac
+
+mkdir -p /tmp/efi
+cd /tmp/efi
+
+# input is a tarball on stdin with kernel and cmdline in /boot
+# output is an iso on stdout
+
+# extract. BSD tar auto recognises compression, unlike GNU tar
+# only if stdin is a tty, if so need files volume mounted...
+[ -t 0 ] || bsdtar xzf -
+
+INITRD="$(find . -name '*.img')"
+KERNEL="$(find . -name kernel)"
+CMDLINE_FILE="$(find . -name cmdline)"
+CMDLINE="$(cat $CMDLINE_FILE )"
+
+# PARTUUID for root
+PARTUUID=$(cat /proc/sys/kernel/random/uuid)
+
+cp /usr/local/share/$BOOTFILE .
+
+mkdir -p EFI/BOOT
+cat >> EFI/BOOT/grub.cfg <<EOF
+set timeout=0
+set gfxpayload=text
+menuentry 'LinuxKit ISO Image' {
+	$LINUX_ENTRY /kernel ${CMDLINE} text
+  $INITRD_ENTRY /initrd.img
+}
+EOF
+
+#
+# calculate sizes
+KERNEL_FILE_SIZE=$(stat -c %s "$KERNEL")
+INITRD_FILE_SIZE=$(stat -c %s "$INITRD")
+EFI_FILE_SIZE=$(stat -c %s "$BOOTFILE")
+# minimum headroom needed in ESP, in bytes
+# 511KiB headroom seems to be enough
+ESP_HEADROOM=$(( 1024 * 1024 ))
+
+# this is the minimum size of our EFI System Partition
+ESP_FILE_SIZE=$(( $KERNEL_FILE_SIZE + $INITRD_FILE_SIZE + $EFI_FILE_SIZE + $ESP_HEADROOM ))
+
+# (x+1024)/1024*1024 rounds up to multiple of 1024KB, or 2048 sectors
+# some firmwares get confused if the partitions are not aligned on 2048 blocks
+# we will round up to the nearest multiple of 2048 blocks
+# since each block is 512 bytes, we want the size to be a multiple of
+# 2048 blocks * 512 bytes = 1048576 bytes = 1024KB
+ESP_FILE_SIZE_KB=$(( ( ($ESP_FILE_SIZE+1024) / 1024 ) / 1024 * 1024 ))
+# and for sectors
+ESP_FILE_SIZE_SECTORS=$(( $ESP_FILE_SIZE_KB * 2 ))
+
+# create a raw disk with an EFI boot partition
+# Stuff it into a FAT filesystem, making it as small as possible.
+mkfs.vfat -v -C $ESP_FILE $(( $ESP_FILE_SIZE_KB )) > /dev/null
+echo "mtools_skip_check=1" >> /etc/mtools.conf && \
+mmd -i $ESP_FILE ::/EFI
+mmd -i $ESP_FILE ::/EFI/BOOT
+mcopy -i $ESP_FILE $BOOTFILE ::/EFI/BOOT/
+mcopy -i $ESP_FILE EFI/BOOT/grub.cfg ::/EFI/BOOT/
+mcopy -i $ESP_FILE $KERNEL ::/
+mcopy -i $ESP_FILE $INITRD ::/
+
+
+# now make our actual filesystem image
+# how big an image do we want?
+# it should be the size of our ESP file+1MB for BIOS boot + 1MB for MBR + 1MB for GPT
+ONEMB=$(( 1024 * 1024 ))
+SIZE_IN_BYTES=$(( $(stat -c %s "$ESP_FILE") + 4*$ONEMB ))
+
+# and make sure the ESP is bootable for BIOS mode
+# settings
+BLKSIZE=512
+MB_BLOCKS=$(( $SIZE_IN_BYTES / $ONEMB ))
+
+# make the image
+dd if=/dev/zero of=$IMGFILE bs=1M count=$MB_BLOCKS
+
+ESP_SECTOR_START=2048
+ESP_SECTOR_END=$(( $ESP_SECTOR_START + $ESP_FILE_SIZE_SECTORS - 1 ))
+
+# create the partitions - size of the ESP must match our image
+# and make sure the ESP is bootable for BIOS mode
+sgdisk --clear \
+    --new 1:$ESP_SECTOR_START:$ESP_SECTOR_END --typecode=1:ef00 --change-name=1:'EFI System' --partition-guid=1:$PARTUUID \
+    --attributes 1:set:2 \
+     $IMGFILE
+
+# copy in our EFI System Partition image
+dd if=$ESP_FILE of=$IMGFILE bs=$BLKSIZE count=$ESP_FILE_SIZE_SECTORS conv=notrunc seek=$ESP_SECTOR_START
+
+)
+
+cat $IMGFILE
--- a/tools/mkimage-raw-efi/patches/0001-TPM-build-issue-fixing.patch
+++ b/tools/mkimage-raw-efi/patches/0001-TPM-build-issue-fixing.patch
@@ -1,0 +1,72 @@
+From 617b08377dbaa9ea3876b5585fe0ba36286fbed6 Mon Sep 17 00:00:00 2001
+From: Dennis Chen <dennis.chen@arm.com>
+Date: Thu, 17 Aug 2017 05:47:55 +0000
+Subject: [PATCH] TPM: build issue fixing
+
+Fix the build issue on arm64 and amd64.
+
+Signed-off-by: Dennis Chen <dennis.chen@arm.com>
+---
+ grub-core/kern/efi/tpm.c | 12 ++++++------
+ include/grub/efi/tpm.h   |  4 ++--
+ 2 files changed, 8 insertions(+), 8 deletions(-)
+
+diff --git a/grub-core/kern/efi/tpm.c b/grub-core/kern/efi/tpm.c
+index c9fb3c1..ed40f98 100644
+--- a/grub-core/kern/efi/tpm.c
++++ b/grub-core/kern/efi/tpm.c
+@@ -175,7 +175,7 @@ grub_tpm1_log_event(grub_efi_handle_t tpm_handle, unsigned char *buf,
+ 		    grub_size_t size, grub_uint8_t pcr,
+ 		    const char *description)
+ {
+-  Event *event;
++  TCG_PCR_EVENT *event;
+   grub_efi_status_t status;
+   grub_efi_tpm_protocol_t *tpm;
+   grub_efi_physical_address_t lastevent;
+@@ -188,15 +188,15 @@ grub_tpm1_log_event(grub_efi_handle_t tpm_handle, unsigned char *buf,
+   if (!grub_tpm_present(tpm))
+     return 0;
+ 
+-  event = grub_zalloc(sizeof (Event) + grub_strlen(description) + 1);
++  event = grub_zalloc(sizeof (TCG_PCR_EVENT) + grub_strlen(description) + 1);
+   if (!event)
+     return grub_error (GRUB_ERR_OUT_OF_MEMORY,
+ 		       N_("cannot allocate TPM event buffer"));
+ 
+-  event->pcrindex = pcr;
+-  event->eventtype = EV_IPL;
+-  event->eventsize = grub_strlen(description) + 1;
+-  grub_memcpy(event->event, description, event->eventsize);
++  event->PCRIndex = pcr;
++  event->EventType = EV_IPL;
++  event->EventSize = grub_strlen(description) + 1;
++  grub_memcpy(event->Event, description, event->EventSize);
+ 
+   algorithm = TCG_ALG_SHA;
+   status = efi_call_7 (tpm->log_extend_event, tpm, buf, (grub_uint64_t) size,
+diff --git a/include/grub/efi/tpm.h b/include/grub/efi/tpm.h
+index e2aff4a..fb3bb0e 100644
+--- a/include/grub/efi/tpm.h
++++ b/include/grub/efi/tpm.h
+@@ -69,7 +69,7 @@ struct grub_efi_tpm_protocol
+ 					    grub_efi_uint32_t TpmOutputParameterBlockSize,
+ 					    grub_efi_uint8_t *TpmOutputParameterBlock);
+   grub_efi_status_t (*log_extend_event) (struct grub_efi_tpm_protocol *this,
+-					 grub_efi_physical_address_t HashData,
++					 grub_efi_uint8_t *HashData,
+ 					 grub_efi_uint64_t HashDataLen,
+ 					 grub_efi_uint32_t AlgorithmId,
+ 					 TCG_PCR_EVENT *TCGLogData,
+@@ -129,7 +129,7 @@ struct grub_efi_tpm2_protocol
+ 				      grub_efi_boolean_t *EventLogTruncated);
+   grub_efi_status_t (*hash_log_extend_event) (struct grub_efi_tpm2_protocol *this,
+ 					      grub_efi_uint64_t Flags,
+-					      grub_efi_physical_address_t *DataToHash,
++					      grub_efi_uint8_t *DataToHash,
+ 					      grub_efi_uint64_t DataToHashLen,
+ 					      EFI_TCG2_EVENT *EfiTcgEvent);
+   grub_efi_status_t (*submit_command) (struct grub_efi_tpm2_protocol *this,
+-- 
+2.7.4
+
--- a/tools/mkimage-raw-efi/patches/0002-video-Allow-to-set-pure-text-mode-in-case-of-EFI.patch
+++ b/tools/mkimage-raw-efi/patches/0002-video-Allow-to-set-pure-text-mode-in-case-of-EFI.patch
@@ -1,0 +1,33 @@
+From ac7afa666cb2b7b133b6e27bcf22c9cd90a2936a Mon Sep 17 00:00:00 2001
+From: Dennis Chen <dennis.chen@arm.com>
+Date: Wed, 6 Sep 2017 09:06:54 +0000
+Subject: [PATCH] video: Allow to set pure 'text' mode in case of EFI
+
+Current code doesn't accept a pure text mode when booting
+from UEFI firmware on i386 platform, this will result in
+below error message even we already have 'set gfxpayload=text'
+configured: "no suitable video mode found". This often happens
+when we boot a VM which UEFI firmware doesn't include 'suitable'
+video modes.
+
+Signed-off-by: Dennis Chen <dennis.chen@arm.com>
+---
+ grub-core/loader/i386/linux.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/grub-core/loader/i386/linux.c b/grub-core/loader/i386/linux.c
+index 5fdfea3..8cf1086 100644
+--- a/grub-core/loader/i386/linux.c
++++ b/grub-core/loader/i386/linux.c
+@@ -49,7 +49,7 @@ GRUB_MOD_LICENSE ("GPLv3+");
+ #include <grub/efi/efi.h>
+ #define HAS_VGA_TEXT 0
+ #define DEFAULT_VIDEO_MODE "auto"
+-#define ACCEPTS_PURE_TEXT 0
++#define ACCEPTS_PURE_TEXT 1
+ #elif defined (GRUB_MACHINE_IEEE1275)
+ #include <grub/ieee1275/ieee1275.h>
+ #define HAS_VGA_TEXT 0
+-- 
+2.7.4
+
--- a/tools/mkimage-raw-efi/Dockerfile
+++ b/tools/mkimage-raw-efi/Dockerfile
@@ -1,0 +1,66 @@
+FROM linuxkit/alpine:8b53d842a47fce43464e15f65ee2f68b82542330 AS grub-build
+
+RUN apk add \
+  automake \
+  make \
+  bison \
+  gettext \
+  flex \
+  gcc \
+  git \
+  libtool \
+  libc-dev \
+  linux-headers \
+  python3 \
+
+  autoconf
+
+# because python is not available
+RUN ln -s python3 /usr/bin/python
+
+ENV GRUB_MODULES="part_gpt fat ext2 iso9660 gzio linux acpi normal cpio crypto disk boot crc64 gpt \
+search_disk_uuid tftp verify xzio xfs video"
+ENV GRUB_COMMIT=6782f6d431d22b4e9ab14e94d263795c7991e160
+
+COPY patches/* /patches/
+RUN mkdir /grub-lib && \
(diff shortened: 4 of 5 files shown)


deitch:
@arm64b, can you validate the EFI works on arm? I used much of your code from `iso-efi`, but I want another validation.

arm64b:
on arm64, for the latest code:
# cd tools/ mkimage-raw-efi
# make ORG=m0926 HASH=test
# tar cvf linuxkit.tar demo-no-dhcp-cmdline demo-no-dhcp-initrd.img demo-no-dhcp-kernel
# cat linuxkit.tar | docker run -i --rm m0926/mkimage-raw-efi:test > linuxkit-efi.raw
stat: can't stat '': No such file or directory

I use below command to generate the `kernel`, `initrd` and `cmdline`:
# bin/moby build demo-no-dhcp.yml

Is there something that I am missing?

deitch:
I have an idea, but I will look. Can you post the contents of `demo-no-dhcp.yml` please?

deitch:
First thing: when `moby build` generates a tar file, it includes the files with fixed names, and only renames them on output. So you need to name the files:

* `kernel`
* `initrd.img`
* `cmdline`

Try renaming them and see?

deitch:
@arm64b on the linuxkit packet ARM server, I was able to get the build to work using files named `kernel` `initrd.imd` `cmdline` in the `.tar` file.

I could not get qemu to boot up, but mostly because I have never launched qemu on arm. Can you modify your tar file, build the `.raw` image, and then boot qemu?

Also, would you mind posting your `qemu` command line here, so I can reuse it?

arm64b:
demo-no-dhcp.yml file:
```
root@LinuxKit:/home/dennis/linuxkit-dev# cat demo-no-dhcp.yml
kernel:
  image: linuxkit/kernel:4.9.50
  cmdline: "console=ttyAMA0"
init:
  - linuxkit/init:851e9c3ad0574d640b733b92fdb26c368d2f7f8f
  - linuxkit/runc:a1b564248a0d0b118c11e61db9f84ecf41dd2d2a
  - linuxkit/containerd:06876ceef325e49e9ba119659357768d5df89075
  - linuxkit/ca-certificates:e44b0a66df5a102c0e220f0066b0d904710dcb10
onboot:
  - name: sysctl
    image: linuxkit/sysctl:154913b72c6f1f33eb408609fca9963628e8c051
onshutdown:
  - name: shutdown
    image: arm64v8/busybox:latest
    command: ["/bin/echo", "so long and thanks for all the fish"]
services:
  - name: getty
    image: linuxkit/getty:797cb79e0a229fcd16ebf44a0da74bcec03968ec
    env:
     - INSECURE=true
  - name: rngd
    image: linuxkit/rngd:558e86a36242bb74353bc9287b715ddb8567357e
  - name: nginx
    image: arm64v8/nginx:1.12
    capabilities:
     - CAP_NET_BIND_SERVICE
     - CAP_CHOWN
     - CAP_SETUID
     - CAP_SETGID
     - CAP_DAC_OVERRIDE
files:
  - path: etc/containerd/config.toml
    contents: |
      state = "/run/containerd"
      root = "/var/lib/containerd"
      snapshotter = "io.containerd.snapshotter.v1.overlayfs"
      differ = "io.containerd.differ.v1.base-diff"
      subreaper = false

      [grpc]
      uid = 0
      gid = 0

      [debug]
      level = "info"

      [metrics]
      address = ":13337"
  - path: etc/linuxkit-config
    metadata: yaml
trust:
  org:
    - linuxkit
    - library
```

deitch:
Got it, thanks. Please try with the filename changes.

arm64b:
Do I need to feed a UEFI firmware to the raw linuxkit image (I've created one)?

arm64b:
When I run the raw image with a UEFI fw specified, the output is:
```
root@LinuxKit:/home/dennis/linuxkit-dev#  bin/linuxkit run qemu -uefi -fw "/home/dennis/qemu/bios.bin" linuxkit-efi.raw
Using existing disk [linuxkit-efi.raw] format
WARNING: Image format was not specified for 'linuxkit-efi.raw' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.

>>Start PXE over IPv4.
```
After a while, it will enter into the UEFI shell:
```
UEFI Interactive Shell v2.2
EDK II
UEFI v2.60 (EDK II, 0x00010000)
Mapping table
     BLK1: Alias(s):
          VenHw(F9B94AE2-8BA6-409B-9D56-B9B417F53CB3)
     BLK0: Alias(s):
          VenHw(8047DB4B-7E9C-4C0C-8EBC-DFBBAACACE8F)
Press ESC in 1 seconds to skip startup.nsh or any other key to continue.
Shell>
```
if I don't specify a UEFI fw, the system will hang:
```
root@LinuxKit:/home/dennis/linuxkit-dev# bin/linuxkit run qemu linuxkit-efi.raw
Using existing disk [linuxkit-efi.raw] format
WARNING: Image format was not specified for 'linuxkit-efi.raw' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
```

deitch:
> Do I need to feed a UEFI firmware to the raw linuxkit image 

I think so. I have yet to run qemu on arm. How do you run with iso?

arm64b:
The QEMU command line is:
```
 /usr/bin/qemu-system-aarch64 -device virtio-rng-pci -smp 1 -m 1024 -uuid 294d5ec8-f16b-4788-af82-3f70c6da4181 -pidfile linuxkit-efi.raw-state/qemu.pid -cpu host -enable-kvm -machine virt,gic_version=host -drive file=linuxkit-efi.raw,index=0,media=disk -drive if=pflash,format=raw,file=/home/dennis/qemu/bios.bin -device virtio-net-pci,netdev=t0,mac=f2:02:f5:05:f3:f2 -netdev user,id=t0 -nographic
```

deitch:
Where did you get the UEFI fw `/home/dennis/qemu/bios.bin`?

arm64b:
From the shell output, we can see that the UEFI doesn't recognize any ESP, so no file system be found.

deitch:
> we can see that the UEFI doesn't recognize any ESP

I can work with that. But right now, I cannot get to running it. Maybe when I get your UEFI fw?

arm64b:
> Where did you get the UEFI fw /home/dennis/qemu/bios.bin?

I download it from the internet, the link can be found at the post "ARM boot tracking", I use the same bios.bin to boot ISO image.

arm64b:
```ARM64 UEFI can be found at https://github.com/tianocore/edk2/tree/master/ArmVirtPkg
A pre-built UEFI binary to boot ARM64 kernel can be found at: https://releases.linaro.org/components/kernel/uefi-linaro/latest/release/qemu64/
the binary name is: QEMU_EFI.fd

We need to convert the QEMU_EFI.fd into pflash format first:
$cat QEMU_EFI.fd /dev/zero | dd iflag=fullblock bs=1M count=64 of=arm64_uefi.bin

thus we can specify the UEFI as the bios bin for the LinuxKit with
$linuxkit run qemu -arch "aarch64' -uefi -fw "your/path/to/arm64_uefi.bin" linuxkit
```

deitch:
Thanks. Will try it myself when I get a chance.

justincormack:
I probably won't get a chance to look at this for a few days as travelling.

deitch:
Worked great for me. Here is what I did (from within `$GOPATH/src/github.com/linuxkit/linuxkit`):

```
$ cd tools/mkimage-raw-efi
$ ORG=deitch HASH=test make tag
$ cd ../..
$ mkdir out
$ cd out
$ moby build -format kernel+initrd ../linuxkit.yml
$ mv linuxkit-kernel kernel
$ mv linuxkit-cmdline cmdline
$ mv linuxkit-initrd.img initrd.img
$ tar cvf ../linuxkit.tar *
$ cd ..
$ rm -rf out
$ cat linuxkit.tar | docker run -i --rm deitch/mkimage-iso-efi:test > linuxkit-efi.raw
$ wget https://releases.linaro.org/components/kernel/uefi-linaro/latest/release/qemu64/QEMU_EFI.fd
$ cat QEMU_EFI.fd /dev/zero | dd iflag=fullblock bs=1M count=64 of=arm64_uefi.bin
$ mkdir -p linuxkit.raw-state/
$ /usr/bin/qemu-system-aarch64 -device virtio-rng-pci -smp 1 -m 1024 -uuid 294d5ec8-f16b-4788-af82-3f70c6da4181 -pidfile linuxkit.raw-state/emu.pid -cpu host -enable-kvm -machine virt,gic_version=host -drive file=./linuxkit-efi.raw,index=0,media=disk -drive if=pflash,format=raw,file=./arm64_uefi.bin -device virtio-net-pci,netdev=t0,mac=f2:02:f5:05:f3:f2 -netdev user,id=t0 -nographic
```

Starts up fine, all the way to the getty shell, just like I expected.

> if I don't specify a UEFI fw, the system will hang

That makes sense. You need to give qemu UEFI firmware with which to work.

arm64b:
> That makes sense. You need to give qemu UEFI firmware with which to work.

So that means for 'raw' format image, we must specify the UEFI firmware, if that's true, I suggest we add a `linuxkit run qemu -uefi -fw "/path/to/fw" -raw linuxkit.raw`, then if the user doesn't specify the firmware, the boot process will stop, while not hang.

deitch:
This is more your expertise than mine (and you even have an `@arm.com` email address! :-) ), but I thought that there is no industry-standard "BIOS" for arm, and thus master boot record loaders (which is what `raw-bios`) boots, has no way to launch (unless you have firmware that knows how to read MBR, in which case might as well use UEFI).

`qemu-system-x86_64` ships with SeaBIOS, so it is able to boot a disk with MBR directly, but it does not ship with UEFI. If you want UEFI, you need to give it alternate firmware.

Unless you have a raw image that is formatted correctly for the specific arm system, it will be kind of stuck with nowhere to go, won't it?

justincormack:
Hmm no, it is `tools/mkimage-*` that needs to go away, not `pkg/mkimage-*` I think.

On the 1GB size, that was intentional, as it is required for AWS and that the raw format is currently used for. It is customisable, and should really have an "as small as possible option", yes.

Spinning up Linuxkit has many advantages:
1. We need to get rid of the requirement to have Docker running to use the `moby` tool, as it makes building in a CI pipeline much easier, and we have had many requests for this. So the idea was to move away from using Docker to build images. LinuxKit can run so long as qemu is installed (or a native hypervisor, but qemu is fine generally for these conversions).
2. Most of the conversions (this one is an exception as it uses mtools) use libguestfs, which spins up a qemu VM to do the conversion anyway, so we are doing exactly the equivalent of running linuxkit anyway, just nested inside a container (often inside another VM if you are running on OSX or Windows). So directly using LinuxKit is less emulation for most cases.

Also for disk images we need some way (possibly optional) of writing out a filesystem not an initramfs, as we have images like Kubernetes that are 1.5GB and therefore problematic as an initramfs. I think we need to think about this while changing these. (I am currently looking at doing a non-bootable squashfs image for another purpose, this might be a sane filesystem to use).

deitch:
> On the 1GB size, that was intentional, as it is required for AWS and that the raw format is currently used for. It is customisable, and should really have an "as small as possible option", yes

Ah, I was wondering why it was so big. I actually think it should be the other way around. By default, as small as possible, but have a `--min-image-size 1G` option to conform it to a certain min size.

> We need to get rid of the requirement to have Docker running to use the moby tool, as it makes building in a CI pipeline much easier, and we have had many requests for this

I don't get this part. I would think having docker installed is simpler and lighter than having linuxkit+qemu. Isn't one of the basic ideas of containers that lighter virtualization is simpler and lighter? I would prefer to generate all of my images in a container on the host rather than spinning up VMs.

> Most of the conversions (this one is an exception as it uses mtools) use libguestfs

Which basically spins up a VM, then mounts and uses the VM kernel to do disk/partition/filesystem-level activities. It seems easier and simpler (and faster, and fewer requirements) to just manipulate the disk images directly (which, as you pointed out, this does).

> Also for disk images we need some way (possibly optional) of writing out a filesystem not an initramfs

Like you did with the iso, I presume? This has been on my mind as well. I was thinking of having flags to `moby build` where it can output either (`kernel`,`cmdline`,`initrd.img`) or (full filesystem with `boot/kernel`, `boot/cmdline`). So I could run `moby build -format tar -layout full` or `moby build -format tar -layout initrd`. I could do the same for `-format iso-efi` and for `-format raw-bios` etc.

We would need to decide on the default format (I vote for `initrd`), but it doesn't really matter as long as both options are available.

deitch:
> We need to get rid of the requirement to have Docker running to use the moby tool

We do have another option. As long as we can build these things from a container (without VM), which we have proven for `iso-bios`, `iso-efi`, `raw-bios`, `raw-efi`, and these are just scripted tools running, there is no reason we cannot just implement them _directly_ in `moby/tool`. All we are doing is manipulating bits in images using standard tools. Most of the layouts are straightforward and the tools haven't changed in years. Then you would need neither lkt nor docker.

justincormack:
The CI issue is that most people do not have Docker available for use inside their CI pipeline.

By "implement them directly" what do you mean exactly? Do you mean compile `mtools` as a library or binary for Linux/Mac/Windows and use that from the host?

deitch:
> The CI issue is that most people do not have Docker available for use inside their CI pipeline.

And so if they make building lkt images part of their pipeline, they might be stuck? Is it better to rely on having the lkt binary and qemu?

> By "implement them directly" what do you mean exactly? Do you mean compile mtools as a library or binary for Linux/Mac/Windows and use that from the host?

Pretty much. If you look at the `raw-*` and `iso-*` images, all we really are doing is manipulating binary images using tools. We can do that in go equally well. I don't _love_ it - I prefer the "pipe to docker image" way some of it works now - but if we _must_ get rid of `docker run` in there, I would rather replace it with just in-binary computation rather than `linuxkit run`.

justincormack:
Yes, its easy to put qemu into your build image, then it should be easily possible to build anything.

Yes, its definitely nicer not to require any vm/qemu at all - I much prefer a simple mode that does everything natively, it should be faster too. It is much harder though if you need to make an ext4 filesystem say, but maybe all the things we really need can be done natively. Testing on Windows would be a concern as these programs are probably compiled less on Windows, but hopefully they are mostly pretty portable by now.

It also fixes issues like having the docker images available (they currently have to be pulled from hub which means you need a network connection if you don't have them cached, which is also annoying as a design).

deitch:
> It also fixes issues like having the docker images available

If we use libguestfs and run it in qemu via linuxkit, then you still need the lkt images available over the network, do you not?

> Yes, its definitely nicer not to require any vm/qemu at all - I much prefer a simple mode that does everything natively,

No argument there. I cannot say how hard or easy it would be to implement these tools. Mostly, not too terrible, although some of it is a bit odd. We would need:

* grub2 pre-compiled with everything in place ~660K (just checked on x86_64, imagine ARM is similar)
* bits for syslinux or extlinux so we can just install it, a few K

More importantly, we would need to replicate in go what the following do:

* `mkfs.vfat`
* mtools utils: `mmd`, `mcopy` (the latter being just copying bits to the right place, not too bad)
* `dd` - just copying bits
* `sgdisk` - a bit more work, but doable
* `sfdisk` - similar
* `xorriso` - if we want to do isos this way as well

Actually really is not too bad. I wouldn't be surprised if some golibs already exist for some of this stuff.

deitch:
Either way, my (admittedly biased) view on this would be:

1. Get this PR working and merged in, so we have raw-efi and raw-bios working universally
2. Merge shared parts (some redundancy here)
3. Migrate out of these and into `moby/tool` over time

justincormack:
Yes, currently the linuxkit images are pulled and saved on the filesystem.

I think the `moby` tool should anyway have a mode that lets you specify any docker (or linuxkit) image to run so it iis easier to test out or use custom images as well.

I am a bit reluctant to merge in raw-bios as that would be a regression from the linuxkit version in some ways; raw-efi doesnt exist now so thats a bit different.

deitch:
> I think the moby tool should anyway have a mode that lets you specify any docker (or linuxkit) image to run so it iis easier to test out or use custom images as well

Something to be said for that. But we would have to have a way to specify the expected output - tar stream of 3 files as kernel/initrd/cmdline or tar stream of entire filesystem - so we need the earlier proposal anyways.

> I am a bit reluctant to merge in raw-bios as that would be a regression from the linuxkit version in some ways

which, I think, we can agree to disagree. I like a docker image over a vm image (unless you are dealing in ultra-light-weight like unikernels, but discussion for another day).

> raw-efi doesnt exist now so thats a bit different

OK, so how do you want to proceed? We can merge both of these in, and then change `moby/tool` to switch only the `raw-efi` one. If we add the "select your image to stream to", then we start having options. It also provides a basis for making it native to `moby/tool`.

justincormack:
The input would always be the full tar with kernel in `boot/` as that is the new canonical format.

deitch:
> The input would always be the full tar with kernel in boot/ as that is the new canonical format

So the code for `tarToInitrd()`, at least as far as feeding other "transformers" like for [raw](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L52) and [gcp](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L63) and [qcow](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L76) and [vhd](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L87) and [dynamic-vhd](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L98) and [vmdk](https://github.com/moby/tool/blob/177969dc4a0fb843f635e45ab8e87d2cd2c531cd/src/moby/output.go#L109) will go away, and _everything_ should expect that format, and create initrd if needed?

Actually, since _everything_ except for iso expects the "3-file-salute", why wouldn't we make that the canonical, and let the rare something that needs to unpackage `initrd.img` do it?

justincormack:
We can't make `initrd` canonical as it basically has very broken hardlink support. It works for that use case but you can't make a filesystem out of it properly.

deitch:
> very broken hardlink support

Where do we use hardlinks?

The only thing that uses expanded filesystem is iso (for now). Everything else uses the "3-file-salute". Does it break something in `iso`?

> we could provide both I guess

I agree that we should have a standard, otherwise you have to figure out what input goes from where. Providing options is great, but we don't want to have to provide `--tar-format 3-file` on every `moby build`!

justincormack:
There are hardlinks if the images the user provides has them; the Alpine images don't generally, but for example the `busybox` ones do. We replace with symlinks which is mostly ok, but could cause problems as they are not semantically identical. `cpio` is weird, it supports hardlinks but you have to duplicate the content, which bloats the image even with compression.

deitch:
> There are hardlinks if the images the user provides has them

Good point. Since moby is a tool that can put together a lot of different images, it is possible hard links could exist.

None of our images formats other than `iso` would support them, then, because all of them (other than `iso`) use initrd, correct?

At this point, this is what I am thinking:

1. Separate this into 2 PRs. This one will have the `raw-efi` code only, the other will have the `raw-bios` code only. We can isolate the discussion about bios there, and include things like min-size.
2. Once this one is in, open a `moby/tool` PR extending the format options to include `raw-efi` and changing `raw` to `raw-bios`.
3. Once that one is in, create a separate `moby/tool` PR adding an option for `-format tar-initrd` (or similar) to enable an output file that is post-`tarToInitrd()`
4. Once that one is in, create a separate `moby/tool` PR adding an option to stream instead of write a file. That should make it possible for anyone to use any post-processor just by doing `moby build -format tar -stdout | docker run -i --rm my image:hash > someoutput.img`. That should fulfill the use case above.
5. Once all of the above is done, I might see if I can write some go libs to implement the basic disk image and filesystem functions, so we can get it all natively right in `moby/tool`.

Thoughts?

deitch:
Where I am supposed to find time for all of the above is beyond me, but fun nonetheless.

justincormack:
Streaming output is already supported for `tar` (most of the other formats can't really stream anyway), you can do `moby -q build -format tar -o - linuxkit.yml | tar tf -` to get a list of files in an image for example right now...

deitch:
> Streaming output is already supported for tar

Oh good! One item off my list already done! 👍

deitch:
@justincormack I updated this PR to include _just_ the EFI work. Opened a separate PR for bios discussion around `raw-bios` at [2558](https://github.com/linuxkit/linuxkit/pull/2558).

deitch:
So of my list:

* 1 (2 separate PRs) is done, we can discuss each separately and get this merged in
* 2 will await this being merged
* 4 already exists with the standard `-o -`
* 5 will do on the side with time
* 3 (2 different tar outputs): would you accept a separate `moby build -format tar-initrd` that outputs the tarfile (or stream) as the "3-file-salute"? I can open a PR for it.

justincormack:
Yeah, a streaming tar-initrd should be very simple (there is a list of which outputs can stream), and seems useful, as I think images making their own initrd is a mistake as getting hardlinks right and doing alignment etc are best done in one place. Happy to merge that. That makes it pretty easy to build custom things.

I was thinking last night that as the CI platforms are all Linux, maybe we could make the docker based converters easily run by people outside docker, eg with multi stage builds its actually not too hard to import the code into your build container (even easier if we maybe have one repo for them and maybe make them binaries not shell scripts?), and then use the tar/tar+initrd output method to build rather than calling docker directly. This should be ok for things that need root so long as they don't need loopback mounts and other things that unprivileged containers can't do.

tools/mkimage-raw-efi/make-efi
```diff
@@ -0,0 +1,144 @@
...
+
+# input is a tarball on stdin with kernel and cmdline in /boot
+# output is an iso on stdout
+
+# extract. BSD tar auto recognises compression, unlike GNU tar
+# only if stdin is a tty, if so need files volume mounted...
+[ -t 0 ] || bsdtar xzf -
+
+INITRD="$(find . -name '*.img')"
+KERNEL="$(find . -name kernel -or -name bzImage)"
```
justincormack: remove bzImage, not been used for months now was just transitional.

tools/mkimage-raw-efi/make-efi
```diff
@@ -0,0 +1,144 @@
...
+
+# copy in our EFI System Partition image
+dd if=$ESP_FILE of=$IMGFILE bs=$BLKSIZE count=$ESP_FILE_SIZE_SECTORS conv=notrunc seek=$ESP_SECTOR_START
+
+)
+
+cat $IMGFILE
+
+# How to build a VHDX. Commented out because we are currently not using it
+# Don't delete: It took too long to figure out how to do this...
```
justincormack: you can delete this... this is a new image does not need this.

tools/mkimage-raw-efi/patches/0001-TPM-build-issue-fixing.patch
```diff
@@ -0,0 +1,72 @@
+From 617b08377dbaa9ea3876b5585fe0ba36286fbed6 Mon Sep 17 00:00:00 2001
+From: Dennis Chen <dennis.chen@arm.com>
```
justincormack: Any chance we can upstream this? Is there a PR? Its annoying duplicating it...

tools/mkimage-raw-efi/patches/0002-video-Allow-to-set-pure-text-mode-in-case-of-EFI.patch
```diff
@@ -0,0 +1,33 @@
+From ac7afa666cb2b7b133b6e27bcf22c9cd90a2936a Mon Sep 17 00:00:00 2001
+From: Dennis Chen <dennis.chen@arm.com>
```
justincormack: ditto

tools/mkimage-raw-efi/patches/0001-TPM-build-issue-fixing.patch
deitch: Not sure. I sort of took it as is from @arm64b

tools/mkimage-raw-efi/patches/0002-video-Allow-to-set-pure-text-mode-in-case-of-EFI.patch
deitch: Ditto

tools/mkimage-raw-efi/make-efi
deitch: Updating.

tools/mkimage-raw-efi/make-efi
deitch: Good point.

deitch:
Updated per your requests.

tools/mkimage-raw-efi/patches/0001-TPM-build-issue-fixing.patch
deitch: I think we should take it as is for now. As we discussed yesterday, our effort probably would be better placed in getting a proper bootable kernel image with initrd built right in, and then the whole grub piece goes away, both for `raw-efi` and `iso-efi`.

deitch:
Updated as follows:

* Use the new `linuxkit pkg ... ` process
* Created docker hub repo (thanks @riyazdf )
* Pushed multi-arch image (arm64 and amd64)

I think this answers everything, @justincormack ?

PR #2541, opened 2017-09-27
wpa_supplicant image for linuxkit

nuald:
**- What I did**

Added wpa_supplicant image for linuxkit.

**- How I did it**

Compiled libnl and wpa_supplicant from the original tagged releases into the image.

**- How to verify it**

Please use a custom kernel with Wi-Fi drivers enabled (I have [one](https://hub.docker.com/r/nuald/kernel/), but it may not work for you) and include the `wpa_supplicant` image into the `services` section:
```
services:
  - name: wpa_supplicant
    image: nuald/wpa_supplicant:a5cef22bd214b2845f65636bc9cf60d805712fe5-amd64
    binds:
     - /etc/wpa_supplicant:/etc/wpa_supplicant
    command: ["/sbin/wpa_supplicant", "-i", "wlan0", "-c", "/etc/wpa_supplicant/wpa_supplicant.conf"]
files:
  - path: etc/wpa_supplicant/wpa_supplicant.conf
    contents: |
      network={
        ssid="<ssid>"
        psk="<host>"
      }
 ```

Added wpa_supplicant image for linuxkit.

### DIFF
--- a/pkg/wpa_supplicant/Dockerfile
+++ b/pkg/wpa_supplicant/Dockerfile
@@ -1,0 +1,42 @@
+FROM linuxkit/alpine:28254e4530703db4caa6b0199a025c30a987dfa1 AS build
+
+RUN apk add --no-cache \
+    musl-dev git build-base openssl-dev linux-headers autoconf automake \
+    libtool bison flex file
+ENV WPA_2_6=a68d8f6a0fd493dca8b7049ea7a6a51b7f95cac0
+ENV LIBNL3_4_0RC1=e01b9df629e2f4f833fdc4fe0bda460bb738d136
+
+RUN mkdir -p /usr/src
+RUN git clone -n git://w1.fi/srv/git/hostap.git /usr/src/hostap && \
+    cd /usr/src/hostap && \
+    git checkout $WPA_2_6 && \
+    git clone -n git://git.infradead.org/users/tgr/libnl.git /usr/src/libnl && \
+    cd /usr/src/libnl && \
+    git checkout $LIBNL3_4_0RC1
+
+COPY config /usr/src/hostap/wpa_supplicant/.config
+
+RUN cd /usr/src/libnl && \
+    export CFLAGS="$CFLAGS -D__GLIBC__" && \
+    autoreconf --install && \
+    ./configure --prefix=/usr --enable-cli=no && \
+    make && \
+    make install && \
+    cd /usr/src/hostap/wpa_supplicant && \
+    make
+
+FROM linuxkit/alpine:28254e4530703db4caa6b0199a025c30a987dfa1 AS mirror
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by rn** [comment]
Thanks for this package and docs. Is there a reason you are not using the wireless tools and `wpa_suplicant` from alpine?

**Comment by nuald** [comment]
Frankly, I'm not sure how to do it. wpa_supplicant requires libnl, but it's not included into linuxkit. But if you refer to building wpa_supplicant image using alpine packages (APKBUILD), it would be the same problem - I would need to build libnl, dbus and smartcard libraries too. It would be the overkill, so I've just used the vanilla wpa_supplicant sources. I guess there is a workaround (using some config options and whatnot), but unfortunately I'm not so familiar with linuxkit and alpine to do it.

If I missed anything, please let me know, maybe there is a way to use wpa_supplicant without custom package/image.

**Comment by rn** [comment]
I have to admit, I have no idea how wireless is set up on Linux (the last time I did this was like 15 years ago...) but looking at https://wiki.alpinelinux.org/wiki/Connecting_to_a_wireless_access_point and running:
```
/ # apk add --no-cache wireless-tools wpa_supplicant
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
(1/7) Installing wireless-tools (30_pre9-r0)
(2/7) Installing dbus-libs (1.10.14-r0)
(3/7) Installing expat (2.2.0-r1)
(4/7) Installing dbus (1.10.14-r0)
Executing dbus-1.10.14-r0.pre-install
Executing dbus-1.10.14-r0.post-install
(5/7) Installing libnl3 (3.2.28-r0)
(6/7) Installing pcsc-lite-libs (1.8.21-r0)
(7/7) Installing wpa_supplicant (2.6-r3)
Executing busybox-1.26.2-r5.trigger
Executing dbus-1.10.14-r0.trigger
OK: 9 MiB in 18 packages
```
seems to install a small number of package which we could add to the alpine base and then turn into a LinuxKit package.

nuald:
Yes, however I couldn't find any instructions how to modify the alpine base, and what's the process of updating it. If the base is updated, I'll resubmit a new package.

rn:
made a PR for the update alpine base

It might also be worth looking into adding the tools to the `ip` package which we already use for more custom network configurations. @justincormack may have a view on this

eyz:
`ip` sounds like a great place to put this. Would we want to simply add some shell scripts to the `ip` package until we can get a more formal config format? As @justincormack suggested in another thread, perhaps we could use a debian-style network config file, if that is agreeable

rn:
Closing this due to inactivity. The `ip` package seems to be a better place to add this functionality.

PR #2542, opened 2017-09-27
linux-firmware image for linuxkit

nuald:
**- What I did**

Added linux-firmware image for linuxkit.

Please note that there is nothing special about linuxkit in this docker image, however you may want to include additional firmwares (like Alpine Linux or other distros do) specific for linuxkit. Otherwise (or for any other reasons), please feel free to disregard this PR, and I'll just keep and maintain my own docker image.

**- How I did it**

Pulled the original linux-firmware blobs into the image. Please notice that linux-firmware Git repo doesn't do any tagging, so the latest commit hash has been used.

**- How to verify it**

Please use the sample configuration for your `linuxkit.yml`:

```
init:
  - nuald/linux-firmware:2837c74cdb79bb5d9ef3359e230ef394b98d0965
```

Root FS will have firmwares available at /lib/firmware.

Added linux-firmware image for linuxkit.

### DIFF
--- a/pkg/linux-firmware/Dockerfile
+++ b/pkg/linux-firmware/Dockerfile
@@ -1,0 +1,11 @@
+FROM linuxkit/alpine:28254e4530703db4caa6b0199a025c30a987dfa1 AS mirror
+
+RUN apk add --no-cache git
+ENV HEAD=a61ac5cf8374edbfe692d12f805a1b194f7fead2
+
+RUN git clone -n git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git /var/linux-firmware && \
+    cd /var/linux-firmware && \
+    git checkout $HEAD
+
+FROM scratch
+COPY --from=mirror /var/linux-firmware/ /lib/firmware/
--- a/pkg/linux-firmware/Makefile
+++ b/pkg/linux-firmware/Makefile
@@ -1,0 +1,4 @@
+IMAGE=linux-firmware
+NETWORK=1
+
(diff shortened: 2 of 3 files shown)


rn:
The firmware image is pretty big and I'm not sure it's a good idea to include it in LinuxKit. To me it would make more sense to build a custom package which only contains the firmwares one needs

nuald:
I agree, however for the end-users it's not trivial to figure out how to do it. I see three options here:

- leave everything as is;
- add some options to the toolkit for the automatic grabbing firmware blobs and putting them into the final image;
- just add a note into README that users can download blobs from the linux-firmware git, and add the corresponding entry into "files" section that put those blobs into the final image (that's how I did the first tests with the firmwares).

rn:
I think the best option is to create a `docs/networking.md` (which should also contain general network config options) and then have a section on wireless setup. While the firmware repo is contains non-wireless firmwares, it seems mostly be used for that in our case (from what I can see)

nuald:
I can take on this, no problems. But it would be better if we have:

- wpa_supplicant package available (it's an essential part of the Wi-Fi setup);
- pciutils package in the alpine base (it contains the [PCI ID Database](https://github.com/pciutils/pciids)) - lspci gives only hex codes now, and some users may have troubles reading it.

I'll provide the new PR, and going to close this one. Thanks!

rn:
what do you need `pciutils` for? Just for readability? You can do `apk add --no-cache pciutils` in the `getty` or `sshd` containers for debug if you need to

PR #2543, opened 2017-09-27
tools/alpine: Add wireless-tools and wpa_supplicant

rn:
available as: linuxkit/alpine:a94acee63dc1d0bd5cd5006d66514f62373650f2

Package for x86 and arm64 signed and pushed.

relates #2541 
/cc nuald 

birds-smile

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -99,6 +99,8 @@ tzdata
 util-linux
 util-linux-dev
 vim
+wireless-tools
+wpa_supplicant
 xfsprogs
 xfsprogs-extra
 xorriso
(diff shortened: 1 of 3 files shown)


deitch:
Should we break `linuxkit/alpine` into a separate repo? CI takes a long time to run, and I don't think a single test has a anything to do with an updated `linuxkit/alpine`. Sure, if you _use_ a different hash, things get tested, but this kind of change just creates a new hash, so everything already in-place is entirely unaffected.

Thoughts?

rn:
it's rare that one just updates `linuxkit/alpine` and it doesn't seem worth the hassle of maintaining another repro at this point. I'm fine with waiting for 45mins or whoever long it takes to run CI for changes like this.

deitch:
> it's rare that one just updates linuxkit/alpine and it doesn't seem worth the hassle of maintaining another repro at this point

Yeah, but when we have a bunch of PRs, and stuff is sitting behind it in queue, it costs us.

Is there really a hassle with another repo? Why not separate concerns?

PR #2544, opened 2017-09-27
Initial networking documentation.

nuald:
**- What I did**

Added the initial documentation about LinuxKit networking (mostly covering Wi-Fi functionality). It's a follow up for the #2542 PR.

Note to maintainers: the documentation may have spelling and grammar errors, also I could miss some vital information. I'd recommend to either merge it and fix by the follow up commits, or just copy-paste text, create your own document and close this PR. Also I didn't add any links to it, I'm not sure where to put them.

**- How I did it**

The documentation is based on the real setup I had to do for my Dell XPS 15 9550 (no mention about loading from USB stick though, but it's an offtopic here).

**- How to verify it**

The sample configuration is at the end. I guess only testing on real hardware is a good way to verify it, but as I mentioned above, the configuration does work for me.

Please note, that it refers to not yet merged wpa_supplicant: #2541

Initial networking documentation.

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -1,0 +1,156 @@
+# Networking
+
+While the common pattern of using LinuxKit inside the virtualized environment
+doesn't usually bring any network connectivity issues, they still may occur.
+Moreover, LinuxKit may be used on the real hardware, and in that case the
+network setup could be a major part of the getting the image run.
+
+## General Steps
+
+There is a standard checklist one have to follow to set up the networking:
+
+1. Drivers are loaded by kernel (otherwise, may need a [custom kernel](kernels.md));
+2. Firmware is loaded (otherwise, need to put firmware blobs into the image);
+3. Network interfaces are visible in the system (otherwise, need to revisit items 1 and 2);
+4. Wi-Fi: wpa_supplicant is running (otherwise, need to include it into services section);
+5. DHCP: dhcpcd service is loaded.
+6. May need to set up static IP addressed and routes, but it's out of scope of this document.
+
+Usually, `lspci` provides the information about the detected devices, and
+`dmesg` prints the information whether the drivers for those devices
+are loaded or not (usually if nothing mentioned in the log, then no driver is loaded).
+
+### Wi-Fi Notes
+
+Vanilla LinuxKit kernel doesn't provide the Wi-Fi support, therefore
+the custom kernel should be used. The major options to enable:
+ - `CONFIG_WIRELESS=y` (Wireless);
+ - `CONFIG_CFG80211=y` (Improved wireless configuration API);
+ - `CONFIG_MAC80211=y` (Generic IEEE 802.11 Networking Stack);
+ - `CONFIG_WLAN=y` (Wireless LAN).
+
+`wpa_supplicant` requires CONFIG_CFG80211 to configure the Wi-Fi connections.
+CONFIG_CFG80211_WEXT may be useful too for the old hardware.
+
+The particular driver is better to compile as a module (and it's mandatory if it
+requires a userspace firmware), and enable it with the modprobe later.
+Please note that enabling the driver family is not always enough, for example,
+Broadcom FullMAC device (`CONFIG_BRCMFMAC=m`) may need to have PCIE bus interface
+support (`CONFIG_BRCMFMAC_PCIE=y`) enabled too to work.
+
+In the image configuration please add the modprobe item
+(with the example of the Broadcom FullMAC device):
+```
+onboot:
+  - name: modprobe
+    image: linuxkit/modprobe:<hash>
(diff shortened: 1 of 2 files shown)


deitch:
Overall, I like this. Some minor comments inline.

docs/networking.md
```diff
@@ -0,0 +1,156 @@
+# Networking
+
```
deitch: I think I would start with something more basic, sort of:

```
By default, linuxkit recognizes network devices but does not configure them. You must provide either static configuration or dhcpd, onboot or as a service, to get an IP.

The normal and simplest use case is just to include dhcpd in onboot services <include example here from linuxkit.yml>. This will cause dhcpd to startup *once* and assign an IP.

The alternative method is to make it a service. This ensures dhcpd catches any updates, but may cause issues with services that expect an IP address on startup, since service launch order is *not* guaranteed.

Similarly, using WiFi requires wpa_supplicant in addition to dhcpd.

Finally, if you require any specialized firmware or device drivers for your network card, you may need to load it.

While the common setup .... <continue from here>
```

docs/networking.md
```diff
@@ -0,0 +1,156 @@
...
+```
+services:
+  - name: dhcpcd
+    image: linuxkit/dhcpcd:<hash>
+    command: ["/sbin/dhcpcd", "wlan0"]
+```
+
+You may want to use a configuration file too - the sample above provides just basic functionality.
+
+## Sample linuxkit.yml
```
deitch: I would remove this from here, and make a `wifi.yml` in the `examples/` directory. Just reference it in the docs.

deitch:
Looks good to me, but let's not do anything here until #2541 is in

examples/wifi.yml
```diff
@@ -0,0 +1,37 @@
+kernel:
```
rn: All the other examples can be used as is, ie you can do `moby build <example>.yml`, but you can't do it with this one. I know @deitch suggested to move it here, but I'm not in favour as this is more of a template than a fully functioning example (like the others). I would prefer if this in the documentation instead.

docs/networking.md
```diff
@@ -0,0 +1,134 @@
...
+files:
+  - path: /lib/firmware/brcm/brcmfmac43602-pcie.bin
+    source: brcmfmac43602-pcie.bin
+```
+
+Please note that `dmesg` may still give a warning that firmware is missing, but
+it can be ignored. Please use `lsmod` to verify that the module is actually loaded.
+Also you could see the network interfaces available using `ifconfig -a`.
+
+### WPA Supplicant
```
rn: This should really be a subsection of the WiFi notes

docs/networking.md
```diff
@@ -0,0 +1,134 @@
...
+  - name: modprobe
+    image: linuxkit/modprobe:<hash>
+    command: ["modprobe", "-a", "brcmfmac"]
+```
+
+The next step is getting an optional firmware. Please use the official
+[Linux Wireless wiki](https://wireless.wiki.kernel.org/welcome) and search for the
+driver instructions there (`dmesg` may give the helpful messages too).
+
+### Userspace firmware
```
rn: I would move this maybe above the WiFi notes as this is generic, ie applies to other NICs as well (but mostly WiFi)

docs/networking.md
```diff
@@ -0,0 +1,134 @@
...
+3. Network interfaces are visible in the system (otherwise, need to revisit items 1 and 2);
+4. Wi-Fi: wpa_supplicant is running (otherwise, need to include it into services section);
+5. DHCP: dhcpcd service is loaded.
+6. May need to set up static IP addressed and routes, but it's out of scope of this document.
+
+Usually, `lspci` provides the information about the detected devices, and
+`dmesg` prints the information whether the drivers for those devices
+are loaded or not (usually if nothing mentioned in the log, then no driver is loaded).
+
+### Wi-Fi Notes
```
rn: I would make this a heading one level up, ie `##` and have the wifi specific bits be under this heading

nuald:
Surely, I'll squash the commits next time. As for `wifi.yml`, we can still make it works if it's allowed to refer to 3rd-party images (I already have Docker images with fat kernel and full firmware database). If we include some console network manager (like `wicd`), we won't even need any custom `wpa_supplicant` configuration.

Or maybe we can meet somewhere in the middle - if you're willing to have one more LinuxKit kernel to support with additional device drivers. As for firmwares, we can just add a note that for testing purposes, users can use publicly available Docker image if they want (it's an automated build with the simple [Dockerfile](https://hub.docker.com/r/nuald/linux-firmware/~/dockerfile/), so they can build their own image if they have any concerns).

rn:
closing this issue due to inactivity and because it relies on #2541 which was also close due to inactivity

PR #2545, opened 2017-09-27
kubernetes: update to latest cri-containerd

ijc:
This new version vendors containerd v1.0.0-beta.1 which is our current base.

It also fixes the build for seccomp support so this PR enables that at build time.

Lastly this new version requires `/dev/disk/by-uuid` so it can report the UUID of the container storage filesystem back to kubernetes, implement support for that and ensure `/dev` is bound into the cri containerd.

Note that kube 1.7 does not currently use `/dev/disk/by-uuid` directly itself, but code in the master branch does, so this is future proof for that too (the kubelet container already binds `/dev`).

Also a bonus fix to the `ssh_to_kubelet` script which wasn't using a unique exec id.

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -22,7 +22,7 @@ onboot:
   - name: format
     image: linuxkit/format:158d992b7bf7ab984100c697d7e72161ea7d7382
   - name: mount
-    image: linuxkit/mount:4fe245efb01384e42622c36302e13e386bbaeb08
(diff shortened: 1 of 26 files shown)


ijc:
@rn I know what you mean. I figured that unless you had mounted the filesystem you would be unlikely to be interested in the uuid.

Perhaps `format` might have been a better home and thinking about it further even if it stayed in `mounts` maybe the links should have been made before doing the mount, since then you could potentially mount by UUID (but that likely involves more work than just moving the call)

Part of me just thinks we should switch from mdev to eudev (in Alpine) rather than slowly reimplementing that functionality it provides step by step.

PR #2554, opened 2017-10-02
Add 4.13.x and remove 4.12.x kernels

rn:
- 4.12 has been EOLed, remove it
- Add support for 4.13 kernels
  - Patches from 4.12 apply cleany
  - Kernel config derived from 4.12
- Enable `FORTIFY_SOURCE`, a new kernel config in 4.13 which enables bound checks on string/memory functions

I've tested the 4.13 kernel with LCOW and it seems to work but I will leave the 4.12 for LCOW for now as this is still the recommended kernel in the OpenGCS repro.

x86 kernels already build and pushed for x86, arm64 kernels are being built

take-a-leap


kernel/kernel_config-4.13.x-x86_64
```diff
@@ -3817,7 +3817,7 @@ CONFIG_SECURITY_PATH=y
 CONFIG_HAVE_HARDENED_USERCOPY_ALLOCATOR=y
 CONFIG_HARDENED_USERCOPY=y
 # CONFIG_HARDENED_USERCOPY_PAGESPAN is not set
-# CONFIG_FORTIFY_SOURCE is not set
+CONFIG_FORTIFY_SOURCE=y
```
justincormack: cool

PR #2556, opened 2017-10-02
tools/alpine: Explicitly add blkid

rn:
On x86_64, the blkid package gets installed as a dependency, but
not on arm64. Explicitly add it as the new format package depends
on it.

bull-dog


eyz:
Needed for #2532

PR #2558, opened 2017-10-03
Add support to make minimal raw bios images

deitch:

**- What I did**
Created a new `tools/` entry and therefore images:

* `tools/mkimage-raw-bios/`: create raw images that are bootable BIOS

The bios images are intended as a replacement for the `pkg/mkimage/`, but provides 2 advantages:

1. Much smaller builds (using `linuxkit.yml` gives ~64MB image vs 1.0GB)
2. Simpler builds: everything is done in a container, no need to spin up a VM in linuxkit to do filesystem manipulation

**- How I did it**
1. Created new directory
2. Created entrypoint that creates correct image given the `tar` input on stdin from moby

The intent, once merged, is to update moby build to remove the output format `raw` and replace it with `raw-bios`. We then can remove `pkg/mkimage` from the linuxkit repo entirely.

**Note:** There is some overlap between `raw-bios` and `iso-bios`. I would _like_ to extract common code, perhaps simplifying to just 1 docker image, perhaps merging common code with `raw-efi` but it was enough work to get this here and working fully. Let's merge it in first and get it working, then remove duplication.

**- How to verify it**
1. Create a `linuxkit.tar` as the usual input format for builders. This is *not* the output of `moby build -format tar ...`, since that puts everything in the tar as if on an iso. Instead, recreate the output of `tarToInitrd()` by either getting kernel+cmdline+initrd and tarring up, or doing `moby build -format raw ...` and taking the 3 files off the iso image and tarring them up.
2. Build the image using `make tag`
3. Create a raw image the way `moby build` would: `cat linuxkit.tar | docker run -i --rm <image_name> > linuxkit-bios.raw`
4. Run the image

To run the image, you can load it into qemu, vbox or a bare metal server.

For qemu, my command-lines (shamelessly hijacked and modified from `linuxkit run qemu ...`  are:

```
$ /usr/local/bin/qemu-system-x86_64 -device virtio-rng-pci -smp 1 -m 1024 -uuid 940c5cec-0640-47cf-aa48-feea2dd23037 -pidfile linuxkit.raw-state/qemu.pid -drive file=./linuxkit-bios.raw,index=0,media=disk  -device virtio-net-pci,netdev=t0,mac=12:7b:f1:13:29:2a -netdev user,id=t0 --nographic
```

Simplify and shrink raw BIOS images

Marmot

NOTE: This is a split from #2540. Some of the discussion with @justincormack will be replicated in a separate comment.

### DIFF
--- a/tools/mkimage-raw-bios/make-bios
+++ b/tools/mkimage-raw-bios/make-bios
@@ -1,0 +1,105 @@
+#!/bin/sh
+
+set -e
+
+# for debugging
+[ -n "$DEBUG" ] && set -x
+
+IMGFILE=$PWD/disk.img
+
+
+# we want everything except the final result to stderr
+( exec 1>&2;
+
+ESP_FILE=$PWD/boot.img
+
+
+mkdir -p /tmp/bios
+cd /tmp/bios
+
+# input is a tarball of filesystem on stdin with kernel in /boot
+# output is an iso on stdout
+
+# extract. BSD tar auto recognises compression, unlike GNU tar
+# only if stdin is a tty, if so need files volume mounted...
+[ -t 0 ] || bsdtar xzf -
+
+INITRD="$(find . -name '*.img')"
+KERNEL="$(find . -name kernel)"
+CMDLINE_FILE="$(find . -name cmdline)"
+CMDLINE="$(cat $CMDLINE_FILE )"
+
+
+cat >> syslinux.cfg <<EOF
+DEFAULT linux
+LABEL linux
+    KERNEL /kernel
+    INITRD /initrd.img
+    APPEND ${CMDLINE}
+EOF
+
+# now build an img file
+KERNEL_FILE_SIZE=$(stat -c %s "$KERNEL")
+INITRD_FILE_SIZE=$(stat -c %s "$INITRD")
+# minimum headroom needed in ESP, in bytes
+# 511KiB headroom seems to be enough
+ESP_HEADROOM=$(( 1024 * 1024 ))
+
+
+ESP_FILE_SIZE=$(( $KERNEL_FILE_SIZE + $INITRD_FILE_SIZE + $ESP_HEADROOM ))
+# (x+1024)/1024*1024 rounds up to multiple of 1024KB, or 2048 sectors
+# some firmwares get confused if the partitions are not aligned on 2048 blocks
+# we will round up to the nearest multiple of 2048 blocks
+# since each block is 512 bytes, we want the size to be a multiple of
+# 2048 blocks * 512 bytes = 1048576 bytes = 1024KB
+ESP_FILE_SIZE_KB=$(( ( ($ESP_FILE_SIZE+1024) / 1024 ) / 1024 * 1024 ))
+ESP_FILE_SIZE_MB=$(( $ESP_FILE_SIZE_KB / 1024 ))
+# and for sectors
+ESP_FILE_SIZE_SECTORS=$(( $ESP_FILE_SIZE_KB * 2 ))
+
+# create a raw disk with an EFI boot partition
+# Stuff it into a FAT filesystem, making it as small as possible.
+mkfs.vfat -v -C $ESP_FILE $(( $ESP_FILE_SIZE_KB )) > /dev/null
+echo "mtools_skip_check=1" >> /etc/mtools.conf && \
+mcopy -i $ESP_FILE syslinux.cfg ::/
+mcopy -i $ESP_FILE $KERNEL ::/kernel
+mcopy -i $ESP_FILE $INITRD ::/initrd.img
+
+# install syslinux
+syslinux --install $ESP_FILE
+
+# now make our actual filesystem image
+# how big an image do we want?
+# it should be the size of our ESP file+1MB for BIOS boot + 1MB for MBR + 1MB for backup
+ONEMB=$(( 1024 * 1024 ))
+SIZE_IN_BYTES=$(( $(stat -c %s "$ESP_FILE") + 4*$ONEMB ))
+
+# and make sure the ESP is bootable for BIOS mode
+# settings
+BLKSIZE=512
+MB_BLOCKS=$(( $SIZE_IN_BYTES / $ONEMB ))
+
+# make the image
+dd if=/dev/zero of=$IMGFILE bs=1M count=$MB_BLOCKS
+
+
+echo "w" | fdisk $IMGFILE || true
+# format one large partition of the apprropriate size
+echo ","$ESP_FILE_SIZE_SECTORS",;" | sfdisk $IMGFILE
+
+# of course, "one large partition" means minus the 2048 at the beginning
+ESP_SECTOR_START=2048
+
+# copy in our EFI System Partition image
+dd if=$ESP_FILE of=$IMGFILE bs=$BLKSIZE count=$ESP_FILE_SIZE_SECTORS conv=notrunc seek=$ESP_SECTOR_START
+
+# install mbr
+#dd if=/usr/share/syslinux/mbr.bin of="$IMGFILE" bs=440 count=1 conv=notrunc
+dd if=/usr/share/syslinux/altmbr.bin bs=439 count=1 conv=notrunc of=$IMGFILE
+printf '\1' | dd bs=1 count=1 seek=439 conv=notrunc of=$IMGFILE
+
+sfdisk -A "$IMGFILE" 1
+
+)
+
+cat $IMGFILE
--- a/tools/mkimage-raw-bios/Dockerfile
+++ b/tools/mkimage-raw-bios/Dockerfile
@@ -1,0 +1,14 @@
+FROM linuxkit/alpine:87a0cd10449d72f374f950004467737dbf440630
+
+RUN \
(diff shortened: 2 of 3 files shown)


deitch:
The discussion in #2540 centred around whether it is preferred to run this post-processing stage as a docker container, or as s linuxkit run (essentially full vm).

Notable comments below.

@justincormack wrote:

> On the 1GB size, that was intentional, as it is required for AWS and that the raw format is currently used for. It is customisable, and should really have an "as small as possible option", yes

> Spinning up Linuxkit has many advantages:
> 1. We need to get rid of the requirement to have Docker running to use the moby tool, as it makes building in a CI pipeline much easier, and we have had many requests for this. So the idea was to move away from using Docker to build images. LinuxKit can run so long as qemu is installed (or a native hypervisor, but qemu is fine generally for these conversions).
> 2. Most of the conversions (this one is an exception as it uses mtools) use libguestfs, which spins up a qemu VM to do the conversion anyway, so we are doing exactly the equivalent of running linuxkit anyway, just nested inside a container (often inside another VM if you are running on OSX or Windows). So directly using LinuxKit is less emulation for most cases.

Also from @justincormack as to why some people want docker dependency removed:

> The CI issue is that most people do not have Docker available for use inside their CI pipeline.

My concerns were:

1. Running lkt still needs to have the images available over network at run-time, whether you do `docker run` or `linuxkit run`, you need to get them.
2. I find the dependency to be smaller if all you need is docker vs linuxkit and qemu
3. It is faster to run a docker container than spin up a VM.

tools/mkimage-raw-bios/make-bios
```diff
@@ -0,0 +1,106 @@
...
+
+# input is a tarball of filesystem on stdin with kernel in /boot
+# output is an iso on stdout
+
+# extract. BSD tar auto recognises compression, unlike GNU tar
+# only if stdin is a tty, if so need files volume mounted...
+[ -t 0 ] || bsdtar xzf -
+
+INITRD="$(find . -name '*.img')"
+KERNEL="$(find . -name kernel -or -name bzImage)"
```
justincormack: we can drop the `bzImage` now - the rename was months ago.

justincormack:
We can't currently use this instead of `raw` as it is not going to work for AWS.

1. AWS requires a 1GB image, and you do not provide a way to override the size. The `moby` LinuxKit based version is maybe the only one to extend this functionality but the idea was to roll it out more generally. So you need to add a `size` option to the shell script that sets the total device size, or we add an `aws` option thats like this but pads to whole gigabytes. (there are other issues here; the LinuxKit version will generally manage to make a sparse file in this case, this could be done if you have an option to provide an output file instead of stdout).
2. Current Docker4AWS images assume you can extend the ext4 filesystem if a larger block device is used. As discussed this will need to be changed to create a new partition instead and format it.

So I think there is some work that needs to be done before we can replace the original `raw` with this.

justincormack:
Maybe we should have an `aws` target that calls this and then writes it into a multiple of 1GB filesystem image instead of adding complexity here?

tools/mkimage-raw-bios/make-bios
deitch: Good point. Cleaning up.

deitch:
Removed the `-or bzImage`

deitch:
> Maybe we should have an aws target that calls this and then writes it into a multiple of 1GB filesystem image instead of adding complexity here?

I like that. Create a new `raw-bios` option, then replace the current `raw` with `aws` that does:

1. Multiples of 1GB (if I understood you correctly, it must always round up to the nearest 1GB?)
2. Uses `ext4`

justincormack:
I need to check if it is minimum 1GB or multiples. But yes.

deitch:
Sounds like a plan. Once we get `raw-efi` and `raw-bios` in, and have `moby/tool` using them, I will put together a `raw-aws` variant.

deitch:
@justincormack I am going to open a new issue to serve as the focus for ami formatting.

deitch:
> So I think there is some work that needs to be done before we can replace the original raw with this

I was thinking about this as I opened #2562. I think we should do the following:

1. _add_ this as `raw-bios`
2. _rename_ the existing `raw` to `aws` (using the existing lkt-based post-processor)
3. Eventually replace the existing lkt-based post-processor with a variant of this one
4. Slowly move the codebase into go and a single binary or even into moby itself from these docker images

So we don't have 1 build (`raw` which is AWS), we have 3: `raw-efi`, `raw-bios`, `aws`

justincormack:
Sounds reasonable.

deitch:
> Sounds reasonable

Excellent. Next steps for this `raw-bios` and `raw-efi` PRs?

justincormack:
I think we should maybe rename `aws` first. Possibly could dig out the pre-LinuxKit implementation from the history and restore it? Athough the ordering is not that critical. I am just looking at sorting out a tar+initrd streaming output from `moby` but it needs a few changes...

deitch:
> Possibly could dig out the pre-LinuxKit implementation from the history and restore it?

I think I would leave it as is. We have the lkt-based one, it works for aws images. Let's leave it. I would sooner get the new `raw-efi` and `raw-bios` merged in, then we can PR `moby/tool` to:

* add `raw-efi`
* add `raw-bios`
* rename `raw` to `aws`

deitch:
This PR has been updated:

* use the new `linuxkit pkg ...` process
* docker registry repo created (thanks @riyazdf)
* image has been pushed out to registry

I think this answers everything you were asking, @justincormack

tools/mkimage-raw-bios/Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY: push forcepush tag forcetag show-tag
+
```
justincormack: needs removing now

tools/mkimage-raw-bios/make-bios
```diff
@@ -0,0 +1,106 @@
+#!/bin/sh
+
+set -e
+
+set -e
```
justincormack: `set -e` is duplicated

deitch: Fixed

tools/mkimage-raw-bios/Makefile
deitch: Why?

justincormack: I think all the Makefiles went in Ian's update?

deitch: There was a discussion about it. In the end, came down to philosophy. I like them as an "entrypoint" or "wrapper" that makes it obvious what to do and simplifies it. I still find it easier to do `make push` and `make tag` than `lkt pkg push .`

Actually, I took the exact contents of the `Makefile` from one of them in there.

Doesn't hurt, does it?

ijc: Either all packages should have it or none should and certainly we shouldn't be reproducing a dozen lines of boilerplate here and in your other PR, it should be a single line include.

deitch: > Either all packages should have it or none should

I like make and a global entrypoint. But I get that it is a personal preference.

> it should be a single line include

I don't understand? Single line include?

ijc: `include ../../../something.mk` which contains the lines you are currently duplicating all over the place.

deitch: oh right. Let me see if that works. You did put a `Makefile` in `tools/Makefile`, although that has logic to hit all directories below.

ijc: Although ased on discussions on the slack channel where the existing `pkg/Makefile` has already caused confusion about how to override options I don't think I'm in favour of propagating that to all the indivdual packages -- otherwise the docs have to be "`run make «foo», unless you want to pass another options, in which case run `linuxkit pkg foo --option`". I don't think splitting our instructions like that is wise, they should just talk about how to run `linuxkit`. If there are `Makefiles` then people will expect to use them even if they need something other than the default.

So having thought about it I'm pretty against having needless Makefiles.

deitch: Well, including `../Makefile` sort of works. Sort of.

Let's not get into the deeper philosophical discussion here. I prefer `Makefile` even (or especially) when they wrap saner tools (like `lkt pkg ...`).  I have them for every one of my `go` and `node` and whatever projects, even if `go build .` or `npm start` is painfully obvious. But I don't care enough to hold these PRs up.

@ijc and @justincormack do we want to remove the `Makefile` here and the other PRs and move ahead?

ijc: > You did put a Makefile in tools/Makefile, although that has logic to hit all directories below.

Right that is just for the iteration and as I said I'm now realising it wasn't a very good idea because it has confused at least one person so far.

I'm thinking about other more flexible ways that could be achieved. I don't think I want `linuxkit bulk-pkg build «dir1» «dir2» .. «dirN»`, but perhaps I'll think of something better.

> do we want to remove the Makefile here and the other PRs and move ahead?

Yes, please.

deitch: Done in all 3 PRs.

PR #2559, opened 2017-10-03
kubernetes: bump to v1.8.0

ijc:
Update to Kubernetes v1.8.0.

This requires updating some integration scripting to deal with slightly modified ways of doing things and bumping the bits of the image cache.


ijc:
yml updates will be along shortly, still building those.

PR #2560, opened 2017-10-03
Add squashfs support to the kernel config

justincormack:
This is a useful read only filesystem for images that is efficient and
small as it supports compression.

For many use cases, when you are writing to media, it makes more sense than
using an initramfs as it does not require RAM, and it is more suitable for
disk media than ISO.

puppies-495x294


justincormack:
I am not in a hurry, so was going to wait...

rn:
I merge now and pull it in with the next kernel update

PR #2565, opened 2017-10-04
kubernetes: add optional support for a kubeadm.yaml

djs55:
**- What I did**

Allow more `kubeadm` options to be selected by supplying a `yaml` file. Unfortunately not all options are available via the CLI interface.

**- How I did it**

In the `kubeadm-init.sh` if the `/etc/kubeadm/kubeadm.yaml` file exists, then it uses it. Otherwise it falls back to the original behaviour.

**- How to verify it**

Supply an `/etc/kubeadm/kubeadm.yaml` with some new options, and observe they take effect on a new cluster.

Allow custom `kubeadm` configuration files to be provided via `/etc/kubeadm/kubeadm.yaml`

### DIFF
--- a/projects/kubernetes/kubernetes/kubeadm-init.sh
+++ b/projects/kubernetes/kubernetes/kubeadm-init.sh
@@ -1,7 +1,15 @@
 #!/bin/sh
 set -e
 touch /var/lib/kubeadm/.kubeadm-init.sh-started
-kubeadm init --skip-preflight-checks --kubernetes-version @KUBERNETES_VERSION@ $@
+if [ -f /etc/kubeadm/kubeadm.yaml ]; then
+    echo Using the configuration from /etc/kubeadm/kubeadm.yaml
+    if [ $# -ne 0 ] ; then
+        echo WARNING: Ignoring command line options: $@
+    fi
+    kubeadm init --skip-preflight-checks --config /etc/kubeadm/kubeadm.yaml
+else
+    kubeadm init --skip-preflight-checks --kubernetes-version @KUBERNETES_VERSION@ $@
+fi
 for i in /etc/kubeadm/kube-system.init/*.yaml ; do
     if [ -e "$i" ] ; then
 	echo "Applying "$(basename "$i")


projects/kubernetes/kubernetes/kubeadm-init.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/sh
 set -e
 touch /var/lib/kubeadm/.kubeadm-init.sh-started
-kubeadm init --skip-preflight-checks --kubernetes-version @KUBERNETES_VERSION@ $@
+if [ -f /etc/kubeadm/kubeadm.yaml ]; then
+  echo Using the configuration from /etc/kubeadm/kubeadm.yaml
```
ijc: I wonder if at this point something like this might be appreciated by someone in the future:
```
if [ $# -ne 0 ] ; then
    echo WARNING: Ignoring command line options: $@
fi
```

djs55: good idea. I'll add it.

ijc: :+1: once you've pushed that give me a prod and I'll build the package so you can update `kube.yml`

ijc:
@djs55 I've pushed out `linuxkitprojects/kubernetes:eeb6227af856f0cfab8a5af1e5b97ee39f9ccd39`.

ijc:
```
- CONFIG_IOSCHED_CFQ: enabled
- CONFIG_CFQ_GROUP_IOSCHED: enabled
- CONFIG_CGROUP_PINFO:root:Error getting serial output (backendError): TIMEOUT
INFO:root:Error getting serial output (resourceNotReady): The resource 'projects/moby-datakit-ci/zones/europe-west1-b/instances/linuxkit-ci-5' is not ready
[2017-10-04 16:38.42] Removing test VM...
[2017-10-04 16:38.42] Running "/gcloud/destroy.py" "linuxkit-ci-5"
INFO:root:Destroying VM
```
WIll restart

PR #2566, opened 2017-10-05
Update kernels to 4.13.5/4.9.53/4.4.90 and also wireguard tools

rn:
- Update kernels as usual
- The wireguard version bump needed an synchronised update of kernel and userspace.
  - The userspace tools were already updated in the alpine base image
  - Update the getty/ip/sshd packages to the new userspace

crane


rn:
This failed on the wireguard test.

Weirdly the wireguard tests was on kernel tag `4.9.43` so the test failed with the update. That package then was not update for quite some time...

Force pushed an update

PR #2567, opened 2017-10-05
tools/alpine: Add nfs-utils

errordeveloper:
Needed for Kubernetes to pass more e2e tests.

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -69,6 +69,7 @@ mtools
 multipath-tools
 musl-dev
 ncurses-dev
+nfs-utils
 open-iscsi
 openntpd
 openrc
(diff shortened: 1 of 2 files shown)


justincormack:
If someone has a chance to rebuild and push the package that would be great (tied up right now)

justincormack:
Its probably ok to open a PR with just the `packages` file changes but not the versions.

rn:
that works too, but given that whoever then pushes has to do another PR, it's probably easier to just do everything in one commit

errordeveloper:
> that works too, but given that whoever then pushes has to do another PR, it's probably easier to just do everything in one commit

@rn makes sense, thanks!

rn:
see #2573

Issue #2564, opened 2017-10-04
Introduce "linuxkit pkg" sub-command for build and push


ijc:
Currently the LinuxKit package build/tag/push machinery is encapsulated in a `make` snippet (`pkg/package.mk`) which is included by each package's `Makefile` after overriding various configuration options.

There is also a helper script for signing/pushing in `scripts/push-manifest.sh` which is referenced by the packaging snippet.

This arrangement is one of the bigger blockers to having LinuxKit packages hosted in repositories other than this one. It's also a bit fragile.

I propose to replace this functionality with a new subcommand of the `linuxkit` binary. The existing per-package `Makefile`s would be replaced by a `linuxkit-build.yml` file encapsulating the various options/overrides available to the packages (as `Makefile` variables) today.

I propose to initially do a fairly straight conversion from the Makefile to go, e.g. using `os.exec` to call out to `docker build`, `git` etc as the `Makefile` does today, rather than attempt any sort of closer integration with the underlying tools. Initially I may do the same for the `push-manifest.sh` helper but I'd like to move its functionality into the tool sooner rather than later.

Command mapping:
* `make -C pkg/foo tag` » `linuxkit pkg build pkg/foo` ( or `cd pkg/foo ; make tag` » `linuxkit pkg build .`)
* `make -C pkg/foo forcetag` » `linuxkit pkg build --force pkg/foo`
* `make -C pkg/foo push` » `linuxkit pkg push pkg/foo`
* `make -C pkg/foo forcepush` » `linuxkit pkg push --force pkg/foo`
* `make --no-print-directory -C pkg/foo show-tag` » `linuxkit pkg show-tag pkg/foo`
* `make --no-print-directory -C pkg/foo show-tag HASH_COMMIT=DEADBEEF` » `linuxkit pkg show-tag --commit DEADBEEF pkg/foo`

Makefile variables to be supported by `linuxkit-build.yml` (with more yamly names):
* `ORG` (hub org to push to, default is linuxkit)
* `HASH` (hash to tag with, default is tree-sh)
* `ARCHES` (valid arches, default x86_64 + arm64)
* `REPO` (git repo this is in, default https://github.com/linuxkit/linuxkit)
* `NETWORK` (availability of network to build, default n)
* `TRUST` (enable content trust, default y)
* `CACHE` (some builds use `--no-cache` directly, need to decide if we should expose this)
* `DEPS` and `BASE_DEPS` will not be supported.

Future enhancements could include:
[...]

PR #2571, opened 2017-10-06
Implement package build, push and show-tag in linuxkit tool

ijc:
**- What I did**

This implements the proposal in #2564 and converts a handful of representative or especially interesting (from a build PoV) packages to use it.
    
For now those pkg/* affected get a stub-`Makefile`, once all packages are converted then `pkg/Makefile` can be adjusted and those stubs can be removed.
    
For now only `pkg/package.mk`'s functionality is implemented. In particular:
    
- `push-manifest.sh` remains a separate script, to enable calling it on systems
      with just the LinuxKit tools installed arrange to install it under a less
      generic name.
- `kernel` and `tools/alpine` do not use `pkg/package.mk` and those cases are
      not yet fully considered/covered.
    
I have updated the documentation assuming that the existing uses of `pkg/package.mk` will be removed quite soon in a follow up PR rather than trying to document the situation which results after just this commit.
    
Due to cmd/linuxkit now gaining a library the build needs adjusting slightly to allow both `make bin/linuxkit` and `go build` to work.
    
`go vet` has forced me to write some rather asinine comments for things that are rather obvious from the name.
    
**- How I did it**

**- How to verify it**

```
$ make bin/linuxkit
$ linuxkit pkg build pkg/init
_[etc as described in #2564]_
```

Package build is now integrated into the `linuxkit` tool.


ijc:
Once we are agreed on the shape of this I think we should merge it and I will quickly followup with a new PR which does a bulk update of all the packages. @rn has suggested bumping everything to the latest alpine mirror base at the same time which I am happy to do.

docs/packages.md
```diff
@@ -11,6 +11,22 @@ All LinuxKit packages are:
...
+- `Dockerfile`: contains the steps to build the package.
+
+`build.yml` contains the following fields:
+
+- `image` _(string)_: *(mandatory)* The name of the image to build
+- `org` _(string)_: The hub/registry organisation to which this package belongs
+- `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
+- `gitrepo` _(string)_: The git repository where the package source is kept.
+- `network` _(bool)_: Allow network access during the package build (default: no)
+- `trust` _(bool)_: Enable Docker content trust for this package (default: yes)
```
justincormack: its weird having booleans that default to `true` - should this be `notrust` and `nocache`?

justincormack: also means cli flags can just be `--notrust` not `-trust=0`

justincormack: or `disable-` rather than `no`

ijc: I don't think it's at all weird, IMHO inverting the sense of the name is odd and leads to double negatives in the code `if !foo.notrust` which just make it harder to follow.

ijc: `disable` has similar problems to `no`. The `--notrust` things is IMHO just a short-coming of the golang `flag` package. I've been considering looking for a better one since it also doesn't let you know if something was set on the command line or not leading to hoop jumping.

src/cmd/linuxkit/pkg_build.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/pkgsrc"
```
justincormack: `pkgsrc` is confusingly overloaded term to some of us...

ijc: I struggled with the name since `pkg` also means something in Go directory structures and `package` is a keyword. I'm open to suggestions.

src/cmd/linuxkit/pkgsrc/docker.go
```diff
@@ -0,0 +1,90 @@
...
+	dct   bool
+	cache bool
+}
+
+func newDockerRunner(dct, cache bool) dockerRunner {
+	return dockerRunner{dct: dct, cache: cache}
+}
+
+func (dr dockerRunner) command(args ...string) error {
+	cmd := exec.Command("docker", args...)
```
justincormack: I would find `docker` with lookpath and error up front of the docker binary is not found in a more helpful way.

src/cmd/linuxkit/pkgsrc/docker.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := dr.push(img + suffix); err != nil {
+		return err
+	}
+
+	dctArg := "0"
+	if dr.dct {
+		dctArg = "1"
+	}
+
+	cmd := exec.Command("linuxkit-push-manifest", img, dctArg)
```
justincormack: and here, as someone could easily not have this installed

docs/packages.md
ijc: The CLI needs to be able to overload the yml, which could have an explicit `true`, so it's not as simple as inverting the option only. I could add both `--trust` and `--notrust` and handle reconciling when both are used (since `flagset.Visit` goes in cli order it should be reasonably easy to implement last-one-wins semantics).

src/cmd/linuxkit/pkgsrc/docker.go
ijc: I broke the name and the result is:
```
$ linuxkit pkg build pkg/init
exec: "dockerx": executable file not found in $PATH
```

justincormack: yes but it doesnt give you a helpful message like "linuxkit pkg requires docker to be installed"

ijc: Is that not readily apparent from "docker not found in $PATH" ?

docs/packages.md
justincormack: I still prefer boolean defaults to always be false. Having both cli flags is ok.

justincormack: I dont think the code argument is any good, you need both cases anyway so dont invert the sense.

justincormack: If defaults randomly vary between true and false I have to either be really verbose and write everything out or RTFM every time I make a config file. If they are always false its obvious.

src/cmd/linuxkit/pkgsrc/docker.go
ijc: BTW I debated doing something really gross here to save installing a new tool, until the script could be reimplemented in go. Namely:

```
pushManifest := `
«108 lines of scripts/push-manifest.sh»
`
....
cmd := exec.Command("/bin/sh", "-c", pushManifest, «other args»)
```

src/cmd/linuxkit/pkg_build.go
justincormack: maybe `pkglib`?

src/cmd/linuxkit/pkgsrc/docker.go
justincormack: No. Because I didnt try to run docker and I dont know the implementation currently shells out, this is unexpected.

src/cmd/linuxkit/pkg_build.go
ijc: Sure.

src/cmd/linuxkit/pkgsrc/docker.go
justincormack: yeah its not too terrible. You could use `go generate` to keep the original in a seperate file but it hardly seems wirth it

src/cmd/linuxkit/pkg_build.go
ijc: I think that goes best with `Pkg` rather than `PkgLib` for the type so I shall do so.

src/cmd/linuxkit/pkgsrc/docker.go
ijc: `go generate` would be nice because then syntax highlighting would work for anyone editing it. I think I will go this route then.

src/cmd/linuxkit/pkgsrc/docker.go
rn: I don't understand the argument. To me `executable file not found in $PATH` is sufficient

ijc: I mostly agree with @rn but I think it would be sufficient to check for `exec.ErrNotFound` rather than doing a manual `LookPath` so the impact might not be too bad.

docs/packages.md
ijc: I'll try it and see how it looks

rn: I think `-disable-content-trust` for the CLI and `disable-content-trust` for the YAML option is better. It's what `docker` and `moby` is using and it's sufficiently lengthy and awkward to type for someone to really want to disable it. For this option it doesn't seem to make sense, where someone would disable content trust in the YAML and the you want to override it in the CLI to enable it. That seems like an odd use case.

docs/packages.md
```diff
@@ -43,9 +61,13 @@ they can't be build in parallel.
 To build a package on an architecture:
 
 ```
-DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="<passphrase>" make
+DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="<passphrase>" linuxkit pkg push «path-to-package»
 ```
 
+`«path-to-package»` is the path to the package's source directory
+(containing `build.yml` and `Dockerfile`). It can be `.` if the
```
rn: containing at least `build.yml`...

projects/kubernetes/image-cache/Makefile.pkg
```diff
@@ -1,11 +1,22 @@
-ORG?=linuxkitprojects
-IMAGE=kubernetes-image-cache-$(CACHE)
-NOTRUST=1
-SOURCE=$(BUILDDIR)
-DEPS=$(BUILDDIR)/Dockerfile
-ARCHES=x86_64
+$(BUILDDIR)/build.yml: build.yml.in
+	@sed -e 's/@@CACHE@@/$(CACHE)/g' < $< > $@
```
rn: Instead of `sed` would it make sense to provide variables to `linuxkit pkg` via the command line? Could be a improvement done later, though

src/cmd/linuxkit/pkgsrc/build.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if _, ok := os.LookupEnv("DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE"); !ok && ps.trust && bo.push {
+		return fmt.Errorf("Pushing with trust enabled requires $DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE to be set")
+	}
+
+	arch := runtime.GOARCH
+
+	if !ps.archSupported(arch) {
+		return fmt.Errorf("Arch %s not supported by this package", arch)
```
rn: erroring here is a a different to the current behaviour where we just print out a message but return with exit code 0. This will return (eventually) exit code 1 (if i read the code correctly).

I think printing a warning/info and exit with 0 is better in this case because we want to  just run `linuxkit pkg build` across all `./pkg` on all arches. If one of them errors because the package is not supported for the current arch then the rest of the build would fail.

rn:
go syntax is certainly more readable than Makefile syntax :)

docs/packages.md
```diff
@@ -11,6 +11,22 @@ All LinuxKit packages are:
 - Derived from well-known (and signed) sources for repeatable builds.
 - Build with multi-stage builds to minimise their size.
 
+## Package source
+
+A package source consists of a directory containing two files:
```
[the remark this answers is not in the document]
ijc: It doesn't say "only", but OK.

docs/packages.md
```diff
@@ -11,6 +11,22 @@ All LinuxKit packages are:
...
+
+- `build.yml`: contains metadata associated with the package
+- `Dockerfile`: contains the steps to build the package.
+
+`build.yml` contains the following fields:
+
+- `image` _(string)_: *(mandatory)* The name of the image to build
+- `org` _(string)_: The hub/registry organisation to which this package belongs
+- `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
+- `gitrepo` _(string)_: The git repository where the package source is kept.
```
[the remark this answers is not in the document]
ijc: This is hardcoded to `https://github.com/linuxkit/linuxkit` in `pkg/package.mk` today because you can't find it by shelling out to git, even if you assume the relevant remote is named `origin` this may point to some private name (e.g. `git@github.com:linuxkit/linuxkit`) rather than the desired name (https://github.com/linuxkit/linuxkit).

I eventually intend to deal with this by being able find a `.build-defaults.yml` in the root of the git tree which can override this and other things, such as `org` with local defaults.

docs/packages.md
ijc: Like that name, I think I might as well support the odd override since it's not impossible, just unusual and I'll have the pattern from the other bools anyway. I could imagine someone not wanting to routinely sign stuff but only specific builds which they decide are good by some metric.

docs/packages.md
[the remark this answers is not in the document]
rn: Ah, good point

projects/kubernetes/image-cache/Makefile.pkg
ijc: That would involve interpolating all the values and figuring out the semantics of how each one should be dealt with, and handling quoting/escaping. Might be nice but I'm not inclined to do that.

docs/packages.md
riyazdf: agree with `disable-content-trust` consistency, and it makes the default boolean case much cleaner

docs/packages.md
```diff
@@ -11,6 +11,22 @@ All LinuxKit packages are:
...
+- `network` _(bool)_: Allow network access during the package build (default: no)
+- `trust` _(bool)_: Enable Docker content trust for this package (default: yes)
+- `cache` _(bool)_: Enable build cache for this package (default: yes)
```
riyazdf: relatedly: should this be `-disable-cache`?

docs/packages.md
```diff
@@ -24,6 +40,8 @@ Before you can build packages you need:
   bin/manifest-tool`, or `go get github.com:estesp/manifest-tool`, or
   via the LinuxKit homebrew tap with `brew install --HEAD
   manifest-tool`. `manifest-tool` must be in your path.
+- The LinuxKit tools `linuxkit` and `linuxkit-push-manifest` which
```
rn: I think you now embed the script so it does not need to be in the path anymore.

ijc:
I think I've covered all the comments. Shall we press ahead?

src/cmd/linuxkit/pkglib/manifest_push_script.go
```diff
@@ -0,0 +1,112 @@
+package pkglib
```
rn: This file is auto-generated using, right? I have never used the gogen stuff, but wouldn't expect that this file needs to be checked in...

justincormack: Yeah, better to check in, Go expects that. Otherwise you need to add a Makefile to generate first.

ijc: This is just short term. Once all other uses of `packaging.mk` are removed I can move `scripts/push-manifest.sh` into this source dir, which will result in it being in the docker build context used for the building the binary and I can switch to using `go gen` at build time.

PR #2578, opened 2017-10-09
Bulk conversion from `pkg/package.mk` to `linuxkit pkg`

ijc:
Please review but do not merge, there will be a set of bulk updates added once enough LGTMs are applied.

**- What I did**

Updated everything which used `pkg/package.mk` to `linuxkit pkg` (introduced in #2571, relates to #2564). This was mostly a mechanical translation of the `Makefile` into a `build.yml`.

**- What I plan to do after review but before merge**

- [x] Rebuild `tools/alpine`.
- [x] `./scripts/update-component-sha.sh --image linuxkit/alpine «new hash»`.
- [x] Rebuild the world.
- [x] Update all the assemblies to use the resulting images.

**- How I did it**

Mostly by hand, although I did script the pkg case a bit (see the commit message).

**- How to verify it**

CI?

#2571 already has:
> Package build is now integrated into the linuxkit tool.
Which is sufficient.

https://upload.wikimedia.org/wikipedia/en/1/16/Deftones_-_White_Pony-greycoverart.jpg


ijc:
I forgot to say: Contrary to https://github.com/linuxkit/linuxkit/pull/2571#discussion_r143431225 `scripts/push-manifest.sh` remains where it is since it is still used by `kernel/Makefile`.

I plan to look at porting the kernel build over shortly, but only in a followup PR.

deitch:
One question. Do we not want a basic `Makefile` in the `pkg/` and `tools/` dirs? We obviously don't really need them with `linuxkit pkg build` and `linuxkit pkg push`, but a `Makefile` usually is a good clean entrypoint.

ijc:
@deitch there is (still) a `Makefile` in both `pkg` and `tools` after this PR.

deitch:
@ijc I meant in each subdir, i.e. in each "tool" or "pkg".

deitch:
Yeah, a bit of a pain, but each `pkg/<subdir>/` or `tools/<subdir>/` is its own mini-project. `Makefile` does give a universal set of entry points.

ijc:
I see. I guess I had been considering this to be a packaging format or sorts, none of deb, rpm or abuild etc include a Makefile in their source layout to hook to the build tool so why would LinuxKit.

Ultimately I suppose I don't really mind one way or another, the Makefile would be a oneliner (to include some common set of targets).

deitch:
> Ultimately I suppose I don't really mind one way or another, the Makefile would be a oneliner (to include some common set of targets).

Which ends up recreating some of what we already have. Yeah, it would be basically 2 targets: build and push. With no vars, because that is all in `build.yml`.

More of a philosophy question. Personally, I like having that universal entrypoint. Whether I come to a "build this with C" or "run it with node" or "go compile" or "moby build", a `Makefile` always gives me "these are the basic things you can/should do." 

But, again, philosophy question.

rn:
LGTM, thanks for tackling this

ijc:
CI says:
```
[+ 74m 08s] ERROR: Failed to create usr/lib/go/pkg/linux_amd64/strconv.a: No space left on device
[+ 74m 08s] ERROR: go-1.8.3-r0: No space left on device
[+ 74m 08s] Executing busybox-1.26.2-r5.trigger
[+ 74m 08s] Executing ca-certificates-20161130-r2.trigger
[+ 74m 08s] 1 errors; 180 MiB in 35 packages
[+ 74m 08s] The command '/bin/sh -c apk add --no-cache go musl-dev git build-base' returned a non-zero code: 1
[+ 74m 08s] exit status 1
```
I'll hit restart. /cc @talex5 in case there is more needed.

talex5:
Looks like that error is coming from the build VM. Since we create a fresh one for each build, I'm guessing that rebuilding won't help (unless something earlier in the build uses up a random amount of disk space). Might be worth adding a `df -h` in the Makefile near that point.

ijc:
Thanks. I've just pushed a change with makes everything use the same alpine hash, which should reduce the number of copies of that which are pulled to 1. Hopefully that will be sufficient to make progress, if not I'll start debugging as you suggest.

ijc:
I think the space issue is because this wants to rebuild the world (because I changed everything). Normally it would pull and only build if that fails. Since the pull just pulls in the final (usually `FROM scatch`) images it doesn't need so much room compared with the build environment for everything.

So this should resolve itself once I've built and pushed everything (fingers crossed).

We might like to bump the build VMs to have enough space for a complete build, but not urgently for this PR.

ijc:
Finally got everything built and yml's updated. I will merge once CI goes green.

eyz:
I'm getting the `pkg build` help for `pkg push` -
```
$ linuxkit pkg push --help
USAGE: linuxkit pkg build [options] path

'path' specifies the path to the package source directory.

  -build-yml string
    	Override the name of the yml file (default "build.yml")
  -disable-cache
    	Disable build cache
  -disable-content-trust
    	Enable content trust
  -enable-cache
    	Enable build cache (default true)
  -enable-content-trust
    	Enable content trust (default true)
  -force
    	Force rebuild
  -force-dirty
    	Force the pkg to be considered dirty
  -hash string
    	Override the image hash (default is to query git for the package's tree-sh)
  -hash-commit string
    	Override the git commit to use for the hash (default "HEAD")
  -hash-path string
    	Override the directory to use for the image hash, must be a parent of the package dir (default is to use the package dir)
  -network
    	Allow network use during build
  -nonetwork
    	Disallow network use during build (default true)
  -org string
    	Override the hub org (default "linuxkit")
```

ijc:
@eyz thanks, it's just the word build which is wrong, the options themselves are correct (most are common to build and push). I've got a followup branching coming, will include this.

PR #2588, opened 2017-10-10
linuxkit pkg: Log before building or pushing.

ijc:
Previously there would be a make "entering directory" hint in the logs, but
with the switch to `linuxkit pkg` that no longer occurs.


ijc:
+ bonus typo fix. Thanks @eyz for spotting.

PR #2589, opened 2017-10-10
linuxkit pkg: Support --release option to pkg push.

ijc:
This overrides the default (which is to look for an exact git tag) and releases
just that.

/cc @eyz I can't easily test this without releasing something, are you able to test?


eyz:
LGTM

`linuxkit pkg push --force -disable-content-trust -org MYREGISTRY:5000 -release "${RELEASE_TAG}" my-swap` is building, tagging, and pushing as expected

PR #2591, opened 2017-10-11
fix dns not working in the docker blueprint

dprotaso:
Signed-off-by: dave <dprotaso@gmail.com>

**- What I did**

`docker pull` was failing due to dns issues. We now bind the host's /etc/resolv.conf

**- How to verify it**
From the machine running the VM (in my case a Mac)
```
docker -H unix://docker-for-mac-state/guest.00000948 pull hello-world
```

### DIFF
--- a/blueprints/docker-for-mac/docker-ce.yml
+++ b/blueprints/docker-for-mac/docker-ce.yml
@@ -11,6 +11,7 @@ services:
      - type: cgroup
        options: ["rw","nosuid","noexec","nodev","relatime"]
     binds:
+     - /etc/resolv.conf:/etc/resolv.conf
      - /var/lib/docker:/var/lib/docker
      - /lib/modules:/lib/modules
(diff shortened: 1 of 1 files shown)


justincormack:
Ah sorry I thought I added these everywhere...

@djs55 this is not in the blueprint, should it be there too?

djs55:
In the d4m blueprint I've currently got
```
            "/run/resolvconf/resolv.conf:/etc/resolv.conf", \
```
-- should that be `/etc/resolv.conf:/etc/resolv.conf` instead?

justincormack:
Oh sorry I missed that, yes that is fine. It doesn't matter much as `runc` resolves symlinks before doing mounts.

justincormack:
@djs55 it is not in https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac/docker-ce.yml though

PR #2598, opened 2017-10-12
Add fmt/lint/etc and test to local target

ijc:


ijc:
I'm not sure if the new `Run linuxkit tests in ci-*` commit will work in CI (since I'm not sure if go is installed in the host image). If not we can drop it for now.

ijc:
```
[+  9m 30s] make local-test
[+  9m 30s] make[1]: Entering directory '/tmp/build'
[+  9m 30s] /bin/sh: 1: go: not found
[+  9m 30s] go test 
[+  9m 30s] make[1]: go: Command not found
[+  9m 30s] Makefile:86: recipe for target 'local-test' failed
```
Backing out the final commit now..

PR #2608, opened 2017-10-13
linuxkit: allow user to configure a content trust passphrase command

ijc:
Could be e.g.:
```
pkg:
  content-trust-passphrase-command: "lpass show <key> --password"
```
or
```
pkg:
  content-trust-passphrase-command: "gpg -d ~/.notary/passphrase.gpg"
```


ijc:
I should update `docs/package.md`, one sec.

ijc:
Hrm, this is actually a bit irritating since it runs it (which prompts me) even when I run `linuxkit pkg build` (which doesn't need the key). It caches the answer for a while so not too bad, but I think I need to move the handling around a bit.

PR #2609, opened 2017-10-13
[WIP] experimenting with Travis

ijc:
Experimenting with Travis to build `bin/linuxkit` in a bunch of different configurations (`{linux,darwin,windows} · {go1.9,go master}`).

Compared with the usual build this also runs `go test` on the tool.


ijc:
Well, it works.

justincormack:
@ijc this seems to be working but is it still WIP?

ijc:
Wasn't quite sure if we really wanted to go this route, seems like it would be worth it to me though. I removed the tag.

PR #2610, opened 2017-10-13
tools: Add dtc to alpine base

rn:
The device tree compiler is needed for some of the ongoing
arm64 work

hippo

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -27,6 +27,7 @@ curl
 dhcpcd
 diffutils
 dosfstools
+dtc
 e2fsprogs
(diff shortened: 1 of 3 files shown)


ijc:
I enabled Travis to have a play in https://github.com/linuxkit/linuxkit/pull/2609 but didn't expect it to try and test branches with no `.travis.yml` in it by default &mdash; but it did. I've toggled the setting now so it shouldn't happen again but can't find a way to remove the failure above -- suggest ignoring and merging when there are no other failures. Sorry about that!

rn:
no worries. Tests should pass anyway as nothing changed...

PR #2611, opened 2017-10-13
linuxkit pkg: defer content trust passphrase setup until we know it i…

ijc:
…s needed

Otherwise "linuxkit pkg build" etc will needlessly run the command (which might
prompt the user).

Followup to https://github.com/linuxkit/linuxkit/pull/2608#issuecomment-336474518.


rn:
CI failure seems unrelated. Hit rebuild

PR #2620, opened 2017-10-18
add support for OS_CACERT, OS_INSECURE

mmariani:
Signed-off-by: Marco Mariani <marco.mariani@alterway.fr>

**- What I did**

I've added basic support for self-signed TLS authentication in the OpenStack commands.

**- How I did it**

Since the gophercloud package does not support passing a certificate, I override the HTTPClient instance of the ProviderClient.

**- How to verify it**

Set OS_CACERT to a self-signed certificate, or pass it with --cacert.
To verify --insecure, set OS_AUTH_URL to the ip address instead of the hostname. It should fail unless the --insecure flag is set.

Added OS_CACERT and OS_INSECURE to "run openstack" and "push openstack".

### DIFF
--- a/src/cmd/linuxkit/push_openstack.go
+++ b/src/cmd/linuxkit/push_openstack.go
@@ -31,6 +31,8 @@ func pushOpenstack(args []string) {
 	projectNameFlag := flags.String("project", "", "Name of the Project (aka Tenant) to be used")
 	userDomainFlag := flags.String("domain", "Default", "Domain name")
 	usernameFlag := flags.String("username", "", "Username with permissions to upload image")
+	cacertFlag := flags.String("cacert", "", "CA certificate bundle file")
+	insecureFlag := flags.Bool("insecure", false, "Disable server certificate verification")
 
 	if err := flags.Parse(args); err != nil {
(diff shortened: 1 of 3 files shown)


justincormack:
Can you just do a PR with the one change in? You have a merged branch with the other PRs you did in...

mmariani:
Of course, I pushed the wrong one, sorry.

src/cmd/linuxkit/push_openstack.go
```diff
@@ -59,9 +63,20 @@ func pushOpenstack(args []string) {
...
-	provider, err := openstack.AuthenticatedClient(authOpts)
+
+	provider, err := openstack.NewClient(authOpts.IdentityEndpoint)
+	if err != nil {
+		log.Fatalf("Failed to connect to OpenStack")
+	}
+
+	provider.HTTPClient, err = openstackHTTPClient(cacert, insecure)
+	if err != nil {
+		panic(err)
```
justincormack: I think using `log.Fatalf` as elsewhere is more consistent?

mmariani:
i've replaced panic() by log.Fatalf()

PR #2621, opened 2017-10-18
Fix GCP error printing

dnephin:
Before:

```
$ ./bin/linuxkit push gcp -project ... -bucket ... ./broken.img.tar.gz
...
Creating image: broken
FATA[0013] Error creating Google Compute Image: error running operation: &{[0xc4203a71a0] [] []}
```

After:
```
FATA[0006] Error creating Google Compute Image: error running operation: &{[%!s(*compute.OperationErrorErrors=&{INVALID_IMAGE_TAR  The tar archive is not a valid image. [] []})] [] []}
```
Still not the best error, but at least it includes the relevant details now.

### DIFF
--- a/src/cmd/linuxkit/gcp.go
+++ b/src/cmd/linuxkit/gcp.go
@@ -461,7 +461,7 @@ func (g *GCPClient) pollOperationStatus(operationName string) error {
 			return fmt.Errorf("error fetching operation status: %v", err)
 		}
 		if operation.Error != nil {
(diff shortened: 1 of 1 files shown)


justincormack:
@dnephin we are just looking into the tool changes that broke this (and also speaking to Google Cloud to find out how we can debug).

dnephin:
Do you mean the underlying error? I managed to get the example working, after running `moby build` a second time. The first time it built an invalid tarball, not sure why.

This PR is just to fix the error message.

PR #2622, opened 2017-10-18
add iscsi pkg

garthy:
Signed-off-by: Garth Bushell <garth@garthy.com>

Add iscsi pkg

### DIFF
--- a/pkg/iscsid/Dockerfile
+++ b/pkg/iscsid/Dockerfile
@@ -1,0 +1,17 @@
+FROM linuxkit/alpine:ad35b6ddbc70faa07e59a9d7dee7707c08122e8d AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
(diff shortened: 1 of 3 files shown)


eyz:
@garthy I also have an iscsi LinuxKit pkg, but I haven't submitted it as a PR. Do you have any need to also run `multipathd`, and do you have any config / metadata integration planned for this package?

garthy:
So this is a work in progress. Our cloud boots of iSCSI with handover from IPXE/EFI. so this is needed at boot time. 

* For config the target name/IP/port is static so I can build that in to the image definition
* I'd missed the /dev and net namespace as what I know about linuxkit was from DockerCon today. I'll fix that up

Unfortunately I'm actually blocked from testing this by two other issues (Although maybe me not understandng stuff)
* ixegb[vf] drivers not build into kernel
* EFI QCOW2 image required

justincormack:
@garthy do you want to open an issue for the things you need, and we can discuss in one place?

rn:
Closing due to inactivity. As @justincormack suggested it might be best to first discuss iSCSI boot/root on an issue and then move on from there.

PR #2623, opened 2017-10-18
make Intel ixgbe[vf] drivers built-in

garthy:
Signed-off-by: Garth Bushell <garth@garthy.com>

Makes the Intel ixgbe and ixgbevf kernel drivers built-in. These are used on Amazon(AWS) and Oracle(OCI) Cloud.


rn:
why do they need to be built in if they are avail as modules? In general we prefer to keep hardware drivers as modules

eyz:
@rn I think the goal is to attach the volumes in `onboot`, before services. That is simply my understanding.

rn:
Actually, the kernel modules for drivers should be autoloaded by `mdev` which is run by init [here](https://github.com/linuxkit/linuxkit/blob/master/pkg/init/cmd/rc.init/main.go#L248). We had issues with cold plug, ie loading drivers for devices present at boot time, before. maybe it got broken...

garthy:
So the use case for this Oracle Cloud is we boot off iSCSI via the Ibft and netroot. I believe this needs the network drivers loaded. Having just looked at this I've just realised the CONFIG_ISCSI_IBFT_FIND will also need to be set. hang on..

justincormack:
(needs a rebase)

If you want iscsi root that would also need the iscsi modules built in? Also though, how are you going to do the iscsi attach? As far as I know the kernel cannot attach an iscsi root directly, so you have to do this in userspace, ie from the initramfs anyway, in which case you don't need the moule built in.

It might make sense to get this working in a branch and then work out how to integrate after.

rn:
you could still PXE boot a kernel+initrd where the initrd has the relevant modules and then switch root to iSCSI. That's how this is normally done with other distros.

rn:
closing this issue due to idleness and it's not clear if this is needed. @garthy please re-submit (or open an issue to discuss further),

PR #2624, opened 2017-10-18
Fix broken AWS / OpenStack metadata, add example config for latter

yankcrime:
**- What I did**

Fix a typo that somehow snuck in via 6deabe8a2 *cough* and which broke metadata access for AWS and OpenStack providers.

By way of penance, this commit adds a configuration example for OpenStack.

**- How to verify it**

Booting an image should now result in successful metadata retrieval on AWS and OpenStack, as verified in logs such as:

```
2017/10/18 21:44:52 openstack: Probe succeeded
```

Fix metadata URL typo, add an example configuration file for OpenStack.

img_4480

### DIFF
--- a/examples/openstack.yml
+++ b/examples/openstack.yml
@@ -1,0 +1,36 @@
+kernel:
+  image: linuxkit/kernel:4.9.54
+  cmdline: "console=ttyS0"
+init:
+  - linuxkit/init:6b3755e47f00d6027321d3fca99a19af6504be75
+  - linuxkit/runc:52f92cb577879ce4cfe4e89be2d63af82523fc92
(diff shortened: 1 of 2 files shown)


deitch:
Nice catch. 

I don't understand how it ever worked before, since all metadata URLs would be malformed, e.g. `http://169.254.169.254/latest/meta-datahostname` instead of `http://169.254.169.254/latest/meta-data/hostname`.

I don't see any tests around it accessing the metadata, we probably should add so we don't vet caught in it again.

rn:
I pushed updated packages as `linuxkit/metadata:c385bb2dce56acf8b831bbf7fca3c8fc836ded66`

PR #2627, opened 2017-10-20
Improve `linuxkit pkg` handling of git 

### LINKED ISSUE #2613: linuxkit pkg build has no "ignore git" option
If you try to use `linuxkit pkg build` outside a git repo, you get `fatal: Not a git repository (or any of the parent directories): .git`

I can't actually seem to be able to use it outside a git repo...

I think there should be a `--nogit` option that doesn't try to label it at all.

cc @ijc

ijc:
Fixes #2613 (allowing building etc of packages not in git) but also fixes a more general issue where things wouldn't work if invoking `linuxkit pkg` from outside the git tree containing the package.

Previously:
```
$ cd /not/a/git/dir
$ git status
fatal: Not a git repository (or any of the parent directories): .git
$ linuxkit pkg build $GOPATH/github.com/linuxkit/linuxkit/pkg/swap/
fatal: Not a git repository (or any of the parent directories): .git
exit status 128
```
Even though `$GOPATH/github.com/linuxkit/linuxkit/` is a git clone/

This now works correctly. As does:
```
$ git status
fatal: Not a git repository (or any of the parent directories): .git
$ ls -A
build.yml  Dockerfile
$ cat build.yml 
image: dummy
$ cat Dockerfile 
FROM hello-world:latest
$ linuxkit pkg build .
Building "linuxkit/dummy"
Using default tag: latest
[...]
Successfully tagged linuxkit/dummy-amd64:latest
Tagging hello-world@sha256:07d5f7800dfe37b8c2196c7b1c524c33808ce2e0f74e7aa00e603295ca9a0972 as hello-world:latest
Tagging linuxkit/dummy-amd64 as linuxkit/dummy
Build complete, not pushing, all done.

### DIFF
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -71,7 +71,7 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 	}
 
 	if bo.release == "" {
-		r, err := gitCommitTag("HEAD")
+		r, err := p.git.commitTag("HEAD")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


ijc:
> Tagging linuxkit/dummy-amd64 as linuxkit/dummy

I think I should make the `:latest` explicit in this code rather than relying on the builder.

PR #2629, opened 2017-10-22
Kernel updates

### LINKED ISSUE #2593: drop patches from 4.4 kernels
We have a fairly large number of patches on the 4.4 kernels, mainly around vsock, hvsock. Anyone wanting to use those should be on a more recent kernel. I sugegst we drop our entire patch queue for 4.4 and just ship mainline. Now it is the six year support kernel it is not going away in January so we really need to make a decision on future support...

rn:
- Enable kernel builds without patch queue. This allows us to build un-patched mainline kernels
- Remove 4.4.x patches. The patches were for vsock and hvsock support. If you need those, you should use a newer kernel. resolves #2593
- drop `kernel_` prefix from config files
- kernel updates
  - 4.13.7, 4.13.8, 4.13.9
  - 4.9.57, 4.9.58
  - 4.4.93, 4.4.94

All kernels compiled, pushed, signed for x86_64 and arm64

polar-bear


.gitattributes
```diff
@@ -1,2 +1,2 @@
 **/vendor/** linguist-vendored
-kernel/*.x linguist-language=text
+kernel/config-*.x linguist-language=text
```
ijc: ITYM `kernel/config-*.x-*` due to the arch suffix, you are trying to match e.g. `kernel/config-4.9.x-x86_64`. 

Actually `kernel/config-*` would work better since it would catch `kernel/config-dbg` too.

rn: Ah, forgot about the arch suffix. pushed an update

PR #2640, opened 2017-10-25
Output image_id, server_id in stdout to ease automation

mmariani:
**- What I did**

I've added the image_id, server_id values to the command output, in addition to the log messages.

**- How I did it**

Just a couple of prints.

**- How to verify it**

image_id=$(linuxkit push openstack ...)
server_id=$(linuxkit run openstack ...)

"run openstack" and "push openstack" now print image.id and server.id to stdout to avoid the need of parsing the command result in scripts.


justincormack:
Do you think it makes sense to update the docs at https://github.com/linuxkit/linuxkit/blob/master/docs/platform-openstack.md too?

mmariani:
Actually, from the docs:

> If successful, this will return the image's UUID

and

> This will create a new instance with the same name as the image, and if successful will return the newly-created instance's UUID.

It's exactly what it does with my patch. It _also_ logs stuff on stderr for humans, but that's not "returning".

Issue #2639, opened 2017-10-25
Kubernetes helm not working on cri-containerd


w9n:
**Description**
helm is working fine on linuxkit kubernetes docker version but not on cri-containerd

**Steps to reproduce the issue:**
bootstrap cri-containerd linuxkit cluster
rbac and helm init:
```
kubectl -n kube-system create sa tiller
kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller
helm init --service-account tiller
```

**Describe the results you received:**
on cri-containerd version
```
$ helm version
Client: &version.Version{SemVer:"v2.7.0", GitCommit:"08c1144f5eb3e3b636d9775617287cc26e53dba4", GitTreeState:"clean"}
E1025 12:58:56.258425   32701 portforward.go:331] an error occurred forwarding 38803 -> 44134: error forwarding port 44134 to pod 4cbb8e4ddffbb89b6bc610548a6f5a322305330ed84aa8bddb8244f3e2e099be, uid : failed to find socat: exec: "socat": executable file not found in $PATH
Error: cannot connect to Tiller

```
but this works
```
kubectl port-forward tiller-deploy-5458cb4cc-gztfv  38803:44134 --namespace kube-system
Forwarding from 127.0.0.1:36949 -> 44134
```
**Describe the results you expected:**
on docker version:
```
$ helm version
Client: &version.Version{SemVer:"v2.7.0", GitCommit:"08c1144f5eb3e3b636d9775617287cc26e53dba4", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.7.0", GitCommit:"08c1144f5eb3e3b636d9775617287cc26e53dba4", GitTreeState:"clean"}
```

**Additional information you deem important (e.g. issue happens only occasionally):**
there are many posts about coreos missing socat in the kubelet context, but the linuxkit cri-containerd and docker version actually use the same kubelet container yml which has socat.

justincormack:
Thats weird. Any idea where `portforward.go` is from?

w9n:
there is:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/portforward.go
and 
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/portforward/portforward.go

[...]

Issue #2632, opened 2017-10-23
kubernetes cri-containerd image error on reboot


w9n:
**Description**
cri-containerd kubernetes version doesnt succefully reboot.

**Steps to reproduce the issue:**
```
KUBE_MASTER_AUTOINIT="" ./boot.sh
<wait and check for kubernetes>
poweroff
KUBE_PRESERVE_STATE=true ./boot.sh
```

**Describe the results you received:**
no task in the k8s.io namespace start, lots of missing image errors in the logs.

ctr -n k8s.io images list
```
ERRO[0000] failed calculating size for image docker.io/weaveworks/weave-kube:2.0.4  error="content digest sha256:1ad426d140e5e15b6584f3dbad0e98084993f02ad95e3cba2086b0b054a04033: not found"
ERRO[0000] failed calculating size for image docker.io/weaveworks/weave-kube@sha256:cddb2620620481271763d39c797b2f0b970d18bc5e98ffb7151a3c5f8a923324  error="content digest sha256:1ad426d140e5e15b6584f3dbad0e98084993f02ad95e3cba2086b0b054a04033: not found"
ERRO[0000] failed calculating size for image docker.io/weaveworks/weave-npc:2.0.4  error="content digest sha256:0dad21d6c11e2aad989c0359bd7d2f60470aede01a186dc651c645542dffc28b: not found"
ERRO[0000] failed calculating size for image docker.io/weaveworks/weave-npc@sha256:676b3a815e2cb2477e6d4ed3c6699c3d3a9c23e30b8f5e76af8e5c0b5030d525  error="content digest sha256:0dad21d6c11e2aad989c0359bd7d2f60470aede01a186dc651c645542dffc28b: not found"
ERRO[0000] failed resolving platform for image gcr.io/google_containers/etcd-amd64:3.0.17  error="content digest sha256:243830dae7dd6ff78859fa1d66098a89e2951a9e95af4ef4d4d2c03d97975771: not found"
ERRO[0000] failed resolving platform for image gcr.io/google_containers/etcd-amd64@sha256:d83d3545e06fb035db8512e33bd44afb55dea007a3abd7b17742d3ac6d235940  error="content digest sha256:243830dae7dd6ff78859fa1d66098a89e2951a9e95af4ef4d4d2c03d97975771: not found"
[...]

PR #2645, opened 2017-10-26
kubernetes: add socat to cri-containerd and dependency fixes

w9n:
See #2639

### DIFF
--- a/pkg/containerd/Dockerfile
+++ b/pkg/containerd/Dockerfile
@@ -1,4 +1,4 @@
-FROM linuxkit/alpine:07f7d136e427dc68154cd5edbb2b9576f9ac5213 as alpine
+FROM linuxkit/alpine:69019e422ef1185138e24e49d24330aab7002b74 as alpine
 RUN apk add tzdata
 
 WORKDIR $GOPATH/src/github.com/containerd/containerd
--- a/projects/kubernetes/cri-containerd.yml
+++ b/projects/kubernetes/cri-containerd.yml
@@ -1,6 +1,6 @@
 services:
   - name: cri-containerd
-    image: linuxkitprojects/cri-containerd:afb7c111a40788b1e085856fbddacf8a878385e7
+    image: linuxkitprojects/cri-containerd:e5d36c799a0e3ec495648afb362a7beac56f7417
 files:
   - path: /etc/kubelet.sh.conf
     contents: |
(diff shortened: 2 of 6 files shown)


w9n:
#2632

w9n:
```
[+ 68m 20s] error during connect: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.29/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%22org.opencontainers.image.revision%22%3A%22cc4d2e96fba86ee1816ab2f6376855090a9578bc%22%2C%22org.opencontainers.image.source%22%3A%22https%3A%2F%2Fgithub.com%2Flinuxkit%2Flinuxkit%22%7D&memory=0&memswap=0&networkmode=none&rm=1&shmsize=0&t=linuxkit%2Fcontainerd%3A69968ebd9f66edccceec0344865bf648ad105ea9-amd64&target=&ulimits=null: No trust data for 69019e422ef1185138e24e49d24330aab7002b74
```
alpine base image need to be signed and pushed first

eyz:
@w9n maybe revert the image hash, let the CI go, and then get someone to build and push it first?

w9n:
The image can be build from my branch aswell - no need for 2 PRs

eyz:
@w9n do you mean from w9n:helm-bug? Ideally CI would pass before the package get built and signed, from what I've seen. I think you may want to revert the hash and `git push -f` to get CI to pass, then one of the maintainers can build the package and go from there.

w9n:
ill do if required but the first PR and ci would be pointless and inconsistent because the tests and images would use old hashes and containers. This might should be rethought. @ijc will take care of this PR next week.

rn:
This PR changes too many unrelated things in one commit. From I can see, there are several unrelated changes, which ideally should be broken up into separate commits.
- bump `containerd` to a new non-tagged version and at this point I think we prefer to only use tagged versions.
- Update Kube project to new containerd, Note, if we bump `containerd`, we should bump all uses of `containerd` because otherwise we run multiple versions of `containerd` which seems messy and makes it hard to debug.
- Add `socat` to `cri-containerd` package
- Update `cri-containerd` hashes.

Also, changes to the `linuxkit/alpine` package are best done by a maintainer, because by the time one of us pushes your changes, the versions may have changed again. It also needs to be pushed to `arm64` which would likely change the `aarch64` versions file as well.

ijc:
I added `socat` as part of #2665. I think the containerd bump here can wait until beta.3 comes out (should be quite soon it seems).

PR #2648, opened 2017-10-28
WIP: linuxkit pkg template yml to generate immutable yml for reuse

w9n:
When writing a moby yml I do not care about the container version and hashes (only if I specifically want to stay with a version) and would like to simplify the process of writing them. I also do not want to keep track of all pkg version I use downstream and update them manually when something changes upstream.

The idea is to optionally split the moby config into 1 mutable and n immutable versions. The mutable version describes the desired state which uses image names like ```linuxkit/containerd:mutable```(or the usual hash tags for pinning down on a version), while the immutable configs stay like currently and get created on demand or mutation from a specifc pkg git repo. They can be versioned and stored in the end. 

Therefore ```linuxkit pkg collect --pkgroot /path/to/repo/linuxkit mutable.yml ``` first collects which container in the mutable yml are tagged mutable and then walks recursivly through pkgroot and looks for  ```build.yml```of a package with the same image name.  When found it calculates and stores the package hash (Not sure about error case). At this Point we could compare with the last hash of the image and implement build and push on change. At the end a yml with the hashes gets written versioned into a specific dir near the mutable yml and persist as an Artifact for a possible build.

This could also be used for CI that triggers ```linuxkit pkg collect --push``` on every pull for every mutable image yml. Would need to save the immutable artifacts somewhere.

The big disadvantage is that you probably dont want moby as dependency in linuxkit which is needed to parse the moby config (On the other side the linuxkit git pkg for the hashes is private). Let me know what you think about it!

(Im just getting used to go, so its a bit messy and uncomplete yet)


rn:
I don't understand the explanation for this PR (maybe I'm missing some context).

Could you first describe which problem you try to solve?

w9n:
added some context to the description.

deitch:
> When writing a moby yml I do not care about the container version and hashes (only if I specifically want to stay with a version) and would like to simplify the process of writing them. I also do not want to keep track of all pkg version I use downstream and update them manually when something changes upstream

I have a hard time understanding how that would work with immutable images (which is what LinuxKit is all about). 

I think this might be easier to understand if you gave sample `.yml` files for what it would look like if this were in place?

w9n:
i basically thought of a not buildable config
```
kernel:
  image: linuxkit/kernel:4.9.58
init:
  - linuxkit/init:mutable
  - linuxkit/runc:mutable
  - linuxkit/containerd:82be2bbb7cf83bab161ffe2a64624ba1107725ff
  - linuxkit/ca-certificates:mutable
onboot:
  - name: sysctl
    image: linuxkit/sysctl:mutable
  - name: rngd1
    image: linuxkit/rngd:mutable
    command: ["/sbin/rngd", "-1"]
services:
  - name: getty
    image: linuxkit/getty:mutable
    env:
     - INSECURE=true
  - name: rngd
    image: linuxkit/rngd:mutable
  - name: dhcpcd
    image: linuxkit/dhcpcd:mutable
  - name: sshd
    image: linuxkit/sshd:mutable
files:
  - path: root/.ssh/authorized_keys
    source: ~/.ssh/id_rsa.pub
    mode: "0600"
    optional: true
trust:
  org:
    - linuxkit

```
when used with ```linuxkit pkg collect --pkgroot /path/to/repo/linuxkit ssh-mutable.yml``` creates a new file artifacts/ssh-vXYZ.yml with the most current hashes of pkgroot

```
kernel:
  image: linuxkit/kernel:4.9.58
init:
  - linuxkit/init:42a92119e1ca10380e0d33e26c0cbcf85b9b3558
  - linuxkit/runc:817fdc592eac6cb7804fa1721a43a7f6e23fb50f
  - linuxkit/containerd:82be2bbb7cf83bab161ffe2a64624ba1107725ff
  - linuxkit/ca-certificates:af4880e78edc28743f7c5e262678c67c6add4c26
onboot:
  - name: sysctl
    image: linuxkit/sysctl:a9ad57ed738a31ea9380cd73236866c312b35489
  - name: rngd1
    image: linuxkit/rngd:842e5e8ece7934f0cab9fd0027b595ff3471e5b9
    command: ["/sbin/rngd", "-1"]
services:
  - name: getty
    image: linuxkit/getty:626ccc8e1766c40447f29a790d3a7cfff126f2a2
    env:
     - INSECURE=true
  - name: rngd
    image: linuxkit/rngd:842e5e8ece7934f0cab9fd0027b595ff3471e5b9
  - name: dhcpcd
    image: linuxkit/dhcpcd:48831507404049660b960e4055f544917d90378e
  - name: sshd
    image: linuxkit/sshd:f55ec010619e19178d5daecb4e595e84ecbf7d67
files:
  - path: root/.ssh/authorized_keys
    source: ~/.ssh/id_rsa.pub
    mode: "0600"
    optional: true
trust:
  org:
    - linuxkit

```
I think the main practical reason is to use it in CI where it is necessary to know which pkg has changed to not rebuild everything on a merge. It also gives gives rollback capabilities outside the git history.

But I reevaluated priorities and think its a nice to have but linuxkit is working fine now and i dont think its worth the moby dep. I might create a little external tool for practicing, but low priority. 

There is a different problem that a container ```linuxkit/sshd:f55ec010619e19178d5daecb4e595e84ecbf7d67``` is not really immutable because it pulls latest alpine packages on container build time. The current Hash is needed for multi arch but by creating ymls as artifacts it might be possible to use real hashes of a container and make the yml artifacts platform specific. But this gets out of scope for me.

deitch:
So, is this basically a way of saying, "for some packages, I want the latest available hash at build time"?

w9n:
yep, but still create the immutable ymls so its reproducable what has been build and compare what has been changed.

deitch:
So rather than the usual moby `.yml`, it kind of is a moby "template" `.yml`, that gets parsed to generate the real `.yml` with immutable hashes? Still immutable, but somewhat easier to maintain?

w9n:
exactly. It would enable other possibilities as well as described above but there might be other ideas for CI and real hashes.

deitch:
OK, I get it now; want to think it over somewhat. @rn had commented earlier; let's see his thoughts.

I am going to rename the issue.

rn:
This would be better done in the `moby` tool as it handles the build YAML files. There are two related proposals:
- https://github.com/moby/tool/issues/95
- https://github.com/moby/tool/issues/101

deitch:
Ah yes, the `.mobytags` and overrides proposals.

@w9n does that address the question?

w9n:
yes, it pretty much goes is in the right direction. I will read and think about the related Issues and PRs in the next days.

ijc:
I'm still catching up on my post PTO backlog(s) so I haven't fully grokked this one yet, but one quick though I had was that what is called `foo:mutable` here would be called `foo:latest` in many other contexts.

w9n:
i didnt choose ```foo:latest``` because it can conflict with other contexts. But can maybe be prevented if carefully designed. I created a proper proposal.

w9n:
Since moby became a dependency this works now depending on moby/tool#194 `linuxkit pkg collect --pkgroot ~/go/src/github.com/linuxkit/linuxkit/pkg test.yml`

<details>
 <summary>test,yml</summary>

```
kernel:
  image: linuxkit/kernel:4.9.58
init:
  - linuxkit/init:<latest>
  - linuxkit/runc:<latest>
  - linuxkit/containerd:82be2bbb7cf83bab161ffe2a64624ba1107725ff
  - linuxkit/ca-certificates:<latest>
onboot:
  - name: sysctl
    image: linuxkit/sysctl:<latest>
  - name: rngd1
    image: linuxkit/rngd:<latest>
    command: ["/sbin/rngd", "-1"]
services:
  - name: getty
    image: linuxkit/getty:<latest>
    env:
     - INSECURE=true
  - name: rngd
    image: linuxkit/rngd:<latest>
  - name: dhcpcd
    image: linuxkit/dhcpcd:<latest>
  - name: sshd
    image: linuxkit/sshd:<latest>
files:
  - path: root/.ssh/authorized_keys
    source: ~/.ssh/id_rsa.pub
    mode: "0600"
    optional: true
trust:
  org:
    - linuxkit
```

</details>
<details>
 <summary>result hash of the yml files bytes (5f39869eef39b525e90661e2327ed645c1a9df8a.yml)</summary>

```
kernel:
  image: linuxkit/kernel:4.9.58
  cmdline: console=tty0 console=ttyS0 console=ttyAMA0
init:
- linuxkit/init:9250948d0de494df8a811edb3242b4584057cfe4
- linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
- linuxkit/containerd:82be2bbb7cf83bab161ffe2a64624ba1107725ff
- linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
onboot:
- name: sysctl
  image: linuxkit/sysctl:ce3bde5118a41092f1b7048c85d14fb35237ed45
- name: rngd1
  image: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
  command:
  - /sbin/rngd
  - "-1"
onshutdown: []
services:
- name: getty
  image: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
  env:
  - INSECURE=true
- name: rngd
  image: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
- name: dhcpcd
  image: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
- name: sshd
  image: linuxkit/sshd:ac5e8364e2e9aa8717a3295c51eb60b8c57373d5
trust:
  org:
  - linuxkit
files:
- path: root/.ssh/authorized_keys
  directory: false
  source: ~/.ssh/id_rsa.pub
  optional: true
  mode: "0600"
```

</details>

_
I would like to have a file like `~/.linuxkit.conf` where i can specify multiple pkg roots.
It could also easily be upgraded to support version tags in that file (as proposed in moby/tool#101 )  or in the `build.yml` in the pkg directory.

w9n:
@deitch @rn @ijc Please recheck, adopting this could quit manually updating hashes upstream.
ive added default pkg roots in the global conf e.g.:
```
repos:
  - path: /go/src/github.com/linuxkit/linuxkit/pkg
  - path: /go/src/github.com/linuxkit/kubernetes/pkg
```
and a tags array in `build.yml` (might should be cached in a centralized db)
```
tags:
  - name: <2.1>
    hash: bb0c6ae2f12a1b55df24ebce2067784a931285df
```
if a container has `<latest>` tag it calculates the git hash of the tree and checks if the tree is dirty. 

There is `linuxkit pkg build -c` which renders the template and outputs an image as usual.


I would like to autobuild the tags. it should be possible to recover a tree hash from .git into a temp dir to build pkgs that are not latest. Does anyone has experience with that?

I also would like to be able to override the default config in `build.yml` by the tag field at some point.

deitch:
I remain torn on this. I think I still like the idea of having the ability to tag an image with something special (like `<latest>`) and have it actively go and get the most recent version and populate it. it is less pure, but addresses a real use case.

On the other hand, something still feels a bit funny to me on this. If we are advocating immutable builds, then do we  _want_ the examples to have tags that need be pre-processed? Those are likely to become the most referenced starting points. Do we want to encourage mutable `yml`? The logical next step is just to have that stage 1 yml be accepted by `lkt build` and be pre-processed and built in one step, saving the output yml somewhere (local filesystem? inside image? tag?).

If I understand the implementation correctly, `pkg collect` would then look in local dirs as specified by the default or `--pkroot`? If so, I would need to have all of that installed if I want to run `lkt build` while using mutable tags, which would prevent the majority of users from using it. Most just want to install the binary and run it, not build it like we do. Why wouldn't it just talk to the registry directly?

w9n:
> If I understand the implementation correctly, pkg collect would then look in local dirs as specified by the default or --pkroot? If so, I would need to have all of that installed if I want to run lkt build while using mutable tags, which would prevent the majority of users from using it. Most just want to install the binary and run it, not build it like we do. Why wouldn't it just talk to the registry directly?

Yes, you would need atleast the linuxkit/linuxkit repo and a setup pkgroot parameter or global config to render the template. The rendered template could still get provisioned somewhere and be used as right now, but i dont think its desirable to do this manually in this repo.

> The logical next step is just to have that stage 1 yml be accepted by lkt build and be pre-processed and built in one step, saving the output yml somewhere (local filesystem? inside image? tag?).

Thats what `linuxkit build -collect` is for. It currently only uses the pkg roots of the global config and is not outputting the yml. I would prefer to have it in a dir with the different image formats for that particular build but quite unsure about the naming conventions yet. (It currently not using the rendered template which is pretty weird, ill fix that tonight). Should they all just be named by the hash of the yml + ending or by an paramter + ending?

`linuxkit pkg collect --pkgroot` currently appends the parameters to the global conf. I think overriding is better here.

> Why wouldn't it just talk to the registry directly?

hm, we might could also tag images <latest> and stuff but i dont think this would help make things more reproducable and visible to the user.

ijc:
> if a container has `<latest>` tag

Are the angle-brackets (`<>`) literally part of the tag here or just "syntax"?

I'm also not too clear on the difference between the tags here and those which can be part of the image on registry (i.e. what we currently populate with the hash when we push). Once LinuxKit has a release I was imagining we would push tagged images corresponding to that release to hub.

Maybe the `<>` is precisely to deal with that case?

If there was some documentation added by this PR describing things for the users then I might be less fuzzy on what was going on (I've not tried to work it out from the code).

w9n:
> Are the angle-brackets (<>) literally part of the tag here or just "syntax"?

a Tag is considered a Template and tried to be rendered if its surrounded by [<>](https://github.com/linuxkit/linuxkit/pull/2648/files#diff-54b90923dcb624b9443397970cfcb903R140), so that it doesnt conflict with tag names from normal docker images. `<latest>` is a special case. The Tag name in `build.yml` currently needs to be something like `<2.1>`.

If we just simply would use tags like `linuxkit/init:latest` from the registry, then images build on different times would be different without recognizing or reproducability for the user. `linuxkit/init:2.1` might be viable but having this information only in the registry is not really reproducable if the registry dies or something.

Ill try to document everything.

deitch:
> Yes, you would need atleast the linuxkit/linuxkit repo and a setup pkgroot parameter or global config to render the template
> Thats what linuxkit build -collect is for

Then I am completely confused. I had thought the goal here was to have an ability to have a form of `yml` that can be pre-processed by `lkt` to turn into a "true" `yml` with hash versions etc. The only difference would be that the "pre" `yml` has a special "marker" tag that means, "replace me with the latest available version" (else we would just specify an exact version).

If so, wouldn't `pkg collect` find latest from the registry?

Are there 2 distinct behaviours?

w9n:
> Are there 2 distinct behaviours?

There is `linuxkit pkg collect --pkgroot` which only does the pre-processing and outputs a valid yml and there is `linuxkit build -collect` which pre-processes and directly outputs the image (should output the valid yml aswell). `<latest>` gets replaced with the tree hash of pkg dir. E.g. there are uncommited changes in linuxkit/pkg/init then it gets replaced by `linuxkit/init:9250948d0de494df8a811edb3242b4584057cfe4-dirty`. I would like dirty hashes to be always auto-build for rapid development and other hashes if they are not available so that i dont have to manually build images.

deitch:
> there is linuxkit build -collect which pre-processes and directly outputs the image

So `build -collect` = `lkt pkg collect > valid.yml && lkt build valid.yml` all in one?

I find it somewhat confusing, as the commands don't seem as instinctive. Basically, this all works only  if I am in "developer mode", i.e. I have the local go data installed. It creates a `lkt` run mode where it no longer is 100% standalone (well, docker run as well for creating output images, but slowly working on that).

rn:
`linuxkit pkg collect` seems a misnomer as it is not an operation on a package. It confused me before and I was half way through writing a comment here...

Anyway, I would prefer if there was a distinct step, say:
```
linuxkit transform --pkgroot <org>:<path> --pkgroot <org>:<other path> foo.template.yml > foo.yml
linuxkit build foo.yml
```
The first would translate/tranform (not sure about the name) a *template* file into a valid YAML file by filling in the tags from the packages in the package route directories. And the second is the build as it is today. The `--pkgroot` format supports multiple orgs on hub, so for something like `--pkgroot  linuxkit:~/src/linuxkit/linuxkit/pkg --pkgroot linuxkit:~/src/linuxkit/kubernetes/pkg` means that the hash for `linuxkit/init` in the YAML would be looked up in two places.

The first step could also add the git HEAD hash as a comment at the start for `foo.yml`. If you really want to you can also do `linuxkit transform foo.template.yml | linuxkit build -`  but that way it's harder to have a record of actually got build.

deitch:
> linuxkit pkg collect seems a misnomer as it is not an operation on a package

+1 

That definitely was part of my confusion, but wasn't clear to me until you articulated it.

> I would prefer if there was a distinct step

Yes, but I could see how you might want a convenience of a single step. 

> The --pkgroot format supports multiple orgs on hub

I still am confused about this. I had thought the idea was to have `<latest>` means, "get most recent hash from the hub." Why do I need it locally?

rn:
> Yes, but I could see how you might want a convenience of a single step.

You get a "single" step with `linuxkit tranform | linuxkit build -`. I really do not want `linuxkit build` to do too much magic. It should just build what it is told to build.

> I still am confused about this. I had thought the idea was to have <latest> means, "get most recent hash from the hub." Why do I need it locally?

I don't want `latest` as in `docker run alpine` and you have no idea what you get. It makes no sense and defeats the purpose of LinuxKit. You should always build from known versions. What `linuxkit tranform` does, it would do the equivalent of `linuxkit pkg show-tag` for the package and use that hash and stick it into the YAML.

deitch:
>  I really do not want linuxkit build to do too much magic. It should just build what it is told to build.

OK.

> I don't want latest as in docker run alpine and you have no idea what you get... You should always build from known versions. 

Agreed. I distinctly dislike docker's `latest` (even if I get the convenience).

> What linuxkit tranform does, it would do the equivalent of linuxkit pkg show-tag for the package and use that hash and stick it into the YAML.

In other words, `lkt transform` (`templatify`? :-) ) is a _developer_ tool, not an _end-user_ tool. It isn't used by end-users working with just the `lkt` binary, since they need the local clone. 

If that is true, is it really needed? Wouldn't the scripts already in here to modify the hashes be enough, or maybe a slightly modified version?

I see the benefit for end-users, where there aren't local repos and scripts and heaven-knows-what, so they might want a templating system to generate immutable from "give me the latest". But for devs who already have the entire repo, and thus all of the scripts, is it necessary? I get the convenience, but is it worth all _that_ much?

deitch:
On second thought, I guess I can see it. `lkt` already contains all of the code for parsing the `yml`, so it fits there better than in some scripts that have to regenerate it. I have done that enough times with kube `yml`, e.g. and it always is messy.

OK, withdraw that question. But it _clearly_ is a developer tool for local usage.

rn:
sure, this can be done with a shell script. `grep` image patterns and `linuxkit pkg show-tag` plus sed to update `sed`.

ijc:
> You get a "single" step with linuxkit tranform | linuxkit build -. I really do not want linuxkit build to do too much magic. It should just build what it is told to build.

That magic would also imply that one could not use `moby build` as an alternative/equivalent to `linuxkit build`, and I'm not sure of that is desired or not.

w9n:
dropping `linuxkit build -collect` and `--pkgroot <org>:<path>` sounds good to me.

I would suggest adding `linuxkit transform -o outputfile <filename> | linuxkit build -` or something. 

Can we agree on `transform` in general? 

Is it okay to keep templating the pkg tags for possible autobuild and also use the tags as image tag in the pkg build process later? 

Ill put autobuild on the roadmap for another PR.

rn:
`linuxkit transform -o outputfile <filename> | linuxkit build -` does not seem to make sense to me

alternative naming would be `transform`, `translate`, `generate`. At the moment I think `generate` is my favourite, and we could add a comment to the generated file saying that is was auto-generated etc.

I think there are several use cases, like LCOW, where it would be nice to have a template and then generate the actual YAML and check it in.

I don't know what you mean by autobuild?

deitch:
>  I think generate is my favourite

I like it too. My only concern is that, to some degree, build = generate. We are generating the output img, which is equivalent to building it. Nothing about "generate" implies "generate config file". But you could say the same thing about "transform". Maybe not "translate", but "translate" implies between two equal versions, which this is not.

`compile`? `process`? `preprocess`? `m4`? (ok, last one tongue in cheek....)

deitch:
Or perhaps `lkt template`?

w9n:
generate sounds good but dont mind changing.

Changed to `linuxkit generate --pkgroot <org>:<path> --pkgroot <org>:<path> <filenames> | linuxkit build` and dropped `linuxkit build -collect` 

There are some logs that im now writing to stderr. writing the result always to stdout works well.

I would suggest to add auto build dirty pkgs on `linuxkit generate --auto`.

There should also be a cmd for something like `linuxkit pkg build -t 2.3 ./somepkgdir

rn:
@w9n thanks for the quick change, I have a few comments on the implementation itself, but before you spent more time on this it might be good to hear from @ijc and @justincormack what they think of the current proposal (we had quite a few discussions in the past on similar approaches with different views).

So the current proposal is to have a `foo.template.yml` which does not have all the tags present and then a new `linuxkit build` command which takes a list of prefixes and package directories and generates a `foo.yml` from it. So you can do something like:
```
linuxkit generate --pkgdir linuxkit:~/src/linuxkit/linuxkit/pkg --pkgdir linuxkit:~/src/linuxkit/kubernetes/pkg foo.template.yml > foo.yml
```
This would substitute any `<latest>` tags in `foo.template.yml` with the current hashes.

@ijc @justincormack what's your view on this proposal. I quite like it.

@w9n a few notes on the current implementation:
- you shouldn't need to add anything to `pkglib`. This is a new top level command which doesn't operate in packages at all. Maybe `pkglib` needs extending, but a quick look suggests it may have all the interfaces needed.
- I don't understand why you need to walk directories in the code. If the template contains `linuxkit/foo:<latest>` and you pass in `linuxkit:~/src/linuxkit/linuxkit/pkg` you can just do the equivalent of `linuxkit pkg show-tag ~/src/linuxkit/linuxkit/pkg/foo` to get the tag. No need to walk directories etc.
- Does your current implementation take multiple templates as argument, like `linuxkit generate ... foo.template.yml bar.template.yml`? This might be quite nice to have, but can be added later.

ijc:
> then a new `linuxkit build` command 

ITYM `linuxkit generate`? Not mad enamoured on the name, but no better ideas (`bake` or `substitute` came to mind but not sure they are better).

> I don't understand why you need to walk directories in the code. 

Is that to avoid assuming that the package `linuxkit/foo` is built from `pkg/foo`? Not sure if this holds today (I suspect so). It may not hold for 3rd party repos I guess.

My only concern is whether or not there is some existing YAML templating syntax/library/spec which we could be reusing here instead of inventing my own, and as a corollary to that how `<>` interacts with regular YAML syntax and also how it interacts with the permissible tags for things on hub (i.e. what if I push a literal `linuxkit/foo:<x.y.z>` to hub and want to use it? Are `<>` valid in that context).

ijc:
> My only concern is whether or not there is some existing YAML templating syntax/library/spec which we could be reusing here instead of inventing my own

Couldn't find anything (at first `tags` in the spec looked useful, until I realised that was for _types_ not values). Maybe aliases could be used but the idea of an application populating a default set of aliases doesn't seem to be considered.

Many projects seem to apply Go's `text/template` to their yaml (i.e. the `{{.Foo}}` stuff). Not sure how I feel about that but we could have `{{.Pkg.LinuxKit.Foo.latest}}` be handled that way perhaps (main benefit is we don't have to think about parsing and substitution too hard ourselves). Or probably `<>` is just fine.

deitch:
> Couldn't find anything

deitch:
> ITYM linuxkit generate? Not mad enamoured on the name, but no better ideas (bake or substitute came to mind but not sure they are better).

`linuxkit template`?

w9n:
> you shouldn't need to add anything to pkglib. This is a new top level command which doesn't operate in packages at all. Maybe pkglib needs extending, but a quick look suggests it may have all the interfaces needed.

That was the main reason i ended up with `linuxkit pkg collect` in the beginning; the git methods are not public but i actually would like to replace them with something like https://github.com/src-d/go-git . Is it possible to build a pkg which tree is not part of HEAD atm? Basically pkg/init with some old hash? This should be possible and there is `linuxkit pkg build -hash-commit` but i does seem to make something different.

> I don't understand why you need to walk directories in the code. If the template contains linuxkit/foo:<latest> and you pass in linuxkit:~/src/linuxkit/linuxkit/pkg you can just do the equivalent of linuxkit pkg show-tag ~/src/linuxkit/linuxkit/pkg/foo to get the tag. No need to walk directories etc.

I only did this so that its as flexible as possible in the beginning but ill implemented a fixed sheme like `./*/build.yml` relative to the pkgroot.
> Does your current implementation take multiple templates as argument, like linuxkit generate ... foo.template.yml bar.template.yml? This might be quite nice to have, but can be added later.

yes [move moby config cli parsing from pkg build to util](https://github.com/linuxkit/linuxkit/pull/2648/commits/bceca7732183ef5b26911c372278890b0cb5c15a) moves the build parsing to util so that it can be reused by `linuxkit generate`

> what if I push a literal linuxkit/foo:<x.y.z> to hub and want to use it? Are <> valid in that context

<> are not valid as docker image tag name.


Things actually get interesting with #2817, with a tag like this:
```
tags:
  - name: <dev>
    hash: <latest>
    config:
      build_args:
        kubernetes_repo_url: $KUBERNETES_REPO
```

Where $KUBERNETES_REPO is a keyvalue from either global conf or from cli parsing.

w9n:
bake & build would kind of short and precisely describe whats going on and actually sound nice :)

ijc:
> the git methods are not public 

You appear to be using them to reimplement a subset of the pkglib functionality, which I expect will be rather error prone or at the very least subject to drift.

Instead you should use the public methods to create a `pkglib.Pkg` and use its methods to get the hash etc. If public methods for doing what you need in that regard that are missing we can add them but I think you just need `Tag()`, maybe a `New` which can be used directly rather than trying to go via `NewFromCLI` would be useful, might need some refactoring though.

w9n:
Yes `New` in pkglib.Pkg would be awesome - but i think this needs some more work to build old hashes. [Here](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) are some good information. It would basically require `git cat-file -p hash` and then recursivly load the tree into a tmp dir for build. `src-d/go-git` has a nicer interface for that.

ijc:
Where/why does this code need to deal with old hashes?

Generating the hash corresponding to a package in the an old revision (by tag or commit ID) is already supported by the pkglib code.

w9n:
if there is
```
tags:
  - name: <2.1>
    hash: bb0c6ae2f12a1b55df24ebce2067784a931285df
```
i would like to be able to build the pkg without finding out in which commit the hash is and checking it out. `generate` should not be in pkglib but needs a public interface to do all of its hash stuff and pkglib needs some refactor to be usable without cli. So im thinking if its the right moment to implement `src-d/go-git` along with it.

rn:
Can we address one issue at a time. templating of the YAML file is one thing and improving/changing package builds is another. Let's sort the templating one first.

ijc:
You can already get the package hash from a git tag today already, it's currently exposed as the `hash-commit` in `NewFromCLI`.

Although I'm not really sure I follow your use case, I would expect that if the git repo was tagged X.Y there would be a image on hub tagged as `linuxkit/foo:X.Y` so you don't need the indirection you are adding here in the common case. `linxukit pkg push` should push these automatically if the repo is tagged.

I'm also confused by your reference to figuring out the commit and building it since that (specifically building) should be out of scope here I think.

If you give:
```
tags:
  - name: <2.1>
    hash: bb0c6ae2f12a1b55df24ebce2067784a931285df
```
are you expecting to do something more complex than blindly replacing `<2.1>` with `bb0c6ae2f12a1b55df24ebce2067784a931285df`? If so then that's not what I understood from the most recent summary of what this PR was doing https://github.com/linuxkit/linuxkit/pull/2648#issuecomment-355530363.

This PR sorely need to be also adding some documentation for the feature(s) it is adding, the PR comments are now covering too many alternatives and possibilities I have no idea what it is actually intending to do any more. Please add in-tree documentation as a priority so we have a shared understanding of the (current) intention here.

w9n:
I made `pkgInfo` public and moved it into the config namespace I put GlobalConfig in because `generate` needs to read the tags in the config before deciding which commitHash it is using in pkglib. Also created a constructor in `PkgInfo`, so that it doesnt need to be parsed twice.

This PR should be straightforward templating of `<tags>` in moby files onto tags in `build.yml` and having `<latest>` (could split this into its own commit if you prefer). Sorry for the confusion will tackle the other ideas by time :) generate is named `linuxkit bake` now. I will write some docs in the next days.

ijc:
> I made pkgInfo public 

I don't like this, sorry. This struct should remain internal to pkglib, with the `Pkg` struct/object remaining the public interface.

AFAICT where you currently have
```
+		b, err := ioutil.ReadFile(filepath.Join(repo, pkg, pkglib.PkgConfFile))
[...]
+		var pi config.PkgInfo
+		if err := yaml.Unmarshal(b, &pi); err != nil {
````
In order to pass `pi` to `substitute` and from there to `pkglib.New` you can instead pass the path `filepath.Join(repo, pkg)` all the way up that call chain and have `New` to the read and unmarshall. To avoid the code duplication between `New` and `NewFromCLI` (which I think might have been your motivation) you can instead have an _internal_ helper which both call (`newFromInfo` perhaps).

**Comment by w9n** [comment]
@ijc I did what you suggested and exposed a `SetCommit` `GetHash` and `Image` function in pkglib for the template package.

**Comment by ijc** [comment]
You shouldn't need `SetCommit` since the only caller outside `pkglib` passes a hardcoded `"HEAD"` and this should be the default in al code paths which don't involve `NewFromCLI`.

Please leave `NewFromCommit` in `pkglib.go` it doesn't require a separate file and moving it makes the review harder.

NB I am holding off doing an in depth review here until the feature is documented so I know the expected behaviour, currently I'm just dipping into things you are mentioning in comments. I'd strongly recommend documenting ASAP before working on the code further, otherwise it is hard to say if you are moving in the right direction at all.

**Comment by w9n** [comment]
done

File: README.md [hunk 1]
```diff
@@ -56,6 +56,21 @@ output a raw BIOS bootable disk image, or `linuxkit build -format iso-efi linuxk
 
 Since `linuxkit build` is built around the [Moby tool](https://github.com/moby/tool) the input yml files are described in the [Moby tool documentation](https://github.com/moby/tool/blob/master/docs/yaml.md).
 
+
+### Baking images
+```
+linuxkit bake linuxkit_template.yml
+```
+This will replace all angle bracketed tags in the image field by its hash defined in `build.yml`. If the tag is `<latest>` it gets replaced by the current pkg directorys git subtree hash. The location of the pkg directory is either defined by parameter `--pkgroot org:path` or in the global config, e.g.:
```
ijc: Please document how tags other than `<latest>` are handled, or explicitly state that `<latest>` is the only supported tag today if that is the case.

README.md
```diff
@@ -56,6 +56,21 @@ output a raw BIOS bootable disk image, or `linuxkit build -format iso-efi linuxk
...
+
+```
+repos:
+  - path: /go/src/github.com/linuxkit/linuxkit/pkg
+    org: linuxkit
+  - path: /go/src/github.com/kubernetes/kubernetes/pkg
+    org: linuxkit
```
ijc: Please define the meaning of the fields. I guess that `path` is searched for subdirectories (nested? or one level?) which might be packages (based on them containing a `build.yml`?).

I don't understand the need for `org` here at all, and packages under `path` will have an org as part of their `build.yml` (or will use the default). Perhaps it is an optimisation to avoid searching that path if the org doesn't match? That seems like a premature optimisation to me, but if this field is to exist (and I've inferred the reason correctly) it should accept the possibility that a given repo might have packages targetting multiple orgs, and hence be `orgs` and an array.

README.md
w9n: you put e.g. this into the kubernetes `build.yml` and use `<1.9>` as tag.
```
tags:
    - name: <1.9>
      hash: some_hash
```
I think this is more of a maintainer thing then for the front page.

ijc: Whether it is for maintainers or anyone else it needs documenting, if not here then in `docs/packaging.md`.

I had not appreciated so that that this PR to have anything to do with `build.yml`. Looking back over the comments this does seem to be have been mentioned in the past and I missed it, I was working from https://github.com/linuxkit/linuxkit/pull/2648#issuecomment-355530363 onwards and it wasn't mentioned after that.

As I've mentioned before for that use case we would push tagged images to hub. i.e. rather than associating `<1.9>` with a `linuxkit/pkg:hash` in `build.yml` maintainers would just push `linuxkit/pkg:1.9` to hub.

I thought the examples of `tags` given in various comments were for inputs passed into `linuxkit build` directly (i.e. actual arguments) not to be things which it would go off and found inside a repo somewhere.

README.md
w9n: @rn suggested it. I dont think performance is relevant soon so dont care to much. I can remove, leave it, or make it an array.

updated the docs.

README.md
w9n: i asked that [here](https://github.com/linuxkit/linuxkit/pull/2648#issuecomment-355347992). I still would like to keep them as they could be very interesting with #2817 and variable substitution of parameters.

README.md
rn: yes, I did suggest the `org` part, but that was because I forgot that `build.yml` contains the information. My initial intention was not to use `build.yml` but it does make sense since people may have directory name != package name.

README.md
ijc: I am not convinced that there is a use case of having the tag mapping in `build.yml`, while I could see an argument for having such a mapping elsewhere as an input to `linuxkit build` (which seems to be more what #2817 is about).

In terms of the code and documentation which is in _this_ PR it needs to be self consistent and in particular the documentation completely describe the state of the functionality as it exists after this PR, not how it might look after some future PR.

README.md
ijc: lets remove it for now.

README.md
w9n: done.

README.md
w9n: done.

linuxkit_template.yml
```diff
@@ -0,0 +1,61 @@
+kernel:
+  image: linuxkit/kernel:4.9.75
```
rn: I would prefer this not to be in the top-level. Maybe `./examples` would be a better place. Also having a test which `bake`s the template and then builds it would be useful.

w9n: done, lets see what ci says

w9n:
so i rethought this and the whole templating problem might just not fit into linuxkit. Nevertheless its very tedious to maintain hashes in moby configs somewhere upstream (I thinks its pretty noisy in this repo, too). So I tried to make this PR more statically. I will update my desktopless project this weekend and will use this solution from somewhere, tho i wouldnt be mad if you dont want to merge it here.

ijc:
I don't think there are objections to templating in general, my specific objection currently is that tags should not be defined in `pkg/*/build./yml`, that use case is satisfied by git tags and pushing the actual corresponding tags to hub (or tagging things in the local docker image repo for dev).

The `<latest>` thing seems clearly useful and I could also see the utility in being able to do things like:
```
linuxkit bake -D foo=abcdef foo.yml
```
(substitutes `org/pkg:<foo>` into `org/pkg:abcdef`)
or
```
linuxkit bake foo.yml vars.yml
```
where `vars.yml` includes the definitions in various ways for example something like the `tags` you are currently adding to `build.yml` (this is what I was expecting you were actually implementing here).

If this PR was just laying the ground work while implementing only `<latest>` I think it could be beaten into shape, the other ideas I mention above are subject to a bit more discussion.

w9n:
If git tags or parameter like `foo=abcdef` are used than this would require e.g. v1.0 to be valid for all pkgs in a repo. Why not reversing the relationship into 1 pkg has X release tags instead?

ijc:
We intend to release linuxkit as a whole, with a single consistent set of packages, so the git tagging does (and should) apply to all packages in the repo.

If you want to version packages independently then separate repos would seem to be the right answer, although you could also push tags to hub without tagging in git if you wanted to tag independently. I think `linuxkit pkg push --release v1.1` will do that, but it's not something which gets tested so it might need fixes.

You could also maintain a `versions.yml` with tags (to be passed to `linuxkit bake`) inside the repository but outside the build directory (so editing it doesn't change the package hashes). There is certainly scope for enhancing your proposed `tags` datastructure to allow per-package tagging too.

rn:
I agree with @ijc that we do want some templating. We had several discussions in the past, like https://github.com/moby/tool/issues/101 and https://github.com/moby/tool/pull/96. I like the explicit templating step `bake` now uses instead of substituting on the fly. So I think we agree on that portion and it would be nice to have this available.

Do you have an example for a baked YAML file. It would be useful if it would contain comments at the top stating that the file was generated, and ideally also the `HEAD` of the package repositories used to generate it. I would check in the generated file and this info would be useful.

As for version tags, for core LinuxKit packages, I think version tags on hub (like say `0.2`) should always correspond to git tags and we will release all the packages on LinuxKit as a whole, as @ijc said. As for non core LinuxKit packages, You can tag them in whatever way you like using `-hash` and/or `-release`. So if you want to version your packages that way, you can do that. I'm not sure we then want to support this with the templating mechanism. In any case this would be a separate functionality and ideally won't be rolled into the PR implementing the templating itself.

justincormack:
I don't understand what the output of `linuxkit bake example/linuxkit_template.yml` is. Does it output another file, modify the existing file or do a build?

ijc:
It's a new file which you can pass to `linuxkit build` (after archiving it for reproducibility purposes etc).

justincormack:
Does it output to stdout or what? This needs to be documented properly.

w9n:
Okay, so I could live with this:
```
tags:
    - name: v1.0
      init: some_hash
      dhcp: some_hash
    - name: v2.0
      init: some_other_hash
      dhcp: another_hash
```
Its basically the same information as if the tags would be in build.yml just at a different location (Not sure if its the better position). The advantage is that you can e.g. hotfix v1.0 or what else for one pkg without influencing everything else.

@justincormack yes its writing to stdout and logging to stderr but i could also give it a `-o` multiflag if wished.

rn:
@w9n please read my comment https://github.com/linuxkit/linuxkit/pull/2648#issuecomment-356576666. Could we initially focus on the templating stuff itself (ie replacing `<latest>` with the hash of the package). It seem separate from the version tags and both @ijc and myself are not sure if we should handle the version tags that way, ie by extending `build.yml`

w9n:
> Do you have an example for a baked YAML file. It would be useful if it would contain comments at the top stating that the file was generated, and ideally also the HEAD of the package repositories used to generate it. I would check in the generated file and this info would be useful.

should be no problem to prefix `###----------- AUTO-GENERATED--------###`  to the output. Here is the output of `linuxkit_template.yml`. I dont think i could easily print a comment above an image field in a moby structure when marshalling the yml.
```
template: linuxkit/ca-certificates:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/ca-certificates
  hashed: linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
--------------------------
template: linuxkit/containerd:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/containerd
  hashed: linuxkit/containerd:e58a382c33bb509ba3e0e8170dfaa5a100504c5b
--------------------------
template: linuxkit/dhcpcd:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/dhcpcd
  hashed: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
--------------------------
template: linuxkit/getty:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/getty
  hashed: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
--------------------------
template: linuxkit/init:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/init
  hashed: linuxkit/init:5a577d070817b4f17821657823082651baafd4ed
--------------------------
template: linuxkit/rngd:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/rngd
  hashed: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
--------------------------
template: linuxkit/runc:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/runc
  hashed: linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
--------------------------
template: linuxkit/sysctl:<latest>
    path: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/sysctl
  hashed: linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21
--------------------------
kernel:
  image: linuxkit/kernel:4.9.75
  cmdline: console=tty0 console=ttyS0 console=ttyAMA0
init:
- linuxkit/init:5a577d070817b4f17821657823082651baafd4ed
- linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
- linuxkit/containerd:e58a382c33bb509ba3e0e8170dfaa5a100504c5b
- linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
onboot:
- name: sysctl
  image: linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21
- name: dhcpcd
  image: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
  command:
  - /sbin/dhcpcd
  - --nobackground
  - -f
  - /dhcpcd.conf
  - "-1"
onshutdown:
- name: shutdown
  image: busybox:latest
  command:
  - /bin/echo
  - so long and thanks for all the fish
services:
- name: getty
  image: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
  env:
  - INSECURE=true
- name: rngd
  image: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
- name: nginx
  image: nginx:alpine
  capabilities:
  - CAP_NET_BIND_SERVICE
  - CAP_CHOWN
  - CAP_SETUID
  - CAP_SETGID
  - CAP_DAC_OVERRIDE
  binds:
  - /etc/resolv.conf:/etc/resolv.conf
trust:
  org:
  - linuxkit
  - library
files:
- path: etc/containerd/config.toml
  directory: false
  contents: |
    state = "/run/containerd"
    root = "/var/lib/containerd"
    snapshotter = "io.containerd.snapshotter.v1.overlayfs"
    differ = "io.containerd.differ.v1.base-diff"
    subreaper = false

    [grpc]
    uid = 0
    gid = 0

    [debug]
    level = "info"

    [metrics]
    address = ":13337"
  optional: false
- path: etc/linuxkit-config
  directory: false
  metadata: yaml
  optional: false
```

w9n:
i dont really trust the test time, something must be wrong...
```
[+ 24m 58s] [PASS    ] linuxkit.build.examples.bake 0.04s
```

w9n:
@ijc 
```
linuxkit bake -D foo=abcdef foo.yml
```
This can actually be done with `envsubst` right now and also be extended with `bake`

w9n:
updated the last comment, otherwise i guess i should revert setting global config in its own namespace and pass the repos as strings into `Bake` for better encapsulation. Let me know if you would like one of the proposed solutions in this PR. I could implement it tonight or leave it.

w9n:
Sorry for all these changes
 - dropped putting GlobalConfig in its own namespace
 - squashed the git commit stuff
 - moved logging from Template to bake cli and writing it as comment into the result
 - add `Time.Now()` to log comment
 - gave Template a more defined surface based on a `Template`, `Result` and `Substitution` structures. `Template.Bake() (Result,error)` is the only public function. 

here is a result yml:

<details>

```

###-------------------------------AUTO-GENERATED-------------------------------###
#     time: 2018-01-12 08:28:29.070212358 +0100 CET m=+0.383653088
#-----------------------------------------------------------------------------------
# template: linuxkit/ca-certificates:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/ca-certificates
#   result: linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
#-----------------------------------------------------------------------------------
# template: linuxkit/containerd:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/containerd
#   result: linuxkit/containerd:e58a382c33bb509ba3e0e8170dfaa5a100504c5b
#-----------------------------------------------------------------------------------
# template: linuxkit/dhcpcd:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/dhcpcd
#   result: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
#-----------------------------------------------------------------------------------
# template: linuxkit/getty:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/getty
#   result: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
#-----------------------------------------------------------------------------------
# template: linuxkit/init:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/init
#   result: linuxkit/init:5a577d070817b4f17821657823082651baafd4ed
#-----------------------------------------------------------------------------------
# template: linuxkit/rngd:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/rngd
#   result: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
#-----------------------------------------------------------------------------------
# template: linuxkit/runc:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/runc
#   result: linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
#-----------------------------------------------------------------------------------
# template: linuxkit/sysctl:<latest>
#   source: /home/user/go/src/github.com/linuxkit/linuxkit/pkg/sysctl
#   result: linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21
#-----------------------------------------------------------------------------------
kernel:
  image: linuxkit/kernel:4.9.75
  cmdline: console=tty0 console=ttyS0 console=ttyAMA0
init:
- linuxkit/init:5a577d070817b4f17821657823082651baafd4ed
- linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
- linuxkit/containerd:e58a382c33bb509ba3e0e8170dfaa5a100504c5b
- linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
onboot:
- name: sysctl
  image: linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21
- name: dhcpcd
  image: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
  command:
  - /sbin/dhcpcd
  - --nobackground
  - -f
  - /dhcpcd.conf
  - "-1"
onshutdown:
- name: shutdown
  image: busybox:latest
  command:
  - /bin/echo
  - so long and thanks for all the fish
services:
- name: getty
  image: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
  env:
  - INSECURE=true
- name: rngd
  image: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
- name: nginx
  image: nginx:alpine
  capabilities:
  - CAP_NET_BIND_SERVICE
  - CAP_CHOWN
  - CAP_SETUID
  - CAP_SETGID
  - CAP_DAC_OVERRIDE
  binds:
  - /etc/resolv.conf:/etc/resolv.conf
trust:
  org:
  - linuxkit
  - library
files:
- path: etc/containerd/config.toml
  directory: false
  contents: |
    state = "/run/containerd"
    root = "/var/lib/containerd"
    snapshotter = "io.containerd.snapshotter.v1.overlayfs"
    differ = "io.containerd.differ.v1.base-diff"
    subreaper = false

    [grpc]
    uid = 0
    gid = 0

    [debug]
    level = "info"

    [metrics]
    address = ":13337"
  optional: false
- path: etc/linuxkit-config
  directory: false
  metadata: yaml
  optional: false

```

</details>


It should be easier to review and extend now. I didnt implement 1:1 substitution because its not needed yet and pipes can be used.

w9n:
```
[+ 24m 27s] [PASS    ] linuxkit.build.examples.bake 32.57s
```

fixed test. Should cover everything but global config parsing.

PR #2649, opened 2017-10-28
Kubernetes: Autoinit and persistent state by default

w9n:
Autoinit does not harm after initial bootstrap and state should not be deleted by default.

### DIFF
--- a/projects/kubernetes/README.md
+++ b/projects/kubernetes/README.md
@@ -18,6 +18,10 @@ Boot Kubernetes master OS image using `hyperkit` on macOS: or `qemu` on Linux:
 ```
 ./boot.sh
 ```
+or, to automatically initialise the cluster upon boot with no additional options
+```
+KUBE_MASTER_AUTOINIT="" ./boot.sh
+```
 
 Get IP address of the master:
 ```
@@ -29,7 +33,7 @@ Login to the kubelet container:
 ./ssh_into_kubelet.sh <master-ip>
 ```
 
-Manually initialise master with `kubeadm`:
+Manually initialise master with `kubeadm` if booted without `KUBE_MASTER_AUTOINIT`:
 ```
 kubeadm-init.sh
 ```
--- a/projects/kubernetes/boot.sh
+++ b/projects/kubernetes/boot.sh
@@ -15,7 +15,7 @@ set -e
 : ${KUBE_RUN_ARGS:=}
 : ${KUBE_EFI:=}
 : ${KUBE_MAC:=}
-: ${KUBE_PRESERVE_STATE:=}
+: ${KUBE_CLEAR_STATE:=}
 
 [ "$(uname -s)" = "Darwin" ] && KUBE_EFI=1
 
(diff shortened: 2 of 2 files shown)


justincormack:
cc @ijc

projects/kubernetes/boot.sh
```diff
@@ -15,7 +15,8 @@ set -e
 : ${KUBE_RUN_ARGS:=}
 : ${KUBE_EFI:=}
 : ${KUBE_MAC:=}
-: ${KUBE_PRESERVE_STATE:=}
+: ${KUBE_PRESERVE_STATE:=true}
```
ijc: With this change how can I now disable this? (I don't think I can). I think it would be best to invert the name and leaving it defaulting to blank, i.e. `KUBE_RESET_STATE`,  `KUBE_NUKE_STATE` or `KUBE_CLEAR_STATE` or somesuch.

projects/kubernetes/boot.sh
```diff
@@ -15,7 +15,8 @@ set -e
...
+: ${KUBE_MASTER_AUTOINIT:=""}
```
ijc: I think this now cannot be unset by the user either, not sure if this one can be sensibly inverted though, might need to split into two options, one to disable auto init and another to allow the user to provide the options to use for autoinit.

projects/kubernetes/boot.sh
w9n: done

projects/kubernetes/boot.sh
w9n: i just put the option into the README.

projects/kubernetes/README.md
```diff
@@ -29,7 +33,7 @@ Login to the kubelet container:
 ./ssh_into_kubelet.sh <master-ip>
 ```
 
-Manually initialise master with `kubeadm`:
+Manually initialise master with `kubeadm` if bootet without `KUBE_MASTER_AUTOINIT`:
```
ijc: typo: `bootet`

projects/kubernetes/boot.sh
```diff
@@ -73,7 +73,7 @@ else
     exit 1
 fi
 set -x
-if [ -z "${KUBE_PRESERVE_STATE}" ] ; then
+if [ "${KUBE_CLEAR_STATE}" == "true" ] ; then
```
ijc: Please use `[ -n "${KUBE_CLEAR_STATE}" ]` so e.g. `KUBE_CLEAR_STATE=y` (which is what I do) works.

w9n: sorry... should be fine now!

projects/kubernetes/README.md
```diff
@@ -18,6 +18,10 @@ Boot Kubernetes master OS image using `hyperkit` on macOS: or `qemu` on Linux:
 ```
 ./boot.sh
 ```
+or
```
ijc: This should give some sort of indication as to what the following alternative does. Something like `or, to automatically initialise the cluster upon boot with no additional options` would be enough.

projects/kubernetes/boot.sh
```diff
@@ -58,7 +58,11 @@ elif [ $# -gt 1 ] || [ $# -eq 1 -a -n "${KUBE_PRESERVE_STATE}" ] ; then
     img="kube-node"
     name="node-${1}"
     shift
-    data="{\"kubeadm\": {\"join\": \"${*}\"} }"
+
+    if [ $# -ge 1 ] ; then
+    	data="{\"kubeadm\": {\"join\": \"${*}\"} }"
+    fi
```
ijc: Took me a moment to get my head around the inversion of the logic here, but I think it's correct.

ijc:

w9n:
done, i tried ```    data="{\"kubeadm\": {\"join\": \"--node-name ${name} ${*}\"} }" ``` but it just ignored the node name parameter. Would be nice to specify better host names.

I also think it would be a good point to move to infrakit from here since linuxkit is not designed to manage the lifetime of a cluster.

justincormack:
@w9n yeah the infrakit integration needs some updating and integration but the basics are there to work on.

PR #2650, opened 2017-10-28
Support static ppc64le executables in qemu

oflebbe:
The /sbin/ldconfig.real command in ubuntu for ppc64le cannot be executed
in docker images with the help of qemu since the binfmt-misc configuration does not match the
executable file format of static executables. This patch is a port of
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=794737 to linuxkit

Signed-off-by: Olaf Flebbe <of@oflebbe.de>

**- What I did**
I run on docker for mac on a ppc64le ubuntu image. Running ldconfig.real traps an error:

```
docker run --rm -it ppc64le/ubuntu:16.04   /sbin/ldconfig.real
standard_init_linux.go:195: exec user process caused "exec format error"

```
**- How I did it**

I edited the /Applications/Docker.app/Contents/Resources/linuxkit/docker-for-mac.iso 
and applied the patch manually with emacs

**- How to verify it**
started docker for mac again and run
`docker run --rm -it ppc64le/ubuntu:16.04   /sbin/ldconfig.real`
-- no output -- ok.

Support static ppc64le executables in qemu

images


justincormack:
Thanks, yes this was a port from the Debian values.

We need to rebuild the package as well.

justincormack:
I will merge this and do a package update in a new PR.

PR #2653, opened 2017-10-29
tools: Use alpine multi-arch image for linuxkit/alpine

rn:
https://github.com/docker-library/official-images/pull/3623
introduced alpine multi-arch images. Use it instead of
separate base images.

dog-wind


justincormack:
finally!

PR #2661, opened 2017-10-31
Fix KB calculation for raw images

deitch:

**- What I did**
Fix the raw image calculation. Because of the shell's integer calc, the image size in KB sometimes would return a size smaller than it should be. This ensures it always rounds **up** to the nearest whole KB, and then again **up** to the nearest multiple of 1024KB. Fixed in raw-efi and raw-bios.

**- How I did it**
Changed a line in each mkimage.

**- How to verify it**
Build the image (or not, since I already pushed with latest hash), run it. Easiest is to run `moby build -format tar-kernel-initrd ...` and then `cat <output>.tar | docker run -i --rm -e DEBUG=true <image> > output.img`

For `<image>`, the following are available as pushed:

* `linuxkit/mkimage-raw-bios:eac4dcb78f837618e7e009060146977c2adabe19`
* `linuxkit/mkimage-raw-efi:f2bae7a502598d6b78c2f0b6de77f207df27add7`

Improve size calculation for raw-bios and raw-efi

ferret


deitch:
Thanks @rn. Lots of rows in an Excel sheet comparing various sizes and calc. :-)

I will merge as soon as CI passes.

PR #2667, opened 2017-11-02
Kubernetes: Bump to 1.8.2 and Weave 2.0.5

ijc:
Also renamed a Makefile variable for consistency with all the others.

When I first did the builds the image cache pushes seemed to work but only pushed the `-amd64` suffixed mages and not the manifest, which seems to be something which has happened before. In the course of investigating it seemed to "just work" (the manifests) so I don't know what is going on there, I'll keep looking into that one.


ijc:
Thanks. I've kicked travis to have another go.

ijc:
@riyazdf. Thanks for the review.

The kube images are all x86_64 only:
```
$ tail -n2 kubernetes/build.yml
arches:
  - amd64
```

riyazdf:
ah ok!  I misunderstood your original comment then.  This makes sense 👍

PR #2668, opened 2017-11-02
Kubernetes e2e

errordeveloper:

Add Kubernetes end-to-end test runner

baby giraffe outside-0010-2239_1492202848627_9500116_ver1 0_640_360


projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
+ORG?=docker
```
ijc: Should be `linuxkitprojects` I think.

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
...
+IMAGE=kubernetes-for-desktop-e2e
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+
+kubernetes_version=v1.8.0
```
ijc: Does this need to remain in sync with `kubernetes/Dockerfile`?

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
+ORG?=docker
+IMAGE=kubernetes-for-desktop-e2e
```
ijc: Is it specific to `desktop` though? Couldn't it just as well run on a LinuxKit VM running on a Linux host?

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
...
+build: _cache/kubernetes-test.tar.gz _cache/kubectl
+	docker build --network=none -t $(ORG)/$(IMAGE):$(HASH) .
+
+.PHONY: run
+run: build
+	docker run -ti -v /var/lib/kubeadm:/etc/kubernetes $(ORG)/$(IMAGE):$(HASH)
+
+_cache/kubectl:
+	mkdir -p _cache
+	curl -fSL -o $@ https://dl.k8s.io/$(kubernetes_version)/bin/linux/amd64/kubectl
```
ijc: Does d4m at least not already supply a `kubectl`?

projects/kubernetes/e2e/e2e.sh
```diff
@@ -0,0 +1,19 @@
...
+exec /usr/bin/ginkgo \
+  -nodes=4 \
+  -progress \
+  -flakeAttempts=2 \
+  -skip='\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[HPA\]|Dashboard|Services.*functioning.*NodePort|.*NFS.*|.*Volume.*|\[sig-storage\]|.*StatefulSet.*|should\ proxy\ to\ cadvisor\ using\ proxy\ subresource' \
+  /usr/bin/e2e.test -- \
+    -provider=local \
+    -num-nodes=1 \
+    -kubeconfig=/etc/kubernetes/admin.conf \
+    -host=https://192.168.65.2:6443 \
```
ijc: Ah, I guess this IP is a dependency on desktop.

projects/kubernetes/e2e/Makefile
ijc: Or perhaps since this is not really a linuxkit "package" but rather a service container of sorts maybe it belongs somewhere else, not sure.

projects/kubernetes/e2e/e2e.sh
justincormack: Yeah it would be nice to make this not just work on desktop, but we can always do later potentially.

errordeveloper:
> Perhaps the image is already suitable to be run as a service so it would be possible to construct a yml file which could be passed to `moby build` to produce a VM which when run would execute the tests and report to the console?

Perhaps, although it would have to wait for control plane to start, which isn't too hard. However, alternative approach would be to run it as a pod. Up until now I was using`docker run`, which is was just because I was lazy.

> How long do these tests take to run?

I depends, last time on my desktop it ran for 2h+, but we can tweak the subset...

> If it isn't too long we could perhaps even consider adding them to CI?

Maybe something to discuss later. Kubernetes CI runs various [subsets of the e2e suite](https://github.com/kubernetes/test-infra/blob/master/jobs/config.json).

errordeveloper:
> Would be nice to have some basic doc on how to use...

I'll add a readme and document how I've been using it so far on desktop.

> ...and where to look for information about the test cases etc.

Good question! I think the answer is to grep the Kubernetes repo, I'm happy to document it as such.

projects/kubernetes/e2e/Makefile
errordeveloper: Yes, but we need it inside the image, as we have namespace cleanup script and `e2e.test` shells out to it too.

projects/kubernetes/e2e/Makefile
errordeveloper: Yes, sounds like it's better to keep the versions in sync. Right now this uses release binaries, but if one builds from source it'd be ideal if freshly built artefacts get used here too.

projects/kubernetes/e2e/Makefile
errordeveloper: Practically, it's not, apart from the hardcoded IP.

projects/kubernetes/e2e/Makefile
errordeveloper: Right now this image doesn't get published, not sure if it will need to be published at some point.

errordeveloper:
I feel somewhat uncomfortable with this subtree approach. Justin suggested
to cherry-pick, but I couldn't find a way to do it... Anyway, I would
rather squash this, and the other PR.

errordeveloper:
@ijc I've squashed and add basic documentation.

ijc:
> run it as a pod

Now that you mention it -- might that be something worth contributing to kube somewhere rather than LinuxKit and just have some machinery for running it (i.e. a yml with the appropraite addition to `/etc/kubeadm/kube-system.init/NN-e2e.conf` in place) here?

errordeveloper:
> might that be something worth contributing to kube somewhere

Sounds like a good idea, but it's complicated.

There are already many ways to run the e2es, e.g. github.com/heptio/sonobuoy.
The upstream effort is mostly around https://github.com/kubernetes/test-infra/tree/master/kubetest, which is complicated (for good reasons, of course), it knows of ways to provision clusters and then calls `hack/run-e2e.sh` in main Kubernetes repo (so using it involves reading Go program that calls a shell scritp that calls the test framework).
So I've just gone the simplest approach really and run test framework directly.
I'm sure the sig-testing folks would appreciate contributions, but I cannot see how we could land this thing right away.

errordeveloper:
Alternatively, we can look into using heptio/sonobuoy, but it just felt a little complicated as I was trying to wrap my head (again!) around this stuff.

ijc:
I much prefer squashing to sub-tree stuff (in fact I'd prefer just about anything to using subtree in this case ;-)).

AIUI heptio/sonobuoy are (partially?) a hosted service thing, I don't think that meets our needs here.

> I'm sure the sig-testing folks would appreciate contributions, but I cannot see how we could land this thing right away.

Might be worth seeing what they think?

ijc:

errordeveloper:
> AIUI heptio/sonobuoy are (partially?) a hosted service thing, I don't think that meets our needs here.

Yes, that was my other concern.

errordeveloper:
> Might be worth seeing what they think?

Yes, but I'd first like to make sure we have a thing in place that we can run in CI nightly or even just manually and identify more clearly the types of tests that fail and have some basic way of tracking progress towards reasonable level of conformance.

errordeveloper:
Rebased to fix Travis.

w9n:
From what i understand does Sonobuoy pretty much the same thing (https://github.com/heptio/kube-conformance) and has a plugin architecture to e.g. implement systemd logs aggregation.
 Here is a an example to run it as a pod in a cluster (https://github.com/heptio/sonobuoy/blob/master/examples/quickstart.yaml). No hosted service from what i have seen. Its on my list but didnt have time to properly check it out yet.

w9n:
ok, I have found that creepy frontend that allows optional remote execution through a heptio server... 

Sonobuoy as a Framework seems nice and also creates a [snapshot](https://github.com/heptio/sonobuoy/blob/master/docs/snapshot.md) of all kubernetes objects, but i dont think this is necessary for ci. The e2e testing plugin does pretty much exactly the same as this container.

errordeveloper:
> The e2e testing plugin does pretty much exactly the same as this container.

Yes. Maybe we could borrow it, but I don't know how much value there is vs us having full control over how it runs.

> Skips other tests `GINKOGO_SKIP="Alpha|Disruptive|Feature|Flaky|Kubectl"`

Is it what sonobuoy does or a suggestion of what we should skip?

Currently, what's on the skip list is up for a debate.

w9n:
Thats what they skip by [default](https://github.com/heptio/kube-conformance/blob/master/Dockerfile#L32). They also set ```ENV E2E_FOCUS="Conformance"``` by default which only runs a schedule one pod Test. This needs to be overriden. Im running the test for the last 1 hour+. I can share the result but its not a fresh cluster (trying to debug some problems...).

w9n:
The variables end up [here](https://github.com/heptio/kube-conformance/blob/master/run_e2e.sh#L18) in the script but are mostly what the names intend.

w9n:
Im fine with this solution, control is better.

errordeveloper:
@ijc ping

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
...
+IMAGE=kubernetes-e2e
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+
+kubernetes_version=v1.8.0
+
+kube_release_artefacts=https://dl.k8s.io/$(kubernetes_version)/bin/linux/amd64
+
+.PHONY: build
+build: _cache/kubernetes-test.tar.gz _cache/kubectl
+	docker build --network=none -t $(ORG)/$(IMAGE):$(HASH) .
```
ijc: Please can we switch the actual build phase over to the `linuxkit pkg build` scheme. I appreciate this will still need wrapping in a `Makefile` to fetch deps etc.

I thought I'd mentioned this before but clearly I forgot, sorry.

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
+ORG?=linuxkitprojects
+IMAGE=kubernetes-e2e
+HASH?=$(shell git ls-tree HEAD -- ../$(notdir $(CURDIR)) | awk '{print $$3}')
+
+kubernetes_version=v1.8.0
```
ijc: v1.8.3 now?

projects/kubernetes/e2e/Makefile
```diff
@@ -0,0 +1,29 @@
...
+local: build
+	docker run -ti -v /var/lib/kubeadm:/etc/kubernetes $(ORG)/$(IMAGE):$(HASH)
+
+_cache/kubectl:
+	mkdir -p _cache
+	curl -fSL -o $@ https://dl.k8s.io/$(kubernetes_version)/bin/linux/amd64/kubectl
+
+_cache/kubernetes-test.tar.gz:
+	mkdir -p _cache
+	curl -fSL -o $@ https://dl.k8s.io/$(kubernetes_version)/kubernetes-test.tar.gz
```
ijc: Should we be checking a sha256 or something for this (and `kubectl` for that matter)?

projects/kubernetes/e2e/README.md
```diff
@@ -0,0 +1,27 @@
+# Kubernetes end-to-end test suite (e2e)
+
+In this subdirectory you can build Kubernetes e2e image, that you can
+use with `docker run` on Desktop or bundle as a job to run on a cluster
+elsewhere.
```
ijc: Can you give an example of how to achieve the second case please.

projects/kubernetes/e2e/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Building the image
+
+```
+make build HASH=current
+```
```
ijc: Why recommend overriding the hash in this way rather than using the usual hash?

projects/kubernetes/e2e/e2e.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# cleanup resources created by previous runs
+kubectl get namespaces \
+  --output="jsonpath={range .items[?(.status.phase == \"Active\")]}{.metadata.name}{\"\n\"}{end}" \
+  | grep '^e2e-.*' \
+  | xargs -r kubectl delete namespaces
+
+skip="${GINKOGO_SKIP:-'\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[HPA\]|Dashboard|Services.*functioning.*NodePort|.*NFS.*|.*Volume.*|\[sig-storage\]|.*StatefulSet.*|should\ proxy\ to\ cadvisor\ using\ proxy\ subresource'}"
+nodes="${GINKOGO_NODES:-"4"}" \
+flakeAttempts="${GINKOGO_FLAKE_ATTEMPTS:-"2"}" \
```
ijc: A bunch of the uses of `\` in this file seem odd to me, why are they needed? (I mean the ones on `nodes=`, `flakeAttempts=`, `provide=` and `host=`, I understand the ones for the cleanup and exec bits).

Also the quoting in constructs such as `"${GINKOGO_FLAKE_ATTEMPTS:-"2"}"` is a bit odd (the effect is that `2` is unquoted I think). I'd suggest dropping the inner quotes.

justincormack:
Sorry can you reopen this on https://github.com/linuxkit/kubernetes

PR #2670, opened 2017-11-02
Kubernetes vpnkit-expose-port-controller

errordeveloper:

Add `vpnkit-expose-port-controller`

baby_giraffe_leaps_for_joy_at_fota_wildlife_park


ijc:
4,670 files -- I presume that is mostly vendoring? Is this the minimal set required or could we look at a better tool (`vndr` perhaps)?

errordeveloper:
Yes, that's mostly what `k8s.io/client-go` brings with it. I've had trouble making things work with `vndr`, but I think it may be still possible. I also think it may be possible to make `dep` trim what we have here.

errordeveloper:
@djs55 you might want to review 7e0e69e.

errordeveloper:
@ijc I've managed to get rid of some unused code in `vendor/` by using `dep prune` see 564b082. There is initial commit where `vendor/` is added, but it's a little bit tedious to clean this stuff up while preserving history.

errordeveloper:
Probably better to squash, this subtree approach doesn't seem right. I can
also do the vendor dir in a separate commit easily that way.

ijc:
squashing is probably easiest. With `git filter-branch` might produce a better output but the complexity is likely not worth it. I'm pretty sure `git subtree` is not the answer unless we are intending to keep the old upstream active and just merge from that -- but I don't think you are proposing that and don't think we want that.

I was thinking overnight -- perhaps it would be better to put this in the vpnkit repo since it is in effect a clientg/glue of that to kube (rather than something particular to LinuxKit as such) it is perhaps better to keep them together.

errordeveloper:
> ...perhaps it would be better to put this in the vpnkit repo since it is in effect a clientg/glue of that to kube (rather than something particular to LinuxKit as such) it is perhaps better to keep them together.

@ljc sounds like a good idea, and then we might even decide to call vpnkit as a library instead of spawning children and managing them.

justincormack:
It could go in vpnkit, although was thinking of moving linuxkit/projects/kubernetes to eg linuxkit/kubernetes anyway in which case maybe it fits better here for now?

errordeveloper:
I'd rather merge and avoid 🚲shedding ;)

ijc:
I think linuxkit/kubernetes makes good sense (I think the last major wrinkle was solved by the `linuxkit pkg` stuff), but I think that even after doing that it would make more sense to put the vpnkit related stuff in the vpnkit repo.

I should have tagged @djs55 when first suggesting that -- perhaps he has an opinion one way or the other.

ijc:

justincormack:
We decided it makes more sense to open this against `vpnkit` as it depends on vpnkit APIs, so closing here.

PR #2671, opened 2017-11-02
Kubernetes: disable fail on swap

pgayvallet:
The kubelet command often fails with the following error :

`Error: failed to run Kubelet: Running with swap on is not supported, please disable swap! or set --fail-swap-on flag to false. /proc/swaps contained: [Filename Type Size Used Priority /dev/dm-1 partition 4001788 0 **-1]**`

PR disables the swap check by adding -fail-swap-on=false to the kubelet arguments

Signed-off-by: pgayvallet <pierre.gayvallet@gmail.com>

### DIFF
--- a/projects/kubernetes/kubernetes/kubelet.sh
+++ b/projects/kubernetes/kubernetes/kubelet.sh
@@ -59,4 +59,5 @@ exec kubelet --kubeconfig=/etc/kubernetes/kubelet.conf \
 	      --cni-conf-dir=/var/lib/cni/etc/net.d \
 	      --cni-bin-dir=/var/lib/cni/opt/bin \
 	      --cadvisor-port=0 \
+	      --fail-swap-on=false \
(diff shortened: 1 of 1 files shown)


ijc:
You say "often" but I have never seen this and the in-tree yml files do not include the swap component.

I infer that you must be using your own yml assemblies with this image which includes the swap image. Given that I would suggest you also arrange that `/etc/kubelet.sh.conf` on the host sets `KUBELET_ARGS=--fail-on-swap=false`. You could also override the entrypoint/command for the service in your yml, but since that would mean duplicating all the default options that might be a bit more annoying.

I'm reluctant to just add this option by default since I suppose the Kube folks have good reasons for recommending to not run swap (of course, given that, removing your swap device would be another alternative).

ijc:
I'll close, but feel free to continue the discussion if you think my reasoning above is bogus.

PR #2675, opened 2017-11-03
Do not test on Go master

justincormack:

lint-roller-uses-10


justincormack:
This should fix the `go fmt` issues that are breaking CI.

PR #2677, opened 2017-11-03
linuxkit pkg: pull the actual tag before build

ijc:
Pulling the arch suffixed version does not provide any assurance that a
previous build was actually completed.

Relates a little to the issue with failing to push the cache images I mentioned in #2667, since I discovered it while investigating, but I don't think it is a fix for that.


ijc:
Rebased onto Travis fix.

PR #2679, opened 2017-11-03
linuxkit pkg: add `--dev` to force local org and hash

ijc:
$USER and "dev" respectively.

Might be too specific to my personal workflow but figured I'd throw it out there in case there was interest.

### DIFF
--- a/src/cmd/linuxkit/pkglib/pkglib.go
+++ b/src/cmd/linuxkit/pkglib/pkglib.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"gopkg.in/yaml.v2"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -90,6 +92,13 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 		return Pkg{}, err
 	}
 
+	if devMode {
+		pi.Org = os.Getenv("USER")
```
rn: This does not seem to honour the `--org` override but probably should. On some of my systems my user name is different to my hub name. Or is the intention to not push in `--dev` mode, in which case we might want to explicitly disable this

justincormack: ah, on all my systems my username is different to my hub name yeah...

rn:
thanks for the comment. Makes things clearer. I sometimes do push local package to my hub org to share them between computers

PR #2680, opened 2017-11-03
Remove accidental example commits

justincormack:
From #2672

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

foobear

PR #2681, opened 2017-11-03
Delete accidentally committed files

rn:
I had an out-of-sync git tree on a different machine where I hadn't picked up a force push to f14f418e5efa ("tools/alpine: Update to get new version of WireGuard tools"). So, we ended up with temporary  files. This commit removes them.

kitten-hat


rn:
close in favour of #2680

Issue #2636, opened 2017-10-25
Performance is very slow when multiple LinuxKit instance running simultaneously


arm64b:
**Description**

**Steps to reproduce the issue:**
1. open 2 ssh terminal to connect the same remote arm64 server, then
2. run 2 linuxkit instance at the same time in the above terminal with below command for each one:
terminal a: `bin/linuxkit run qemu -uefi -state "mysql" -fw /home/dennis/qemu/bios.bin -networking "tap,tap0" -iso arm64-efi.iso`
terminal b. `bin/linuxkit run qemu -uefi -state "apache" -fw /home/dennis/qemu/bios.bin -networking "tap,tap1" -iso arm64-efi.iso`

**Describe the results you received:**
We find that the LinuxKit will boot in a very slow speed comparing only one instance launched.

**Describe the results you expected:**
We hope the 2 LinuxKit instances will run as the same speed as we run only one LinuxKit instance each time.

**Additional information you deem important (e.g. issue happens only occasionally):**
I use `perf record` to record the performance related data for the entire life cycle of a LinuxKit instance(from `linuxkit run` to `poweroff` the LinuxKit system , the data as below (see the attached screen snapshot):
![linuxkit](https://user-images.githubusercontent.com/29669302/31983510-1dc151ce-b990-11e7-9ae6-e1df70303cc0.PNG)

rn:
I can not repro this. I've booted two linukit instances, one after the other, using the `minimal.yml` using `qemu/kvm` on a arm64 server and both boot in about 4s.

In any case this is unlikely a LinuxKIt issue, but more likely an issue either with your system, the version of KVM or qemu.

I used:
```
[master]linuxkit-dev-arm64:t/linuxkit/examples: qemu-system-aarch64 --version
QEMU emulator version 2.5.0 (Debian 1:2.5+dfsg-5ubuntu10.16), Copyright (c) 2003-2008 Fabrice Bellard
[master]linuxkit-dev-arm64:t/linuxkit/examples: uname -a
Linux linuxkit-dev-arm64 4.10.0-26-generic #30~16.04.1-Ubuntu SMP Tue Jun 27 09:40:29 UTC 2017 aarch64 aarch64 aarch64 GNU/Linux
```

arm64b:
I can reproduce this issue on both Qualcomm and Huawei arm64 server.
Rename the linuxkit.yml to arm64.yml, then:
`#bin/moby build -format iso-efi arm64.yml`
[...]

PR #2685, opened 2017-11-06
qemu: Fix performance downgrade when running multiple VM instances

arm64b:
This PR is used to fix the issue of #2636. 

On AArch64 platform, it doesn't have an architected methods to obtain
entropy, e.g, via instructions, nor does it have real hardware rng
device, so we don't need to provide a "virtio-rng-pci" as a paravirtualized
device to expose entropy inside the virtual machine.

**- What I did**
Fix the performance downgrade when running multiple LinuxKit instances on Qualcomm and Huawei AArch64 platforms.
**- How I did it**
Remove the unnecessary 'virtio-rng-pci' device from the AArch64 platform, since this device has no contribution to the entropy on that platform.

**- How to verify it**
Run multiple LinuxKit instances simultaneously as we do triggering the symptoms. 

summit

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -491,7 +491,9 @@ func runQemuContainer(config QemuConfig) error {
 func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
 	// Iterate through the flags and build arguments
 	var qemuArgs []string
-	qemuArgs = append(qemuArgs, "-device", "virtio-rng-pci")
+	if config.Arch != "aarch64" {
+		qemuArgs = append(qemuArgs, "-device", "virtio-rng-pci")
+	}
 	qemuArgs = append(qemuArgs, "-smp", config.CPUs)
 	qemuArgs = append(qemuArgs, "-m", config.Memory)
 	qemuArgs = append(qemuArgs, "-uuid", config.UUID.String())


deitch:
My testing confirms that this works. Removing ` -device virtio-rng-pci` runs cleanly on Huawei and Cavium. I didn't have a chance to test on Qualcomm. @arm64b , I assume you did?

arm64b:
Hello @deitch, yes I tested it on both Qualcomm and Huawei platform.

deitch:
Looks good to me, then. @rn, any comments?

rn:
@arm64b why does using `-device virtio-rng-pci` only affect HiSilicon/Qualcomm but works fine on Cavium?

deitch:
> why does using -device virtio-rng-pci only affect HiSilicon/Qualcomm

And why does it only affect the 2nd (and later) concurrent running VM?

arm64b:
@rn, I prefer to think that the difference is the HW difference between Cavium and HiSilicon/Qualcomm in terms of the RNG entroy emulated by the 'virtio-rng-pci'.

deitch:
> entroy emulated by the 'virtio-rng-pci'.

rn:
@arm64b I would prefer if we root cause this properly before unconditionally disabling `-device virtio-rng-pci` on all `aarch64` platforms. If some platforms emulate it properly, then ideally we would detect it.

The other difference (at least in my tests on #2636) was that I was using different kernels, which may also be relevant.

arm64b:
> But isn't it emulated? Why should it matter?

The architecture should looks like this:

       QEMU emulated device (virtio-rng-pci)
                             ^
                              |
                             V
                     virtio layer
                             ^
                              |
                             V
                      PCI device (Host)

I think the PCI implementation or IP used in Cavium is different than the other two. 

As far as the kernel, I tested both 4.9.x and 4.13.x on Qualcomm, the symptom of this issue is the same, so I think it should not be related with kernel.

deitch:
> The architecture should looks like this

Makes sense. I had thought it was completely emulated. How would we monitor the lower layers?

arm64b:
> How would we monitor the lower layers?

What do you mean the lower layer? is it the PCI layer in host?

deitch:
> is it the PCI layer in host?

According to the qemu docs (not an expert on qemu...), `virtio-rng-pci` goes to the host to get its entropy. By default, it uses the host's `/dev/random`, unless you explicitly give it an alternate device.

arm64b:
Actually on AArch64, we don't have an architecture method or real hardware device to provide the entropy of the RNG, so it doesn't make sense to provide a paravirtualized device 'virtio-rng-pci' to the guest. /dev/random is a kernel generated, not related with the 'virtio-rng-pci', `cat /dev/random` is a normal VM emulated behavior just we do in host.

deitch:
> Actually on AArch64, we don't have an architecture method or real hardware device to provide the entropy of the RNG, so it doesn't make sense to provide

You are saying that not just to fix this, but in general, it doesn't make sense to create the paravirtualized device on arm?

arm64b:
> You are saying that not just to fix this, but in general, it doesn't make sense to create the paravirtualized device on arm?

No. it does make sense to create a paravirtualized device to the VM on AArch64. but in this case, I guess the kernel will detect the information of the 'virtio-rng-pci' during its boot, but the back end of this device has different behavior in Cavium than the Hisilicon or Qualcomm for the underlying PCI implementation in HW layer.

deitch:
Aha! @arm64b, very nice. I no longer need qemu to replicate the problem. It is a problem with `/dev/random` on Huawei/Qualcomm vs Cavium.

The Cavium appears to have a hardware rng, which the kernel picks up. This in turn makes calls to `/dev/random` very quick.

The Huawei and Qualcomm do not. This makes calls to them extremely slow, as there may not be enough entropy for it. Calls to `/dev/random` are supposed to block until enough entropy is available.

This also explains why some waits are very slow, others are relatively quicker.

To replicate, on each of Cavium/Qualcomm/Huawei:

1. open a shell (no need for two)
2. `head -1 /dev/random` (or `dd` or any way you feel like getting random bits)
3. When done (or in another shell), repeat above

On the cavium, with a hardware generator, it returns immediately. On the others, the second call blocks.

If you really want to see it, just do `cat /dev/random`. On the Cavium, it will start dumping data. On Qualcomm/Huawei, it will generate a few bytes at best and block, eventually a few more, etc.

To see the hardware:

```
$ cat /sys/devices/virtual/misc/hw_random/rng_available
```

The Cavium gives a device; the others do not.

### Options

1. disable `virtio-rng-pci` as this PR does
2. detect if there is an underlying hwrng, and, if so, use `virtio-rng-pci`, else disable

I tend towards number 2, but my opinion.

arm64b:
Hello @deitch, Nice! I prefer the OP2 to the others(but really this a very 'weird' behavior on Cavium, since all the AArch64 platforms I see before don't have a real HW RNG entropy provider, Cavium is the excepted one), because if the host does have a HW RNG entropy, we need to pass it to the VM as the entropy generator inside the VM. 
I will update this RP to adopt OP2, it make sense more!

deitch:
> I will update this RP to adopt OP2

Before you put in the effort, let's have input from @rn

rn:
There is `HW_RANDOM_MSM` for Qualcomm and `HW_RANDOM_HISI` for HiSilicon. On the 4.12.13 kernel on packet.net's HiSilicon machine there is no kernel config file (and not `/proc/config.gz` so can't tell if it is compiled. On the ThunderX on packet.net `cavium_rng` is loaded. Don't have access to Qualcomm gear. I'll check with packet.net on their kernel config.

@arm64b can you check on your systems/kernels if the hardware rng is enabled in the kernel and if the module is loaded?

deitch:
@rn, nonetheless, do we not need a way to handle it, whether the hardware is available (physically and driver loaded) or not?

ijc:
Can we tell `qemu` to use `/dev/urandom` instead of `/dev/random`?

arm64b:
I am not sure if the RNG HW device on Cavium can provide **REAL** random entropy, but IMO in any case it does make sense to pass this device in form of para-virtualization to the guest since the VM can make use of the RNG feature from the host in this case.  

@rn When I upgrade the kernel on Cavium using the upstream stable kernel, the RANDOM related items  in the .config file I generated as below:
```
CONFIG_HW_RANDOM=y
CONFIG_HW_RANDOM_TIMERIOMEM=m
CONFIG_HW_RANDOM_BCM2835=m
CONFIG_HW_RANDOM_IPROC_RNG200=m
CONFIG_HW_RANDOM_VIRTIO=m
CONFIG_HW_RANDOM_TPM=m
CONFIG_HW_RANDOM_HISI=m
CONFIG_HW_RANDOM_MSM=m
CONFIG_HW_RANDOM_XGENE=m
CONFIG_HW_RANDOM_CAVIUM=m
CONFIG_HW_RANDOM_MTK=y
```
But I don't find the relevant kernel module loaded which means the kernel doesn't detect corresponding HW device exist in the system.

arm64b:
On Cavium:
```
root@Xenial:/sys/devices/virtual/misc/hw_random# cat rng_available 
cavium-rng-0000:00:09.1 cavium-rng-0004:00:09.1 
root@Xenial:/sys/devices/virtual/misc/hw_random# lspci
0000:00:01.0 PCI bridge: Cavium, Inc. THUNDERX PCC Bridge (rev 08)
0000:00:09.0 Processing accelerators: Cavium, Inc. THUNDERX Random Number Generator (rev 08)
0000:00:09.1 Processing accelerators: Cavium, Inc. Device a033 (rev 08)
```
There is a PCI device a033 as the HW RNG, but I believe there're no such devices on both Qualcomm and Huawei.

justincormack:
Using `/dev/urandom` for rng passthrough makes sense in all cases, including x86. This is just `linuxkit run` not designed for production so it is fine. In production cases for virt on arm64 you probably want to decide to pass through based on if the host has a real entropy source. For linuxkit on physical hardware we do want to support the common hw rand devices.

Also, please Arm, add a CPU based rng, its like being back in the old days of per motherboard configs for x86 back in the early 2000s it was really annoying. Is still have some rng driver in the Linux kernel left over from then...

arm64b:
Using the `/dev/urandom` or `/dev/random` is up to the RNG pkg of LinuxKit. But if the host really has the hardware device to provide a rng entropy source, we definitely want to passthru it to the VM, I think nobody will have objection for this.

ijc:
@arm64b Of course Huawei and Qualcom SoC's don't have the Cavium RNG! I'd be surprised if they didn't have their own (although maybe with no driver yet).

I certainly don't think we should cripple all guests on all ARM platforms (even those which have a hardware RNG) by removing the virtio RNG by default, that would be completely backwards.

I'm shocked that ARMv8 doesn't include a RNG instruction, but that does seem to be the case.

arm64b:
@ijc  Yes it doesn't have a RNG instruction from the perspective of ARMv8 spec, so we only rely on the 3rd party partner can provide an external hw device to do that like Cavium does in their own SoC.

ijc:
@arm64b no, the use of `urandom` or `random` for the backend of the virt devcie is down to QEMU on the host, not to LinuxKit's `rng` package -- you are thinking of within the guest, but it is the host level which matters here.

We are not going to remove the virtio pci devicde from VMs started by LinuxKit. I would rather see the platforms which lack a HW RNG be crippled than cripple those platforms which have done the right thing and do have an RNG, but it won't come to that because we can *and should* just tell QEMU to use `/dev/urandom` on the backend.

Of course we should also enable any kernel drivers we can find.

rn:
```
qemu-system-aarch64 -smp 1 -m 1024 -nographic     -cpu host -enable-kvm -machine virt,gic_version=host  -object rng-random,filename=/dev/urandom,id=rng0   -device virtio-rng-pci,rng=rng0     -device virtio-net-pci,netdev=t0 -netdev user,id=t0     -kernel minimal-kernel -initrd minimal-initrd.img     -append console=ttyAMA0
```
seems to work on the HiSIlicon system. I'll test a bit more

rn:
closing in favour of #2686

arm64b:
Oops! I am so surprised to get this info today, and after testing(with the latest codes) the issue is still there, I even suspect I am missing something from last night, so I get the UEFI fw by using @rn 's method:
```
sudo apt-get install qemu-efi
dd if=/dev/zero of=flash0.img bs=1M count=64
dd if=/usr/share/qemu-efi/QEMU_EFI.fd of=flash0.img conv=notrunc
```
I think the merged PR is wrong to solve this problem, `virtio-rng-pci` is just a para-virtualized device to fake a HW RNG to the guest, so the guest will have a 'real' /dev/hwrng as the RNG entropy, but it's source is not trust-able, because actually there's no **REAL** hardware RNG device in the host.

As far as the /dev/urandom and /dev/random, from the entropy generation perspective, there's no any relationship with /dev/hwrng. Both /dev/urandom and /dev/random are the character devices to get the entropy from the 'noise of the system', they are software based PRNG entropy source.

So the correct solution to solve this problem is:
If we do have a REAL RNG hardware device in the system (as Cavium does), we need to passthru this device to the guest via 'virtio-rng-pci', else we don't need to provide a para-virtualized device to the VM.

arm64b:
Based on the latest code, if I spawn a VM with `bin/linuxkit run`, the problem is still there.

deitch:
> Based on the latest code, if I spawn a VM with bin/linuxkit run, the problem is still there

I can confirm this, unfortunately. 

@rn the problem appears to be the command-line. I ran `lkt run qermu test` and look at the process:

```
# ps -ef | grep qemu
root     32491 32486  5 06:22 pts/0    00:00:06 /usr/bin/qemu-system-aarch64 -device virtio-rng-pci -smp 1 -m 1024 -uuid 86b0679a-81f4-44e3-8539-94a7fe8e74c0 -pidfile test-state/qemu.pid -cpu host -enable-kvm -machine virt,gic_version=host -object rng-random,id=rng0,filename=/dev/urandom -device virtio-rng-pci,rng=rng0 -kernel test-kernel -initrd test-initrd.img -append console=tty0 console=ttyS0 console=ttyAMA0 -device virtio-net-pci,netdev=t0,mac=e6:b4:26:75:bf:29 -netdev user,id=t0 -nographic
```

Looks like we pass `-device virtio-rng-pci` twice. Once with the correct `,rng=rng0` and once without. If you run it without, all works well.

deitch:
Fixed in #2693

PR #2686, opened 2017-11-06
qemu: Use /dev/urandom on Linux for virtio-rng-pci

### LINKED ISSUE #2636: Performance is very slow when multiple LinuxKit instance running simultaneously

rn:
On Linux the default is /dev/random which may block if there
is not enough entropy (and it determines lack of entropy
conservatively). /dev/urandom never blocks but is fed from
the same entropy source as /dev/random so this change should
not affect platforms which have hardware RNGs but helps to
prevent stalls on platforms, in particular some arm64 platforms,
where the hardware RNG is either not present or not enabled.

If you really care about proper entropy inside a VM do not use
'linuxkit run', it's mostly for testing and shouldn't be used
for production anyway.

fixes #2636

goat-happy

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -521,6 +521,12 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
 		}
 	}
 
+	rng := "rng-random,id=rng0"
+	if runtime.GOOS == "linux" {
+		rng = rng + ",filename=/dev/urandom"
+	}
+	qemuArgs = append(qemuArgs, "-object", rng, "-device", "virtio-rng-pci,rng=rng0")
+
 	var lastDisk int
 	for i, d := range config.Disks {
 		index := i


rn:
crap, will fix this properly, forgot to use `urandom` after lunch

rn:
pushed an update

justincormack:
I don't see any reason not to use `/dev/urandom` on all platforms. On OSX they are the same device.

deitch:
> I don't see any reason not to use /dev/urandom on all platforms

It does draw from hardware RNG via `/dev/random` when available, right? And since `lkt` isn't intended for prod, we would prioritize speed and functionality over true entropy, so that makes sense.

justincormack:
Yes, `/dev/urandom` is seeded from the same sources.

ijc:
I had no idea MacOS ahd `/dev/urandom` it certainly makes sense to use it if it exists everywhere which has `/dev/random` -- do all BSDs have it too?

If we have to keep a conditional then I think it would be better for things to be identical apart from the presence/absence of `filename=`, i.e. to still add the `rng0` stuff and `-object` stuff.

rn:
@justincormack didn't know `/dev/urandom` was avail on macOS, but I'd much rather keep it the way it is, ie only use `urandom` on Linux where we know it may cause issues and use the platform default elsewhere

rn:
@ijc you mean like this?
```
        if runtime.GOOS == "linux" {
                qemuArgs = append(qemuArgs, "-object", "rng-random,filename=/dev/urandom,id=rng\
0", "-device", "virtio-rng-pci,rng=rng0")
        } else {
                qemuArgs = append(qemuArgs, "-object", "rng-random,id=rng0", "-device", "virtio\
-rng-pci,rng=rng0")
        }
```

justincormack:
Actually on OSX they are different devices but behave the same, its FreeBSD where they are the same actual device. It seems OpenBSD may not have `/dev/urandom` though...

ijc:
More like:
```
rng := "rng-random,id=rng0"
if runtime.GOOS == "linux" { 
        rng = rng + ",filename=/dev/urandom"
}
qemuArgs = append(qemuArgs, "-object", rng, "-device", "virtio-rng-pci,rng=rng0")
```

rn:
@ijc doesn't make much of a difference to me, but pushed your suggestion

deitch:
> on Linux where we know it may cause issues and use the platform default elsewhere

We actually don't know for sure where it will or will not cause issues. @vielmetti pointed out that if we run `rngd`, it feeds `/dev/random`, so even without hardware RNG it works more quickly.

So our psuedo-code would need to be: `if !(present hwrng), && !(running rngd) then use /dev/urandom else use /dev/random`.

I imagine there are more cases where it might fail and it would miss it.

@justincormack's suggestion is to use `/dev/urandom`, which (according to what he explained) _always_ returns quickly. It basically does, `if (/dev/random returns quickly) use it, else generate`

Also, from `man 4 urandom`:

> If you are unsure about whether you should use /dev/random or /dev/urandom, then probably you want to use the latter. As a general rule, /dev/urandom should be used for everything except long-lived GPG/SSL/SSH keys

PR #2687, opened 2017-11-06
metadata: handle multiple levels in json

### LINKED ISSUE #2641: metadata: tree is limited to only 2 levels
**Description**

The json metadata represents a disk. A disk can have more than 2 levels of depth.

Mount a json file like this: 

```
{
  "etc": {
    "ssl": {
      "certs": {
        "ca-certificates.crt": {
          "perm": "0644",
          "content": "some content"
        }
      }
    }
  }
}
```

Directory etc is empty

The tree described in the json exists.

guillaumerose:
This PR intends to fix #2641

### DIFF
--- a/pkg/metadata/main.go
+++ b/pkg/metadata/main.go
@@ -111,7 +111,7 @@ func main() {
 	}
 
 	if userdata != nil {
-		if err := processUserData(userdata); err != nil {
+		if err := processUserData(ConfigPath, userdata); err != nil {
 			log.Printf("Could not extract user data: %s", err)
 		}
 	}
@@ -139,9 +139,9 @@ func main() {
 //        }
 // }
 // Will create foobar/foo with mode 0644 and content "hello"
-func processUserData(data []byte) error {
+func processUserData(basePath string, data []byte) error {
 	// Always write the raw data to a file
-	err := ioutil.WriteFile(path.Join(ConfigPath, "userdata"), data, 0644)
+	err := ioutil.WriteFile(path.Join(basePath, "userdata"), data, 0644)
 	if err != nil {
 		log.Printf("Could not write userdata: %s", err)
 		return err
@@ -160,7 +160,7 @@ func processUserData(data []byte) error {
 		return nil
 	}
 	for d, val := range cm {
-		dir := path.Join(ConfigPath, d)
+		dir := path.Join(basePath, d)
 		if err := os.MkdirAll(dir, 0755); err != nil {
 			log.Printf("Failed to create %s: %s", dir, err)
 			continue
--- a/pkg/metadata/main.go
+++ b/pkg/metadata/main.go
@@ -154,55 +154,44 @@ func processUserData(basePath string, data []byte) error {
 		// This is not an error
 		return nil
 	}
-	cm, ok := fd.(map[string]interface{})
-	if !ok {
-		log.Printf("Could convert JSON to desired format: %s", fd)
-		return nil
-	}
-	for d, val := range cm {
-		dir := path.Join(basePath, d)
-		if err := os.MkdirAll(dir, 0755); err != nil {
-			log.Printf("Failed to create %s: %s", dir, err)
-			continue
-		}
-		files, ok := val.(map[string]interface{})
-		if !ok {
-			log.Printf("Could convert JSON for files: %s", val)
-			continue
-		}
-		for f, i := range files {
-			p := uint64(0644)
-			var c string
 
-			switch fi := i.(type) {
-			case map[string]interface{}:
-				if _, ok := fi["perm"]; !ok {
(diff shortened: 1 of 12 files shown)


pkg/metadata/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func TestWriteConfig(t *testing.T) {
+	basePath, err := ioutil.TempDir(os.TempDir(), "metadata")
+	if err != nil {
+		t.Fatalf("can't make a temp rootdir %v", err)
+	}
+	defer os.RemoveAll(basePath)
+
+	json := `{"ssh":{"sshd_config":{"perm":"0600","content":"PermitRootLogin yes\nPasswordAuthentication no"}},"foo":{"bar":"foobar","baz":{"perm":"0600","content":"bar"}}}`
```
ijc: AIUI using backticks to quote in go gives you the opportunity to format things a bit more readable, e.g. to use raw newlines and indentation.

pkg/metadata/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+)
+
+func TestWriteConfig(t *testing.T) {
+	basePath, err := ioutil.TempDir(os.TempDir(), "metadata")
```
ijc: Passing `""` (the empty string) as the first argument has the same effect as manually passing `os.TempDir()` but is more usual.

ijc:
Does anything anywhere run this test in CI? Would be good to add that at the same time.

> handle multiple levels of nesting.

From #2641 there doesn't seem to be much appetite from us maintainers on that front, please do discuss it further (on that ticket) before working on that.

guillaumerose:
I updated the PR with your comments. I added the handling of multiple levels. It also fixes the hostname file with this method.

rn:
I missed the update somehow. I though the idea was to discuss this a bit further on #2641...there is one concern raised by @ijc regarding permissions/ownership on subdirectories, which wasn't that important in the existing implementation, but with the new use case of basically mirroring directory hierarchies might be more important.

However, I guess this can be added later by *extending* the JSON. Or, alternatively, packages which do care can copy/chmod/chown the files appropriately.

I've build and pushed the updated image. Could you update the hashes in a separate commit to:
```
linuxkit/metadata:087a5cb4314ca7f02e61931d2ae0982f9dfae88b
```
That way CI will pick them up.

pkg/metadata/main.go
```diff
@@ -154,55 +154,44 @@ func processUserData(data []byte) error {
...
+	default:
+		log.Printf("Couldn't convert JSON for items: %s", input)
 	}
+}
 
-	return nil
+func isFile(json map[string]interface{}) bool {
+	_, perm := json["perm"]
+	_, content := json["content"]
+	return perm && content
```
ijc: Does this mean I cannot create a directory simultaneously containing two files named `perm` and `content`?

It would be good to see `docs/metadata.md` updated to reflect the new structure/behaviour.

guillaumerose: Yes that's not good. Should we enforce type of objects by adding a new field ? Like directory or file. It will be better in the future if we want to add directory permissions.

```json
{
  "type": "directory",
  "name": "etc",
  "content": [
    {
      "type": "directory",
      "name": "ssl",
      "content": [
        {
          "type": "directory",
          "name": "certs",
          "content": [
            {
              "type": "file",
              "name": "ca-certificates.crt",
              "content": "some content"
            }
          ]
        }
      ]
    },
    {
      "type": "file",
      "name": "resolv.conf",
      "content": "some content"
    }
  ]
}
```

It's a breaking change now, but maybe better.

ijc: I'd suggest either `content` becomes bimodal (so contains either a JSON string and is a file or contains a JSON map and is a directory, which can be detected at runtime) or add a new field `entries` which would be an map (i.e. a directory) but mutually exclusive with the existing `content` field so you can tell the cases apart that way.

I think the latter option (adding `entries`) is likely to be nicer from a schema PoV (fields which a dynamically typed are a bit of a pain).

I don't think you need the `type` entry to distinguish the cases, which I think means it can be backwards compatible.

guillaumerose: I will write an other PR for that, thanks.

ijc: If we merge this  PR and then update/fix in a followup then that will be two breaking changes rather than zero. Or maybe you meant for that PR to supersede/replace this one?

ijc: :+1:, thanks!

justincormack:
Given the conversation above, going to close this for now and review the new version.

PR #2688, opened 2017-11-06
Kubernetes : add capability not to start kubelet

pgayvallet:
For kubernetes integration in D4D, we need to be able to enable or disable the kubelet at startup in the linuxkit VM (to start it or not).

Currently, there is no way to do this properly with the `linuxkit/kubernetes` image. We could create a custom image for our needs, but it's possible that other uses of the `linuxkit/kubernetes` image get the same needs, so the original image is probably the place to do the improvement.

PR adds checks in the kubelet.sh script to be able to instantly exit the script if `/var/config/kubelet/disabled` file or `KUBELET_DISABLED` envvar is present. 

Default script behaviour is unaltered : if nothing is set, the kubelet script will execute normally.

### DIFF
--- a/projects/kubernetes/kubernetes/kubelet.sh
+++ b/projects/kubernetes/kubernetes/kubelet.sh
@@ -2,11 +2,15 @@
 # Kubelet outputs only to stderr, so arrange for everything we do to go there too
 exec 1>&2
 
+if [ -e /etc/kubelet.sh.conf ] ; then
+    . /etc/kubelet.sh.conf
+fi
+
 if [ -f /var/config/kubelet/disabled ] ; then
     echo "kubelet.sh: /var/config/kubelet/disabled file is present, exiting"
     exit 0
 fi
-if ! [ -z "$KUBELET_DISABLED" ] ; then
+if [ -n "$KUBELET_DISABLED" ] ; then
     echo "kubelet.sh: KUBELET_DISABLED environ variable is set, exiting"
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


ijc:
> PR follows the discussion started here docker/editions.moby#325

Since that is a closed repo please summarise the reasons for this change here.

(in fact, IMHO, it is always worth explaining a PR inline rather than referencing some other discussion even if it is also public, just for clarities sake)

pgayvallet:
PR description updated

projects/kubernetes/kubernetes/kubelet.sh
```diff
@@ -2,6 +2,15 @@
 # Kubelet outputs only to stderr, so arrange for everything we do to go there too
 exec 1>&2
 
+if [ -f /var/config/kubelet/disabled ] ; then
+    echo "kubelet.sh: /var/config/kubelet/disabled file is present, exiting"
+    exit 0
+fi
+if ! [ -z "$KUBELET_DISABLED" ] ; then
```
ijc: `if ! [ -z "FOO" ]` is the same as `if [ -n "FOO" ]` and I think the latter is clearer.

I think this all needs to go after `. /etc/kubelet.sh.conf` since that might be how one wants to set `KUBELET_DISABLED`.

That in turn seems to suggest the `. /etc/kubelet.sh.conf` should be moved above the CNI unpack, since if we are exiting we want to skip that too.

So the order should be:
```
if [ -e /etc/kubelet.sh.conf ] ; then
    . /etc/kubelet.sh.conf
fi

<both blocks of your new code>

if [ ! -e /var/lib/cni/.opt.defaults-extracted ] ; then
    mkdir -p /var/lib/cni/opt/bin
    tar -xzf /root/cni.tgz -C /var/lib/cni/opt/bin
    touch /var/lib/cni/.opt.defaults-extracted
fi
```

ijc:
Please ignore the circleci failure (it's not ready yet), can be merged once everything else is green

ijc:
I built and pushed `linuxkitprojects/kubernetes:392d3a96d298711b5e5d53a7fddf83991db27759` from master after this merge.

The in-tree yml has *not* been updated to the new hash, I guess there is no hurry and I can take care of that during the next natural bump (if you need it sooner please open a followup PR).

PR #2690, opened 2017-11-06
Add Circle CI runes

ijc:
Advantage over Travis for this usecase is the automatic artifact handling
without needing to mess with S3 or similar, making it simple to offer
downloads.


.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
+version: 2
+jobs:
+  build:
+    working_directory: /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}
```
justincormack: I think you want to always make this `linuxkit/linuxkit` as if I have forked it the gopath should not change

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      - run: mkdir -p ./dist
+      - run:
+          name: Lint
+          command: make local-check
+      - run:
+          name: Build x86_64/linux
+          environment:
+            GOOS: linux
+            GOARCH: x86_64
+          command: go build -i -o dist/linuxkit-$GOOS-$GOARCH github.com/linuxkit/linuxkit/src/cmd/linuxkit
```
justincormack: any way to add the commits and version to this sanely?

.circleci/config.yml
ijc: Oh yeah, I had that and then read the docs which uses this and mislead me!

.circleci/config.yml
ijc: I'll investigate. Might be that we can use `make local-build` if Circle doesn't have the issue described in the travis.yml comment (which I should replicate here if it does)

ijc: Turns out that not only does CircleCI not suffer from that issue, but the workaround (using `go build -i`) is broken, while not having the workaround works...

ijc:
OK, I'm pretty happy with it now, please review.

I debated using the Circle2.0 workflows stuff to parallelise the builds, but I think that might be overengineering it a bit at this point, it seems pretty quick anyways.

ijc:
My changes to enable static building seem to have broken x86_64-linux on Travis (but not circle). Rather than try to reconcile the two I think I'd be inclined to remove/stop the travis builds since this circle stuff completely replaces it (and adds artefacts too)

Makefile
```diff
@@ -72,6 +72,14 @@ test-cross:
 	$(MAKE) -j 3 GOOS=linux tmp_moby_bin.tar tmp_rtf_bin.tar tmp_mt_bin.tar tmp_linuxkit_bin.tar
 	$(MAKE) clean
 
+ifeq ($(GOARCH)-$(GOOS),amd64-linux)
+LOCAL_BUILDMODE?=pie
```
justincormack: I guess cross building static pie for arm64-linux doesnt work as there is no cross linker?

ijc: The failure was in https://circleci.com/gh/linuxkit/linuxkit/25 and I think your diagnosis is correct.

ijc:
I've now removed the travis config, I'll disable travis as this PR gets merged.

ijc:
I couldn't find a travis master switch, but I've disabled "Build branch updates" and "Build pull request updates" in the [settings for linuxkit/linuxkit](https://travis-ci.org/linuxkit/linuxkit/settings).

ijc:
Ah, I found it right after in the [org profile](https://travis-ci.org/profile/linuxkit?offset=0) and turned it off fully (I think)

PR #2694, opened 2017-11-07
pkglib: Enable runtime debug output

rn:
Log commands executed when '-v' is used on the commandline.

By default we do not dump the environment as this may contain
sensitive information, such as passwords and passphrases.

fox-golf

### DIFF
--- a/src/cmd/linuxkit/pkglib/docker.go
+++ b/src/cmd/linuxkit/pkglib/docker.go
@@ -8,9 +8,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-)
 
-const debugDockerCommands = false
+	log "github.com/sirupsen/logrus"
+)
 
 const dctEnableEnv = "DOCKER_CONTENT_TRUST=1"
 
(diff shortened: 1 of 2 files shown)


ijc:
> By default we do not dump the environment as this may contain sensitive information, such as passwords and passphrases.

I don't think we ever log the environment, before or after this command.

Are you referring to this:
```
	if debugFull {
		var dct string
		if dr.dct {
			dct = dctEnableEnv + " "
		}
        log.Debugf("Executing: %s %v", dct, cmd.Args)
```
??? Remember that `dctEnableEnv` is just the string `"DOCKER_CONTENT_TRUST=1"` not a full env.

rn:
misread the code in a hurry. Pushed and update

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -39,13 +39,13 @@ func (dr dockerRunner) command(args ...string) error {
...
-	if debugDockerCommands {
-		var dct string
-		if dr.dct {
-			dct = dctEnableEnv + " "
-		}
-		fmt.Fprintf(os.Stderr, "+ %s%v\n", dct, cmd.Args)
+
+	dct := ""
+	if dr.dct {
+		dct = dctEnableEnv + " "
```
ijc: Unless the diff is misleading me with the removal of the `debugDockerCommands` conditional it now makes sense to merge the to `if dr.dct` blocks I think.

PR #2700, opened 2017-11-09
Kubernetes updates

ijc:
**- What I did**

Various updates:
* Add an option to do simple bridged networking. The way the files are injected is a bit hacky (alpine image doing echo) but will do for now, I'd like to figure out a way to do this with a kube yaml file (perhaps a `Job`) or some other way.
* Only build the images relevant to the build platform (but also make it configurable)
* Improvements to the handling of `/etc/kubeadm/kube-system.init/` with possibly invalid content.
* Bump to newer cri-containerd master, with vendoring updated to containerd beta3 and other fixes.

**- How I did it**

Individual commit messages have more info.

**- How to verify it**

`make KUBE_FORMATS="iso-efi iso-bios" KUBE_RUNTIME=docker KUBE_NETWORK=bridge kube-master.iso`

and variants.

Support for simple bridged networking

[](https://www.metal-archives.com/albums/Sanzu/Heavy_over_the_Home/576429)

### DIFF
--- a/projects/kubernetes/Makefile
+++ b/projects/kubernetes/Makefile
@@ -4,6 +4,13 @@ KUBE_NETWORK ?= weave-v2.0.5
 INIT_YAML ?=
 INIT_YAML += network.yaml
 
+ifeq ($(shell uname -s),"Darwin")
+KUBE_FORMATS ?= iso-efi
+endif
+KUBE_FORMATS ?= iso-bios
+
+KUBE_FORMAT_ARGS := $(patsubst %,-format %,$(KUBE_FORMATS))
+
 all: build-container-images build-vm-images
 
 build-container-images:
@@ -20,10 +27,10 @@ build-vm-images: kube-master.iso kube-node.iso
 
 # NB cannot use $^ because $(INIT_YAML) is not for consumption by "moby build"
 kube-master.iso: kube.yml $(KUBE_RUNTIME).yml $(KUBE_RUNTIME)-master.yml $(INIT_YAML)
-	moby build -name kube-master -format iso-efi -format iso-bios kube.yml $(KUBE_RUNTIME).yml $(KUBE_RUNTIME)-master.yml
+	moby build -name kube-master $(KUBE_FORMAT_ARGS) kube.yml $(KUBE_RUNTIME).yml $(KUBE_RUNTIME)-master.yml
 
 kube-node.iso: kube.yml $(KUBE_RUNTIME).yml
-	moby build -name kube-node -format iso-efi -format iso-bios $^
+	moby build -name kube-node $(KUBE_FORMAT_ARGS) $^
 
 network.yaml: $(KUBE_NETWORK).yaml
 	ln -nf $< $@
--- a/projects/kubernetes/Makefile
+++ b/projects/kubernetes/Makefile
@@ -1,8 +1,7 @@
 KUBE_RUNTIME ?= docker
-KUBE_NETWORK ?= weave-v2.0.5
+KUBE_NETWORK ?= weave
 
-INIT_YAML ?=
-INIT_YAML += network.yaml
+KUBE_NETWORK_WEAVE ?= v2.0.5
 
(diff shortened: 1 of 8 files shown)


projects/kubernetes/kubernetes/kubeadm-init.sh
```diff
@@ -11,9 +11,19 @@ else
     kubeadm init --skip-preflight-checks --kubernetes-version @KUBERNETES_VERSION@ $@
 fi
 for i in /etc/kubeadm/kube-system.init/*.yaml ; do
+    n = $(basename "$i")
```
ijc: There is a typo here which I somehow missed, need to lose the spaces around the `=` to be valid shell.

projects/kubernetes/Makefile
```diff
@@ -18,21 +22,19 @@ push-container-images:
...
+	moby build -name kube-node $(KUBE_FORMAT_ARGS) $^
 
-network.yaml: $(KUBE_NETWORK).yaml
-	ln -nf $< $@
+weave.yml: kube-weave.yaml
 
-weave-%.yaml:
-	curl -L -o $@ https://cloud.weave.works/k8s/v1.8/net?v=$*
+kube-weave.yaml:
+	curl -L -o $@ https://cloud.weave.works/k8s/v1.8/net?v=v2.0.5
```
justincormack: can we put the weave version at the top of the file? Hard to find down here

projects/kubernetes/bridge.yml
```diff
@@ -0,0 +1,6 @@
+onboot:
+  - name: bridge
+    image: alpine:3.6
```
justincormack: busybox is smaller than alpine, unless you are already using alpine elsewhere

projects/kubernetes/bridge.yml
```diff
@@ -0,0 +1,6 @@
...
+    command: ["/bin/sh", "-c", "set -ex; mkdir -p /var/lib/cni/etc/net.d; echo '{\"cniVersion\":\"0.3.1\",\"name\":\"default\",\"plugins\":[{\"type\":\"bridge\",\"bridge\":\"cni0\",\"isDefaultGateway\":true,\"ipMasq\":false,\"hairpinMode\":true,\"ipam\":{\"type\":\"host-local\",\"subnet\":\"10.1.0.0/16\",\"gateway\":\"10.1.0.1\"},\"dns\":{\"nameservers\":[\"10.1.0.1\"]}},{\"type\":\"portmap\",\"capabilities\":{\"portMappings\":true},\"snat\":true}]}' > /var/lib/cni/etc/net.d/10-default.conflist; echo '{\"cniVersion\":\"0.2.0\",\"type\":\"loopback\"}' > /var/lib/cni/etc/net.d/99-loopback.conf"]
```
justincormack: the `mkdir` could be added to the runtime config for kubelet? This does seem a bit ugly...

projects/kubernetes/Makefile
ijc: Sure.

I had debated supporting `KUBE_NETWORK="weave v2.0.5"` like syntax and playing games like `$(first $(KUBE_NETWORK)).yml` to get `weave.yml` and `net?v=$(word 2,$(KUBE_NETWORK))` for the URL. Seemed overly clever to me though.

projects/kubernetes/bridge.yml
ijc: I'm not super happy with it either. I'd been thinking of https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/ so it could be a yaml to apply.

I'd also been considering having a directory somewhere which was the template/skeleton of this directory which could be applied at the same time as we unpack the default CNI things in `kubelet.sh`.

I think I'd like to get this batch of changes (particularly the `cri-containerd` bump) in now though and rework more in the future.

justincormack: Yeah sure, its not too terrible and its a new option anyway.

PR #2701, opened 2017-11-09
Add support for publish port on localhost to the hyperkit backend

rn:
This series of patches adds support for the `-publish` command line option to the hyperkit backend, which allows exposing selected ports on `localhost` by configuring the port forwarding feature of VPNKit. This mode is supported both in `docker-for-mac` and `vpnkit` networking modes. See the updated documentation for details

It also includes a fair bit of refactoring:
- common functions for parsing published ports and multiple occurrences of flags was from `run_qemu.go` to the common `utils.go`
- Switch to use `github.com/google/uuid` for UUID handling. We used `github.com/satori/go.uuid` before, but the VPNKit go bindings use the former so we may as well use it everywhere.
- Fix to the `vpnkit-forwarder` example

dog-dig


rn:
pushed an update to the hyperkit vendoring to pick up the better/speedier disk creation using ftruncate

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -337,3 +368,78 @@ func launchVPNKit(etherSock string, vsockSock string, portSock string) (*os.Proc
...
+		}
+	}
+	log.Debugf("VPNKit UUID:%s IP: %v", vpnkitUUID, vif.IP)
+
+	log.Debugf("Connecting to VPNKit on %s", portSocket)
+	c, err := vpnkit.NewConnection(context.Background(), portSocket)
+	if err != nil {
+		return nil, fmt.Errorf("Connection to VPNKit failed: %v", err)
+	}
+
```
djs55: Is it worth a comment here explaining the reason why there's no `defer c.Close()`? Or could we put the `Close` in the cleanup function?

djs55:
Looks good to me!

src/cmd/linuxkit/run_hyperkit.go
rn: type `vpnkit.Connection` does not seem to have a `Close()` method. I'm pretty sure I tried to call it and it's not in https://github.com/moby/vpnkit/blob/master/go/pkg/vpnkit/connection.go

djs55: Ah oops! I've filed https://github.com/moby/vpnkit/issues/307

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -44,16 +46,18 @@ func runHyperKit(args []string) {
...
-	networking := flags.String("networking", hyperkitNetworkingDefault, "Networking mode. Valid options are 'default', 'docker-for-mac', 'vpnkit[,socket-path]', 'vmnet' and 'none'. 'docker-for-mac' connects to the network used by Docker for Mac. 'vpnkit' connects to the VPNKit socket specified. If socket-path is omitted a new VPNKit instance will be started and 'vpnkit_eth.sock' will be created in the state directory. 'vmnet' uses the Apple vmnet framework, requires root/sudo. 'none' disables networking.`")
+	networking := flags.String("networking", hyperkitNetworkingDefault, "Networking mode. Valid options are 'default', 'docker-for-mac', 'vpnkit[,eth-socket-path[,port-socket-path]]', 'vmnet' and 'none'. 'docker-for-mac' connects to the network used by Docker for Mac. 'vpnkit' connects to the VPNKit socket specified. If socket-path is omitted a new VPNKit instance will be started and 'vpnkit_eth.sock' and 'vpnkit_port.sock' will be created in the state directory. 'vmnet' uses the Apple vmnet framework, requires root/sudo. 'none' disables networking.`")
```
MagnusS: This may need an update now that there are two socket parameters:
`vpnkit` connects to the VPNKit socket specified. If socket-path is omitted a new VPNKit instance will be started

MagnusS:
LGTM!

src/cmd/linuxkit/run_qemu.go
```diff
@@ -705,7 +655,7 @@ func buildQemuForwardings(publishFlags multipleFlag, containerized bool) (string
 func buildDockerForwardings(publishedPorts []string) ([]string, error) {
 	pmap := []string{}
 	for _, port := range publishedPorts {
-		s, err := splitPublish(port)
+		s, err := NewPublishedPort(port)
```
ijc: the switch from `s` to `N` also means this is now a public function. Probably harmless but checking if it was intended and/or deemed ok.

src/cmd/linuxkit/run_hyperkit.go
ijc: Was this update done?

I wonder if (separately) maybe this screed should be moved into `flags.Usage` where it can be more sanely/readably formatted?

src/cmd/linuxkit/run_qemu.go
rn: yes it was intended primarily because `DiskConfig` just above it is also public. But then some other functions in `util.go` are not, hey ho

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -337,3 +368,78 @@ func launchVPNKit(etherSock string, vsockSock string, portSock string) (*os.Proc
...
+	ctx := context.Background()
+
+	vpnkitUUID, err := uuid.Parse(h.VPNKitUUID)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse VPNKit UUID %s: %v", h.VPNKitUUID, err)
+	}
+
+	localhost := net.ParseIP("127.0.0.1")
+	if localhost == nil {
+		return nil, fmt.Errorf("Failed to parse 127.0.0.0")
```
ijc: Error message and the actual IP are mismatched...

rn:
Force pushed and update for the error message and also updated the help message.

PR #2703, opened 2017-11-10
metadata: handle json with more than 2 levels

guillaumerose:
This PR intends to fix https://github.com/linuxkit/linuxkit/issues/2641

It introduces a new recursive JSON structure.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -22,19 +22,25 @@ directories to be created and the directories are populated with files.
 For example, the following userdata file:
 ```JSON
 {
-    "ssh" : {
-        "sshd_config" : {
-            "perm" : "0600",
-            "content": "PermitRootLogin yes\nPasswordAuthentication no"
-        }
-    },
-    "foo" : {
-        "bar" : "foobar",
-        "baz" : {
-            "perm": "0600",
-            "content": "bar"
-        }
+  "ssh": {
+    "entries": {
+      "sshd_config": {
+        "perm": "0600",
+        "content": "PermitRootLogin yes\nPasswordAuthentication no"
+      }
     }
+  },
+  "foo": {
+    "entries": {
+      "bar": {
+        "content": "foobar"
+      },
+      "baz": {
+        "perm": "0600",
+        "content": "bar"
+      }
+    }
+  }
 }
 ```
 will generate the following files:
@@ -44,16 +50,15 @@ will generate the following files:
 /var/config/foo/baz
 ```
 
-Each file can either be:
+The JSON file consists of a map from `name` to an entry object. Each entry object has the following fields:
+- `content`: if present then the entry is a file. The value is a string containing the desired contents of the file.
+- `entries`: if present then the entry is a directory. The value is a map from `name` to entry objects.
+- `perm`: the permissions to create the file with.
 
-- a simple string (as for `foo/bar` above) in which case the file will
-  be created with the given contents and read/write (but not execute)
-  permissions for user and read permissions for group and everyone else (in octal format `0644`).
-- a map (as for `ssh/sshd_config` and `foo/baz` above) with the
-  following mandatory keys:
-  - `content`: the contents of the file.
-  - `perm`: the permissions to create the file with.
-
+The `content` and `entries` fields are mutually exclusive, it is an error to include both, 
+one or the other _must_ be present. 
+The file or directory's name in each case is the same as the key which referred to that entry.
+ 
 This hierarchy can then be used by individual containers, who can bind
 mount the config sub-directory into their namespace where it is
 needed.
--- a/pkg/metadata/main.go
+++ b/pkg/metadata/main.go
@@ -111,7 +111,7 @@ func main() {
 	}
 
 	if userdata != nil {
-		if err := processUserData(userdata); err != nil {
+		if err := processUserData(ConfigPath, userdata); err != nil {
 			log.Printf("Could not extract user data: %s", err)
 		}
 	}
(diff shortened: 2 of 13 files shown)


ijc:
Was it necessary to move from maps (where the key is the name) to an array with an explicit name field?

The maps method seemed nicer to me and changing away from it is a breaking change.

Rewriting the example you added here I would envisage something like:
```
{
  "ssh": {
    "entries": {
      "sshd_config": {
        "perm": "0600",
        "content": "PermitRootLogin yes\nPasswordAuthentication no"
      }
    }
  },
  "foo": {
    "entries": {
      "bar": {
        "content": "foobar"
      },
      "baz": {
        "perm": "0600",
        "content": "bar"
      }
    }
  }
}
```
Could that be made to work?

guillaumerose:
Yes I can do it, I think. 

What I like in the current JSON is that it uses a recursive structure. So I can use it both to create and parse the JSON. This structure can be then reused in Docker for Mac for instance. Easier and safer to work with this than with map[string]interface{}, and we can probably validate it with a JSON schema. 
You're example is more concise and human-readable. What should we do ?

ijc:
I think you can have a recursive type in go, something like:
```
type entry struct {
      perm string
      content string
      entries map[string]entry
}
```

Or at least a stunt version of that looks to work: https://play.golang.org/p/GdE4w2Bmna

Maybe its not so simple/possible in the real world?

guillaumerose:
I updated the PR with your struct. That's better, thanks. Regarding the docs, I tried something :)

pkg/metadata/main.go
```diff
@@ -139,70 +139,80 @@ func main() {
...
+	return defaultMode, nil
+}
 
-	return nil
+type entry struct {
+	path string
+
+	Perm    string           `json:"perm"`
+	Content *string          `json:"content"`
+	Entries map[string]entry `json:"entries"`
```
ijc: I wonder if these should all be `,omitempty`?

docs/metadata.md
```diff
@@ -44,16 +50,13 @@ will generate the following files:
...
-  be created with the given contents and read/write (but not execute)
-  permissions for user and read permissions for group and everyone else (in octal format `0644`).
-- a map (as for `ssh/sshd_config` and `foo/baz` above) with the
-  following mandatory keys:
-  - `content`: the contents of the file.
-  - `perm`: the permissions to create the file with.
-
+The JSON file contains two types of entry :
+- directory, described with a name, permissions and entries that will be included in this directory,
+- file, with name, permissions and content.
```
ijc: Can we wrap the field names in \` backticks please so they are formatted.

Also the field is called `perm` not `permissions`, it would be better to document the real name. It might be better to keep the previous structure, i.e. a list of key names like.

Lastly, the way this is written it reads like `name` is a field when actually it is (now) the key in the map, can that be clarified please.

How about:
```
The JSON file consists of a map from `name` to an entry object. Each entry object has the following fields:
- `content`: if present then the entry is a file. The value is a string containing the desired contents of the file.
- `entries`: if present then the entry is a directory. The value is a map from `name` to entry objects.
- `perm`: the permissions to create the file with.

The `content` and `entries` fields are mutually exclusive, it is an error to include both, one or the other _must_ be present. The file or directory's name in each case is the same as the key which referred to that entry.
```

pkg/metadata/main.go
```diff
@@ -139,70 +139,80 @@ func main() {
...
-			continue
+
+	var queue []entry
+	var current entry
+	queue = enqueue(queue, root, basePath)
+	for len(queue) != 0 {
+		current, queue = queue[0], queue[1:]
+		if current.Content != nil {
+			writeFile(current)
+		} else {
```
ijc: I think we should check that exactly one of `Content` and `Entries` is non nil, i.e. that one but not both of them is set. That's reduce the possibility for confusion.

pkg/metadata/main.go
```diff
@@ -139,70 +139,80 @@ func main() {
...
-			if err := ioutil.WriteFile(path.Join(dir, f), []byte(c), os.FileMode(p)); err != nil {
-				log.Printf("Failed to write %s/%s: %s", dir, f, err)
-				continue
+func enqueue(queue []entry, entries map[string]entry, parent string) []entry {
+	for dir, entry := range entries {
+		entry.path = path.Join(parent, dir)
+		queue = append(queue, entry)
+	}
+	return queue
+}
```
ijc: I must confess I don't quite follow the logic here and in the caller, why not use a simple recursive function call?

docs/metadata.md
```diff
@@ -22,19 +22,25 @@ directories to be created and the directories are populated with files.
...
-        "baz" : {
-            "perm": "0600",
-            "content": "bar"
-        }
+  "ssh": {
+    "entries": {
+      "sshd_config": {
+        "perm": "0600",
+        "content": "PermitRootLogin yes\nPasswordAuthentication no"
+      }
```
ijc: This turns out not to be backwards compatible (sorry, I somehow convinced myself before that it would be).

I can't see a way to fix that without reintroducing the issue with having directory entries named `perm` etc, so I think we should just bite the bullet, but I leave the final determination on that to @rn .

rn: I'm fine with making this change not being backwards compat at this stage. We haven't done a release yet...I'll give a heads up on the community slack

pkg/metadata/main.go
guillaumerose: Entries is, with this struct, never nil. I have to use `*map[string]entry` for that, which is a weird type, no ?

pkg/metadata/main.go
guillaumerose: I thought you would prefer an iterative version in fact. I can revert.

pkg/metadata/main.go
ijc: I thought you had said you wanted to be able to write the json out with this too? Perhaps I misremember, but it seems like a reasonable usecase anyway and perhaps would even justify making it `type Entry` so it can be imported?

pkg/metadata/main.go
ijc: I don't think the iterative version has any benefits here and it is harder to reason about it's correctness.

The main reason to go with an iterative version would be malicious inputs trying to overflow the stack, but I think overflowing the stack is not as dangerous in Go as it is in, say, C. As far as I can tell overflowing the stack is a runtime panic, rather than an opportunity to corrupt the heap for fun and profit.

Also the stack is fairly plentiful in userspace and the size of an input file needed to recurse deeply enough to cause an issue would be prohibitively large in the first place even before one got to parsing it, let alone traversing it.

pkg/metadata/main.go
guillaumerose: I do the change for that also

justincormack: no, the empty map is `nil` but you have to initialize before writing. Go is weird. So you should be able to check if it is `nil` still...

pkg/metadata/main.go
guillaumerose: Yes, I will do that also. You're totally right. `Entry` instead of `entry`.

pkg/metadata/main.go
ijc: Sorry, I meant `null`/not present in the JSON sense, i.e. an empty map rather than a nil in the Go sense. (Is an uninitialised map == `nil` in Go? I'm not sure)

ijc: (I hadn't seen Justin's comment yet when I wrote the above)

justincormack: yes, uninitialised map == `nil` thats what I was saying, you dont need a pointer

ijc: we raced a second time ;-)

pkg/metadata/main.go
justincormack: if the metadata importer crashes due to bad input its will just not have imported some metadata, which seems an ok failure. A stack overflow is not really a security risk I dont think.

guillaumerose:
I exported what is needed to build the configuration file and added a serialization test. The code is recursive now. I also included @ijc changes in the documentation.

pkg/metadata/main.go
```diff
@@ -139,70 +139,82 @@ func main() {
...
-			default:
-				log.Printf("Couldn't convert JSON for items: %s", i)
-				continue
-			}
+	} else if isDirectory(current) {
+		filemode, err := parseFileMode(current.Perm, 0755)
+		if err != nil {
+			log.Printf("Failed to parse permission %s: %s", current, err)
+			return
+		}
```
ijc: Not terribly important but you could pull this out of both halves of the `if` if you wanted to.

ijc:
Hrm, did we drop the ability to just write a raw string (rather than a map) as an entry if you just want a basic file? That's a shame, but oh well RIP.

rn:
Thanks. I've build and pushed the package. Could you, in a separate commit in this PR update the YAML files to use `linuxkit/metadata:026aca5c08c22589a7e319f79449bef2c65f04c5`

If the CI is green afterwards it's good to merge.

MagnusS:
Looks like this also needs an update: https://github.com/linuxkit/linuxkit/blob/master/blueprints/docker-for-mac/metadata.json

guillaumerose:
Thanks @MagnusS, I've updated the file.

projects/kubernetes/kube.yml
```diff
@@ -18,7 +18,7 @@ onboot:
     image: linuxkit/dhcpcd:48831507404049660b960e4055f544917d90378e
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: metadata
-    image: linuxkit/metadata:9506d124d0a3ff645c9781c47f207423abf6154d
+    image: linuxkit/metadata:026aca5c08c22589a7e319f79449bef2c65f04c5
```
ijc: Please could you drop this one, I'll bump it when I sort out `projects/kubernetes/boot.sh` to match.

rn:
I force rebuild the package. the old/new image should be `inuxkit/metadata:026aca5c08c22589a7e319f79449bef2c65f04c5`. Could you update the YAML files (except the kube on as requested by @ijc)

guillaumerose:
I've updated the yaml. Thanks

PR #2705, opened 2017-11-10
linuxkit pkg: Handle build --dev later

ijc:
The placement in #2679 can never have worked since we parse the yml over the
struct later on.


rn:
I'm using `linuxkit pkg build -dev -org rneugeba` and get local images `rneugeba/<foo>:dev-amd64` just fine.

ijc:
This fixes `linuxkit pkg build -dev` which should produce `$USER/<foo>:dev`. Adding `-org` would indeed work today.

rn:
ah, ok. out of habit I always seem to spec `--org` anyway.

PR #2716, opened 2017-11-13
Add support for 4.14 kernels

rn:
This PR adds initial support for the recently released 4.14 kernels. The config file for 4.14 was derived from 4.13 and I kept the conversion in separate commits to make it easier to spot what has changed. This PR also tweaks the `Dockerfile.kconfig` as the untar started failing.

4.14, 4.14-dbg, 4.14-perf and 4.14-dbg-perf are build and pushed for x86_64 and arm64.

I've lightly tested the 4.14 kernel on:
- `x86_64`: HyperKit, qemu, packet.net type 0 and type 1
- `arm64`: qemu, kvm

There seems to be an issue with PXE booting this kernel on packet.net type 2a which requires some more investigation but this should not stop this PR from being merge.

pengus


kernel/Dockerfile.kconfig
```diff
@@ -17,7 +17,7 @@ RUN set -e && for VERSION in ${KERNEL_VERSIONS}; do \
         echo "Downloading/Unpacking $VERSION" && \
         KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${VERSION}.tar.xz && \
         [ -f sources/linux-${VERSION}.tar.xz ] || curl -fSLo sources/linux-${VERSION}.tar.xz ${KERNEL_SOURCE} && \
-        tar xf sources/linux-${VERSION}.tar.xz; \
+        bsdtar xf sources/linux-${VERSION}.tar.xz; \
```
justincormack: this should work with GNU tar with `-xjf`

justincormack:
lgtm

kernel/Dockerfile.kconfig
rn: It doesn't (with `-xJf`). I get the same error. I think this may be an issue with `overlay2` as google comes up with a bunch of hits with the same error

rn:
I added the remaining commit (tests) and update the subject/description of the PR. If folks (and CI) are happy this is ready to be merged

PR #2717, opened 2017-11-13
various typography

friism:
this is purely docs/README-related


docs/kernels.md
```diff
@@ -37,7 +37,6 @@ In summary, LinuxKit offers a choice of the following kernels:
 - [linuxkit/kernel-centos](https://hub.docker.com/r/linuxkit/kernel-centos/): Selected CentOS kernels.
 - [linuxkit/kernel-fedora](https://hub.docker.com/r/linuxkit/kernel-fedora/): Selected Fedora kernels.
 
-
```
rn: why are you deleting these empty lines? The file mostly uses two new-line before a section heading to make it visually clearer that a new section starts (though there are a couple of case where this is not followed).

friism: Those extra newlines waste valuable bandwidth when doing git-pull, etc.. I can make it consistent at 2 newline if you prefer though.

friism: @rn fixed

justincormack:
Commits need to be signed off, we can't review this otherwise.

PR #2725, opened 2017-11-15
update boot.sh metadata

w9n:
**- What I did**
use the new metadata scheme and combine it with KUBE_METADATA which can be either a file or json string. NOTE: KUBE_METADATA must be without root brackets.
load cni configuration on every boot from metadata directory and before the kubelet starts.
`kube-weave.yaml` gets created by makefile not `weave-sa.yaml`
**- How I did it**
start the data variable with `{` and add external metadata and/or init specific metadata. end data with `}`.
**- How to verify it**
set KUBE_METADATA to a string which is not a filename or a filename to load the file. Combine with `KUBE_MASTER_AUTOINIT`.

update Kubernetes metadata

### DIFF
--- a/projects/kubernetes/kubernetes/kubeadm-init.sh
+++ b/projects/kubernetes/kubernetes/kubeadm-init.sh
@@ -11,9 +11,6 @@ else
     kubeadm init --skip-preflight-checks --kubernetes-version @KUBERNETES_VERSION@ $@
 fi
 
-if [ -d /var/config/cni/etc/net.d ]; then
-  cp /var/config/cni/etc/net.d/* /var/lib/cni/etc/net.d/
-fi
 # sorting by basename relies on the dirnames having the same number of directories
 YAML=$(ls -1 /var/config/kube-system.init/*.yaml /etc/kubeadm/kube-system.init/*.yaml 2>/dev/null | sort --field-separator=/ --key=5)
 for i in ${YAML}; do
--- a/projects/kubernetes/kubernetes/kubelet.sh
+++ b/projects/kubernetes/kubernetes/kubelet.sh
@@ -21,6 +21,11 @@ if [ ! -e /var/lib/cni/.opt.defaults-extracted ] ; then
     touch /var/lib/cni/.opt.defaults-extracted
 fi
 
+if [ -d /var/config/cni/etc/net.d ]; then
+    mkdir -p /var/lib/cni/etc/net.d
+    cp /var/config/cni/etc/net.d/* /var/lib/cni/etc/net.d/
+fi
+
 await=/etc/kubernetes/kubelet.conf
 
 if [ -f "/etc/kubernetes/kubelet.conf" ] ; then


projects/kubernetes/boot.sh
```diff
@@ -16,26 +16,36 @@ set -e
 : ${KUBE_EFI:=}
 : ${KUBE_MAC:=}
 : ${KUBE_CLEAR_STATE:=}
+#: ${KUBE_MASTER_AUTOINIT:=}
```
ijc: Spurious change?

projects/kubernetes/boot.sh
```diff
@@ -16,26 +16,36 @@ set -e
...
 
 [ "$(uname -s)" = "Darwin" ] && KUBE_EFI=1
 
 suffix=".iso"
 [ -n "${KUBE_EFI}" ] && suffix="-efi.iso" && uefi="--uefi"
 
+data={
+if [ -n "${external_data}" ] ; then
+	if [ -s "${external_data}" ] ; then
+		external_data=$(tr '\n' ' ' <$external_data | tr -d '\t' )
```
ijc: Why is this transformation important? Perhaps a comment would be useful?

projects/kubernetes/boot.sh
```diff
@@ -16,26 +16,36 @@ set -e
...
+	fi
+	data+=$external_data
```
ijc: Is this `+=` a bashism? I can't find it described in http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html. Note that this script uses `/bin/sh`, `dash` is a more POSIXly correct shell and:
```
$ dash
$ foo="1"
$ foo+="1"
dash: 2: foo+=1: not found
```

projects/kubernetes/boot.sh
```diff
@@ -16,26 +16,36 @@ set -e
...
+	kubeadm_data="${kubeadm_data+$kubeadm_data, }\"init\": {\"content\": \"${KUBE_MASTER_AUTOINIT}\" }"
     fi
     if [ "${KUBE_MASTER_UNTAINT}" = "y" ] ; then
-	kubeadm_data="${kubeadm_data+$kubeadm_data, }\"untaint-master\": \"\""
+	kubeadm_data="${kubeadm_data+$kubeadm_data, }\"untaint-master\": {\"content\": \"\" }"
     fi
 
     if [ -n "${kubeadm_data}" ] ; then
-	data="{ \"kubeadm\": { ${kubeadm_data} } }"
+	data="${data+$data, } \"kubeadm\": {\"entries\": { ${kubeadm_data} } }"
```
ijc: `$data` is now always at least `{` and in that case I think this will produce invalid JSON since it will be `{, \"kubeadm...`

ijc:
Please rebase onto #2732 which fixes the metadata formatting.

w9n:
Accidentally deleted the branch. mostly done in #2732.

PR #2728, opened 2017-11-15
alpine: Remove the 'content trust build' workaround

arm64b:
Before the alpine image can be multi-arched and signed, the
DOCKER_CONTENT_TRUST=1 doesn't work on AArch64 based image build.
Now since we've used 'push_manifest.sh' to sign the multi-arch
supported alpine base image, we need to always enable
DOCKER_CONTENT_TRUST for the 'docker build' and DOCKER_CONTENT_PULL
for the 'docker pull' from the Perspective of security.

**- What I did**
Remove the workaround of 'content trust build' issue when build AArch64 docker image with `--arg`
**- How I did it**
Remove the NOTRUST and always exported both DOCKER_CONTENT_TRUST and DOCKER_CONTENT_TRUST as 1.
**- How to verify it**
Build the individual alpine image on both AArch64 and amd64 platform.

summit

### DIFF
--- a/tools/alpine/Makefile
+++ b/tools/alpine/Makefile
@@ -4,25 +4,10 @@ ORG?=linuxkit
 IMAGE=alpine
 DEPS=packages
 
-# The logic for content trust is a bit convoluted because:
-# - The arm64 base image is currently not signed so we need to pull it
-#   with content trust disabled. This is controlled by
-#   DOCKER_CONTENT_PULL.
-# - 'docker build' with the FROM image supplied as environment
-#   variable *and* with DOCKER_CONTENT_TRUST=1 currently does not work
-#   (https://github.com/moby/moby/issues/34199). We therefor build
-#   with DOCKER_CONTENT_TRUST explicitly set to 0. However, we pull
-#   the base image just before with content trust enabled (if
-#   supported, see above).
-# - By default we always pull and push the linuxkit/alpine image with
-#   content trust, unless explicitly disabled with NOTRUST. Once the
-#   above issues are resolved, this will be the only mechanism to control
-#   content trust.
-ifdef NOTRUST
-DOCKER_CONTENT_PULL=0
-else
-DOCKER_CONTENT_PULL=1
+ifeq ($(DOCKER_CONTENT_TRUST),)
+ifndef NOTRUST
 export DOCKER_CONTENT_TRUST=1
+endif
 endif
 
 ARCH := $(shell uname -m)
@@ -41,7 +26,7 @@ show-tag:
 	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions.$(ARCH)
 
 iid: Dockerfile Makefile $(DEPS)
-	DOCKER_CONTENT_TRUST=1 docker build --no-cache --iidfile iid .
+	docker build --no-cache --iidfile iid .
 
 hash: Makefile iid
 	docker run --rm $(shell cat iid) sh -c 'echo Dockerfile /lib/apk/db/installed $$(find /mirror -name '*.apk' -type f) $$(find /go/bin -type f) | xargs cat | sha1sum' | sed 's/ .*//' | sed 's/$$/$(SUFFIX)/'> $@


arm64b:
Interesting, why this PR will affect the build of iso-efi image?
`[+ 12m 33s] [FAIL    ] linuxkit.build.outputs.iso-efi 10.39s`

justincormack:
Actually half the code got fixed in https://github.com/linuxkit/linuxkit/commit/181096f93515ee5feeb65e834988f0d89264de14#diff-6e2b22c01cc5078acc6ae98a99dc3117

we should be able to remove all of this part at the top now I think, `DOCKER_CONTENT_PULL` is not used, and `DOCKER_CONTENT_TRUST` is explicitly set further down for build.

arm64b:
> we should be able to remove all of this part at the top now I think, DOCKER_CONTENT_PULL is not used, and DOCKER_CONTENT_TRUST is explicitly set further down for build.

Hello @justincormack, so I understand that you suggest to remove all the `DOCKER_CONTENT_*`, is that right?

justincormack:
Yes, I don't think any of it is now needed.

arm64b:
OK, I also think so. I will force an update.

tools/alpine/Makefile
```diff
@@ -4,27 +4,6 @@ ORG?=linuxkit
...
-#   supported, see above).
-# - By default we always pull and push the linuxkit/alpine image with
-#   content trust, unless explicitly disabled with NOTRUST. Once the
-#   above issues are resolved, this will be the only mechanism to control
-#   content trust.
-ifdef NOTRUST
-DOCKER_CONTENT_PULL=0
-else
-DOCKER_CONTENT_PULL=1
-export DOCKER_CONTENT_TRUST=1
```
rn: This seems to completely disable DCT

arm64b: I can't see the necessary that we need to disable the DCT in any case given the issue we encountered before has been resolved (refer to the https://github.com/moby/moby/issues/34199), at that time, the Alpine base image is not multi-arch and not signed for AArch64, so we introduce the NO_TRUST as AArch64 specific workaround , but now it's not the case.

rn: I don't understand. `alpine:3.6` *is* signed. So is the `linuxkit/alpine`. So everything should have content trust enabled, and if my reading of this patch is correct it disables it

arm64b: Ah, if the default `docker build` behavior is DCT disabled, then you're right. is it?
But anyway, set DOCKER_CONTENT_TRUST=1 before the following `docker build` in this Dockerfile will pretty ease your concern here.  I give the previous comments because you quoted code is not in `docker build` followed. Will force update soon.

arm64b: Actually IMO we don't need an explicit `DOCKER_CONTENT_TRUST=1' prefixed before `docker build --no-cache --iidfile iid .`, just as you pointed: both `alpine:3.6` and `linuxkit/alpine` are or will be signed.

ijc: AIUI without `DOCKER_CONTENT_TRUST=1` nothing will check the signatures, so it is needed. Please do the same as `kernel/Makefile` which is:
```
ifeq ($(DOCKER_CONTENT_TRUST),)
ifndef NOTRUST
export DOCKER_CONTENT_TRUST=1
endif
endif
```
And remove the other settings of `DOCKER_CONTENT_TRUST` in this `Makefile`.

`NOTRUST` is useful for people doing local or development builds, perhaps in other orgs or with other base images.

arm64b: Just googled the DCT purpose, `Once you set this environment variable, every Docker operation is secure. This means that any future Docker commands are executed based on signatures.`, so my understanding is, we need to use 
```
ifeq ($(DOCKER_CONTENT_TRUST),)
ifndef NOTRUST
export DOCKER_CONTENT_TRUST=1
endif
endif
```
the same block as kernel  into this PR's and remove other DCT variable prefixed.
Right? @ijc

ijc: > Right? @ijc

Since you repeated exactly what I said I of course agree.

PR #2732, opened 2017-11-15
kubernetes: Update to new metadata format, rename packages

ijc:
**- What I did**

Updated `projects/kubernetes/boot.sh` post #2703 and bumped yml to the latest metadata package.

rename images to better reflect what they are.
* `linuxkitprojects/kubernetes` → `linuxkitprojects/kubelet`
* `linuxkitprojects/kubernetes-image-cache-*` → `linuxkitprojects/kubernetes-docker-image-cache-*`
    
`kubernetes` subdirectory is also renamed to `kubelet`.
    
`image-cache` subdirectory is not renamed since we may want to build other sorts of image cache at some point.

**- How I did it**

emacs mostly

**- How to verify it**

Having built some images I tested the metadata fixes in particular with

**For master:**
* `KUBE_MASTER_UNTAINT=y ./boot.sh` and confirm `/var/config/kubeadm/untaint-master` is present but empty.
* `KUBE_MASTER_AUTOINIT='' ./boot.sh` and confirm `/var/config/kubeadm/init` is present but empty.
* `KUBE_MASTER_UNTAINT=y KUBE_MASTER_AUTOINIT='--node-name=linuxkit-master' ./boot.sh` and confirm /var/config/kubeadm/untaint-master is present but empty and /var/config/kubeadm/init exists and contains '--node-name=linuxkit-master'.

**For node:**
* `./boot.sh 1 --token «etc»` and confirm /var/config/kubeadm/join exists and contains '--token «etc»'

[](https://morefuzz.net/reviews/kamchatka-long-road-made-gold/)


ijc:
Added an extra commit to pass the metadata via a file rather than inline, which I meant to do the first time around and forgot.

ijc:
Added one more fix which @justincormack just alerted me to (regression from #2700)

PR #2734, opened 2017-11-15
kubernetes: fix cni config by metadata for nodes

w9n:
**- What I did**
copy cni settings in `kubelet.sh`
**- How I did it**
vim
**- How to verify it**
write some metadata into "cni/etc/net.d/"
kubernetes: fix cni config by metadata for nodes

### DIFF
--- a/projects/kubernetes/kubelet/kubelet.sh
+++ b/projects/kubernetes/kubelet/kubelet.sh
@@ -21,6 +21,12 @@ if [ ! -e /var/lib/cni/.opt.defaults-extracted ] ; then
     touch /var/lib/cni/.opt.defaults-extracted
 fi
 
+if [ ! -e /var/lib/cni/.cni.configs-extracted && -d /var/config/cni/etc/net.d ] ; then
+    mkdir -p /var/lib/cni/etc/net.d
+    cp /var/config/cni/etc/net.d/* /var/lib/cni/etc/net.d/
+    touch /var/lib/cni/.cni.configs-extracted
+fi
+
 await=/etc/kubernetes/kubelet.conf
 
 if [ -f "/etc/kubernetes/kubelet.conf" ] ; then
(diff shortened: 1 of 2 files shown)


ijc:
The description here is missing the "why?" part.

The reason this was initially done in `kubeadm-init.sh` since it is setup which is done once on cluster initialisation (which corresponds to `kubeadm-init` time) and not repeated on every reboot, otherwise changes made subsequently (e.g. by deploying some relevant daemonset) would potentially be overwritten or conflicting. I don't think reboot is an appropriate time to be reconfiguring a cluster.

So if this is to move it would need to add a gate on a stamp file I think, and given the above I'm not sure it should move at all.

w9n:
it doesnt work on nodes because `elif [ -e /var/config/kubeadm/join ] ; then` doesnt use `kubeadm-init.sh`. Im fine with init time here and changed appropriately, I hope. Not sure if cni configuration on runtime or init-time in general is better.

w9n:
i wasnt sure about some behavior in the beginning. Simplified as suggested and should be fine now.

projects/kubernetes/kubelet/kubelet.sh
```diff
@@ -21,6 +21,12 @@ if [ ! -e /var/lib/cni/.opt.defaults-extracted ] ; then
     touch /var/lib/cni/.opt.defaults-extracted
 fi
 
+if [ ! -e /var/lib/cni/.opt.configs-extracted ] ; then
+    mkdir -p /var/lib/cni/etc/net.d
+    cp /var/config/cni/etc/net.d/* /var/lib/cni/etc/net.d/
```
ijc: What happens if `/var/config/cni/etc/net.d` doesn't exist. (Hint: nothing good)

Please name the stamp `.etc-net.d-configs-extracted` (the `opt` in the example you cribbed from is because it is populating `/opt/cni`.

justincormack:
Sorry can you reopen this on https://github.com/linuxkit/kubernetes

PR #2736, opened 2017-11-15
cAdvisor

mmariani:
**- What I did**

Added a cAdvisor package with example file.

### DIFF
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -1,0 +1,59 @@
+kernel:
+  image: linuxkit/kernel:4.9.62
+  cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
+init:
+  - linuxkit/init:42a92119e1ca10380e0d33e26c0cbcf85b9b3558
+  - linuxkit/runc:1b0741d07949c0acc444cd6a04ee7f833443579d
+  - linuxkit/containerd:bfb61cc1d26c39cd4b2bc08f7a9963fefa0ef3bf
+  - linuxkit/ca-certificates:af4880e78edc28743f7c5e262678c67c6add4c26
+onboot:
+  - name: sysctl
+    image: linuxkit/sysctl:a9ad57ed738a31ea9380cd73236866c312b35489
+  - name: dhcpcd
+    image: linuxkit/dhcpcd:48831507404049660b960e4055f544917d90378e
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+  - name: sysfs
+    image: linuxkit/sysfs:5367b46211882278b84a9e8048855ca5df65beda
+  - name: format
+    image: linuxkit/format:6b46d0450082f397177da36be6b4d74d93eacd1e
+  - name: mount
+    image: linuxkit/mount:41685ecc8039643948e5dff46e17584753469a7a
+    command: ["/usr/bin/mountie", "/var/lib/docker"]
+
+services:
+  - name: getty
+    image: linuxkit/getty:6af22c32c98536a79230eef000e9abd06b037faa
+    env:
+      - INSECURE=true
+  - name: rngd
+    image: linuxkit/rngd:842e5e8ece7934f0cab9fd0027b595ff3471e5b9
+  - name: ntpd
+    image: linuxkit/openntpd:07a80c3e3e816658318ac027e1253ff9a228b8de
+
+  - name: docker
+    image: docker:17.10.0-ce-dind
+    capabilities:
+      - all
+    net: host
+    mounts:
+      - type: cgroup
+        options: ["rw", "nosuid", "noexec", "nodev", "relatime"]
+    binds:
+      - /etc/resolv.conf:/etc/resolv.conf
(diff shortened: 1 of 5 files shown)


pkg/cadvisor/waitfordocker.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/ash
```
justincormack: please use `/bin/sh` not explicitly choosing `ash`

pkg/cadvisor/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM google/cadvisor:v0.28.0@sha256:09c8d73c9b799d30777763b7029cfd8624b8a1bd33af652ec3b51a6b827d492a
```
rn: This is x86_64 only, so `build.yml` below muct restrict the arch to that. Alternatively, it could build from source and do so for x86_64 and arm64

mmariani: building from source is not done even in the upstream Dockerfile. What about "go get github.com/google/cadvisor" ?

mmariani: the problem in this case, would be repeatability. i'm not familiar enough with go tools to install a specific commit

pkg/cadvisor/waitfordocker.sh
mmariani: oh sure, I tried too hard to silence shellcheck.

pkg/cadvisor/Dockerfile
mmariani: I've looked it up. I suppose go get -d ... git branch ... make

rn: @mmariani you may want to model it after https://github.com/linuxkit/linuxkit/tree/master/pkg/vpnkit-expose-port

mmariani: ok, will do. thanks

examples/cadvisor.yml
```diff
@@ -0,0 +1,60 @@
+---
```
riyazdf: can we remove this line?

mmariani: Sure, my yamllint was picky

pkg/cadvisor/Dockerfile
mmariani: Would the current version be ok? I did a static binary, in case the shell script is not required anymore in the future, although it seems to use dlopen() for some feature (nvidia monitoring, etc)

rn: I'd be fine with the current version but then it also needs to have:
```
arches:
  - amd64
```
in the `build.yml`

rn:
This also needs a rebase as it seems to have picked up some other commits

pkg/cadvisor/Dockerfile
mmariani: Sorry, I don't understand. It builds from sources and should work on arm64 too, I only used the linuxkit/alpine image. I see others use pretty much the same dockerfile for both: https://github.com/kontena/cadvisor-docker , am I missing something?

rn: My apologies. The GH UI showed me the old diff and I didn't spot the refresh

mmariani:
I've just found this commit, reverting the static build (which I reinstated in the dockerfile):

https://github.com/google/cadvisor/commit/318f28bef6dc49efb547b0d22398d942d803baab

It should work fine except for the nvidia library, which afaik is not available for ARM:

https://developer.nvidia.com/gpu-deployment-kit

pkg/cadvisor/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM linuxkit/alpine:07f7d136e427dc68154cd5edbb2b9576f9ac5213 as build
+
+RUN apk add --no-cache bash go git g++ linux-headers make
```
rn: are `g++` and `linux-headers` needed? They look like they were copy/pasted from somewhere...

pkg/cadvisor/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ADD /static.patch /tmp/
+
+RUN go get -d ${GITREPO} \
+    && cd /go/src/${GITREPO} \
+    && git checkout ${COMMIT} \
+    && patch -p1 build/build.sh </tmp/static.patch \
+    && make build \
+    && mv cadvisor /usr/bin/
+
+# -----------------------------------------------------------
```
rn: similar to my comment on https://github.com/linuxkit/linuxkit/pull/2754 I prefer to not having these dividers. Minor nit, though

pkg/cadvisor/Dockerfile
mmariani: Yes they are required, or at least musl-dev and linux-headers. It looks for netlink, pthreads etc.

pkg/cadvisor/Dockerfile
mmariani: of course

rn:
Thanks for the updates. They LGTM but I need to set up `linuxkit/cadvisor` with signing and am running into some issues. Will merge once they are resolved (and hopefully no rebase required)

PR #2737, opened 2017-11-15
Only add in ttyS1 to kernel cmdline on Packet if not there already

### LINKED ISSUE #2735: console not working correctly on packet
Getting garbage along with what I type. Investigating...

justincormack:
This is not in most examples, but is in the Packet example, and
causes a duplicated console.

fix #2735

term_cat

### DIFF
--- a/src/cmd/linuxkit/run_packet.go
+++ b/src/cmd/linuxkit/run_packet.go
@@ -14,6 +14,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


ijc:
I think `pkg/getty` (which I suppose is the thing acting on this to produce the multiple gettys) ought to be dedupping (perhaps in addition to this fix)

justincormack:
@ijc yeah, I will open another issue for that, it would be hygenic.

justincormack:
See also #2741

PR #2738, opened 2017-11-15
moby no longer uses efi in EFI ISO suffix

justincormack:
This was changed a while back and the boot method is no longer
encoded in any of the output names.

cat-iso

### DIFF
--- a/projects/kubernetes/boot.sh
+++ b/projects/kubernetes/boot.sh
@@ -20,7 +20,7 @@ set -e
 [ "$(uname -s)" = "Darwin" ] && KUBE_EFI=1
 
 suffix=".iso"
-[ -n "${KUBE_EFI}" ] && suffix="-efi.iso" && uefi="--uefi"
+[ -n "${KUBE_EFI}" ] && uefi="--uefi"
 
 if [ $# -eq 0 ] ; then
     img="kube-master"


justincormack:
Yeah I was wondering about that, just looking at another issue, so probably will rebase on #2732 anyway

justincormack:
Although in future we will want to support non ISO outputs...

ijc:
It could well be that adding non-iso will push this script to breaking point...

Are you sure the suffix is gone though?
```
$ moby build -format iso-efi -format iso-bios linuxkit.yml 
...
Create outputs:
  linuxkit-efi.iso
  linuxkit.iso
$ ls -l linuxkit*.iso
-rw-rw-r-- 1 ijc ijc 170323968 Nov 15 16:36 linuxkit-efi.iso
-rw-rw-r-- 1 ijc ijc 170917888 Nov 15 16:36 linuxkit.iso
$ moby version
moby version 0.0
commit: 99480b5dd01b18ff2c80a2ce33ad46a436ccdb25
$ git grep MOBY_COMMIT= origin/master:Makefile
origin/master:Makefile:MOBY_COMMIT=99480b5dd01b18ff2c80a2ce33ad46a436ccdb25
```

justincormack:
no, thats what I am looking at the platform selection is borked... I did think we had changed it (and we should...)

PR #2740, opened 2017-11-15
wireguard: update to 0.0.20171111

zx2c4:
Routine version bump that also removes the necessity of carrying that extra patch. Changes:

```
  * Kconfig: remove trailing whitespace
  * allowedips: rename from routingtable
  * tools: remove ioctl cruft
  * global: revert checkpatch.pl changes

  Cleanliness.

  * device: please lockdep
  * device: wait for all peers to be freed before destroying

  These make the various checkers happy.

  * netlink: plug memory leak
  * qemu: check for memory leaks

  There was a small memory leak on the netlink configuration layer that's now
  been fixed.

  * receive: hoist fpu outside of receive loop

  Should be a small speedup on x86_64.

  * qemu: more debugging
  * qemu: bump kernel version

  Significantly more debugging checkers have been turned on.

  * wg-quick: stat the correct enclosing folder of config file
  * wg-quick: allow for tabs in keys

  Minor fixups for wg-quick(8).

  * compat: 4.4.0 has strange ECN function

  Nobody actually runs base 4.4.0, but this is more correct anyway.

  * netlink: make sure we reserve space for NLMSG_DONE

  A rather important change - due to an upstream kernel bug, that's existed
  since the advent of netlink itself, sometimes wg(8) failed to receive valid
  data back from kernelspace, resulting in "ENOBUFS" when trying to dump all
  peers. This patch works around it while we wait for upstream to commit the
  fix.

  * curve25519: reject deriving from NULL private keys
  * tools: allow for NULL keys everywhere

  A null 25519 private point isn't a valid point (prior to normalization), which
  is why we use it as the "unsetting" value. Conversely, however, except for
  psk, we should be using the existence of it in the netlink message being an
  indication of whether or not it's set, for the tools.
```

image

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -37,11 +37,9 @@ ENV KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VE
 ENV KERNEL_SHA256_SUMS=https://www.kernel.org/pub/linux/kernel/v4.x/sha256sums.asc
 ENV KERNEL_PGP2_SIGN=https://www.kernel.org/pub/linux/kernel/v4.x/linux-${KERNEL_VERSION}.tar.sign
 
-ENV WIREGUARD_VERSION=0.0.20171101
-ENV WIREGUARD_SHA256=096b6482a65e566c7bf8c059f5ee6aadb2de565b04b6d810c685f1c377540325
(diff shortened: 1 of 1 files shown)


zx2c4:
Corresponding Alpine PR is at https://github.com/alpinelinux/aports/pull/2731

rn:
Thanks. Will pick this up with the next kernel update (possibly tomorrow)

Issue #2741, opened 2017-11-15
getty package should check for duplicate tty


justincormack:
if you end up with (perhaps for tooling issues) `console=ttyS1 console=ttyS1` on the kernel command line the console will be unusable for entering data, as the two `agetty` processes will complete. The script should try to de-dup.

deitch:
I am confused. Are you saying that if the same `console=` appears twice, then it will start two getty processes, which will trounce each other?

rn:
@deitch yes

deitch:
Ok. Will take care of this.

deitch:
See #2742

PR #2742, opened 2017-11-15
Keep track of processed ttys and only start same one once

deitch:

**- What I did**

Added deduplication logic to `linuxkit/getty` so that if the same line appears twice as `console=`, it will only start each once.

Also logs what it is doing.

Fixed [2741](https://github.com/linuxkit/linuxkit/issues/2741)

**- How I did it**
Updated `/usr/bin/rungetty.sh`

**- How to verify it**

1. Run a linuxkit `.yml` file with duplicate `console=` lines

```yml
  cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttyS0 console=tty0"
```

2. `moby build`
3. `linuxkit run`
4. Log in, see that `/var/log/getty.out.log` shows it starting only once. Also verify with `ps -ef` if you want.

Prevent `getty` from starting getty twice for same tty

baboob

### DIFF
--- a/pkg/getty/usr/bin/rungetty.sh
+++ b/pkg/getty/usr/bin/rungetty.sh
@@ -15,6 +15,14 @@ start_getty() {
 	term="linux"
 	[ "$speed" = "$1" ] && speed=115200
 
+	# did we already process this tty?
+	if $(echo "${PROCESSEDTTY}" | grep -q -w "$tty"); then
+		echo "getty: already processed tty for $tty, not starting twice" | tee /dev/console
+		return
+	fi
+	# now indicate that we are processing it
+	PROCESSEDTTY="${PROCESSEDTTY} ${tty}"
+
 	# does the device even exist?
 	if [ ! -c /dev/$tty ]; then
 		echo "getty: cmdline has console=$tty but /dev/$tty is not a character device; not starting getty for $tty" | tee /dev/console
@@ -44,6 +52,7 @@ start_getty() {
 		echo "getty: cmdline has console=$tty but does not exist in $securetty; will not be able to log in as root on this tty $tty." | tee /dev/$tty
 	fi
 	# respawn forever
+	echo "getty: starting getty for $tty"  | tee /dev/$tty
 	infinite_loop setsid.getty -w /sbin/agetty $loginargs $line $speed $tty $term &
 }
 
@@ -54,6 +63,8 @@ if [ -z "$INITGETTY" ]; then
 export PS1="(ns: getty) $PS1"
 EOF
 fi
+
+PROCESSEDTTY=
 
 # check if we have /etc/getty.shadow
 ROOTSHADOW=/hostroot/etc/getty.shadow


deitch:
Pushed out in arm64 and x86

pkg/getty/usr/bin/rungetty.sh
```diff
@@ -15,6 +15,14 @@ start_getty() {
 	term="linux"
 	[ "$speed" = "$1" ] && speed=115200
 
+	# did we already process this tty?
+	if $(echo "${PROCESSEDTTY}" | grep -q -w "$tty"); then
+		echo "getty: already processed tty for $tty, not starting twice" | tee /dev/console
+		return
+	fi
+	# now indicate that we are processing it
+	PROCESSEDTTY="${PROCESSEDTTY} ${tty}"
```
ijc: One way to achieve a similar thing without quite so much shell gymnastics and grep etc would be `touch /var/lock/getty.${tty}` and `if [ -e /var/lock/getty.${tty} ]`.

deitch: No preference. I am happy to leave it this way or change it. @ijc and @justincormack do you care?

justincormack: I think files are nicer than env vars. I would use `/tmp` though I think as its really only temporary to a short run of this program.

deitch: It isn't really an env var, just a var in the context of the script (probably should not have named it in caps...).

deitch: The more I think about it, more I think I would prefer to leave it as is. This isn't intended to survive process restarts or anything, as you said:

> its really only temporary to a short run of this program.

so why not just leave it as memory (var) inside program?

ijc: The machinations to extract things from the var are the thing I don't like, rather than the fact of the var itself. Files would be cleaner in that regard.

deitch: > machinations to extract things from the var

deitch:
This one is open 6 weeks later, can we just take it as is?

justincormack:
I think there was a preference for using files...

deitch:
> I think there was a preference for using files...

I was neutral, but I kind of dislike polluting the filesystem to keep track of things that are tracked in a single process. Why not keep it in memory?

rn:
Is there a desire to move this forward or shall we close this? It's over a year old?

deitch:
I’m neutral on it. It’s the right thing to do, but no one really has said anything about it in a year. If we still want it, I’ll redo it for memory based tracking.

nustiueudinastea:
I also prefer the current way it's implemented (using grep in the script) because it is more robust. There is not need to leave state in the filesystem, which can lead to situation where the file exists but the process does not. Any chance this will be merged? Thanks!

deitch:
Is this over a year and a half old already? Wow.

Reviewing this, I _definitely_ prefer env vars over files. Even better would be if it actually could check the tty state itself, but this is good enough for now. I am going to rebase and push out to kick off CI again. If it passes, I would like to merge it in.

deitch:
Rebased. Waiting for CI to finish.

PR #2743, opened 2017-11-16
After partitioning we now retry the stat prior to formatting

dprotaso:
**- What I did**
After partitioning make waiting for the device durable again

**- How I did it**

stat can fail - simply don't exit when it does

**- How to verify it**

This is hard to reproduce but we were getting this bug with fresh disks numerous times today

format pkg has retries `stat` again when waiting for the partition device

### DIFF
--- a/pkg/format/format.go
+++ b/pkg/format/format.go
@@ -100,13 +100,15 @@ func refreshDevicesAndWaitFor(awaitedDevice string) error {
 	exec.Command("mdev", "-s").Run()
 
 	// wait for device
-	var done bool
+	var (
+		done bool
+		err  error
+		stat os.FileInfo
+	)
+
 	for i := 0; i < timeout; i++ {
-		stat, err := os.Stat(awaitedDevice)
-		if err != nil {
-			return err
-		}
-		if isBlockDevice(&stat) {
+		stat, err = os.Stat(awaitedDevice)
+		if err == nil && isBlockDevice(&stat) {
 			done = true
 			break
 		}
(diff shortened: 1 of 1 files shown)


pkg/format/format.go
```diff
@@ -100,21 +100,27 @@ func refreshDevicesAndWaitFor(awaitedDevice string) error {
...
 		time.Sleep(100 * time.Millisecond)
 		exec.Command("mdev", "-s").Run()
 	}
 	if !done {
-		return fmt.Errorf("Error waiting for device %s", awaitedDevice)
+		var statMsg string
+		if err != nil {
+			statMsg = fmt.Sprintf(" - stat returned: %v", err)
+		}
+		return fmt.Errorf("Error waiting for device %s%s", awaitedDevice, statMsg)
```
eyz: Maybe change to this instead? 
`return fmt.Errorf("Failed to find block device %s%s", awaitedDevice, statMsg)`

justincormack:
@dave-tucker is in SF at the minute so might be a delay reviewing...

justincormack:
Which error are you getting from `stat` when it fails?

rn:
I would also be good to add a bit more detail to the commit message (instead of the PR). This way they are stored in git.

dprotaso:
/dev/vda1 not found

rn:
@dprotaso thanks for the update. The changes look good to me. Could you update all uses of the `format` package in the linuxkit repo *in a separate commit*. I've build/pushed/signed the udapte and the new tag is: `linuxkit/format:6b46d0450082f397177da36be6b4d74d93eacd1e`

There is a script which helps you updating the YAML files. From the top-level:
```
./scripts/update-component-sha.sh linuxkit/format:10e75e78e1f134d310c62e9a0352df1c67b0dd81 linuxkit/format:6b46d0450082f397177da36be6b4d74d93eacd1e
```
should do the right thing and allow CI to test your change.

justincormack:
Will build the new package...

rn:
@justincormack new package is already build/pushed. Just needs update of YAMLs

PR #2746, opened 2017-11-16
Create state dir before trying to touch file in it

justincormack:

flying-monkey-yes-it-is-possible-10-pictures-5

### DIFF
--- a/projects/kubernetes/boot.sh
+++ b/projects/kubernetes/boot.sh
@@ -82,6 +82,7 @@ if [ -n "${KUBE_CLEAR_STATE}" ] ; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


ijc:
Doh, I considered this but overlooked the `$KUBE_CLEAR_STATE` is set but directory doesn't already exist case!

I think this needs to be `mkdir -p` to account for the case where `$KUBE_CLEAR_STATE` being set has already made the dir, and given that perhaps in the `else` half of that code block is a better place for this?

justincormack:
Ah yes, added `-p`. It looked a bit weird in an `else` clause so I left it...

Issue #2349, opened 2017-08-03
linuxkit/node_exporter does not build


rn:
With recent changes we pull with content trust enabled, and the prometheus image is not signed so it fails. It's disabled for now, but should get fixed and re-enabled.

ijc:
Is this used anymore? Does this predate the addition of TRUST=0 support in the build? Or do we not want to disable that globally just to disable it for the FROM?

rn:
probably pre-dates `TRUST=0` but we want to push the image with content trust, we just can't pull the base image with content trust.

Another option is to remove this package for now and add a replacement later which just builds it form source.

ijc:
I was looking at it due to https://github.com/linuxkit/linuxkit/issues/2564. I think I'll add some sort of bodge rather than actually fix this package or add anything to the `linuxkit pkg build` I'm running (like I might just misname the yml file so the directory is skipped)

PR #2754, opened 2017-11-17
Node exporter

mmariani:
**- What I did**

Built node-exporter from sources. Static, minimal image. Used 0.15.0 since the 0.15.1 tag exists only on dockerhub.

This aims to fix https://github.com/linuxkit/linuxkit/issues/2349 but the current PR does not pass the test-pr target. Any idea why? It's not clear to me from the logs.

### DIFF
--- a/pkg/node_exporter/Dockerfile
+++ b/pkg/node_exporter/Dockerfile
@@ -1,7 +1,27 @@
-FROM prom/node-exporter:v0.15.1@sha256:88c602650e861bb8045ac83421edecbd056fc58708d47554a0d1a10f19ed7556
+FROM linuxkit/alpine:07f7d136e427dc68154cd5edbb2b9576f9ac5213 as build
+
+RUN apk add --no-cache go git musl-dev make
+
+ENV GOPATH=/go PATH=$PATH:/go/bin
+ENV GITREPO=github.com/prometheus/node_exporter
+ENV COMMIT=v0.15.1
+
+RUN go get -d ${GITREPO} \
+    && cd /go/src/${GITREPO} \
+    && git checkout ${COMMIT} \
+    && CGO_ENABLED=0 make build \
+    && mv node_exporter /bin/
+
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=build /bin/node_exporter /bin/node_exporter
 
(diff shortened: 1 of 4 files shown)


pkg/node_exporter/Dockerfile
```diff
@@ -1,7 +1,28 @@
...
+
+ENV GOPATH=/go PATH=$PATH:/go/bin
+ENV GITREPO=github.com/prometheus/node_exporter
+ENV COMMIT=v0.15.0
+
+RUN go get -d ${GITREPO} \
+    && cd /go/src/${GITREPO} \
+    && git checkout ${COMMIT} \
+    && CGO_ENABLED=0 make build \
+    && mv node_exporter /usr/bin/
```
rn: minor nit, but here you move it to `/usr/bin` and below it is then copied to `/bin`. Either move int to `/bin` or something like `/out`.

pkg/node_exporter/Dockerfile
```diff
@@ -1,7 +1,28 @@
-FROM prom/node-exporter:v0.15.1@sha256:88c602650e861bb8045ac83421edecbd056fc58708d47554a0d1a10f19ed7556
+FROM linuxkit/alpine:07f7d136e427dc68154cd5edbb2b9576f9ac5213 as build
+
+RUN apk add --no-cache go git g++ make
```
rn: is `g++` needed? Other pure go packages tend to just install `go musl-dev`, but of course you also need `make` here because it's used below.

pkg/node_exporter/Dockerfile
```diff
@@ -1,7 +1,28 @@
...
+ENV GITREPO=github.com/prometheus/node_exporter
+ENV COMMIT=v0.15.0
+
+RUN go get -d ${GITREPO} \
+    && cd /go/src/${GITREPO} \
+    && git checkout ${COMMIT} \
+    && CGO_ENABLED=0 make build \
+    && mv node_exporter /usr/bin/
+
+# -----------------------------------------------------------
```
rn: Super minor nit, but I'm not a huge fan of these type of dividers and we don't use them elsewhere

mmariani:
PR updated. style consistency is never "minor" :)

rn:
Thanks. I've build/pushed/signed the images and hit the rebuild button on the CI

PR #2755, opened 2017-11-17
Add support for creating cgroups in runtime section

justincormack:
Implements https://github.com/moby/tool/pull/181

Design for things like Kubernetes setup that requires some cgroups to
exist when the service starts but it is not running in these, other
services are, so there would be a race if they are not created in each.

Essentially it is just a sugared `mkdir` in all the cgroup dirs.

group of wolves

### DIFF
--- a/pkg/init/cmd/service/prepare.go
+++ b/pkg/init/cmd/service/prepare.go
@@ -14,15 +14,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const (
-	wgPath      = "/usr/bin/wg"
-	nsenterPath = "/usr/bin/nsenter-net"
-)
-
 // Note these definitions are from moby/tool/src/moby/config.go and should be kept in sync
 
 // Runtime is the type of config processed at runtime, not used to build the OCI spec
 type Runtime struct {
+	Cgroups    []string      `yaml:"cgroups" json:"cgroups,omitempty"`
 	Mounts     []specs.Mount `yaml:"mounts" json:"mounts,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/init/cmd/service/prepare.go
```diff
@@ -14,15 +14,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const (
-	wgPath      = "/usr/bin/wg"
-	nsenterPath = "/usr/bin/nsenter-net"
-)
-
```
ijc: It built in CI so I suppose this is intentional/harmless?

justincormack: yeah I just happened to spot that, its not used any more...

pkg/init/cmd/service/prepare.go
```diff
@@ -122,7 +118,27 @@ func parseMountOptions(options []string) (int, string) {
 	return flag, strings.Join(data, ",")
 }
 
-// prepareFilesystem sets up the mounts, before the container is created
+// newCgroup creates a cgroup (ie directory) under all directories in /sys/fs/cgroup
+// we could use github.com/containerd/cgroups but it has a lot of deps and this is just a sugary mkdir
+func newCgroup(cgroup string) error {
+	dirs, err := ioutil.ReadDir("/sys/fs/cgroup")
```
ijc: The docs weirdly don't say, but I suppose this omits `.` and `..`?

justincormack: yeah it does... not sure how you get them if you need them, use `x/sys/unix` I guess...

justincormack:
ping @ijc

PR #2761, opened 2017-11-20
Restore linuxkit build

justincormack:
Restore the `linuxkit build` functionality by using `moby/tool` as a library.

This is simpler for users as they can just use `linuxkit build` and `linuxkit run`, and also means that there are fewer issues if people update out of sync (eg if they are using homebrew). Using `moby` directly is of course still supported, and is the best option if you are modifying the `moby` functionality for example.

wandering albatross


src/cmd/linuxkit/build.go
```diff
@@ -0,0 +1,215 @@
+package main
```
rn: This file is more or less a verbatim copy of https://github.com/moby/tool/blob/master/cmd/moby/build.go and I'm not too keen on the duplication...

src/cmd/linuxkit/vendor.conf
```diff
@@ -1,34 +1,53 @@
...
 github.com/vmware/govmomi 6f8ebd89d521d9f9af7a6c2219c4deee511020dd
+github.com/xeipuuv/gojsonpointer 6fe8760cad3569743d51ddbb243b26f8456742dc
+github.com/xeipuuv/gojsonreference e02fc20de94c78484cd5ffb007f8af96be030a45
+github.com/xeipuuv/gojsonschema 702b404897d4364af44dc8dcabc9815947942325
 golang.org/x/crypto 573951cbe80bb6352881271bb276f48749eab6f4
 golang.org/x/net a6577fac2d73be281a500b310739095313165611
 golang.org/x/oauth2 1611bb46e67abc64a71ecc5c3ae67f1cbbc2b921
-golang.org/x/sys 739734461d1c916b6c72a63d7efda2b27edb369f
+golang.org/x/sys 95c6576299259db960f6c5b9b69ea52422860fce
+#golang.org/x/sys 739734461d1c916b6c72a63d7efda2b27edb369f
```
rn: Why is this commented out? Remove it?

justincormack: oh, yeah will do

README.md
```diff
@@ -19,21 +19,19 @@ LinuxKit, a toolkit for building custom minimal, immutable Linux distributions.
 
 ## Getting Started
 
-### Build the `moby` and `linuxkit` tools
+### Build the `linuxkit` tool
 
-LinuxKit uses the `moby` tool for image builds, and the `linuxkit` tool for pushing and running VM images.
+LinuxKit uses the `linuxkit` tool for ibuilding, pushing and running VM images.
```
rn: typo

src/cmd/linuxkit/build.go
justincormack: It is identical, not more-or-less, so it is easy to update. Not ideal, maybe can move some more code in to the library later.

rn: OK, identical is better than almost identical. Having this moved to some common piece of code somewhere would be good, but could be done later

rn:
This is quite a change, maybe getting an approval from one or two additional maintainer would be good.

/ping @riyazdf @ijc @deitch

projects/etcd/README.md
```diff
@@ -85,7 +85,7 @@ docker run --rm -ti etcd.local etcdctl member list
 docker run --rm -ti etcd.local etcdctl cluster-health
 ```
 
-You can perform rolling updates, by for example, switching the kernel version in `etcd.yml`, build a new moby, e.g., `moby build -name etcd-4.10 etcd`, update `infrakit.json`, and then commit the new configuration to InfraKit: `infrakit group commit infrakit.json`.
+You can perform rolling updates, by for example, switching the kernel version in `etcd.yml`, build a new moby, e.g., `linuxkit build -name etcd-4.10 etcd`, update `infrakit.json`, and then commit the new configuration to InfraKit: `infrakit group commit infrakit.json`.
```
riyazdf: s/moby/linuxkit image ?

src/cmd/linuxkit/build.go
```diff
@@ -0,0 +1,215 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/moby/tool/src/moby"
+	log "github.com/sirupsen/logrus"
+)
+
+const defaultNameForStdin = "moby"
```
riyazdf: nit: do we want to change this?

reports/2017-03-19.md
```diff
@@ -37,7 +37,7 @@ Welcome everyone, looking forward to your contrinutions and working with you. Pl
 
 ## This week
 
-We are working towards a simpler and better documented basic flow. Early next week the `moby` tool will be changed to have `moby build` and `moby run` subcommands. At least initially the `moby run` flow will only be targeted at running locally (hyperkit and kvm, qemu). We may add a third cluster run option later, using infrakit. @rneugeba has been working on Go hyperkit bindings to make this easier in [#1327](https://github.com/docker/moby/pull/1327).
+We are working towards a simpler and better documented basic flow. Early next week the `moby` tool will be changed to have `linuxkit build` and `moby run` subcommands. At least initially the `moby run` flow will only be targeted at running locally (hyperkit and kvm, qemu). We may add a third cluster run option later, using infrakit. @rneugeba has been working on Go hyperkit bindings to make this easier in [#1327](https://github.com/docker/moby/pull/1327).
```
riyazdf: we can probably revert all changes to `reports/` since it was correctly specified at its point of writing

src/cmd/linuxkit/build.go
riyazdf: I missed [@rn's comment](https://github.com/linuxkit/linuxkit/pull/2761#discussion_r152032892), I agree that keeping this file identical makes sense for now so no need to change

justincormack:
@riyazdf removed all the changes in `reports/` thanks

deitch:
Looks good here as well. It also makes testing significantly easier, as you don't have to update one, then the other, then test on the first, etc.

I am curious what the motivation was to break it out initially, when it was `lkt build`, and why that rationale no longer applies?

justincormack:
ok its green, merging as it touches a lot of files

Issue #2721, opened 2017-11-14
metadata parameter crashes on long inline content


w9n:
**Description**
if  `-data inline_metadata` paramter is too long then linuxkit crashes.

**Steps to reproduce the issue:**
```
linuxkit run qemu -data 'Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?'  some_image
```

**Describe the results you received:**
```
FATA[0000] Cannot read user data: open Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?: file name too long 
```

**Describe the results you expected:**

starting vm.
**Additional information you deem important (e.g. issue happens only occasionally):**
[...]

PR #2762, opened 2017-11-20
Handle empty metadata file better (by ignoring)

ijc:
Current if `-data ""` is used we treat it as a raw string and do not create any a metadata iso to pass to the VM. However if `-data` instead refers to an empty file we currently error out due to a check for empty files in the `iso9660wrap` package.

This PR:
* Vendors the latest `iso9660wrap` package, which allows zero sized files (and also drops the text of The Raven)
* Refactors the code which creates the ISO into a function common to Hyperkit and QEMU
* Causes that function to skip creating a file if the file is empty, as it already does if the argument is an empty string.

Related to https://github.com/linuxkit/kubernetes/issues/4

Per #2721 we should, IMHO, also separate the file vs raw string functionality of the `-data` argument, but I've not tackled that here.

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -141,22 +141,11 @@ func runHyperKit(args []string) {
 		log.Fatalf("Could not create state directory: %v", err)
 	}
 
-	if *data != "" {
-		var d []byte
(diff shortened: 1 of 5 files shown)


deitch:
> if you also want to add the proposal from #2721 to have separate arguments for string and file that would also be fine with me

Addressed in https://github.com/linuxkit/linuxkit/pull/2861

PR #2765, opened 2017-11-21
docs: minor fixes for use of `-data` 

RobbKistler:
Replace `--data` with `-data`, add a hint for reading metadata from a file.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -65,7 +65,7 @@ needed.
 
 # Metadata image creation
 
-Run `linuxkit run` backends accept a `--data=STRING` option which will
+Run `linuxkit run` backends accept a `-data=STRING` option which will
 cause the given string to be passed to the VM in a platform specific
 manner to be picked up by the `pkg/metadata` component.
 


docs/metadata.md
```diff
@@ -65,14 +65,19 @@ needed.
...
 cause the given string to be passed to the VM in a platform specific
 manner to be picked up by the `pkg/metadata` component.
 
 Alternatively `linuxkit metadata create meta.iso STRING` will produce
 a correctly formatted ISO image which can be passed to a VM as a CDROM
 device for consumption by the `pkg/metadata` component.
 
+To read `STRING` from a file:
+```
+-data="$(<FILENAME)"
```
rn: This is a really odd (and confusing) syntax. could you change this to `-data=<FILENAME>` or something?

RobbKistler: That's meant to be literal, but I just realized its also a bash-ism.  Would `-data="$(cat FILENAME)"` make it more clear?  Or should I change implementation to actually read a specified file instead of using the command line arguments directly?  I don't know if anyone would need to pass whats usually  multi-line json as a command line argument.

rn: no, you can actually pass in a file name instead of a string, though there is an issue with that (see https://github.com/linuxkit/linuxkit/issues/2721)

RobbKistler:
@rn I removed the commit about specifying a filename for `-data`.  It's very clear in the CLI help that it can be a string or filename.

PR #2770, opened 2017-11-22
Update cadvisor and add build test

rn:
`eamples/cadvisor` overlapped with a kernel update. This brings it up-to-date. Also add a build test for the example while at it
owls


test/cases/000_build/100_examples/080_cadvisor/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# SUMMARY: Test the wireguard example
```
justincormack: update the comment!

rn:
doh, force pushed

PR #2775, opened 2017-11-27
Makefile: Fix binaries generation issue when 'docker run' failed

arm64b:
Currently we will always create a temp tar file even the 'docker run'
failed for some reason,e.g, the docker daemon doesn't run. As a result,
we'll get the 2nd error even we've fixed 'docker run' failure.

**- What I did**
Fix binaries generation failure
**- How I did it**
Add some interim target names to mitigate the `docker run` failure case. 
**- How to verify it**
1. `systemctl stop docker` (Actually I happened to build the binaries in case of docker daemon is not running and get the error information)
2. `make bin/rtf`
3. The `docker run` will fail in this case, but it generate a invalid temp file `tmp_rtf_bin.tar` (size = 0)
4. `systemctl start docker`
5. `make bin/rtf`, since the invalid `tmp_rtf_bin.tar` target is already there(invalid though), so the `make bin/rtf` will continue the following operations, result in below error message:
```
mkdir -p bin
tar xf tmp_rtf_bin.tar
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
Makefile:26: recipe for target 'bin/rtf' failed
make: *** [bin/rtf] Error 2
```

summit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.DELETE_ON_ERROR:
+
 .PHONY: default all
 default: bin/linuxkit bin/rtf
 all: default


ijc:
This will leak the files `_FOO` on fail in the same way.

Maybe we should just add `.DELETE_ON_ERROR` to the `Makefile? I'm not sure whether or not that is a GNU-makeism though and so wether it works on MacOS.

justincormack:
@ijc that does work on OSX (which uses an old version of GNU make)

ijc:
Sounds like `.DELETE_ON_ERROR` is the way to go then.

I guess on Windows we can easily mandate GNU make while we are mandating a whole pile of other Unix-ish stuff.

arm64b:
@ijc  Yes, this is the right way to go 👍 I just test it on both Linux and MacOS, it does works.

Makefile
```diff
@@ -27,6 +27,7 @@ bin/rtf: tmp_rtf_bin.tar | bin
 	rm $<
 	touch $@
 
+.DELETE_ON_ERROR:
```
ijc: This placement suggests it applies only to `tmp_rtf_bin.tar` when actually this has global effect. I suggest putting it right at the top before the `.PHONY: default all` and with a blank line between `.DELETE_ON_ERROR` and `.PHONY`.

arm64b: Thanks. Done!

PR #2779, opened 2017-11-28
Bind mount /etc/resolv.conf into nginx container

justincormack:
Also make them more consistent in small ways

baby-otter-sleeps-mother-belly-monterey-bay-aquarium-fb__700-png

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -29,6 +29,8 @@ services:
      - CAP_SETUID
      - CAP_SETGID
      - CAP_DAC_OVERRIDE
+    binds:
+     - /etc/resolv.conf:/etc/resolv.conf
 trust:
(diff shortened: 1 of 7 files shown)


rn:
Might be a stupid question, but how does this work with things like `dhcpcd` which write `/etc/resolv.conf`. Shouldn't `nginx` pick these up?

justincormack:
yeah it can be made to work but its kind of complicated right now, via symlinking. At present if dns servers are updated by a dhcp service they wont be seen, but the one shot setup will still work.

(it is complicated as we cant actually set `/etc/resolv.conf` in the current method of extracting containers as docker always overwrites it).

rn:
Ok, let's first make them more consistent anyway

PR #2780, opened 2017-11-29
Add support for newer AWS networking types

justincormack:
- ENA networking is what the very new (and bare metal) machines have
- sriov "simple" mode is Intel VF passthrough

I reflect the underlying APIs, which is bool for ENA and a string
for sriov even though there is currently only one valid sriov option...

hyenas


justincormack:
Yeah, I havent tested it yet, but will try to, but the docs say thats how to do it...

PR #2783, opened 2017-11-29
ci: remove gcp-hack and ltp make targets

RobbKistler:
These GCP-related targets are causing linuxkit.datakit.ci runs to fail,
and might not be needed.

Signed-off-by: Robb Kistler <robb.kistler@docker.com>

**- How to verify it**
Tests on merge to master do not fail in `make test` with:
```
 make[1]: *** No rule to make target 'test-ltp.img.tar.gz', needed by 'ltp'.  Stop.
```

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -3,7 +3,7 @@
 default: check-deps test 
 pr: check-deps test-pr
 # TODO: should have a separate target
-all: check-deps test-pr ltp
+all: check-deps test-pr
 
 LINUXKIT:=$(shell command -v linuxkit 2> /dev/null)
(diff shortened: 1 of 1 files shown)


RobbKistler:
I'm not sure the CI failure is related to my changes -- ssh host verification failure.

rn:
`gcp-hack` is actually needed and the comment removed in this PR explains this somewhat. The CI builds a `gcp` images and then the linux kit CI tries to boot it. This is not done as part of `rtf` because we haven't figured out a sane way to manage the GCP credential.

The LTP target is there to run manually.

PR #2784, opened 2017-11-30
Added initial Vagrant support.

Manouchehri:
**- What I did**

Added Vagrant support (VMware Fusion, VMware Workstation, VirtualBox, and Hyper-V).

**- How I did it**

Added a `Vagrantfile`.

**- How to verify it**

```bash
mbp:~ dave$ git clone -b vagrant https://github.com/Manouchehri/linuxkit
mbp:~ dave$ cd linuxkit/
mbp:linuxkit dave$ vagrant up # Wait for a few minutes and a couple hundred lines of output. 
mbp:linuxkit dave$ vagrant ssh 
vagrant@vagrant:~$ linuxkit build ./go/src/github.com/linuxkit/linuxkit/examples/minimal.yml
Extract kernel image: linuxkit/kernel:4.9.65
....
Create outputs:
  minimal-kernel minimal-initrd.img minimal-cmdline
vagrant@vagrant:~$ ls -l minimal*
-rw-r--r-- 1 vagrant vagrant       42 Nov 30 02:43 minimal-cmdline
-rw-r--r-- 1 vagrant vagrant 41308847 Nov 30 02:43 minimal-initrd.img
-rw-r--r-- 1 vagrant vagrant  6570576 Nov 30 02:43 minimal-kernel
```

Added Vagrant support.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,8 +5,6 @@
 Vagrant.configure("2") do |config|
 	config.vm.box = "bento/ubuntu-16.04"
 
-	config.vm.synced_folder ".", "/vagrant", disabled: true
-
 	config.vm.provision "shell", inline: <<-SHELL
 		# apt-get update
 		# DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
@@ -17,7 +15,11 @@ Vagrant.configure("2") do |config|
 	config.vm.provision "docker"
 
 	config.vm.provision "shell", privileged: false, inline: <<-SHELL
-		go get -u github.com/linuxkit/linuxkit/src/cmd/linuxkit
+		mkdir -p ~/go/src/github.com/linuxkit/
+		ln -s /vagrant ~/go/src/github.com/linuxkit/linuxkit
+		cd ~/go/src/github.com/linuxkit/linuxkit/src/cmd/linuxkit
+		go get -d
+		go install
 		echo "export PATH=${PATH}:${HOME}/go/bin" >> ~/.bashrc
 	SHELL
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,10 +8,10 @@ Vagrant.configure("2") do |config|
 	config.vm.synced_folder ".", "/vagrant", disabled: true
 
 	config.vm.provision "shell", inline: <<-SHELL
-		apt-get update
-		DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+		# apt-get update
+		# DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
 		snap install --classic go
-		apt-get clean
+		# apt-get clean
 	SHELL
 
 	config.vm.provision "docker"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,10 +6,7 @@ Vagrant.configure("2") do |config|
 	config.vm.box = "bento/ubuntu-16.04"
 
 	config.vm.provision "shell", inline: <<-SHELL
-		# apt-get update
-		# DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
 		snap install --classic go
-		# apt-get clean
 	SHELL
 
 	config.vm.provision "docker"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,33 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+# Author: David Manouchehri
+
+Vagrant.configure("2") do |config|
+	config.vm.box = "bento/ubuntu-16.04"
+
+	config.vm.synced_folder ".", "/vagrant", disabled: true
+
+	config.vm.provision "shell", inline: <<-SHELL
+		apt-get update
+		DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+		snap install --classic go
+		apt-get clean
+	SHELL
+
+	config.vm.provision "docker"
+
+	config.vm.provision "shell", privileged: false, inline: <<-SHELL
+		go get -u github.com/linuxkit/linuxkit/src/cmd/linuxkit
+		echo "export PATH=${PATH}:/home/`whoami`/go/bin" >> ~/.bashrc
+	SHELL
+
+	%w(vmware_fusion vmware_workstation vmware_appcatalyst).each do |provider|
+		config.vm.provider provider do |v|
+			v.vmx["vhv.enable"] = "TRUE"
+			v.vmx['ethernet0.virtualDev'] = 'vmxnet3'
+		end
+	end
+
+	config.vm.provider "virtualbox"
+	config.vm.provider "hyperv"
+end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,7 +18,7 @@ Vagrant.configure("2") do |config|
 
 	config.vm.provision "shell", privileged: false, inline: <<-SHELL
 		go get -u github.com/linuxkit/linuxkit/src/cmd/linuxkit
-		echo "export PATH=${PATH}:/home/`whoami`/go/bin" >> ~/.bashrc
+		echo "export PATH=${PATH}:${HOME}/go/bin" >> ~/.bashrc
 	SHELL
 
 	%w(vmware_fusion vmware_workstation vmware_appcatalyst).each do |provider|
--- a/scripts/Vagrantfile
+++ b/scripts/Vagrantfile
@@ -5,6 +5,9 @@
 Vagrant.configure("2") do |config|
 	config.vm.box = "bento/ubuntu-16.04"
 
+	config.vm.synced_folder ".", "/vagrant", disabled: true
+	config.vm.synced_folder "../", "/vagrant"
+
 	config.vm.provision "shell", inline: <<-SHELL
 		snap install --classic go
 	SHELL
--- a/scripts/Vagrantfile
+++ b/scripts/Vagrantfile
@@ -13,14 +13,14 @@ Vagrant.configure("2") do |config|
 	SHELL
 
 	config.vm.provision "docker"
+	config.vm.provision "shell", inline: "ps aux | grep 'sshd:' | awk '{print $2}' | xargs kill"
 
 	config.vm.provision "shell", privileged: false, inline: <<-SHELL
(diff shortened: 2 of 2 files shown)


rn:
Thanks for the patch. Could you explain what this is for and how to use it? Some documentation would be good. In general we also prefer to keep files at the top-level to a minimum and the commits would need to be squashed as well.

But first could you maybe comment here what this is for?

Vagrantfile
```diff
@@ -0,0 +1,33 @@
...
+		# apt-get update
+		# DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+		snap install --classic go
+		# apt-get clean
+	SHELL
+
+	config.vm.provision "docker"
+
+	config.vm.provision "shell", privileged: false, inline: <<-SHELL
+		go get -u github.com/linuxkit/linuxkit/src/cmd/linuxkit
```
ijc: Seems a bit odd to reclone rather than reusing (or even sharing if such a thing is possible with Vagrant) the clone on the host where the Vagrant file is being run.

If that isn't possible then it may as well be in some subdirectory (if not some other repo altogether).

Manouchehri: I'm not too familiar with go, is there something like `go install ~/linuxkit` I could use? (I'm not sure if that handles dependencies, so I went with the safe bet of `go get [URL]`.)

Manouchehri:
@rn I'm using it to bootstrap linuxkit on Windows with VMware Workstation (Hyper-V disabled). I also use it on my Mac (VMware Fusion), since I don't find it very clean to have build related packages being installed on my host.

Manouchehri:
Oh, and for those who use Docker Toolbox with VirtualBox, that should work. (Thanks to @adamancini for testing the vbox build.)

Vagrantfile
ijc: My point was you already have this exact repo cloned on the host, because you ran `vagrant up` from it, so fetching it again (whther with `go get`, or `go install` or whatever) inside the VM seems odd.

Vagrantfile
```diff
@@ -0,0 +1,33 @@
...
+		echo "export PATH=${PATH}:/home/`whoami`/go/bin" >> ~/.bashrc
```
justincormack: you should be able to use `${HOME}` not `/home/`whoami`. Also adding this to `~/.bashrc` unconditionally seems wrong...

Manouchehri: Done.

Vagrantfile
Manouchehri: I understood what you were saying in the first message, was explaining that dealing with golang's package dependencies wasn't trivial for someone who hasn't used it before.

Anyway, I figured it out. Done.

Manouchehri:
Let me know if everything looks good now and I'll squash the commits.

Vagrantfile
```diff
@@ -0,0 +1,35 @@
...
+# Author: David Manouchehri
+
+Vagrant.configure("2") do |config|
+	config.vm.box = "bento/ubuntu-16.04"
+
+	config.vm.provision "shell", inline: <<-SHELL
+		# apt-get update
+		# DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+		snap install --classic go
+		# apt-get clean
```
ijc: I guess these comments are obsolete and can be removed?

Vagrantfile
```diff
@@ -0,0 +1,35 @@
...
+	SHELL
+
+	config.vm.provision "docker"
+
+	config.vm.provision "shell", privileged: false, inline: <<-SHELL
+		mkdir -p ~/go/src/github.com/linuxkit/
+		ln -s /vagrant ~/go/src/github.com/linuxkit/linuxkit
+		cd ~/go/src/github.com/linuxkit/linuxkit/src/cmd/linuxkit
```
ijc: Is something somewhere implicitly setting `$GOPATH`?

Vagrantfile
```diff
@@ -0,0 +1,35 @@
...
+		go get -d
+		go install
```
ijc: If you use `make install` instead of these two then the binaries will be correctly tagged with the git commit sha and version. You can override `PREFIX` if you don't want (or can't) install into `/usr/local/bin`. Using `PREFIX=$HOME/go` would  be compatible with the following.

Vagrantfile
```diff
@@ -0,0 +1,35 @@
...
+		echo "export PATH=${PATH}:${HOME}/go/bin" >> ~/.bashrc
```
ijc: Might this be clearer as `${GOPATH}/bin`?

Manouchehri: I might be wrong, but is `${GOPATH}` even set here? I can check, but not much of a difference anyway.

Vagrantfile
Manouchehri: Okay, if I do this, then we won't be reusing the existing cloned repository like you asked me in https://github.com/linuxkit/linuxkit/pull/2784#pullrequestreview-80121346. Looks like in https://github.com/linuxkit/linuxkit/blob/13adecbc3e136fcb1894f0604db04468dd01a4b3/Makefile#L81 the repository is being cloned again.

I don't mind doing either, but spending time testing changes only to have them reverted in a couple hours is a bit frustrating. (Rebuilding the entire VM takes a few minutes, regardless of how many lines are changed. It's not like Docker where every RUN layer is cached.)

Vagrantfile
Manouchehri: No.

Vagrantfile
Manouchehri: They're good to leave in; they're optional in case someone needs/wants to update all their packages. (This adds a couple minutes onto the build though.)

Vagrantfile
Manouchehri: No, $GOPATH is not set.

```bash
vagrant@vagrant:~$ echo $GOPATH

vagrant@vagrant:~$
```

Vagrantfile
ijc: They look like cruft, if they aren't please leave a descriptive comment.

Vagrantfile
ijc: It seems the default `$GOPATH` is `$HOME/go`, I didn't realise that (I thought it was something one was required to set). OK then.

Vagrantfile
ijc: `make install`'s reclone of the repo to build the tool is orthogonal to the desire to leave the user with a local working tree (where they can edit `pkg/*`, build packages etc which I think is the point of this Vagrant file, isn't it?) which is not a duplicate of the one of the host.

So please both reuse the host repo (as requested in previous review) and use `make install`, there is no contradiction there AFAICT, sorry if it appeared that way.

rn:
It seems to me that this adds support to build linuxkit via vagrant and then being able to invoke linuxkit commands within the VM created by Vagrant. To me this seems a bit of a niche use case. We already support Linux, macOS, and Windows natively and provide pre-build binaries for them (they are available as artefacts on CircleCI). For this reason, having the `Vagrantfile` at the top level seems a bit OTT. It might be better to move it to `./scripts`

Manouchehri:
@rn Correct, this is only for developers who want a clean/isolated environment that's also easy to set up (https://github.com/linuxkit/linuxkit/pull/2784#issuecomment-348144083).

Generally Vagrantfiles (like Dockerfiles) are expected to be at the top level, but yeah, I can move it to  `/scripts` if you want. Would a soft link to the top level be acceptable, or no?

Vagrantfile
Manouchehri: I replaced `go get -d` and `go install` with `make install` like you asked, and it doesn't work.

```
mkdir -p ~/go/src/github.com/linuxkit/
ln -s /vagrant ~/go/src/github.com/linuxkit/linuxkit
cd ~/go/src/github.com/linuxkit/linuxkit
make install
echo "export PATH=${PATH}:${HOME}/go/bin" >> ~/.bashrc
```

```bash
vagrant@vagrant:~/go/src/github.com/linuxkit/linuxkit$ make install
cp -R ./bin/* /usr/local//bin
cp: cannot stat './bin/*': No such file or directory
Makefile:92: recipe for target 'install' failed
make: *** [install] Error 1
```

Vagrantfile
Manouchehri: I don't think basic `apt-get` commands needs to explained, but I've removed them anyway.

justincormack:
I think `scripts/` makes more sense to me as it is fairly niche.

Manouchehri:
Alright, done.

Vagrantfile
ijc: Weird, looks like `install` misses the dependency to actually build the binaries first. You should insert a `make all` in front of the `make install` (or you could look into fixing up the `Makefile`).

Manouchehri: Okay, done.

When possible, could you test changes locally before requesting them?

>30 comments about 35 lines of code is a bit excessive..

smebberson:
This would be awesome to have in LinuxKit. Would make it super easy to have isolated development  environments running in Vagrant with a clean and lightweight linux OS; perfect for running Docker from :)

scripts/Vagrantfile
```diff
@@ -0,0 +1,35 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+# Author: David Manouchehri
```
justincormack: can you remove `Author` - things in the repo are collective responsibility

Manouchehri: You're welcome to remove it.

Manouchehri:
I'm a bit confused on where review process is going; my initial _optional_ script (i.e. not a core change) was a simple 20 line text file (https://github.com/linuxkit/linuxkit/pull/2784/commits/bb73dab8d3529e4f0ba6754be2c0b6c81d5ea6fa). The noise vs. code ratio seems way off here.

justincormack:
@Manouchehri sorry, I think part of the issue is that none of the maintainers are Vagrant users, and having it at the top level seemed a bit annoying (it tends to accumulate cruft). I am happy to have it in `scripts/` but the name thing is a policy that have had elsewhere, as it discourages other people making changes.

Manouchehri:
@justincormack I understand that, but requesting another user to delete one line is still unusual. It's your repository, why not save everyone a bit of time and do it instead?

PR #2785, opened 2017-11-30
pkg: rngd service support on AArch64

arm64b:
Enable the rngd service in case of a real hardware RNG
device is available on AArch64 platform

**- What I did**
Enable the `rngd` service on AArch64 in case of a hardware RNG device available
**- How I did it**
Check the `/dev/hwrng` and use it as the entropy source to feed the `/dev/random` if `/dev/hwrng` is the real character device of a RNG hardware.
**- How to verify it**
1. `linuxkit pkg build -org m1130 rngd/`
2.  `docker run --privileged --rm --device /dev/hwrng m1130/rngd:842e5e8ece7934f0cab9fd0027b595ff3471e5b9-dirty`
3. On an AArch64 platform which has a real RNG device such as Cavium ThunderX, we can see the `/sbin/rngd` is running inside the docker.

summit

### DIFF
--- a/pkg/rngd/cmd/rngd/main.go
+++ b/pkg/rngd/cmd/rngd/main.go
@@ -6,6 +6,8 @@ package main
 import "C"
 
 import (
+	"encoding/binary"
+	"errors"
 	"flag"
 	"log"
 	"os"
@@ -14,7 +16,74 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type rng struct {
+	rngName  string
+	rngDev   string
+	disabled bool
+	rngInit  xinit
+	rngRead  xread
+}
+
+type xinit func(ctx *rng) bool
+type xread func(ctx *rng) (uint64, error)
+
+// Check if hardware RNG device is valid
+func initEntropySource(ctx *rng) bool {
+	f, err := os.Open(ctx.rngDev)
+	defer f.Close()
+	if err != nil {
+		ctx.disabled = true
+		return false
+	}
+
+	// Try to read some data from the entropy source. If it doesn't
+	// return an error, assume it's ok to use
+	buf := make([]byte, 16)
+	n, err := f.Read(buf)
+	if err != nil || n != 16 {
+		ctx.disabled = true
+		return false
+	}
+
+	return true
+}
+
+// Read entropy from the hardware RNG device
+func readHwRNG(ctx *rng) (uint64, error) {
+	f, err := os.Open(ctx.rngDev)
+	defer f.Close()
+	if err != nil {
+		return 0, err
+	}
+	buf := make([]byte, 8)
+	_, err = f.Read(buf)
+	if err != nil {
+		return 0, err
+	}
+
+	r := binary.LittleEndian.Uint64(buf)
+
+	return r, nil
+}
+
+var entropySources = []rng{
+	// Hardware RNG device
+	{
+		rngName: "Hardware RNG Device",
+		rngDev:  "/dev/hwrng",
+		rngInit: initEntropySource,
+		rngRead: readHwRNG,
+	},
+	// RNG instruction support
+	{
+		rngName: "Instruction RNG",
+		rngInit: initDRNG, // Arch specific
+		rngRead: readDRNG, // Arch specific
+	},
+}
+
 func main() {
+	var entSource bool
 	oneshot := flag.Bool("1", false, "Enable oneshot mode")
 	flag.Parse()
 
@@ -23,8 +92,12 @@ func main() {
 		timeout = 0
 	}
 
-	supported := initRand()
-	if !supported {
+	for _, e := range entropySources {
+		if e.rngInit != nil && e.rngInit(&e) {
+			entSource = true
+		}
+	}
+	if entSource == false {
 		log.Fatalf("No random source available")
 	}
 
@@ -81,16 +154,33 @@ type randInfo struct {
 }
 
 func writeEntropy(random *os.File) (int, error) {
-	r, err := rand()
-	if err != nil {
-		// assume can fail occasionally
-		return 0, nil
+	var i int
+	var r uint64
+	var err error
+	var ret uintptr
+	for _, e := range entropySources {
+		if e.disabled {
+			continue
+		}
+		if e.rngRead != nil {
+			r, err = e.rngRead(&e)
+			if err != nil {
+				continue
+			}
+		}
+
+		const entropy = 64 // they are good random numbers, Brent
+		info := randInfo{entropy, 8, r}
+		ret, _, _ = unix.Syscall(unix.SYS_IOCTL, uintptr(random.Fd()), uintptr(C.rndaddentropy), uintptr(unsafe.Pointer(&info)))
+		if ret == 0 {
+			i += 8
+		} else {
+			continue
+		}
 	}
-	const entropy = 64 // they are good random numbers, Brent
-	info := randInfo{entropy, 8, r}
-	ret, _, err := unix.Syscall(unix.SYS_IOCTL, uintptr(random.Fd()), uintptr(C.rndaddentropy), uintptr(unsafe.Pointer(&info)))
-	if ret == 0 {
-		return 8, nil
+	if i == 0 {
+		return 0, errors.New("No entropy added")
 	}
-	return 0, err
+
+	return i, nil
 }
--- a/pkg/rngd/cmd/rngd/rng_linux_arm64.go
+++ b/pkg/rngd/cmd/rngd/rng_linux_arm64.go
@@ -10,12 +10,12 @@ import (
 	"errors"
 )
 
-// No standard RNG on arm64
-
-func initRand() bool {
+// No standard RNG instruction on arm64
+func initDRNG(ctx *rng) bool {
+	ctx.disabled = true
 	return false
 }
 
-func rand() (uint64, error) {
+func readDRNG(_ *rng) (uint64, error) {
 	return 0, errors.New("No randomness available")
 }
--- a/pkg/rngd/cmd/rngd/rng_linux_amd64.go
+++ b/pkg/rngd/cmd/rngd/rng_linux_amd64.go
@@ -44,13 +44,17 @@ var disableRdseed = flag.Bool("disable-rdseed", false, "Disable use of RDSEED")
 
 var hasRdrand, hasRdseed bool
 
-func initRand() bool {
+func initDRNG(ctx *rng) bool {
 	hasRdrand = C.hasrdrand() == 1 && !*disableRdrand
 	hasRdseed = C.hasrdseed() == 1 && !*disableRdseed
-	return hasRdrand || hasRdseed
+	b := hasRdrand || hasRdseed
+	if b == false {
+		ctx.disabled = true
+	}
+	return b
 }
 
-func rand() (uint64, error) {
+func readDRNG(_ *rng) (uint64, error) {
 	var x C.uint64_t
 	// prefer rdseed as that is correct seed
 	if hasRdseed && C.rdseed(&x) == 1 && !*disableRdseed {


arm64b:
@justincormack @ijc @rn, currently we always pass a `virtio-rng` device backed by host's `/dev/urandom` to the guest, that makes the guest's `/dev/hwrng` always valid, in this case the entropy from the `/dev/hwrng` maybe not good enough for some security-sensitive applications. My proposal here is, if there's a real RNG hardware device in the host, we can para-virtualise that to the guest, else we do nothing, thus the `/dev/hwrng` in the guest will give the effective information of the quality of the entropy source. Any comments about that?

deitch:
@arm64b I am a bit confused by this. If I understood you correctly, the problem is as follows.

We always pass `/dev/urandom` of host to guest as `/dev/hwrng`. This makes the guest always _think_ that it has a real `hwrng`, when it may, in fact, be backed by a lower-entropy device.

Essentially, the expectation on Linux is:

* use `hwrng` to discover if you really have a hardware rng
* use `random` if you want reliable entropy and are willing to wait for it
* use `urandom` if you want some entropy, filled in with lower quality if there isn's sufficient available

By promoting `hwrng` in the guest to `urandom` in the host, we are "tricking" the guest into thinking there is a real entropy when there isn't.

Is that it?

arm64b:
@deitch Correct. With `virtio-rng`, we fake a RNG device in the guest but it actually backed by `/dev/urandom`, the latter doesn't block if it does not generate enough entropy, eg, output the same number repeatedly. The essential is we need to make sure guest realize this.

deitch:
In that case, I don't get how this PR helps. Doesn't this change `pkg/rngd` which runs _inside_ the guest and thus has no knowledge? Would we not need to change how `lkt run` calls hyperkit or qemu command-line?

arm64b:
That's different, this PR just uses the entropy from `/dev/hwrng` to feed `/dev/random` in guest, so some security applications can take use of `/dev/random` to generate like keys, the PR doesn't care about the entropy quality in `/dev/hwrng`. But if the guest can realize its `/dev/hwrng` is backed by a real RNG device or not, we can make better decision how to handle it.

deitch:
Ah, so you are not fixing that issue, but rather intelligently checking if `hwrng` exists in the guest? But again, won't it exist in the guest and be (possibly) fake anyways, because it is backed by `urandom` in the underlying host?

arm64b:
> so you are not fixing that issue

Right, that's another topic, it's my proposal.

> won't it exist in the guest and be (possibly) fake anyways, because it is backed by urandom in the underlying host?

For our current solution, the `/dev/hwrng` always valid, as you said, it's backed by `/dev/urandom` in the host. So this PR doesn't care about the backend of the `/dev/hwrng`, it just establishes a bridge between the `/dev/random` and `/dev/hwrng` in guest: read entropy from `/dev/hwrng` and feed it to `/dev/random`.

deitch:
> So this PR doesn't care about the backend of the /dev/hwrng, it just establishes a bridge between the /dev/random and /dev/hwrng in guest: read entropy from /dev/hwrng and feed it to /dev/random

And your plan then is to have a future PR that will replace the blind matching of `urandom` (host) -> `hwrng` (guest) with something more intelligent like `hwrng` (host if it exists) -> `hwrng` (guest)?

This would bring `rngd` in line with its usual job, which is `hwrng`/`hwrandom` -> `random`.

Looks good to me, but like to get input from another someone with lower-level understanding. I believe @justincormack did the original `rngd` implementation?

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -7,15 +7,47 @@ package main
...
+	f, _ := os.Open("/dev/hwrng")
+	defer f.Close()
+
+	buf := make([]byte, 8)
+	_, err := f.Read(buf)
+	if err != nil {
+		return 0, err
+	}
+
+	r = binary.LittleEndian.Uint64(buf)
```
ijc: OOI why is the endianess of the random data relevant?

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -7,15 +7,47 @@ package main
...
-
+// Check if there is a hardware RNG device on AArch64, will use that
+// device as the RNG entropy source if we do have
 func initRand() bool {
-	return false
+
+	f, err := os.Open("/dev/hwrng")
+	defer f.Close()
+	if err != nil {
+		log.Fatalf("Cannot open /dev/hwrng: %v", err)
```
rn: `log.Fatalf` will exit so the return below will not happen and `rngd` will not run. What is the intention here? Exit or return?

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -7,15 +7,47 @@ package main
...
+		return false
+	}
+
+	// Try to read some data from the entropy source. If it doesn't
+	// return an error, assume it's ok to use
+	buf := make([]byte, 16)
+	n, err := f.Read(buf)
+	if err != nil || n != 16 {
+		log.Fatalf("Read /dev/hwrng failed: %v", err)
```
rn: ditto

justincormack:
We should not make this code arm64 specific - `/dev/hwrng` is generic and will cover a tpm rng on amd64 platform too, which I have been meaning to add support for, so we should move it to generic code that is used when this device exists.

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -7,15 +7,47 @@ package main
...
+	defer f.Close()
+	if err != nil {
+		log.Fatalf("Cannot open /dev/hwrng: %v", err)
+		return false
+	}
+
+	// Try to read some data from the entropy source. If it doesn't
+	// return an error, assume it's ok to use
+	buf := make([]byte, 16)
+	n, err := f.Read(buf)
```
rn: Is this a read on `/dev/hwrng` blocking if there is no entropy, like `/dev/random`? If so we want a timeout on this,

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -7,15 +7,47 @@ package main
...
+// Check if there is a hardware RNG device on AArch64, will use that
+// device as the RNG entropy source if we do have
 func initRand() bool {
-	return false
+
+	f, err := os.Open("/dev/hwrng")
+	defer f.Close()
+	if err != nil {
+		log.Fatalf("Cannot open /dev/hwrng: %v", err)
+		return false
```
justincormack: log.Fatalf will exit, so no point returning.

pkg/rngd/cmd/rngd/rng_linux_arm64.go
justincormack: endianness of random data doesnt matter, it is still random bits.

justincormack:
@arm64b in the virtual case, the passthrough that `linuxkit` does is not necessarily meant to be what you use in production, it is meant as a good developer default. Generally you should use qemu/kvm directly and decide what to pass depending on what the host rng situation is and what the requirements are.

deitch:
> Generally you should use qemu/kvm directly and decide what to pass depending on what the host rng situation is and what the requirements are

Probably still is a good thing to "do the right thing" when we run qemu and pass through matching devices, though. If @arm64b wants to put in the work to make that happen, probably a good thing, no?

In prod though, yes, definitely should set qemu command as you want (or maybe don't use qemu?)

justincormack:
@deitch I don't think in general you want to pass through hw rng, you probably want to mix them into host entropy and pass that. Eitehr way we should not be providing policy for production in the `linuxkit` tool, people should make those decisions themselves and use the tools directly.

deitch:
> you probably want to mix them into host entropy and pass that

Didn't understand this. What is mapped to what? Does `hwrng` exist on the guest and if so, mapped to where? If it is to `random` (or `urandom`), then it is not a true `hwrng`.

Or are you saying `lkt run` isn't meant for production use, ever, so it doesn't matter?

deitch:
> I am also slightly confused about /dev/hwrng in the case of virtio rng as it seems to exist and work, but in fact the kernel talks to it directly and AFAIK rngd does not need to read it from userspace, but I am not 100% sure about this, need to look into it.

Oh, so it isn't just me who is confused? I feel better now.

I always thought `rngd`'s job was to read from a hardware random device (like `/dev/hwrng`) and feed `/dev/random`. Perhaps not?

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: Right. Will consolidate that with Justin's comments followed.

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: Yes

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: Will consolidate with ijc's comment above

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: Will return false indicate `rngd` will not be launched with `log.Fatalf` removed.

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: I am not pretty sure since I can't test all the hardware RNG devices, but if that case happens, probably it's hardware bug.

arm64b:
> I am also slightly confused about /dev/hwrng in the case of virtio rng as it seems to exist and work, but in fact the kernel talks to it directly and AFAIK rngd does not need to read it from userspace, but I am not 100% sure about this, need to look into it.

IMO, /dev/hwrng generate the entropy saved in kernel space, while the entropy pool that  `/dev/random` used is also in kernel space, technically we can transmit the entropy generated by `/dev/hwrng` into the entropy pool of `/dev/random`, but actually there isn't a channel to do that in current kernel implementation, that's why we need the `rngd` to do that.

arm64b:
> I always thought rngd's job was to read from a hardware random device (like /dev/hwrng) and feed /dev/random. Perhaps not?

No. Kernel doesn't have such a data tx channel between `/dev/hwrng` and `/dev/random`, so we need a service like `rngd` to read entropy from `/dev/hwrng` and feed it into the entropy pool of `/dev/random`, the latter will be used by security-sensitive applications.

deitch:
> we need a service like rngd to read entropy from /dev/hwrng and feed it into the entropy pool of /dev/random

Uh, isn't that precisely what I asked?

> rngd's job was to read from a hardware random device (like /dev/hwrng) and feed /dev/random.

arm64b:
> Uh, isn't that precisely what I asked?

Ah, so I think your thought is correct 😄

arm64b:
Hello @justincormack @ijc @rn and @deitch , I just forced an update of this PR. In this PR, I addressed some comments from you guys and the important is to move the hardware RNG code as arch agnostic. This PR also has a framework to support both hardware RNG device and some RNG instruction such as it does on x86_64.  For endlessness, since the random number is not related with that, I still use it since this will make the code easy, but you guys have some better one to do that, I am happy to embrace that 😸

arm64b:
Hi Maintainers, I didn't get any feedback for the latest update, so don't know if it's ready or something still need to be changed. Would you please give some comments so I know which way to go. Thanks!
/cc @justincormack

deitch:
> I didn't get any feedback for the latest update

Mix of everyone either busy or at KubeCon (not I, sigh...). Or both.

Going to reread the code changes.

deitch:
Took some time to figure out what this is doing. I do have some comments, but they are mostly stylistic, so I will wait for the responses from the others involved.

A few inline, though.

pkg/rngd/cmd/rngd/main.go
```diff
@@ -14,7 +15,76 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type rng struct {
+	rngName  string
+	rngDev   string
+	disabled bool
```
deitch: Should this not be `enabled`, so the default is false, and then you can enable it when an `init` succeeds?

pkg/rngd/cmd/rngd/main.go
```diff
@@ -14,7 +15,76 @@ import (
...
+	// Intel RDRAND instruction RNG
+	{
+		rngName: "Intel RDRAND Instruction RNG",
+		rngInit: initDRNG, // Arch specific
+		rngRead: readDRNG, // Arch specific
+	},
+}
+
 func main() {
+	var entSource int
```
deitch: Why is this an `int`? Why not a `bool` which defaults to `false`, and then you can set `entSource = true` when you have one at L98? You don't use the count elsewhere.

pkg/rngd/cmd/rngd/main.go
```diff
@@ -14,7 +15,76 @@ import (
...
+}
+
+var entropySources = []rng{
+	// Hardware RNG device
+	{
+		rngName: "Hardware RNG Device",
+		rngDev:  "/dev/hwrng",
+		rngInit: initEntropySource,
+		rngRead: readHwRNG,
+	},
```
deitch: Does this mean that it tries to read the generic (i.e. cross-platform) `/dev/hwrng` first, and thus will work on arm64 and amd64 if available, and only after try `DRNG`, which is Intel-specific?

pkg/rngd/cmd/rngd/main.go
```diff
@@ -81,16 +155,29 @@ type randInfo struct {
...
-	const entropy = 64 // they are good random numbers, Brent
-	info := randInfo{entropy, 8, r}
-	ret, _, err := unix.Syscall(unix.SYS_IOCTL, uintptr(random.Fd()), uintptr(C.rndaddentropy), uintptr(unsafe.Pointer(&info)))
-	if ret == 0 {
-		return 8, nil
+	var i int
+	var r uint64
+	var err error
+	var ret uintptr
+	for _, e := range entropySources {
```
deitch: So, it takes from all enabled sources?

justincormack:
Sorry, been travelling to Kubecon, will try to review soon!

pkg/rngd/cmd/rngd/main.go
```diff
@@ -14,7 +15,76 @@ import (
...
+
+var entropySources = []rng{
+	// Hardware RNG device
+	{
+		rngName: "Hardware RNG Device",
+		rngDev:  "/dev/hwrng",
+		rngInit: initEntropySource,
+		rngRead: readHwRNG,
+	},
+	// TPM support will be here
```
justincormack: I believe tmp uses `/dev/hwrng` too, so this comment is not needed

pkg/rngd/cmd/rngd/rng_linux_arm64.go
```diff
@@ -10,12 +10,12 @@ import (
 	"errors"
 )
 
-// No standard RNG on arm64
-
-func initRand() bool {
+// No standard RNG instruction on arm64
+func initDRNG(ctx *rng) bool {
```
justincormack: maybe we should just change this file to be compiled when  `!linux,!amd64` as this will then be fine if we add other platforms or someone else wants to reuse the code.

pkg/rngd/cmd/rngd/main.go
deitch: I believe you are right.

pkg/rngd/cmd/rngd/main.go
arm64b: Yes. enabled==true means a valid entropy source which can be used.

pkg/rngd/cmd/rngd/main.go
arm64b: entSource stands for the total number of valid/enabled entropy sources in the system, if we use a bool variable, think about below sequence:
Source A: enabled      entSource=true
Source B: disabled     entSource=false
So we can't flip this variable back and forth.

arm64b: Yes, make sense.

pkg/rngd/cmd/rngd/main.go
arm64b: Right, First we try to use hardware RNG device, then the arch specific method such as RNG instruction if that arch provides like amd64 does.

pkg/rngd/cmd/rngd/main.go
arm64b: Yes, The more entropy sources available, the more better quality of the entropies in the pool, the more safer to use it.

pkg/rngd/cmd/rngd/main.go
arm64b: OK, I will remote that.

pkg/rngd/cmd/rngd/rng_linux_arm64.go
arm64b: I don't quite understand, if we use `!linux,!amd` in this file to prevent it compiling, then we will get `undefined initDRNG and readDRNG` error in 'linux+arm64' environment, so my alternative is: if we want to support other platforms, e.g. PowerPC, we can add a file like `rng_linux_ppc.go` to support RNG instruction case if that arch does have one.

arm64b:
Hi Guys,

How about the CI test failure:
...
[+ 77m 46s] Executing busybox-1.26.2-r7.trigger
[+ 77m 46s] OK: 319 MiB in 33 packages
[+ 77m 49s] Error processing tar file(exit status 1): write /usr/lib/go/pkg/tool/linux_amd64/pprof: no space left on device
[+ 77m 49s] exit status 1
[+ 77m 49s] Makefile:11: recipe for target 'build' failed
...
I am pretty sure the `rngd` pkg has been built successfully. `no space left on device` for what? 😢

pkg/rngd/cmd/rngd/main.go
deitch: Got it. Just a nitpick, but it is the reverse now, because the default for `disabled` is `false`

arm64b: Ah, the code will set `disabled` as `true` for an invalid/malfunction entropy source, so when try to write the entropy to the pool, we have below code:
```
if e.disabled {
     continue
}
```
My comments is confusing 😄

deitch: As I said, "just a nitpick". :-)

arm64b:
Seems a stale PR, since we don't need `rngd` to feed entropy to `virtio-rng` device, then let's close it.

PR #2787, opened 2017-12-01
Include tool version info in package/image labels

ijc:
Since the tool can have some impact on the resulting image it seems like a good idea to track these so add `org.mobyproject.linuxkit.version` with the version and `org.mobyproject.linuxkit.revision` with the git commit.

Not sure if I should guard with `!= "unknown"` in both cases, I didn't but maybe it makes sense.


justincormack:
I think adding a label saying "unknown" is fine

PR #2789, opened 2017-12-01
remove subreaper config item

estesp:
containerd v1.0.0-rc.0 changed the subreaper config entry to the
negative given subreaper "on" is the default. However, linuxkit
does not need to change the default subreaper setting so removing
the now invalid config line.

Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com>

Ref: containerd/containerd#1822
It looks like this missed the release notes :(

### DIFF
--- a/pkg/containerd/etc/containerd/config.toml
+++ b/pkg/containerd/etc/containerd/config.toml
@@ -2,7 +2,6 @@ state = "/run/containerd"
 root = "/var/lib/containerd"
 snapshotter = "io.containerd.snapshotter.v1.overlayfs"
 differ = "io.containerd.differ.v1.base-diff"
-subreaper = false
 
 [grpc]
   address = "/run/containerd/containerd.sock"


ijc:
SGTM, although I wonder why we need/want to deviate from the default here, /cc @justincormack.

estesp:
I assumed since it was explicitly there that it was on purpose, but would be good to know

justincormack:
Hmm, the original commit message said "use the default" but it may well be that I accidentally used the default on OSX... I can't see any reason why enabling the subreaper shouldnt work.

ijc:
I guess if we don't merge this PR we'll find out }:-)

estesp:
exactly :)

justincormack:
I think we should just remove this instead...

estesp:
@justincormack makes sense; PR updated

ijc:
I was just about to LGTM but then I thought "I wonder if we ought to optimise for diffability with the output of `./bin/containerd config default`" and went to look, it seems this file is even more out of date than this field. Overwriting the file with the current output produces this diff:
```diff
@@ -1,8 +1,7 @@
-state = "/run/containerd"
 root = "/var/lib/containerd"
-snapshotter = "io.containerd.snapshotter.v1.overlayfs"
-differ = "io.containerd.differ.v1.base-diff"
-subreaper = false
+state = "/run/containerd"
+no_subreaper = false
+oom_score = 0
 
 [grpc]
   address = "/run/containerd/containerd.sock"
@@ -11,7 +10,12 @@ subreaper = false
 
 [debug]
   address = "/run/containerd/debug.sock"
+  uid = 0
+  gid = 0
   level = "info"
 
 [metrics]
   address = ""
+
+[cgroup]
+  path = ""
```

In particular apart from the reaper change it looks like the `snapshotter` and `differ` fields moved in https://github.com/containerd/containerd/commit/3db8adc5d7f7f79ac1fc2f7c2ebb423a888575f3. I'm not sure offhand what the new syntax is, since it seems to relate to how plugin options are specified, I remember not being able to figure this out before...

I wonder if the Go toml bindings have a mode to reject unknown keys and whether containerd should be using it.

justincormack:
Hmm, there is a `cgroup` option now, interesting.

PR #2792, opened 2017-12-04
Add dependent images to build.yml

### LINKED ISSUE #2766: Proposal: add `prebuild` field to `build.yml`
The [kubernetes `pkg/image-cache`](https://github.com/linuxkit/kubernetes/tree/master/pkg/image-cache) build wraps a `Makefile` around the usual `build.yml` machinery in order to run some commands prior to the build (downloading and exporting the Docker images). This means that unlike the majority of other packages (and all the other packages in the `linuxkit/kubernetes` repo) they cannot be built with a simple `linuxkit pkg build «path»`.

I propose adding a `prebuild` field to the `build.yml` format which would be a simple string to be passed to `/bin/sh -c «prebuild»` which could be used to do this preparatory work (I'd either embed a short script using the `|` yaml syntax or call out to a separate script).

Thoughts?

NB: with this in place I would also drop the `build.yml.in` stuff and the dynamic creation of the two package source trees and just duplicate the small amount of content into `pkg/kubernetes-docker-image-cache-@@CACHE@@` directly and get rid of the `Makefile` shenanigans entirely.

NB2: there are some other `Makefile` based builds in tree (at least `tools/guestfs`, `tools/alpine`, `kernel`) but these are using the old scheme and not wrapping `linuxkit build` and would not be helped by this change (at least not directly). Those are covered by issues linked from https://github.com/linuxkit/linuxkit/issues/2587

ijc:
This implements (close to) the second proposal in #2766. There are a couple of preparatory patches but the main work is in the final patch, so I reproduce that commit message here:

>    linuxkit: implement docker image dependencies for pkg build.
>    
>    This allows the `linuxkit/kubernetes` "image-cache" packages to use a standard
>    `linuxkit pkg build` based flow rather than requiring surrounding scaffolding.
>    
>    Fixes #2766. Compared with the original (actually, the second) proposal made in
>    issue #2766, the field is `docker-images` rather than `images` to allow for
>    future inclusion of e.g. `containerd-images`.
>    
>    Signed-off-by: Ian Campbell <ijc@docker.com>

I will shortly be raising a WIP PR against https://github.com/linuxkit/kubernetes which uses this (I need the binary to be built by CI first) and will xpost here.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -30,6 +30,10 @@ A package source consists of a directory containing at least two files:
 - `disable-content-trust` _(bool)_: Disable Docker content trust for this package (default: no)
 - `disable-cache` _(bool)_: Disable build cache for this package (default: no)
 - `config`: _(struct `github.com/moby/tool/src/moby.ImageConfig`)_: Image configuration, marshalled to JSON and added as `org.mobyproject.config` label on image (default: no label)
+- `depends`: Contains information on prerequisites which must be satisfied in order to build the package. Has subfields:
+    - `docker-images`: Docker images to be made available (as `tar` files via `docker image save`) within the package build context. Contains the following nested fields:
+        - `from-file` and `list`: _(string and string list respectively)_. Mutually exclusive fields specifying the list of images to include. Each image must include a valid digest (`sha256:...`) in order to maintain determinism. If `from-file` is used then it is a path relative to (and within) the package directory with one image per line (lines with `#` in column 0 and blank lines are ignore). If `list` is used then each entry is an image.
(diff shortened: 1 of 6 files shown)


ijc:
```
[+ 73m 09s] Building "linuxkit/swap:8daabe644b37ae2fe0d58ab4cdad5847e84c8499"
[+ 73m 13s] No trust data for 8daabe644b37ae2fe0d58ab4cdad5847e84c8499
[+ 73m 13s] No image pulled, continuing with build
[+ 73m 13s] "docker image save" requires at least 1 argument(s).
[+ 73m 13s] See 'docker image save --help'.
[+ 73m 13s] 
[+ 73m 13s] Usage:  docker image save [OPTIONS] IMAGE [IMAGE...]
[+ 73m 13s] 
[+ 73m 13s] Save one or more images to a tar archive (streamed to STDOUT by default)
[+ 73m 13s] exit status 1
```

Looks like I forgot to test the case where this new option wasn't used, oops! Pushed an update.

PR #2793, opened 2017-12-04
Ensure VPNKit process is properly killed when errors occur

xtreme-stevehiehn:
**- What I did**

Ensure VPNKit process is properly killed when errors occur

**- How I did it**

We changed all the log.Fatal* calls to be log.Panic*. We ensure
the program exits with a status code of 1 by using recover()

**- How to verify it**
1. Start LinuxKit (hyperkit backend) with VPNKit networking 
2. Send a KILL signal to hyperkit
3. Ensure vpnkit process exits

VPNKit process will exit when hyperkit fails to start or is terminated

cute animal


xtreme-stevehiehn:
Need to make some edits

PR #2796, opened 2017-12-05
Ensure VPNKit process is properly killed when errors occur

xtreme-stevehiehn:
**- What I did**

Ensure VPNKit process is properly killed when errors occur

**- How I did it**

We added an ExitHandler to logrus so that VPNKit will be shutdown
when log.Fatal* is invoked

**- How to verify it**
1. Start LinuxKit (hyperkit backend) with VPNKit networking 
2. Send a KILL signal to hyperkit
3. Ensure vpnkit process exits

VPNKit process will exit when hyperkit fails to start or is terminated

cute animal

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -291,6 +291,7 @@ func runHyperKit(args []string) {
 				log.Fatalf("Publish ports failed with: %v", err)
 			}
 			defer f()
+			log.RegisterExitHandler(f)
 		default:
 			log.Fatalf("Port publishing requires %q or %q networking mode", hyperkitNetworkingDockerForMac, hyperkitNetworkingVPNKit)
 		}
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -240,14 +240,10 @@ func runHyperKit(args []string) {
 			if err != nil {
 				log.Fatalln("Unable to start vpnkit: ", err)
 			}
-			defer func() {
-				if vpnkitProcess != nil {
-					err := vpnkitProcess.Kill()
-					if err != nil {
-						log.Println(err)
-					}
-				}
-			}()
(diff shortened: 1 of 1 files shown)


rn:
This looks OK to me though I can't say that I'm a fan of using the logger exit function for clean up. However, the alternatives would be to potentially mess around with process groups or enhance VPNKit with some form of dead man handle or some such. @ijc do you have any strong opinion on this? Otherwise I'm inclined to merge.

ijc:
I think given we are "misusing" (well, not really, but YSWIM) the logger to exit (via `log.Fatal`) it seems appropriate to also do the cleanup via a log based hook, even if it does seem a bit weird.

I don't have much appetite for switching away from `log.Fatal` or messing with process groups etc, a dead man's handle in vpnkit sounds nice but I think this PR is ok, so lets go with it.

I wondered why it was necessary to both `defer` and register as an exit handler, but figured out that the latter is not run unless you exit via `log.Fatal` and due to the nature of `os.Exit` (called by `log.Fatal`) there is no chance of the callback running twice, so that's good.

dprotaso:
Our first approach was to change `log.Fatal` to `log.Panic` - and include a defer/recover at the top of runHyperKit in order to exit the process with a status code of 1. You can view the diff here: https://github.com/linuxkit/linuxkit/pull/2793/files This had the problem of masking panics generated outside of the logger.

My ideal solution would be to remove `log.Fatal`s from all the run[Backend] functions and return errors instead. Defers would just work as normal in this situation. The logic to log and exit could be consolidated in `run.go`. This would be a big diff so we decided for the simpler less elegant solution. In hindsight I realize we could have just done this for the hyperkit backend as a start to illustrate the pattern.

rn:
Thanks, I think this is fine for now. At some point we might want to consolidate `linuxkit run` as, over time and with many different contributors, it has grown organically. It started off as a quick way to just test your LinuxKit image. We don't really recommend it for anything else (ie not for production), but that doesn't mean the code couldn't be cleaner...

Issue #2683, opened 2017-11-04
kubernetes: nfs-utils dependencys for pvcs


w9n:
the kubelet container needs to implement fuse dependencies for pvc. RancherOS is building [theirs](https://github.com/rancher/kubernetes-package/blob/master/package/Dockerfile) on debian. There are at least problems with glusterfs in alpine gluster/glusterfs#268 . 
kubernetes-incubator/external-storage#432 Might be related but didnt throw an explicit error.

justincormack:
Do you actually mean `fuse`? There don't seem to be fuse deps there unless they are dependencies of gluster or ceph (maybe they are)? What do you actually want to run?

w9n:
Im not 100% sure about the terminology, but there are volume [provisioner](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#provisioner) in kubernetes and if I want to mount a nfs volume into a pod I have to mount the nfs filesystem in userspace.

I got this after provisioning a pod with a nfs volume claim
```
-11e7-8167-929c9436d3b0" : mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t nfs 10.106.239.49:/export/pvc-8734649f-c163-11e7-8167-929c9436d3b0 /var/lib/kubelet/pods/8740b7c9-c163-11e7-8167-929c9436d3b0/volumes/kubernetes.io~nfs/pvc-8734649f-c163-11e7-8167-929c9436d3b0
Output: mount: wrong fs type, bad option, bad superblock on 10.106.239.49:/export/pvc-8734649f-c163-11e7-8167-929c9436d3b0,
       missing codepage or helper program, or other error
       (for several filesystems (e.g. nfs, cifs) you might
       need a /sbin/mount.<type> helper program)

       In some cases useful info is found in syslog - try
       dmesg | tail or so.

```
after installing nfs-utils
```
 failed: exit status 32
Mounting command: mount
Mounting arguments: -t nfs 10.106.239.49:/export/pvc-8734649f-c163-11e7-8167-929c9436d3b0 /var/lib/kubelet/pods/8740b7c9-c163-11e7-8167-929c9436d3b0/volumes/kubernetes.io~nfs/pvc-8734649f-c163-11e7-8167-929c9436d3b0
Output: mount.nfs: rpc.statd is not running but is required for remote locking.
mount.nfs: Either use '-o nolock' to keep locks local, or start statd.

```

justincormack:
[...]

PR #2800, opened 2017-12-06
[WIP] Contain

w9n:
**- What I did**
Contain is a program to easily execute commands in predefined containers on runtime on containerd. I would like to use the moby service config for container definition at some point.

It can be used as a more sound implementation of a containerized mounter ( linuxkit/kubernetes#16 ) or to basically start any predefined command/container pair.

**- How I did it**
Contain checks each provided container definition in order if its X command parameters are equal to the first X Args of contain and then starts the container with the full args.

**- How to verify it**
~There is a test provided that starts a containerd process, which sometimes weirdly doesnt properly shutdown.~ It seems hard to gracefully shutdown containerd so i just remove the socket file after a SIGKILL

~The test currently fails with:~  The test environment cant be overlay on overlay.
```
failed to mount rootfs component &{overlay overlay [workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/10/work upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/10/fs lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/9/fs]}: invalid argument: unknown
```
~Which I dont know why...~ Feedback is much appreciated. If you want to merge this i would move the dependencies into a build container.

Introduce contain

contain


deitch:
> Contain is a program to easily execute predefined commands in predefined containers

I am confused (but perhaps missed an earlier conversation around this?). Doesn't every `service` and `onboot` already have a `command` which executes it? And if I wanted a different command, could I not just reuse the image and change the `command: ["something"]`?

A sample use case might help, e.g. "I want to be able to do X, here is how I do it, here is how the config changes, etc."

w9n:
I forgot to mention that its for starting container on runtime. so basically the kubernetes use case is `ctn mount ...`, `ctn mount -t nfs ...`, `ctn mount -t glusterfs ...` executes the mount command in different containers, which either already exist or get pulled and created by the contain provided config. I would also like to use it as some kind of docker-compose replacement for desktop apps.

deitch:
These commands are meant to be run to do a command _other_ than the listed `command` for the container, e.g. an init process (like `ctn mount`) or parallel (like a sidecar)? Who runs `ctn mount`?

w9n:
The kubelet invokes ` ctn mount` from the `--experimental-mounter-path` parameter. You can have a look at linuxkit/linuxkit#2683 and linuxkit/kubernetes#16

deitch:
> The kubelet invokes ctn mount from the --experimental-mounter-path parameter. You can have a look at linuxkit/linuxkit#2683 and linuxkit/kubernetes#16

Ah, wasn't as involved with that one (wanted to be, but head down on building and manipulating filesystems from within `moby/tool`), so I probably am missing the use case. Will watch this but refrain until the people involved there @ijc and @justincormack comment.

w9n:
I developed this into the direction of a containerd cli and dont think it fits in as a linuxkit pkg.

PR #2803, opened 2017-12-08
Update to Alpine 3.7

rn:
Update everything to alpine:3.7 starting with the `linuxkit/alpine`.

All packages should be pushed for x86 and arm64. I have not rebuild the kernels (but checked that the kernels can still be build).

I've verified manually for a few packages that they are signed/multiarch images on hub.

Let's see what CI thinks about this

eagle-walk


pkg/mkimage/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM linuxkit/alpine:07f7d136e427dc68154cd5edbb2b9576f9ac5213 AS mirror
+FROM linuxkit/alpine:d307c8a386fa3f32cddda9409b9687e191cdd6f1 AS mirror
```
ijc: This package is in turn used as `FROM` in `mkimag-{gcp,vhd,vmdk,dynamic-vhd}` but those aren't updated here.

rn: They seem to use `linuxkit/guestfs` which in turns is based on `debian:stretch`

ijc: Oh yes, my mistake. I somehow misread `pkg/mkimage` as `tools/guestfs`! Sorry for the noise.

rn:
I managed to find a fix for the `linuxkit/qemu-ga` package. It was some slight change in the alpine package package dependency which I fixed in https://github.com/linuxkit/linuxkit/pull/2803/commits/a41dd90a646de5fccafe87a10091c4187eac6307

Re projects, I can remove the update from all of them if you prefer. I think we have been fairly inconsistent updates of them...

ijc:
I guess there is a reasonable chance that most of the projects were either already out of date and broken (as you say we haven't been terribly consistent about updates) or are being maintained and hence should be fine because they are already reasonably up to date. So I suppose given you have done the update we may as well do it.

rn:
The `linuxkit.kernel.kmod_4.14.x` test is failing with:
```
[STDOUT  ] 2017-12-08T17:25:37.294133Z:   CC [M]  /kmod/hello_world.o
[STDOUT  ] 2017-12-08T17:25:37.300917Z: cc1: error: incompatible gcc/plugin versions
[STDOUT  ] 2017-12-08T17:25:37.301017Z: cc1: error: fail to initialize plugin ./scripts/gcc-plugins/structleak_plugin.so
[STDOUT  ] 2017-12-08T17:25:37.301068Z: cc1: error: incompatible gcc/plugin versions
[STDOUT  ] 2017-12-08T17:25:37.301104Z: cc1: error: fail to initialize plugin ./scripts/gcc-plugins/randomize_layout_plugin.so
[STDOUT  ] 2017-12-08T17:25:37.304657Z: make[2]: *** [scripts/Makefile.build:321: /kmod/hello_world.o] Error 1
[STDOUT  ] 2017-12-08T17:25:37.305007Z: make[1]: Leaving directory '/usr/src/linux-headers-4.14.4-linuxkit'
[STDOUT  ] 2017-12-08T17:25:37.30514Z: make[1]: *** [Makefile:1503: _module_/kmod] Error 2
[STDOUT  ] 2017-12-08T17:25:37.305177Z: make: *** [Makefile:4: all] Error 2
```
So, will be compiling a new 4.14 kernel for arm and x86 and then hit rebuild

rn:
x86 4.14.x kernel pushed and restarted tests

PR #2818, opened 2017-12-18
Improve error message for new packages

errordeveloper:
When I tried to use `linuxkit pkg build` for the first time on a new package, it showed `Unable to parse ls-tree output: ""` error, which I could not parse before I looked at the code. Hopefully, this change would make it easier for people to use the tool for the first time on a new package.

### DIFF
--- a/src/cmd/linuxkit/pkglib/git.go
+++ b/src/cmd/linuxkit/pkglib/git.go
@@ -89,6 +89,10 @@ func (g git) treeHash(pkg, commit string) (string, error) {
 		return "", err
 	}
 
+	if out == "" {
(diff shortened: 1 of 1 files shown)


src/cmd/linuxkit/pkglib/git.go
```diff
@@ -89,6 +89,10 @@ func (g git) treeHash(pkg, commit string) (string, error) {
 		return "", err
 	}
 
+	if len(out) == 0 {
```
ijc: We generally prefer `out == ""` in this code, but otherwise LGTM.

errordeveloper: Ah, that's what I wrote at first :)

PR #2819, opened 2017-12-18
Kernel update to 4.14.7/4.9.70/4.4.106

rn:
Mechanic update of kernels, except I also rolled in a change to the kernel build to add `forcepush` and `forcebuild` targets. I was meant to add that a while, but I had screwed up a push of the 4.14.7 x86 kernel package and needed a force push, which is now there.

boobies


rn:
arm64/x86_64 kernel build and pushed

PR #2822, opened 2017-12-25
Mitigate ebpf privilege escalation

rn:
Due to ebpf verifier bugs introduced in 4.9.x and 4.14.x ebpf programs can access (read/write) memory and therefor can corrupt memory and in some case even gain higher privileges. Setting the `kernel.unprivileged_bpf_disabled=1` mitigates this by preventing unprivileged user to load ebpf programs.

See: https://lwn.net/Articles/742170

We should probably keep this sysctl even when the verifier is fixed upstream.

bear-miss


justincormack:
LGTM. If people want unpriv they can override. I believe thus is also
stopoed by docker seccomp.

PR #2823, opened 2017-12-25
pkg/sysctl fixes: support commented KV lines, additional seperator and comment character, no post-crit KV set skips

eyz:
**- What I did**
- Correctly ignoring all commented KV lines, not only lines without equal signs
- Removing all whitespace in KVs (such as tabs as well) instead of only single spaces
- Added semicolon as a new comment prefix
- Added forward slash as separator character
- KV set fails are now logged as warnings with a prefix `WARN:`, and do not skip the rest of the KVs from being attempted (no silent fails/skips)

**- How I did it**
- Trimming the KV line and checking for hash (or semicolon) as first character and skipping line earlier in code block.
- Using `TrimSpace` to trim any whitespace, not only single space characters.
- Logging sysctl KV set error with `log.Println` instead of failing on first error with `log.Fatal`, so the rest of the KV sets can be attempted (instead of skipping them silently)

**- How to verify it**
- Lines such as the following now aren't attempted to be set: `#a.b.c=xyz`
- Lines such as the following are now skipped as comments `;a.b.c=xyz`
- Keys can be composed of periods `a.b.c=xyz`
- Keys can be composed of forward slashes `a/b/c=xyz`
- Disable IPv6 via cmdline `ipv6.disable=1` in kernel, and now `pkg/sysctl` won't skip the KV sets after `net.ipv6.conf.all.accept_redirects = 0`

pkg/sysctl fixes: support commented KV lines, additional seperator and comment character, no post-crit KV set skips

media-20171119

### DIFF
--- a/pkg/sysctl/main.go
+++ b/pkg/sysctl/main.go
@@ -22,17 +22,19 @@ func init() {
 }
 
 func sysctl(line []byte) error {
+	sysctlLineTrimmed := strings.TrimSpace(string(line[:]))
+	// skip any commented lines
+	if len(sysctlLineTrimmed) >= 1 && (sysctlLineTrimmed[:1] == "#" || sysctlLineTrimmed[:1] == ";") {
+		return nil
+	}
 	// parse line into a string of expected form X.Y.Z=VALUE
-	sysctlLineKV := strings.Split(string(line[:]), "=")
+	sysctlLineKV := strings.Split(sysctlLineTrimmed, "=")
 	if len(sysctlLineKV) != 2 {
-		if len(sysctlLineKV) >= 1 && len(sysctlLineKV[0]) >= 1 && strings.Trim(sysctlLineKV[0], " ")[:1] == "#" {
-			return nil
-		}
-		return fmt.Errorf("Cannot parse %s", string(line))
+		return fmt.Errorf("Cannot parse %s", sysctlLineTrimmed)
 	}
 	// trim any extra whitespace
-	sysctlSetting, sysctlValue := strings.Trim(sysctlLineKV[0], " "), strings.Trim(sysctlLineKV[1], " ")
-	sysctlFile := filepath.Join(sysctlDir, filepath.Join(strings.Split(sysctlSetting, ".")...))
+	sysctlSetting, sysctlValue := strings.TrimSpace(sysctlLineKV[0]), strings.TrimSpace(sysctlLineKV[1])
+	sysctlFile := filepath.Join(sysctlDir, filepath.Join(strings.FieldsFunc(sysctlSetting, splitKv)...))
 	file, err := os.OpenFile(sysctlFile, os.O_WRONLY, 0)
 	if err != nil {
 		return fmt.Errorf("Cannot open %s: %s", sysctlFile, err)
@@ -43,6 +45,10 @@ func sysctl(line []byte) error {
 		return fmt.Errorf("Cannot write to %s: %s", sysctlFile, err)
 	}
 	return nil
+}
+
+func splitKv(r rune) bool {
+	return r == '.' || r == '/'
 }
 
 func main() {
@@ -65,7 +71,7 @@ func main() {
 			}
 			err = sysctl(line)
 			if err != nil {
-				log.Fatal(err)
+				log.Println(fmt.Errorf("WARN: %v", err))
 			}
 		}
 	}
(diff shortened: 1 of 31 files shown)


pkg/sysctl/main.go
```diff
@@ -65,7 +68,7 @@ func main() {
 			}
 			err = sysctl(line)
 			if err != nil {
-				log.Fatal(err)
+				log.Println(fmt.Errorf("WARN: %v", err))
```
w9n: I think this should be `log.Warnln(err.Error())` or `log.Fatalln(err.Error())`, otherwise LGTM to me.

eyz: Thanks, @w9n! I didn't see Warnln

eyz: Actually, @w9n, I just looked again - I don't see ant options other than `Println`, `Fatalln`, or `Panicln`, as this isn't using one of the third party logging libraries. See https://golang.org/pkg/log/

w9n: sorry, i looked at `log "github.com/sirupsen/logrus"` in my project, so your Println seems good in the standard pkg

pkg/sysctl/build.yml
```diff
@@ -1,6 +1,5 @@
 image: sysctl
 config:
   pid: "host"
-  readonly: true
```
w9n: Why did you remove this one?

eyz: @w9n see the last line of each bullet above. I got errors about adding confs in via the image scoped YAML when read-only

pkg/sysctl/main.go
eyz: No problem. Thanks for the review!

pkg/sysctl/build.yml
eyz: Perhaps we need an issue for changing the filesystem to readonly only after the image file adds from the YAML?

w9n: You can `mount --bind` files on a read only filesystem, this should work in your case aswell.

eyz: OK. I'll get more information and report back.

w9n:
should not be necessary as successfully working in https://github.com/linuxkit/kubernetes/blob/master/yml/kube.yml#L53 and Bound in L13

w9n:
everything going into the direction of https://www.freedesktop.org/software/systemd/man/sysctl.d.html as e.g. comments should be appreciated, i think.

eyz:
@w9n  I'll update the code to support `;` as a comment prefix and `/` as a separator

eyz:
The new comment and separator characters have been added, and `readonly: true' reverted back to before the PR

pkg/sysctl/main.go
```diff
@@ -22,18 +22,21 @@ func init() {
...
-		return fmt.Errorf("Cannot parse %s", string(line))
+		return fmt.Errorf("Cannot parse %s", sysctlLineTrimmed)
 	}
 	// trim any extra whitespace
-	sysctlSetting, sysctlValue := strings.Trim(sysctlLineKV[0], " "), strings.Trim(sysctlLineKV[1], " ")
-	sysctlFile := filepath.Join(sysctlDir, filepath.Join(strings.Split(sysctlSetting, ".")...))
+	sysctlSetting, sysctlValue := strings.TrimSpace(sysctlLineKV[0]), strings.TrimSpace(sysctlLineKV[1])
+	sysctlFile := filepath.Join(sysctlDir, filepath.Join(strings.FieldsFunc(sysctlSetting, splitKv)...))
 	file, err := os.OpenFile(sysctlFile, os.O_WRONLY, 0)
+
```
justincormack: can you remove this blank line, as the test below is for this `err` so a blank is a bit confusing.

justincormack:
One very minor comment but otherwise LGTM

pkg/sysctl/main.go
eyz: Yes, agreed that was a bit odd. Thanks. Should be good now to proceed as of green CI.

rn:
@eyz I've build and pushed the new package for x86 and arm64. Could you add a commit to update the YAMLs to the new hash, ie: `linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21`. That way the CI will pick up the new package

eyz:
@rn I had to rebase against linuxkit/linuxkit upstream master. Once CI passes again, could you please rebuild the image and give me the new tag?

rn:
The tag doesn't change from a rebase as it is based on the content hash. It's just the commit hash in the metadata/label which is out of date. I've force pushed an update

eyz:
Updated image tag to linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21

eyz:
@rn CI has completed, and is green

PR #2824, opened 2017-12-27
Fix sriov flag on AWS

justincormack:
This needs to be set to nil if empty.

Signed-off-by: Justin Cormack <justin@specialbusservice.com>

arctic_hare_1


justincormack:
Also added a small doc improvement

PR #2825, opened 2017-12-28
Update kernels

rn:
- Update kernels to 4.14.8/4.9.71/4.4.107
- Update kernels to 4.14.9/4.9.72/4.4.108
- 4.14.9 had some renames of config options, adjust our config
- Update YAMLs

The 4.14.9 and 4.9.72 kernels include fixes to the eBPF verifier referred to in https://github.com/linuxkit/linuxkit/pull/2822 but we should probably keep the sysctl setting which mitigates the priv escalation anyway.

eagle-walk-2


rn:
x86_64 and arm64 kernels build and pushed

PR #2832, opened 2018-01-04
Use pure Go unix.Reboot function in rc.init

tklauser:
Use unix.Reboot from golang.org/x/sys/unix for poweroff and reboot
instead of relying on external commands.

Signed-off-by: Tobias Klauser <tklauser@distanz.ch>

### DIFF
--- a/blueprints/docker-for-mac/base.yml
+++ b/blueprints/docker-for-mac/base.yml
@@ -4,7 +4,7 @@ kernel:
   cmdline: "console=ttyS0 page_poison=1"
 init:
   - linuxkit/vpnkit-expose-port:728e5fe9e6b818d9825b28826b929ae75a386e9e # install vpnkit-expose-port and vpnkit-iptables-wrapper on host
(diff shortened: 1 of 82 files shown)


justincormack:
Thanks, LGTM once CI is happy and we rebuild the container.

tklauser:
Oops, sorry. Now changed, PTAL.

rn:
Thanks for this improvement. I've build, pushed, signed the update packages. Could you update this PR with an *additional* commit to update the YAML files.

Something like this in the root directory should do the trick:
```
scripts/update-component-sha.sh linuxkit/init:9250948d0de494df8a811edb3242b4584057cfe4 linuxkit/init:5a577d070817b4f17821657823082651baafd4ed
```

That way CI will actually test with the update packages.

tklauser:
YAML files now updated and pushed in an additional commit. Sorry for missing this previously and thanks for your help @rn !

PR #2833, opened 2018-01-04
Update Moby tool

justincormack:
- containerd 1.0 vendoring
- yaml omitempty
- null pointer dereference

actic_noaa_walrus_formatted


ijc:
Might be worth propagating the relevant changes from `src/cmd/linuxkit/vendor/github.com/moby/tool/vendor.conf` into `src/cmd/linuxkit/vendor.conf`.

justincormack:
Ah yes

justincormack:
Added commit to bring the two into line

justincormack:
ok, all match except x/sys which is newer here and needs to be, will update moby/tool to match later.

PR #2845, opened 2018-01-10
Add an example of setting up a writeable overlay of a host path

ijc:

### DIFF
--- a/examples/hostmount-writeable-overlay.yml
+++ b/examples/hostmount-writeable-overlay.yml
@@ -1,0 +1,47 @@
+kernel:
+  image: linuxkit/kernel:4.9.76
+  cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
+init:
+  - linuxkit/init:81fc00e879cf56fa6f058f6c891b012bc867de7f
+  - linuxkit/runc:abc3f292653e64a2fd488e9675ace19a55ec7023
+  - linuxkit/containerd:e58a382c33bb509ba3e0e8170dfaa5a100504c5b
+  - linuxkit/ca-certificates:de21b84d9b055ad9dcecc57965b654a7a24ef8e0
+onboot:
+  - name: sysctl
+    image: linuxkit/sysctl:4c1ef93bb5eb1a877318db4b2daa6768ed002e21
+  - name: dhcpcd
+    image: linuxkit/dhcpcd:0d59a6cc03412289ef4313f2491ec666c1715cc9
+    command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
+onshutdown:
+  - name: shutdown
+    image: busybox:latest
+    command: ["/bin/echo", "so long and thanks for all the fish"]
+services:
+  - name: getty
+    image: linuxkit/getty:22e27189b6b354e1d5d38fc0536a5af3f2adb79f
+    env:
+     - INSECURE=true
+    runtime:
+      mounts:
+        # Makes a writeable (but private and non-persistent) mount of the
+        # host etc into the container.
+        - type: overlay
+          source: overlay
+          destination: writeable-host-etc
+          options: ["rw", "lowerdir=/etc", "upperdir=/run/hostetc/upper", "workdir=/run/hostetc/work"]
+  - name: rngd
+    image: linuxkit/rngd:94e01a4b16fadb053455cdc2269c4eb0b39199cd
+  - name: nginx
+    image: nginx:alpine
+    capabilities:
+     - CAP_NET_BIND_SERVICE
+     - CAP_CHOWN
+     - CAP_SETUID
+     - CAP_SETGID
+     - CAP_DAC_OVERRIDE
+    binds:
+     - /etc/resolv.conf:/etc/resolv.conf
+trust:
+  org:
+    - linuxkit
+    - library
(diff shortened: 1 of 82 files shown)


rn:
Failure seems unrelated. Hit rebuild

ijc:
@justincormack (and I) think that `runtime.mkdir` should be usable to create the mount point, but in my testing it didn't work using:
```
    runtime:
      mkdir:
        - /writeable-host-etc
      mounts:
        - type: overlay
          source: overlay
          destination: /writeable-host-etc
          options: ["rw", "lowerdir=/etc", "upperdir=/run/hostetc/upper", "workdir=/run/hostetc/work"]
```
Results in:

```ERRO[0000] preparing filesystem                          error="Cannot create directory for mount destination /writeable-host-etc: mkdir /writeable-host-etc: read-only file system"
```
The container is not  readonly (default is writeable and I didn't tell it otherwise).

Justin was going to take a look and try to fix, we could either merge this as is (representing the status quo) and update along with that fix, or just defer this file into the fix PR. I figure we may as well merge this now.

justincormack:
`mkdir` happens after the mounts, so no. Before mounts there probably isnt anywhere writeable anyway. However it will `mkdir` any `destination` as well, if it is writeable, I think.

justincormack:
Also, just not setting `readonly` sets up a writeable overlay anyway, so not sure why you need to do this. The build takes care of including the necessary mountpoints.

ijc:
I will rebase onto #2849 (which is now in master) and adjust based on the fixes there.

ijc:
@justincormack Things given in the `runtime.mounts` yml appear to go _before_ the rootfs overlay setup, so the filesystem is still readonly at this point since it is the host rootfs:
```
(ns: sshd) linuxkit-2ae2c420a11c:/containers/services/getty# jq .mounts[].destination runtime.json 
"/writeable-host-etc"
"/containers/services/getty/tmp"
"/containers/services/getty/tmp"
"/containers/services/getty/rootfs"
```
Bug?

ijc:
I've rebased over the `mkdir` change, although it is actually irrelevant/pointless since as @justincormack says the mkdir is after mounts anyway.

justincormack:
The order is
1. runtime mounts
2. mkdir
3. runc mounts

runtime mounts _includes_ the overlay mounts (that is how they are implemented) but I guess the ordering is user specified ones then overlay, and that could be reversed.

The relevant code is

https://github.com/moby/tool/blob/master/src/moby/image.go#L253-L258
which is currently appending not pre-prending.

ijc:
I'm currently editing that code in to test inverting it, I expect that will work and I think it probably makes sense to do so.

justincormack:
Yes I think it makes sense, as the generated mounts for overlay are not dependent on anything so may as well be done first.

ijc:
Possibly with the `mkdir` fix I could now successfully use the runc mounts for this, which might also make sense.

justincormack:
Yes, maybe it doesnt need to be changed... you should be able to do pretty much anything in runc mounts too.

ijc:
There seem to be two workable solutions. With the mkdir fix from #2849 this works:
```
    runtime:
      mkdir:
        - writeable-host-etc
        - /run/hostetc/upper
        - /run/hostetc/work
    mounts:
      - type: overlay
        source: overlay
        destination: /writeable-host-etc
        options: ["rw", "lowerdir=/etc", "upperdir=/run/hostetc/upper", "workdir=/run/hostetc/work"]
```
(that's a runc mount).

If prepending (rather than appending) the rootfs overlay to the `runtime.mounts` then IMHO this should also work:
```
    runtime:
      mounts:
        - type: overlay
          source: overlay
          destination: writeable-host-etc
          options: ["rw", "lowerdir=/etc", "upperdir=/run/hostetc/upper", "workdir=/run/hostetc/work"]
```
However that does not work (even with changing moby to preprend the rootfs mounts instead of appending) because `runtime.mounts[].destination` is also relative to the host root. I think it should be changed to be relative to the container rootfs.

My preference would be for the second one to work since it is more concise (in particular you don't have to mkdir the overlay work and upper dirs manually).

So I think I will:
1. Patch moby to prepend the rootfs to `runtime.mounts`.
2. Fix the runtime.mounts support code in the same way the `runtime.mkdir` was just fixed in #2849.
3. Use the `runtime.mounts` variant in this example.

justincormack:
SGTM

ijc:
https://github.com/moby/tool/pull/195 implements 1. in my list. Eventually vendoring that becomes the number 1 here...

ijc:
Updated with https://github.com/linuxkit/linuxkit/pull/2845#issuecomment-356907482:
> 1. Patch moby to prepend the rootfs to runtime.mounts.
> 2. Fix the runtime.mounts support code in the same way the runtime.mkdir was just fixed in #2849.
> 3. Use the runtime.mounts variant in this example.

examples/hostmount-writeable-overlay.yml
```diff
@@ -0,0 +1,69 @@
...
+     - CAP_SETGID
+     - CAP_DAC_OVERRIDE
+    binds:
+     - /etc/resolv.conf:/etc/resolv.conf
+files:
+  - path: etc/containerd/config.toml
+    contents: |
+      state = "/run/containerd"
+      root = "/var/lib/containerd"
+      snapshotter = "io.containerd.snapshotter.v1.overlayfs"
```
justincormack: you could take this `files` section out as it does not add to the example.

ijc: Removed the `files` section

PR #2849, opened 2018-01-10
Fix runtime mkdir

justincormack:
Make relative paths in runtime mkdir be relative to container rootfs
    
Currently all paths were treated as absolute paths.

Will update docs in moby/tool.

Also fix logging to use logrus.

pandacub

### DIFF
--- a/pkg/init/cmd/service/prepare.go
+++ b/pkg/init/cmd/service/prepare.go
@@ -143,6 +143,7 @@ func prepareFilesystem(path string, runtime Runtime) error {
 	// execute the runtime config that should be done up front
 	// we execute Mounts before Mkdir so you can make a directory under a mount
 	// but we do mkdir of the destination path in case missing
+	rootfs := filepath.Join(path, "rootfs")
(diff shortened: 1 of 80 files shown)


pkg/init/cmd/service/prepare.go
```diff
@@ -164,6 +165,11 @@ func prepareFilesystem(path string, runtime Runtime) error {
 			return fmt.Errorf("Failed to mount %s: %v", mount.Source, err)
 		}
 	}
+	// relative paths should be relative to container rootfs
+	err := os.Chdir(rootfs)
```
ijc: IIRC all the container startup happens in the same go process these days, so might this confuse some later thing which expected to be run from `/`?

I think this bit of service launch is not parallel so we are ok from that PoV.

Maybe it would be safer to use `filepath.IsAbs` and `filepoath.Join` though?

justincormack: Yes, probably better.

PR #2852, opened 2018-01-10
Improve packet.net support

rn:
The main purpose of this PR is to serve the iPXE script via HTTP instead of passing it as userdata. This will allow us to use userdata for other things, like per instance configuration information.

- If the `-serve` option is specified, serve the iPXE scrip from memory and the kernel/initrd from the filesystem
- add a new `linuxkit push packet` which writes the kernel/initrd as well as the iPXE script to a directory
- add a new `linuxkit serve` command to start a local webserver serving the content of a specified directory.

While at it also a bunch of fixes and cleanup as well as update the packet.net go bindings.

saiga


docs/platform-packet.md
```diff
@@ -40,11 +40,17 @@ retry the boot typically fixes this.
...
-via
-[iPXE](https://help.packet.net/technical/infrastructure/custom-ipxe). iPXE
-booting requires a HTTP server on which you can store your images. The
-`-base-url` option specifies the URL to the HTTP server.
+LinuxKit on Packet boots the `kernel+initrd` output from moby via
+[iPXE](https://help.packet.net/technical/infrastructure/custom-ipxe)
+which also requires a iPXE script. iPXE booting requires a HTTP server
+on which you can store your images. The `-base-url` option specifies
+the URL to the HTTP server. The boot process expects `<name>-kernel`,
+`<name>-initrd.img`, and `<name>-packet.ipxe` to be available at the
```
justincormack: I think it probably doesnt make sense to put "packet" in the name - there are other ipxe platforms and they generally work in exactly the same way, so a generic name is probably good and then we can reuse the code.

rn: I used to have `<name>.ipxe` but chose to include `packet` because some of the ipxe options may be specific. I can change it back...

rn: for example serial on `ttyS1` on x86

docs/platform-packet.md
```diff
@@ -62,9 +68,10 @@ PACKET_API_KEY=<API key> PACKET_PROJECT_ID=<Project ID> \
 linuxkit run packet -serve :8080 -base-url <ngrok url> packet
 ```
 
-To boot a `arm64` image for Type 2a machine (`-machine
-baremetal_2a`) you currently need build using `linuxkit build packet.yml packet.arm64.yml` and then un-compress both the kernel and
-the initrd before booting, e.g:
+To boot a `arm64` image for Type 2a machine (`-machine baremetal_2a`)
+you currently need build using `linuxkit build packet.yml
```
ijc: "...currently need to build using..." (extra "to")

docs/platform-packet.md
```diff
@@ -40,11 +40,17 @@ retry the boot typically fixes this.
...
-LinuxKit on Packet boots the `kernel+initrd` output from moby
-via
-[iPXE](https://help.packet.net/technical/infrastructure/custom-ipxe). iPXE
-booting requires a HTTP server on which you can store your images. The
-`-base-url` option specifies the URL to the HTTP server.
+LinuxKit on Packet boots the `kernel+initrd` output from moby via
+[iPXE](https://help.packet.net/technical/infrastructure/custom-ipxe)
+which also requires a iPXE script. iPXE booting requires a HTTP server
+on which you can store your images. The `-base-url` option specifies
+the URL to the HTTP server. The boot process expects `<name>-kernel`,
```
ijc: Maybe this will become clear later but I'm unsure from this whether `base-url` is a local path which will be served by a(n existing) webserver or the public URL from which the directory would be available to others (e.g. packet).

rn: rephrased

docs/platform-packet.md
rn: fixed

src/cmd/linuxkit/push_packet.go
```diff
@@ -0,0 +1,147 @@
...
+	flags.Usage = func() {
+		fmt.Printf("USAGE: %s push packet [options] [name]\n\n", invoked)
+		fmt.Printf("Options:\n\n")
+		flags.PrintDefaults()
+	}
+	baseURLFlag := flags.String("base-url", "", "Base URL that the kernel, initrd and iPXE script are served from (or "+packetBaseURL+")")
+	nameFlag := flags.String("img-name", "", "Overrides the prefix used to identify the files. Defaults to [name] (or "+packetNameVar+")")
+	archFlag := flags.String("arch", packetDefaultArch, "Image architecture (x86_64 or aarch64)")
+	decompressFlag := flags.Bool("decompress", packetDefaultDecompress, "Decompress kernel/initrd before pushing")
+	dstFlag := flags.String("destination", "", "URL where to push the image to. Currently only 'file' is supported as a scheme (which is also the default if omitted)")
```
ijc: The default given here is `""` and later on there is a `log.Fatal` inside an `if *dstFlag == ""` with nothing to set a default of `file` that I see in between.

ijc: Oh, I see what you mean now, i misread.... only the scheme is defaulted, but you do need a path. gotcha.

docs/platform-packet.md
ijc: It is clarified in the following examples, so I think it's ok.

src/cmd/linuxkit/push_packet.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := packetCopy(filepath.Join(dstPath, kernelName), kernelName, decompress); err != nil {
+		log.Fatalf("Error copying kernel: %v", err)
+	}
+
+	initrdName := fmt.Sprintf("%s-initrd.img", name)
+	if err := packetCopy(filepath.Join(dstPath, initrdName), initrdName, decompress); err != nil {
+		log.Fatalf("Error copying initrd: %v", err)
+	}
+
+	ipxeScriptName := fmt.Sprintf("%s-packet.ipxe", name)
```
justincormack: My comment disappeared but werent you going to remove the `packet`?

rn: My reply may have disappeared too. The `iPXE` script is somewhat packet specific, like we add `ttyS1` as the console etc. I can drop it if desired, but think having it more specific might be better

justincormack: Oh ok, we can always revisit later

PR #2853, opened 2018-01-10
Added capability to find a host if -hostname isn't specified

thebsdbox:
**- What I did**

At the request of a user on slack, added in the capability to auto find a host that can be used to push an image to a vCenter cluster.

**- How I did it**

If `-hostname` isn't used it will return all hosts in the DC cluster and then use `rand` to choose one of them.

**- How to verify it**

Tested using the `-hostname` and omitting the flag, both pushed to the datastore successfully.

This adds the capability to auto-find a vSphere host in a vCenter cluster as one is needed for image upload. This doesn't add any additional load to the cluster as no virtual machines are started or created.

### DIFF
--- a/src/cmd/linuxkit/run_vcenter.go
+++ b/src/cmd/linuxkit/run_vcenter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math/rand"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


kleinsasserm:
@thebsdbox thank you!, I've tested this PR and it works now as expected mit our vcenter and *multiple* vmware hosts in the vcenter!
```
[vcenterbranch]$ ./linuxkit -v push vcenter -url=https://aaa:bbb@vcenter/sdk -datastore=_lalelu -datacenter=vvv /home/mario/go/src/linuxkit/bin/docker.iso 
INFO[0007] Uploading LinuxKit file [/home/mario/go/src/linuxkit/bin/docker.iso] 
```

PR #2856, opened 2018-01-12
Support Preempt-RT Linux kernel

TiejunChina:
In order to get such a preempt-rt Linux kerne, we need to grab -rt patch via
https://www.kernel.org/pub/linux/kernel/projects/rt/. So far we just enable it
over 4.14.x, so https://www.kernel.org/pub/linux/kernel/projects/rt/4.14 is our
current repo.

Now we can add "-rt" to build this specific Linux kernel. But by default, we still
get the original kernel, or by setting with "-std" particularly. 

Signed-off-by: Tiejun Chen <tiejunc@vmware.com>

-- I've talked this proposal with Justin Cormack, Amir Chaudhry, and Bobby Zhang.


rn:
Thanks for the patches. I have not looked into them in great detail, but I have a couple of high level comments about the approach.

My main concerns are:
- the `rt` patches lag the mainline and are not available for all kernel release, so we will have, say a 4.14.13 kernel and a 4.14.8-rt kernel.
- I don't want to maintain separate full config files for the `rt` kernels. The config files are carefully constructed to follow best practises for security and it's harder to maintain two more configs. I'd much rather prefer if they use the default kernel config and then we append the diff required for the `rt` kernel, very much like we do for the debug kernels.

I also think the patches would be a lot simpler if the `rt` patches were checked in under `patches-4.14.x-rt/`. Then we could just check if `patches-${KERNEL_SERIES}-${EXTRA}` exist and apply them in favour of `patches-${KERNEL_SERIES}`

Lastly, we are currently preparing a a 0.2 release (targeted for end of January) and I probably need to make some changes to the kernel build before then (likely by the end of next week) so it might be good to hold off till then.

TiejunChina:
Rolf,

Thank you for taking your time on this.

Regard to your two concerns:
#1. "the rt patches lag the mainline and are not available for all kernel release" is true. So far the latest rt version is 4.14.12-rt10. Although I can or I'd like to rebase that to align with linuxkit base kernel, 4.14.13, I just feel it is not much significance. So instead, I tried to maintain this in Makefile.

+$(eval $(call kernel,4.14.12,4.14.x,-rt))
+$(eval $(call kernel,4.14.12,4.14.x,-rt-dbg))

This makes our life easier.

#2. I ever tried to follow debug extra fashion. But you know, some Kconfigs conflict between standard kernel and rt kernel. Like CONFIG_TRANSPARENT_HUGEPAGE, this depends on !CONFIG_PREEMPT_RT_FULL.  So this means we have to remove some Kconfigs from the original config file, and then add other RT Kconfigs to come out that rt config file. If this way is fine to you, I definitely would like to figure out that in detail.  

For rt patches's suggestion, what do you mean? I should always have patches-4.14.x-rt/ including one set of patches specific to one given rt version? But this means we ourselves have to update those rt patches for each Linux RT kernel release.  So I still prefer to download the latest rt patches automatically while building rt kernel. I mean, we just say we support 4.14.12-rt, regardless of which rt sub version in https://www.kernel.org/pub/linux/kernel/projects/rt/4.14/4.14.12-rt[x]. But for us, [x] is always the latest version right there.

Note NP, just let us first review this and then I can rebase this after your change to 2.0 release.

rn:
Apologies for the late reply.

I don't have an issue with the `rt` patches lagging and I'd be ok with building older versions for `rt`. I wouldn't want to go through the effort of rebasing the `rt` patches at least once a week when a new kernel comes out. My comment regarding the patches was mostly about where they are kept. The `Dockerfile` we use for building the kernel is already too complex for my liking, so I would like to keep the addition of `-rt` as simple as possible. By checking in the patches under `patches-4.14.x-rt` makes the logic of applying them a lot easier compared to downloading and verifying etc. This would still handle different kernel versions for `-rt` and non `-rt`. so you could have:
```
$(eval $(call kernel,4.14.15,4.14.x,))
$(eval $(call kernel,4.14.12,4.14.x,-rt))
```

For the config, I don't know `kconfig` well enough, but could you say use `config-4.14.x-x86_64` as is and disable `CONFIG_TRANSPARENT_HUGEPAGE` in `config-4.14.x-x86_64-rt`. If you then concatenate `config-4.14.x-x86_64-rt` to `config-4.14.x-x86_64` would that override the config? Or would the build automatically disable, say `CONFIG_TRANSPARENT_HUGEPAGE`? In that could skip the check that the config did not change, like we do for `-dbg` kernels.

TiejunChina:
@rn NP :)

I'd like to address your comments so now, I create kernel/patches-4.14.x-rt/patches-4.14.15-rt13.tar.gz, and two different extra  rt config files, config-4.14.x-aarch64-rt and config-4.14.x-x86_64-rt, like -dbg.

Now my primary change should be refactored as follows:
diff --git a/kernel/Dockerfile b/kernel/Dockerfile
index 96d3ca7..18ecc91 100644
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -28,6 +28,9 @@ RUN apk add \
     xz-dev \
     zlib-dev
 
+# Update patch utility
+RUN apk update && apk add --update patch
+
 # libunwind-dev pkg is missed from arm64 now, below statement will be removed if the pkg is available.
 RUN [ $(uname -m) == x86_64 ] && apk add libunwind-dev || true
 
@@ -61,8 +64,23 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
     gpg2 --verify linux-${KERNEL_VERSION}.tar.sign linux-${KERNEL_VERSION}.tar && \
     cat linux-${KERNEL_VERSION}.tar | tar --absolute-names -x && mv /linux-${KERNEL_VERSION} /linux
 
-# Apply local patches if present
 WORKDIR /linux
+# Apply Preempt-RT kernel patches in the case of RT Linux Kernel
+RUN set -e && \
+    if [ -n "${EXTRA}" ]; then \
+       KERNEL_TYPE=$( (echo "${EXTRA}") | (grep -oE '\-rt|\-std') ); \
+       if [ -z "${KERNEL_TYPE}" ]; then KERNEL_TYPE="-std"; fi; \
+       echo "Building ${KERNEL_TYPE}"; \
+       if [ "${KERNEL_TYPE}" = "-rt" ] && [ -d /patches-${KERNEL_SERIES}-rt ]; then \
+           tar -zxf /patches-${KERNEL_SERIES}-rt/$(ls /patches-${KERNEL_SERIES}-rt) -C /patches-${KERNEL_SERIES}-rt; \
+           for line in `sed -e "s/#.*//g" /patches-${KERNEL_SERIES}-rt/patches/series`; do \
+               patch -p1 < /patches-${KERNEL_SERIES}-rt/patches/$line; \
+           done; \
+       fi; \
+       rm -rf /patches-${KERNEL_SERIES}-rt/patches/; \
+    fi
+
+# Apply local patches if present
 RUN set -e && \
     if [ -d /patches-${KERNEL_SERIES} ]; then \
         for patch in /patches-${KERNEL_SERIES}/*.patch; do \
@@ -80,17 +98,24 @@ RUN case $(uname -m) in \
         KERNEL_DEF_CONF=/linux/arch/arm64/configs/defconfig; \
         ;; \
     esac  && \
+    KERNEL_TYPE=$( (echo "${EXTRA}") | (grep -oE '\-rt|\-std') ); \
+    if [ -z "${KERNEL_TYPE}" ]; then KERNEL_TYPE="-std"; fi; \
+    echo "Configuring ${KERNEL_TYPE}"; \
     cp /config-${KERNEL_SERIES}-$(uname -m) ${KERNEL_DEF_CONF}; \
-    if [ -n "${EXTRA}" ]; then \
-        sed -i "s/CONFIG_LOCALVERSION=\"-linuxkit\"/CONFIG_LOCALVERSION=\"-linuxkit${EXTRA}\"/" ${KERNEL_DEF_CONF}; \
-        if [ "${EXTRA}" = "-dbg" ]; then \
-            sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' ${KERNEL_DEF_CONF}; \
-        fi && \
-        cat /config${EXTRA} >> ${KERNEL_DEF_CONF}; \
+    cat /config-${KERNEL_SERIES}${KERNEL_TYPE}-$(uname -m); \
+    if [ "${KERNEL_TYPE}" = "-rt" ]; then \
+       cat /config-${KERNEL_SERIES}-$(uname -m)-${KERNEL_TYPE} >> ${KERNEL_DEF_CONF}; \
+    fi; \
+    KERNEL_DEBUG=$( (echo "${EXTRA}") | (grep -oE '\-dbg') ); \
+    sed -i "s/CONFIG_LOCALVERSION=\"-linuxkit\"/CONFIG_LOCALVERSION=\"-linuxkit${KERNEL_TYPE}\"/" ${KERNEL_DEF_CONF}; \
+    if [ -n "${KERNEL_DEBUG}" ]; then \
+        sed -i "s/CONFIG_LOCALVERSION=\"-linuxkit${KERNEL_TYPE}\"/CONFIG_LOCALVERSION=\"-linuxkit${KERNEL_TYPE}${KERNEL_DEBUG}\"/" ${KERNEL_DEF_CONF}; \
+        sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' ${KERNEL_DEF_CONF}; \
+        cat /config${KERNEL_DEBUG} >> ${KERNEL_DEF_CONF}; \
     fi && \
     make defconfig && \
     make oldconfig && \
-    if [ -z "${EXTRA}" ]; then diff .config ${KERNEL_DEF_CONF}; fi
+    if [ -z "${KERNEL_DEBUG}" ] && [ "${KERNEL_TYPE}" = "-std" ]; then diff .config ${KERNEL_DEF_CONF}; fi
 
 RUN mkdir /out
 
@@ -105,7 +130,8 @@ RUN make -j "$(getconf _NPROCESSORS_ONLN)" KCFLAGS="-fno-pie" && \
         ;; \
     esac && \
     cp System.map /out && \
-    ([ "${EXTRA}" = "-dbg" ] && cp vmlinux /out || true)
+    ( KERNEL_DEBUG=$( (echo "${EXTRA}") | (grep -oE '\-dbg') ); \
+      if [ "${KERNEL_DEBUG}" = "-dbg" ]; then cp vmlinux /out || true; fi)
 
 # WireGuard
 RUN curl -sSL -o /wireguard.tar.xz "${WIREGUARD_URL}" && \

So what do you think now?

TiejunChina:
@rn Sorry, my new commit created new issue #2897. Do I need to close this?

TiejunChina:
OOPS! Please go #2897

Issue #2843, opened 2018-01-09
Add functionlity CPU micrcode updates


rn:
For baremetal deployments we should have a package which updates the CPU microcode either from the debian or the alpine base packages. It likely only requires dumping the microcode binary in the right place in the rootfs and the kernel will take care of the rest.

rn:
A bit more prodding around, there are three ways to upgrade the CPU microcode (see [here](https://www.kernel.org/doc/Documentation/x86/early-microcode.txt))
- via `/dev/cpu/microcode`
- via `/sys/devices/system/cpu/microcode/reload`
- via prepended initrd

The first two may happen too late in the boot process. In fact, I build an image with the FW files in `/lib/firmware/intel-ucode` and it did not get loaded on a packet.net type 1e machine, which was on revision `0x6a` with the lates firmware from Intel being `0xc2`.

The [`iucode-tool`](https://gitlab.com/iucode-tool/iucode-tool) has a mode to list details about ucode files (`iucode-tool -L .` in the `intel-ucode` directory downloaded from Intel). The `iucode-tool` also has an option to build a cpio archive with the latest microcode:

```
iucode_tool --normal-earlyfw --write-earlyfw=ucode.cpio ./intel-ucode
```

This `cpio` archive can be prepended to the initrd:
```
mv packet-initrd.img packet-initrd.img.orig
cat ucode.cpio packet-initrd.img.orig > packet-initrd.img
```

And when booted the first line of `dmesg` reads:
```
[    0.000000] microcode: microcode updated early to revision 0xc2, date = 2017-11-16
```

The AMD update works similarily

Bottom line is that the original suggestion of creating a `ucode` package and including it in the `YAML` file will not work.

My current thinking is to build the `cpio` archive as part of the kernel build and stash it in the kernel package (The intel `ucode.cpio` is 1.6MB, so negligible). The we add a `-baremetal` flag to `moby build`/`linuxkit build` which, when specified, will extract prepend the `cpio` archive in front of the `initrd` as shown above. Initially we only support it for building `kernel+initrd` but we should investigate how to add it to the other builds as well.

@justincormack @ijc any better ideas?

ijc:
Yeah, it sucks but the "prepend to initrd" way is the only safe/reliable way to do things.

The `-baremetal` UX is a bit nasty, but it's about as good as it'll get (doing it later e.g. with a `linuxkit push baremetal` fails due to not having the yml with the hash to use at that point).

[...]

PR #2858, opened 2018-01-12
Add the intel ucode cpio archive to the kernel package

rn:
- Add the `iucode_tool` to the alpien abse on x86_64. Rather tedious as we have to patch it...
- Update alpine base
- In the kernel build download and verify the latest Intel ucode. Then convert it to a cpio archive and stash it in the kernel package as `intel-ucode.cpio`
- while at it also fix up the README.

Note, I have not updated the packages on hub, but will pick up the changes with the next kernel update if this get's merged.

wip #2843

skinny-jeans


rn:
I'm currently trying to make the `tools/alpine/Dockerfile` a little nicer and need to push to this branch for it. So please don't merge yet. I'll update once done

rn:
OK, I think this is cleaner now. Already pushed the base images for alpine.

PR #2860, opened 2018-01-15
Move metadata to /run instead of /var

deitch:
Signed-off-by: Avi Deitcher <avi@deitcher.net>

**- What I did**
Moved metadata (and userdata) to `/run/config` instead of `/var/config`. Closes #2829 

Two questions to be addressed:

1. Is `/run/config` the right path? I chose it to match `/var/config`, but wonder if something else might need the space? Should it be `/run/metadata`? I prefer `/run/config`, but want to raise the question.
2. Is there anything in the docker-for-mac blueprint that expects it to be in `/var/config`? In the `docker-ce.yml` I updated the mount of data from `/var/config/docker:/var/config/docker` to `/run/config/docker:/var/config/docker`, so it should provide the data to docker where it expects, but not sure if anything else builds on the base system.

**- How I did it**

* Changed the `ConfigPath` var in [main.go](./pkg/metadata/main.go) . 
* No need to rewrite tests, as they pass in the target path as parameter
* Updated the doc
* Updated various consuming ymls

**- How to verify it**
Run it in an environment and see that metadata shows up where desired.

**- Description for the changelog**

Metadata in `/run/config` instead of `/var/config` to avoid possible mount conflicts, and fit with runtime data.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2864, opened 2018-01-15
Add /run to binds and use newer yml label format

deitch:

**- What I did**
Fix the labeling and binds for `linuxkit/metadata`

**- How I did it**
Moved the label to the new `build.yml` config (with thanks to @justincormack for the help) and added mounts for `/run`

**- How to verify it**
Run it. Now metadata shows up correctly.

Fix binds for metadata under `/run`.

### DIFF
--- a/pkg/metadata/Dockerfile
+++ b/pkg/metadata/Dockerfile
@@ -6,7 +6,7 @@ ENV GOPATH=/go PATH=$PATH:/go/bin
 COPY .  /go/src/metadata/
 RUN go-compile.sh /go/src/metadata
 
-RUN mkdir -p out/tmp out/var out/dev out/etc out/etc/ssl/certs
+RUN mkdir -p out/tmp out/var out/run out/dev out/etc out/etc/ssl/certs
 
 FROM scratch
 ENTRYPOINT []
@@ -15,4 +15,3 @@ WORKDIR /
 COPY --from=mirror /go/bin/metadata /usr/bin/metadata
 COPY --from=mirror /out/ /
 CMD ["/usr/bin/metadata"]
-LABEL org.mobyproject.config='{"binds": ["/dev:/dev", "/var:/var", "/sys:/sys", "/etc/resolv.conf:/etc/resolv.conf", "/etc/ssl/certs:/etc/ssl/certs"], "tmpfs": ["/tmp"], "readonly": true, "capabilities": ["CAP_SYS_ADMIN", "CAP_NET_ADMIN"]}'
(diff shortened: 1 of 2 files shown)


deitch:
Thanks @ijc, very happy to have your sign-off, will merge in. @justincormack and I discussed last night, so I know he is good here. 

> I think perhaps I picked up #2860 (previous metadata change) in #2863 without thinking about it?

Maybe? Some specific file?

I have had issues lately with Atom trying to reformat some files, so I have gotten into doing `git add -i` and hunk-by-hunk commits to avoid changing anything else. Pain.

ijc:
I specifically said "once the hashes are bumped." which you didn't do. Please raise a PR ASAP to bump the hashes of linuxkit/metadata.

> Maybe? Some specific file?

See the changes to `linuxkit/metadata` in c0de474e4f0848b727508fcb9dde194bc021c875, which will have (unexpectedly to me) pulled in your changes from #2860 (where you also didn't update the hashes). I think it is a good policy in general  to bump the hashes in the same PR as the changes rather than leaving it as a potentially surprising change for someone coming along later (did I break things with c0de474e4f0848b727508fcb9dde194bc021c875 by pulling in your change from `/var` to `/run`, perhaps I did, but that would have been unexpected given the changes I was actually making).

deitch:
> I specifically said "once the hashes are bumped." which you didn't do

Missed that. PR is open, need to rebase.

deitch:
> I think it is a good policy in general to bump the hashes in the same PR as the changes rather than leaving it as a potentially surprising change for someone coming along later 

Don't disagree, but as we know, it can be a multi-step pain, especially with the slow CI. Make a PR, update the hashes in the PR, takes an hour to go through CI, decide on a change, update, then do the hashes again... sometimes a _lot_ easier to separate them, even if it makes sense to do together.

ijc:
The time for `PR + CI + add hashes to PR + CI` is approximately the same as for `PR + CI + second PR for hashes + CI, it's 2x CI in either case.

I don't see _any_ reason to routinely rush these things and it just makes things more complex not simpler because now we need to be aware that a hash update in one PR might be picking up other changes which went into master without a hash update (which is exactly what bit me in #2860). If we have too many PRs hitting one place which makes for lots of rebasing and need to make a case by case exception that's fine, but this should not be the routine.

rn:
Also, if the hashes are not updated then CI won't pick up the new package so won't detect any issues. May not be the case here, but just getting into the habit of never merging without hashes is a good thing. For small changes like this you might as well add the hash bump with the initial PR then you only have one CI run. I always do that with kernel updates.

PR #2865, opened 2018-01-15
Vendor latest moby and update packet.net examples to include CPU ucode

### LINKED ISSUE #2843: Add functionlity CPU micrcode updates

rn:
- vendor in `moby` with the changes to handle CPU microcode
- Update `packet.net` to include the Intel CPU microcode for amd64 and not for arm64
- Update docs

resolves #2843

military-macaw-flying-front-on-sh-400ppi-copy

### DIFF
--- a/docs/platform-packet.md
+++ b/docs/platform-packet.md
@@ -34,6 +34,13 @@ an additional YAML for [arm64](../examples/packet.arm64.yml) servers
 which provide both access to the serial console and via ssh and
 configures bonding for network devices via metadata (if supported).
 
+For x86_64 builds for Intel servers we strongly recommend adding
+`ucode: intel-ucode.cpio` to the kernel section in the YAML. This
+updates the Intel CPU microcode to the latest by prepending it to the
+generated initrd file. The `ucode` entry is only recommended when
+booting on baremetal. It should be omitted (but is harmless) when
+building images to boot in VMs.
+
 **Note**: The update of the iPXE configuration sometimes may take some
 time and the first boot may fail. Hitting return on the console to
 retry the boot typically fixes this.
--- a/examples/packet.arm64.yml
+++ b/examples/packet.arm64.yml
@@ -1,3 +1,13 @@
+# This YAML snippet is to be used in conjunction with packet.yml to
+# build a arm64 image for packet.net. It adds a modprobe of the NIC
+# driver and overrides the kernel section to disable prepending the
+# Intel CPU microcode to the initrd. If writing a YAML specifically
+# for arm64 then the 'ucode' line in the kernel section can be left
+# out.
+kernel:
+  image: linuxkit/kernel:4.9.76
+  cmdline: "console=ttyAMA0"
+  ucode: ""
 onboot:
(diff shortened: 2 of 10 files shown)


rn:
I've tested the build on x86 and arm64

docs/platform-packet.md
```diff
@@ -34,6 +34,11 @@ an additional YAML for [arm64](../examples/packet.arm64.yml) servers
 which provide both access to the serial console and via ssh and
 configures bonding for network devices via metadata (if supported).
 
+For x86_64 builds for Intel servers we strongly recommend to add
```
ijc: nit: "...we strongly recommend adding..."

examples/packet.arm64.yml
```diff
@@ -1,3 +1,9 @@
+kernel:
+  image: linuxkit/kernel:4.9.76
```
ijc: I guess it is not possible to set `kernel.foo` without also setting `kernel.image`? This could be fixed up in the moby tool which already does conditional assignment of subfields for various things (runtime is one I remember adding myself).

examples/packet.arm64.yml
```diff
@@ -1,3 +1,9 @@
...
+  cmdline: "console=ttyAMA0"
+  # Note: ucode is only set to "" to override the amd64 setting
+  # Can be left out for arm64 builds.
```
ijc: This file could perhaps benefit from a little context, i.e. a comment at the top which says something like:
```
# This file can be built together with the x86_64 specific settings in packet.yml to produce an image suitable for arm64. See docs/platform-packet.md for detail.
```
This then gives a bit more context for the comment above. I think I'd also make the second line more explicit, something like "If writing a yml specifically for arm64 then this can be left out".

rn:
@ijc thanks for the comments on the YAML (and doc). I've pushed an update, hopefully making the purpose of `packet.arm64.yml` more explicit. Also rebased

As for the overrides for the kernel section. I guess I could have included a bit more custom overrides, but I already find the logic for merging these sections a bit to unintuitive already and rather prefer just overriding the whole entry explicitly.

docs/platform-packet.md
```diff
@@ -34,6 +34,11 @@ an additional YAML for [arm64](../examples/packet.arm64.yml) servers
 which provide both access to the serial console and via ssh and
 configures bonding for network devices via metadata (if supported).
 
+For x86_64 builds for Intel servers we strongly recommend adding
+`ucode: intel-ucode.cpio` to the kernel section in the YAML. This
+updates the Intel CPU microcode to the latest by prepending it to the
+generated initrd file.
```
ijc: Should the word "baremetal" appear in here somewhere?

rn: added more text to make it clear that it is recommended for baremetal and shouldn't be used for VM images (but it's harmless there)

PR #2867, opened 2018-01-16
Add support for driver (and CPU) firmware

### LINKED ISSUE #2714: Alternative for including firmware blobs
Linux 4.14.x has removed support for `make firmware_install` and, according to [this post](https://lkml.org/lkml/2017/9/15/343), the in tree firmware has not been updated since 2013. We need an alternative way of including firmware from [this repository](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git) instead.

rn:
This adds two new packages: `firmware` and `firmware-all`. The former only contains firmware blobs required by the driver modules included in a LinuxKit kernel (plus license files and the AMD CPU microde) while the latter contains a snapshot of the `linux-firmware` repository. `firmware` should be sufficient for use with LinuxKit kernels, while `firmware-all` may be used with custom or other distro's kernels.

Repositories are set up and packages pushed/signed. This should now be good for review.

resolves #2714

shark-smile

### DIFF
--- a/pkg/firmware/Dockerfile
+++ b/pkg/firmware/Dockerfile
@@ -1,0 +1,44 @@
+# Make modules from a recentish kernel available
+FROM linuxkit/kernel:4.14.12 AS kernel
+
+FROM linuxkit/alpine:34518265c6cb63ff02074549cc5b64bef40c336f AS build
+RUN apk add --no-cache git kmod
+
+# Clone the firmware repository
+# Make sure you also update the FW_COMMIT in ../firmware-all/Dockerfile
+ENV FW_URL=git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git
+ENV FW_COMMIT=65b1c68c63f974d72610db38dfae49861117cae2
+WORKDIR /
+RUN git clone ${FW_URL} && \
+    cd /linux-firmware && \
+    git checkout ${FW_COMMIT}
+
+# Copy files we always need/want: Licenses, docs and AMD CPU microcode
+WORKDIR /linux-firmware
+RUN set -e && \
+    mkdir -p /out/lib/firmware && \
+    cp README WHENCE /out/lib/firmware && \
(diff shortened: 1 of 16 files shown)


pkg/firmware/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Make modules from a recentish kernel available
+FROM linuxkit/kernel:4.14.12 AS kernel
+
+FROM linuxkit/alpine:34518265c6cb63ff02074549cc5b64bef40c336f AS build
+RUN apk add --no-cache git kmod
+
+# Clone the firmware repository
+ENV FW_URL=git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git
+ENV FW_COMMIT=65b1c68c63f974d72610db38dfae49861117cae2
```
ijc: You could do `FROM liunxkit/firmware-all AS firmware-all` then use `COPY --from` instead of cloning again, would make it easier to keep them in sync (or would it?).

ijc:
General shape and content looks good, I had one wondering about cloning the thing twice.

pkg/firmware/Dockerfile
rn: yeah, but then we'd need to update the hashes as well...either way it's two updates but if we were to re-use `liunxkit/firmware-all` we also have a dependency between packages which is more annoying to handle

ijc: ok. Might be worth a comment above both FW_COMMIT definitions saying "remember to update the other one"?

rn:
This should now be ready for review

eyz:
@rn do you have any docs on how to use this package? I will be happy to soak test it if so

rn:
It's only useful for baremetal. Just look at `examples/packet.yml`. Adding it to the `init` section dumps the firmware files into the rootfs and the driver load should then pick up the updated FW blobs assuming the driver uses the standard in-kernel FW loading mechanism

PR #2869, opened 2018-01-17
Update Moby tool

justincormack:
Add support for configuring the containerd namespace in the runtime config

magpie


justincormack:
Note the other diff in vendoring is a more recent `vndr` decision...

justincormack:
Needs `init` hash update, but after #2872 is merged for conflict reasons

justincormack:
rebased and hash update added

PR #2870, opened 2018-01-17
Update kernel to 4.14.14/4.9.77/4.4.112

rn:
The 4.4.14 has a number of important fixes/additions:
- New support for retpolines (enabled but requires newer gcc to take advantage of). This provides mitigation for Spectre style attacks.
- Various KPTI fixes including fixes for EFI booting
- More eBPF fixes around out-of-bounds and overflow of maps. These were used for variant 1 of CVE-2017-5753.
- Several KVM related to CVE-2017-5753, CVE-2017-5715, CVE-2017-17741.
- New sysfs interface listing vulnerabilities: /sys/devices/system/cpu/vulnerabilities

The 4.9.77 kernel also has seems to have most/all of the above  back-ported.

Also updated the security note to point out process and that folks should use the ucode option if on baremetal.

panda


rn:
x86_64 kernels already build and aarch64 currently building

PR #2871, opened 2018-01-17
Add option to enable nested-virt on GCP

### LINKED ISSUE #2695: Support nested virt for GCP push/run
We should enable nested virt for the `linuxkit` GCP support by default, maybe with an option to disable.

Instructions: https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances

rn:
[Nested Virtualization](https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances) on GCP is enabled by marking the image with the right license and ensuring that the instance is run on at least a Haswell class CPU. This PR adds the relevant bits. A update of the vendored Google cloud API was required too.

I've booted a VM and confirmed that the `vmx` CPU flag was set.

resolves #2695

jetstream


justincormack:
What a weird API!

errordeveloper:
I do not need this for nested guest VMs, right? Maybe docs should be a little more specific, it is not obvious.

rn:
@errordeveloper I don't understand the comment. If you create a VM on GCP and want to run other VMs inside that VM you need this. The nested VMs (ie VMs inside your GCP VM) don't need any special setting (unless maybe you want to run VMs inside there as well...then you have to use whatever KVM options there are for nested virt)

PR #2875, opened 2018-01-19
Increase size of ARP cache

justincormack:
We already had this increased above the defaults, but even larger values
seem to be recommended.

See https://blog.openai.com/scaling-kubernetes-to-2500-nodes/

"It’s common to tune this setting in HPC clusters, and is particularly
relevant in Kubernetes clusters since every pod has its own IP address
which consumes space in the ARP cache."

seal


rn:
approval conditional on hash updates, of course...

PR #2876, opened 2018-01-19
Re-arrange some files/directories

### LINKED ISSUE #2850: Add a contrib directory
We currently have some files which are scattered which would be better located in a `contrib` directory (ie somewhere where it is clear that they are best effort, provided as is). These don't really fit into the `./projects` space either. Examples:
- Docker for Mac blueprint (only remaining blueprint with LCOW and Kube moved elsewhere)
- Foreign kernels support scripts
- Vagrant file

I have a few things which I'd like to add soon as well which would fit better under contrib.

rn:
This is an attempt to clean up some files/directories

- Create a new `./contrib` directory for unsupported but useful stuff
- Move the foreign kernel scripts, logos,  and Vagrantfile to `./contrib`. I have a few more things which likely will end up in `./contrib` soon, such as terraform examples.
- Move Docker for Mac blueprint to `./examples`. We only had one blueprint left so it seems to make more sense to move it to other examples.
- Rename `vsudd.yml` example to `vsudd-containerd.yml`
- Rename and cleanup `./scripts/signing`. We weren't using the script, so remove it and then rename die directory to `certs`

resolves #2850

hippopotamus-underwater-sh-400ppi-copy


justincormack:
Another option is to move logos to a standalone repo? eg linuxkit/project ?

rn:
sure. I just wanted to clear up the top-level directory a bit. was thinking about moving it to `./docs` as well but own repo works too...

justincormack:
I think a different repo makes sense as they licensing is not necessarily the same and they apply to the project as a whole.

rn:
How about I drop the logo move for now and we move them to a different repo later. I guess the certs could go into the new repo as well...

justincormack:
SGTM

rn:
OK, dropped the log change for now

PR #2881, opened 2018-01-24
Update hyperkit go binding

### LINKED ISSUE #2868: Update hyperkit go bindings
The upstream hyperkit go bindings have changed quite a bit. We should revendor and adjust the code here.

rn:
The upstream hyperkit go bindings have changed a bit. revendor and adjust.

Ran the disk tests manually and they still pass. Also did some light manual testing with raw and qcow2.

resolves #2868

monorail-cat


src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -195,6 +195,7 @@ func runHyperKit(args []string) {
 	if err != nil {
 		log.Fatalln("Error creating hyperkit: ", err)
 	}
+	h.SetLogger(log.StandardLogger())
```
ijc: `h.SetLogger` turns into a call to `SetLogger` which is global in the `hyperkit` package, maybe it would be better to call that from a `func init()`?

The `h.SetLogger` wrapper itself is a bit odd since the state is global not per-`h`. Perhaps we should nuke it?

src/cmd/linuxkit/run_hyperkit.go
```diff
@@ -206,12 +207,15 @@ func runHyperKit(args []string) {
 			id = strconv.Itoa(i)
 		}
 		if d.Size != 0 && d.Path == "" {
-			d.Path = filepath.Join(*state, "disk"+id+".img")
+			d.Path = filepath.Join(*state, "disk"+id+".raw")
```
ijc: There seems to be some tests which use `.img` still. At least  `test/cases/040_packages/005_extend/000_ext4/test.sh`.

Maybe hyperkit should accept (without the warning) `.img` as a raw suffix in `GetDiskFormat`? I can sort of see why it might be nice to avoid that though since `.img` _is_ a bit ambiguous in terms of what it might contain.

rn:
Force pushed an update which moves the logger initialisation to and init function and also updates the hyperkit go bindings to include https://github.com/moby/hyperkit/pull/185 (which accepts `.img` as raw disk image extension

src/cmd/linuxkit/run_hyperkit.go
ijc: Did you mean to keep this now that you've allowed `.img`?

rn: yes, `.raw` is just that tiny bit more explicit

PR #2882, opened 2018-01-24
cmd/hyperv: Use Default Switch if present

### LINKED ISSUE #2373: hyper-v run does not like external switch names with spaces in them
The `linuxkit run hyperv` command requires an external VM switch to be configured and it will pick  the first one if not explicitly specified on the commandline. This code does not work if the switch name has spaces in it.

rn:
The Windows 10 Fall Creators Update added a new 'Default
Switch' (which is a NAT switch). Use it if present and the
user has not specified a switch.

Also, handle External switches with spaces in them.

resolves #2373

bird-smoking


rn:
I don't actually have a non Fall Creators Update win 10 machine so couldn't test switch with space. But did test the new Default Switch and it works

PR #2892, opened 2018-01-26
Do not print git commit if empty

### LINKED ISSUE #2891: don't print COMMIT on linuxkit tool if empty
If building from release tarball there is no git commit; just omit this in that case, we have the release number.

justincormack:
fix #2891

dog

### DIFF
--- a/src/cmd/linuxkit/main.go
+++ b/src/cmd/linuxkit/main.go
@@ -47,7 +47,9 @@ func (f *infoFormatter) Format(entry *log.Entry) ([]byte, error) {
 
 func printVersion() {
 	fmt.Printf("%s version %s\n", filepath.Base(os.Args[0]), version.Version)
-	fmt.Printf("commit: %s\n", version.GitCommit)
+	if version.GitCommit != "" {
+		fmt.Printf("commit: %s\n", version.GitCommit)
+	}
 	os.Exit(0)
 }
 


src/cmd/linuxkit/main.go
```diff
@@ -47,7 +47,9 @@ func (f *infoFormatter) Format(entry *log.Entry) ([]byte, error) {
 
 func printVersion() {
 	fmt.Printf("%s version %s\n", filepath.Base(os.Args[0]), version.Version)
-	fmt.Printf("commit: %s\n", version.GitCommit)
+	if version.GitCommit != "" {
```
ijc: This defaults to `"unknown"` rather than `""` in the code.

When building via `tmp_linuxkit_bin.tar` will the `-X` we pass in override that to blank? Perhaps we should set the in-code default (used with `go build` AIUI) to `""` for consistency?

justincormack: Ah, homebrew (and other package managers) is building with `make local-build`which currently doesnt pass anything, which does appear to set as blank.

justincormack: Which is odd, as I would expect the default to be used

justincormack: Oh, misread, no it is blank if not in a git repo

justincormack: I think maybe just leave it like this; building not from a git repo (ie the release tarball) using the correct makefile is the case I want to cover. `go get` etc really doesnt know...

PR #2897, opened 2018-01-31
Support Preempt-RT Linux kernel

TiejunChina:
In order to get such a preempt-rt Linux kerne, we grab -rt patch via
https://www.kernel.org/pub/linux/kernel/projects/rt/. So far we just enable it
over 4.14.x.

Signed-off-by: Tiejun Chen <tiejun.china@gmail.com>

---
v9:
* Fix push issue.
v8:
* Fix an issue of downloading source
v7:
* Make source check better with $(3)$(4).
* Fix one error of `make fetch`.
v6:
* Check $(4) as well in the case of downloading source.
v5:
* Just rebase
v4:
* Rebase
* New -rt patches version: 4.14.18-rt15
* Clean up some changes to Dockerfile
* Add new parameters in kernel/Makefile.
v3:
* Check in the actual -rt patches instead of the tar ball
* Separate $(3) as ${EXTRA} and ${DEBUG} to minimize this fallout to Dockerfile and Makefile 
v2:
* store -rt patches locally
* just provide two -rt config files like -dbg, config-4.14.x-x86_64-rt, config-4.14.x-aarch64-rt


TiejunChina:
@rn Thanks a lot.

My concerns here is, yes, now we just need to filter out '-dbg' from "-rt-dbg",  so EXTRA = "-rt" as expected.  But what if someone also want to expand EXTRA for another usage in the future? For example, EXTRA="-rt-dbg-AAA-BBB-CCC".  We have to first filter out "AAA", and then "BBB", and then "CCC" separately, right? It's still a special case virtually.

rn:
let's hope we don't need to support `-rt-dbg-AAA-BBB-CCC`...It might be cleaner to add a new flag for `debug` and keep the extra as is. Then it's clear that there is only one support variant (plus debug)

TiejunChina:
@rn Updated as v3.

Mostly I tried to separate $(3) as ${EXTRA} and ${DEBUG} to minimize this fallout to Dockerfile and Makefile.

TiejunChina:
@GordonTheTurtle Sorry, looks my git-review does not work as expected. Additionally, all -rt patches are picked from https://www.kernel.org/pub/linux/kernel/projects/rt/ so based on my previous experience, it's not necessary to add my SOB.

ijc:
@TiejunChina the check is that commits in this repo here have SOB in their commit message (which is required). It is not checking if you have added an SOB to the patch files since that is indeed not needed here (is that what you meant?) since the commit SOB would cover it.

Gordon is a bot BTW.

TiejunChina:
@ijc That's right. I didn't add my SOB to those -rt patch files :)

kernel/Dockerfile
```diff
@@ -28,12 +28,16 @@ RUN apk add \
     xz-dev \
     zlib-dev
 
+# Update patch utility
+RUN apk update && apk add --update patch
```
rn: Why is this needed?

kernel/Dockerfile
```diff
@@ -61,8 +65,17 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
     gpg2 --verify linux-${KERNEL_VERSION}.tar.sign linux-${KERNEL_VERSION}.tar && \
     cat linux-${KERNEL_VERSION}.tar | tar --absolute-names -x && mv /linux-${KERNEL_VERSION} /linux
 
-# Apply local patches if present
 WORKDIR /linux
+# Apply Preempt-RT kernel patches in the case of RT Linux Kernel
```
rn: This comment is now no longer valid as the mechanism is independent of `-rt`

kernel/Dockerfile
```diff
@@ -61,8 +65,17 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
+RUN set -e && \
+    if [ -d /patches-${KERNEL_SERIES}${EXTRA} ]; then \
+	echo "Building ${EXTRA} kernel"; \
```
rn: This is not "building" the kernel, it's patching it

kernel/Dockerfile
```diff
@@ -61,8 +65,17 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
+	for line in `sed -e "s/#.*//g" /patches-${KERNEL_SERIES}${EXTRA}/series`; do \
```
rn: why is this `sed` command needed. Can't the patches just be applied like below?

kernel/Dockerfile
```diff
@@ -61,8 +65,17 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
...
+	    patch -p1 < /patches-${KERNEL_SERIES}${EXTRA}/$line; \
+	done; \
+    fi
+
+# Apply local patches if present
```
rn: This comment is no longer correct. Also, wouldn't it make more sense to apply the common patches first and then apply the extra patches on top?

kernel/Dockerfile
```diff
@@ -82,15 +95,16 @@ RUN case $(uname -m) in \
     esac  && \
     cp /config-${KERNEL_SERIES}-$(uname -m) ${KERNEL_DEF_CONF}; \
     if [ -n "${EXTRA}" ]; then \
-        sed -i "s/CONFIG_LOCALVERSION=\"-linuxkit\"/CONFIG_LOCALVERSION=\"-linuxkit${EXTRA}\"/" ${KERNEL_DEF_CONF}; \
-        if [ "${EXTRA}" = "-dbg" ]; then \
-            sed -i 's/CONFIG_PANIC_ON_OOPS=y/# CONFIG_PANIC_ON_OOPS is not set/' ${KERNEL_DEF_CONF}; \
-        fi && \
-        cat /config${EXTRA} >> ${KERNEL_DEF_CONF}; \
+	cat /config-${KERNEL_SERIES}-$(uname -m)${EXTRA} >> ${KERNEL_DEF_CONF}; \
```
rn: can we make this conditional based on `config-${KERNEL_SERIES}-$(uname -m)${EXTRA}` existing?

kernel/Dockerfile
```diff
@@ -105,7 +119,7 @@ RUN make -j "$(getconf _NPROCESSORS_ONLN)" KCFLAGS="-fno-pie" && \
         ;; \
     esac && \
     cp System.map /out && \
-    ([ "${EXTRA}" = "-dbg" ] && cp vmlinux /out || true)
+    ([ "${DEBUG}" = "-dbg" ] && cp vmlinux /out || true)
```
rn: This should probably just check for it being set

kernel/Makefile
```diff
@@ -98,15 +106,17 @@ build_$(2)$(3): Dockerfile Makefile $(wildcard patches-$(2)/*) $(wildcard config
 		docker build \
 			--build-arg KERNEL_VERSION=$(1) \
 			--build-arg KERNEL_SERIES=$(2) \
-			--build-arg EXTRA=$(3) \
+			--build-arg EXTRA=`echo $(3) | sed -e 's/-dbg//'` \
```
rn: what is the reason for the `sed` gymnastics here? Wouldn't it be either to just another argument to the template?

rn:
Oh, and with https://github.com/linuxkit/linuxkit/pull/2900 merged this now needs a rebase. Actually, there will be another kernel update PR later today so might be worth waiting for that too

kernel/Dockerfile
TiejunChina: When I applied one of those -rt patches,  I cannot built that successfully.

net/core/dev.o: In function `net_rx_action':
/linux/net/core/dev.c:5658: undefined reference to `__raise_softirq_irqoff_ksoft'
net/core/dev.o:(.debug_addr+0x6e50): undefined reference to `__raise_softirq_irqoff_ksoft'

One -rt patch, patches-4.14.x-rt/net-provide-a-way-to-delegate-processing-a-softirq-t.patch, is trying to refine __raise_softirq_irqoff_ksoft in kernel/softirq.c. But there are two similar contexts isolated with #-rt #else #endif by one previous -rt patch. The original `patch` command just puts __raise_softirq_irqoff_ksoft into {#else #endif} so I have to update `patch`.

kernel/Dockerfile
TiejunChina: Sure. What about this? 
# Apply local specific patches if present

kernel/Dockerfile
TiejunChina: Here the 'series' file is extracted by patches-4.14.15-rt13.tar.gz. But it is originally constructed like,
###########################################################
# DELTA against a known Linus release
###########################################################

############################################################
# UPSTREAM changes queued
############################################################
rcu-Suppress-lockdep-false-positive-boost_mtx-compla.patch

############################################################
# UPSTREAM FIXES, patches pending
############################################################

############################################################
# Stuff broken upstream, patches submitted
############################################################

# soft hrtimer patches (v4)
0001-timers-Use-static-keys-for-migrate_enable-nohz_activ.patch
...
# ...
....

So we have to clean out this file. I think this should be good, otherwise we have to reconstruct our own 'series' file each time.

kernel/Dockerfile
TiejunChina: # Apply local common patches if present  Does this work for you?

kernel/Makefile
TiejunChina: $(3) is involved in many places in Makefile where it's used to do tag, build command, or something else.  If we introduce $(4) with '-rt', that means we have to modify a lot of places.

kernel/Dockerfile
TiejunChina: You mean if [ -f "/config-${KERNEL_SERIES}-$(uname -m)${EXTRA}" ] ? But even if ${EXTRA} is NULL, "/config-${KERNEL_SERIES}-$(uname -m)${EXTRA}" still exists. Maybe I can rename this file like "/config-${KERNEL_SERIES}${EXTRA}-$(uname -m)". What do you think?

kernel/Dockerfile
TiejunChina: Like this, ([ -n "${DEBUG}" ] && cp vmlinux /out || true) ?

kernel/Dockerfile
ijc: If patch is doing it wrong then that suggests that the patches need refreshing. For automated things like this using `-F 0` to turn off fuzzing and require patches to apply cleanly should be safer than relying on the quircks of a particular patch implementation to DTRT.

TiejunChina: I can do this but why can the updated patch utility apply that correctly? So I prefer to update patch utility, instead. This kind of thing could happen again in the future.

ijc: Applying with fuzz involves applying heuristics which can differ across implementations (or even versions of the same implementation). Using `-F 0` disables fuzzing and doesn't have that issue.

ijc: FWIW `dpgk-source` seems to use `-t -F0 -N -p1 -u` which might be good to apply throughout `kernel/*` not just here.

TiejunChina: @ijc Let make this clear :) The original patch is one simple version from busybox so it doesn't support -F 0. I can show you as follows:

...
Patching -rt kernel
patch: unrecognized option: F
BusyBox v1.27.2 (2017-12-12 10:41:50 GMT) multi-call binary.

Usage: patch [OPTIONS] [ORIGFILE [PATCHFILE]]

        -p,--strip N            Strip N leading components from file names
        -i,--input DIFF         Read DIFF instead of stdin
        -R,--reverse            Reverse patch
        -N,--forward            Ignore already applied patches
        -E,--remove-empty-files Remove output files if they become empty

rn: Can you then add `patch` to the list of utilities installed just above where you add it?

ijc: That's a shame but the fact remains that the solution to "patch with fuzz is applied incorrectly" is to fix the patch not to change the implementation of patch being used.

There's certainly an argument for switch to a fuller patch so that we can use `-F0` but that would not solve the issue you are having here with the fuzzy patch.

rn: Hmm, I thought I commented on this. Anyway, I agree with @ijc that the patches should apply without fuzz. For the other patches I maintain a linux-stable tree and rebase the branches on every kernel update, so it's guaranteed. I prefer if the `-rt` patches are the same as I don't really want to deal with patch conflicts.

kernel/Dockerfile
rn: ah, I hadn't realised this. This seems pretty specific to `-rt` could we make this section so that we look for `series` and if present do what you do and otherwise, just apply `*.patch`? That way we can re-use the mechanism for patching

kernel/Dockerfile
rn: but you check `-n "${EXTRA}"` above

kernel/Makefile
rn: I don't mind it being modified in a lot of places in the makefile. I find it inconsistent to have have them as separate variables  in the Dockerfile but as one string in the Makefile which then is split up...

kernel/Dockerfile
TiejunChina: @ijc @rn I double checked this. This is not sort of fuzzy patch. In that patch, this change is,
+++ b/kernel/softirq.c
@@ -686,6 +686,27 @@ void __raise_softirq_irqoff(unsigned int
 }

 /*
+ * Same as __raise_softirq_irqoff() but will process them in ksoftirqd
+ */
+void __raise_softirq_irqoff_ksoft(unsigned int nr)
...

686 is exact and correct line in kernel/softirq.c as expected. But that original patch always applies this in wrong place with wrong line and even wrong context. Is this one bug to the original patch?

ijc: It does indeed sound like a bug in `busybox patch` (possibly one worth reporting upstream).

In the end it's a bit moot since @rn's change in #2909 (d0d7738efac6127263827f60b65e22eb34264ee4) to use full patch with `-F0` and friends. If you rebase this whole hunk/discussion should go away.

TiejunChina: @ijc Good! Now it works for me with @rn's change #2909.

justincormack:
Needs a rebase.

TiejunChina:
@justincormack Sorry, I just rebased that this morning... Any change was mssied here?

justincormack:
@TiejunChina I guess it was the kernel update we just merged...

rn:
@TiejunChina yes, unfortunately, your patch is in the same area which changes on every kernel update, which happens at least once if not twice a week

TiejunChina:
@justincormack @rn NP. Let me do one simple build test after this rebase, and then I will resend out this pull request. Thank you.

kernel/Dockerfile
```diff
@@ -62,8 +63,18 @@ RUN curl -fsSLO ${KERNEL_SHA256_SUMS} && \
     gpg2 --verify linux-${KERNEL_VERSION}.tar.sign linux-${KERNEL_VERSION}.tar && \
     cat linux-${KERNEL_VERSION}.tar | tar --absolute-names -x && mv /linux-${KERNEL_VERSION} /linux
 
-# Apply local patches if present
 WORKDIR /linux
+# Apply local specific patches if present
+RUN set -e && \
+    if [ -n "${EXTRA}" ] && [ -d /patches-${KERNEL_SERIES}${EXTRA} ]; then \
+	echo "Patching ${EXTRA} kernel"; \
+	for patch in /patches-${KERNEL_SERIES}${EXTRA}/*.patch; do \
```
rn: So, the series file is no longer needed? I recall an earlier version used that. It's great if we don't need it!

TiejunChina: @rn This time I manually renamed these -rt patch files prefixed by the numer term orderly, so now this can work without that orginal series file. I think this can make our life easy.

rn: OK, thanks. Can this be automated? Or even better is there a kernel git branch somewhere with the RT patches? That way We can just to a `git format-patch` on an update. https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git doesn't seem to be structure that way...

If there isn't a suitable git tree, it might be good to add a script under `./scripts` to convert the patches as there are in the tarball into how we'd like them in the patches directory

TiejunChina:
@justincormack @rn Resent out this with that rebase.

kernel/Dockerfile
TiejunChina: @rn Yes, I use one simple script to rename those -rt patches. Maybe I can update this later with another pull request?

kernel/Makefile
```diff
@@ -79,12 +84,15 @@ sources:
...
 # to the global targets
-# Set $3 to "-dbg", to build debug kernels. This defines targets like
+# Set $3 to "-rt", to build Preempt-RT kernels. This defines targets like
+# build_4.14.x-rt and adds "-rt" to the hub image name.
+# Set $4 to "-dbg", to build debug kernels. This defines targets like
 # build_4.9.x-dbg and adds "-dbg" to the hub image name.
+# Set $3 to "-rt" and $4 to "-dbg" to build debug Preempt-RT kernel.
 define kernel
 
 ifeq ($(3),)
```
rn: I think this now also needs a check on `$(4)` as I get:
```
Makefile:219: warning: overriding commands for target `sources/linux-4.14.19.tar.xz'
Makefile:218: warning: ignoring old commands for target `sources/linux-4.14.19.tar.xz'
Makefile:223: warning: overriding commands for target `sources/linux-4.9.81.tar.xz'
Makefile:222: warning: ignoring old commands for target `sources/linux-4.9.81.tar.xz'
```

TiejunChina: @rn Just like this?
 ifeq ($(3),)
+ifeq ($(4),)
 sources/linux-$(1).tar.xz: Makefile | sources
        curl -fsSLo sources/linux-$(1).tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-$(1).tar.xz 
 KERNEL_VERSIONS+=$(1)
 endif
+endif

TiejunChina: @rn Resent new pull request out.

rn: yes, something like this. Thanks.

rn:
I think this PR is getting pretty close. I take a more detailed look tomorrow but did a skim through today and kicked of a kernel compile. I'm happy with the overall approach now. and yes the script for converting the patches can be done in a separate PR. Ideally we get this merged before the next kernel update to avoid another rebase

kernel/Makefile
ijc: `ifeq ($(3)$(4),)` might be more concise and avoid the nested if.

rn:
`make fetch` is broken...related to the last changes:
```
$ make fetch
mkdir -p sources
curl -fsSLo sources/linux-4.15.3.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.15.3.tar.xz
curl -fsSLo sources/linux-4.14.19.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.14.19.tar.xz
make: *** No rule to make target 'sources/linux-4.14.18.tar.xz', needed by 'fetch'. Stop.
```

kernel/Makefile
TiejunChina: @ijc Sure. It looks good.

TiejunChina:
@rn I missed this +$(eval $(call kernel,4.14.18,4.14.x,$(EXTRA),$(DEBUG))). I will resend this pull request.

TiejunChina:
@rn Done.

rn:
I now get:
```
make fetch
Makefile:220: warning: overriding recipe for target 'build_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'build_4.14.x'
Makefile:220: warning: overriding recipe for target 'forcebuild_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'forcebuild_4.14.x'
Makefile:220: warning: overriding recipe for target 'push_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'push_4.14.x'
Makefile:220: warning: overriding recipe for target 'forcepush_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'forcepush_4.14.x'
Makefile:220: warning: overriding recipe for target 'show-tag_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'show-tag_4.14.x'
Makefile:220: warning: overriding recipe for target 'build_perf_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'build_perf_4.14.x'
Makefile:220: warning: overriding recipe for target 'forcebuild_perf_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'forcebuild_perf_4.14.x'
Makefile:220: warning: overriding recipe for target 'push_perf_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'push_perf_4.14.x'
Makefile:220: warning: overriding recipe for target 'forcepush_perf_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'forcepush_perf_4.14.x'
Makefile:220: warning: overriding recipe for target 'build_zfs_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'build_zfs_4.14.x'
Makefile:220: warning: overriding recipe for target 'push_zfs_4.14.x'
Makefile:218: warning: ignoring old recipe for target 'push_zfs_4.14.x'
```

TiejunChina:
@rn I think we need to only check $(4) in the case of downloading source, and we can remove  $(eval $(call kernel,4.14.18,4.14.x,$(EXTRA),$(DEBUG))) as well. I tried `make fetch` and now looks good.
$ sudo make fetch                         
curl -fsSLo sources/linux-4.15.3.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.15.3.tar.xz 
curl -fsSLo sources/linux-4.14.19.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.14.19.tar.xz 
curl -fsSLo sources/linux-4.14.18.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.14.18.tar.xz 
curl -fsSLo sources/linux-4.9.81.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.9.81.tar.xz 
curl -fsSLo sources/linux-4.4.115.tar.xz https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.4.115.tar.xz

TiejunChina:
@rn Just repush this fix.

rn:
The push of debug kernels fails. I'm doing a `make ORG=rneugeba NOTRUST=1` and `4.15.3` and `4.14.19` get build and pushed fine but the next kernel (`4.14.19-dbg`) fails before it is build:

```
Successfully built db2583050829
Successfully tagged rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
docker pull rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || (docker push rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 && docker tag rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 rneugeba/kernel:4.14.19-amd64 && docker push rneugeba/kernel:4.14.19-amd64 && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4  && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19 )
Error response from daemon: manifest for rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 not found
The push refers to repository [docker.io/rneugeba/kernel]
3767a78f67c6: Pushed
4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64: digest: sha256:1d79ccf36c09a474a925145f0102077b217d6e396c4afc27df21569ea4e979f5 size: 529
The push refers to repository [docker.io/rneugeba/kernel]
3767a78f67c6: Layer already exists
4.14.19-amd64: digest: sha256:1d79ccf36c09a474a925145f0102077b217d6e396c4afc27df21569ea4e979f5 size: 529
time="2018-02-16T13:22:30Z" level=warning msg="Couldn't find or access image reference \"rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-arm64\". Skipping image."
Digest: sha256:1afa5b8580678f0846f191b7542147ed2d090b536521d54420a48e47ee5b353e 433
Not signing rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4
time="2018-02-16T13:22:35Z" level=warning msg="Couldn't find or access image reference \"rneugeba/kernel:4.14.19-arm64\". Skipping image."
Digest: sha256:1afa5b8580678f0846f191b7542147ed2d090b536521d54420a48e47ee5b353e 433
Not signing rneugeba/kernel:4.14.19
docker pull rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || DOCKER_CONTENT_TRUST=0 docker build -f Dockerfile.perf --build-arg IMAGE=rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 --no-cache --network=none  -t rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 .
4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64: Pulling from rneugeba/kernel-perf
95dabf9f01e9: Pull complete
Digest: sha256:b1b4323d737e0a9858039149395e76d6dcdae5a96d6d9765e66b7f6787e2c667
Status: Downloaded newer image for rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
docker pull rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || (docker push rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 && docker tag rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 rneugeba/kernel-perf:4.14.19-amd64 && docker push rneugeba/kernel-perf:4.14.19-amd64 && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4  && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel-perf:4.14.19 )
4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64: Pulling from rneugeba/kernel-perf
Digest: sha256:b1b4323d737e0a9858039149395e76d6dcdae5a96d6d9765e66b7f6787e2c667
Status: Image is up to date for rneugeba/kernel-perf:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
docker pull rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || docker build --build-arg KERNEL_VERSION=4.14.19 --build-arg KERNEL_SERIES=4.14.x --build-arg EXTRA= --build-arg DEBUG=-dbg --label org.opencontainers.image.source=https://github.com/linuxkit/linuxkit --label org.opencontainers.image.revision=1c1145b5ea859a07746a992616396be674d6f866 --no-cache -t rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 .
4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64: Pulling from rneugeba/kernel
Digest: sha256:1d79ccf36c09a474a925145f0102077b217d6e396c4afc27df21569ea4e979f5
Status: Image is up to date for rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
docker pull rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || (docker push rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 && docker tag rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 rneugeba/kernel:4.14.19-dbg-amd64 && docker push rneugeba/kernel:4.14.19-dbg-amd64 && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4  && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19-dbg )
Error response from daemon: manifest for rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 not found
The push refers to repository [docker.io/rneugeba/kernel]
tag does not exist: rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
Makefile:219: recipe for target 'push_4.14.x-dbg' failed
```

So here we see the last bits of the `4.14.19` kernel being pushed and the there is check of `4.14.19-dbg` has already been build etc and that fails

In particular:
```
docker pull rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || docker build --build-arg KERNEL_VERSION=4.14.19 --build-arg KERNEL_SERIES=4.14.x --build-arg EXTRA= --build-arg DEBUG=-dbg --label org.opencontainers.image.source=https://github.com/linuxkit/linuxkit --label org.opencontainers.image.revision=1c1145b5ea859a07746a992616396be674d6f866 --no-cache -t rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 .
4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64: Pulling from rneugeba/kernel
Digest: sha256:1d79ccf36c09a474a925145f0102077b217d6e396c4afc27df21569ea4e979f5
Status: Image is up to date for rneugeba/kernel:4.14.19-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
docker pull rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 || (docker push rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 && docker tag rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 rneugeba/kernel:4.14.19-dbg-amd64 && docker push rneugeba/kernel:4.14.19-dbg-amd64 && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4  && /home/rneugeba/src/linuxkit/linuxkit/scripts/push-manifest.sh rneugeba/kernel:4.14.19-dbg )
Error response from daemon: manifest for rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64 not found
The push refers to repository [docker.io/rneugeba/kernel]
tag does not exist: rneugeba/kernel:4.14.19-dbg-4171d4e406ebdd4dd3e476d19cb0605ee74bd2d4-amd64
Makefile:219: recipe for target 'push_4.14.x-dbg' failed
```
The first like pulls the non-debug version but tries to invoke the build for the debug version. So somewhere the debug stuff is missing

PR #2899, opened 2018-02-01
Allow choice of qemu binary and acceleration

rn:
There are more acceleration backend in the pipeline for qemu. Allow to select the qemu binary being used (makes it easier to select master builds) and the acceleration option.

For example:
```
linuxkit run qemu -qemu ~/src/repos/virt/qemu/x86_64-softmmu/qemu-system-x86_64 -accel hvf minimal
```
boots with a qemu master build on macOS with Hypervisor Framework Acceleration.

Note this also changes the `LINUXKIT_QEMU_KVM` env variable to `LINUXKIT_QEMU_ACCEL`

turtle

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -167,6 +167,7 @@ func runQemu(args []string) {
 
 	// Backend configuration
 	qemuContainerized := flags.Bool("containerized", false, "Run qemu in a container")
+	qemuCmd := flags.String("qemu", "", "Path to the qemu binary (otherwise look in $PATH)")
 
 	// Generate UUID, so that /sys/class/dmi/id/product_uuid is populated
 	vmUUID := uuid.New()
@@ -311,6 +312,7 @@ func runQemu(args []string) {
 		Memory:         *mem,
 		KVM:            *enableKVM,
 		Containerized:  *qemuContainerized,
+		QemuBinPath:    *qemuCmd,
 		PublishedPorts: publishFlags,
 		NetdevConfig:   netdevConfig,
 		UUID:           vmUUID,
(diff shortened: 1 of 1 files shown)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -160,13 +151,14 @@ func runQemu(args []string) {
...
 	// VM configuration
-	enableKVM := flags.Bool("kvm", haveKVM(), "Enable KVM acceleration")
+	accel := flags.String("accel", defaultAccel, "Choose acceleration mode")
 	arch := flags.String("arch", defaultArch, "Type of architecture to use, e.g. x86_64, aarch64")
 	cpus := flags.String("cpus", "1", "Number of CPUs")
 	mem := flags.String("mem", "1024", "Amount of memory in MB")
 
 	// Backend configuration
 	qemuContainerized := flags.Bool("containerized", false, "Run qemu in a container")
+	qemuCmd := flags.String("qemu", "", "Path to the qemu binary (if not in default location)")
```
ijc: "if not in $PATH" might be a more accurate way of putting it?

src/cmd/linuxkit/run_qemu.go
```diff
@@ -499,12 +492,12 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
...
 
-	if config.KVM {
-		qemuArgs = append(qemuArgs, "-enable-kvm")
+	if config.Accel != "" {
+		qemuArgs = append(qemuArgs, "-accel", config.Accel)
 		if config.Arch == "aarch64" {
 			qemuArgs = append(qemuArgs, "-machine", "virt,gic_version=host")
 		} else {
-			qemuArgs = append(qemuArgs, "-machine", "q35,accel=kvm:tcg")
+			qemuArgs = append(qemuArgs, "-machine", "q35")
```
ijc: IIRC `accel=kvm:tcg` meant "try kvm but fallback to tcg" or some such, so should the default for `config.Accel` be `kvm:tcg` rather than just `kvm`?

rn:
turns out `-accel` is relatively new, for example not in 2.5.0 which is shipping with Ubuntu 16.04. So need to rework this to use `-machine blah,accel=`

rn:
Pushed update:
- fixed the nits from @ijc 
- use the `accel` option for system

Also enabled `hvf` as default on macOS and fixed a bug where kvm was selected for non-native architectures, ie running x86 on arm64 and vice versa

Tested on macOS, Linux x86 and arm64 with both x86 and arm64 images

So this is ready to be reviewed

PR #2901, opened 2018-02-02
Make it possible to key the package tags off of top level tree hash

### LINKED ISSUE #2887: Make it possible to key the package tags off of top level tree hash
We are trying to use linuxkit pkg functionality on repositories containing one package per repo. This currently doesn't quite work because the [treeHash](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/pkglib/git.go#L86)  always expects a package to reside in a subfolder so it can find the tree hash corresponding to it.

I don't think there's any reason to restrict it to just the subfolder since it can as easily use the top level tree hash (the one referenced in the commit).

If this sounds like a good idea in general, I can provide a pretty simple patch

rvs:
Closes #2887

Signed-off-by: Roman Shaposhnik <rvs@zededa.com>

**- What I did**

Added a logic that determines whether the package is at the top level of the git repo. If it is we use the tree hash from the commit itself.

**- How I did it**

Modified treeHash function to take above into account

**- How to verify it**

I built the current packages in LK (to make sure there are no
regressions) plus I also built this:
    https://github.com/rvs/zen-dnsmasq

It is now possible to have linuxkit package definition at the very top of the git repository (one package per git repo model)

### DIFF
--- a/src/cmd/linuxkit/pkglib/git.go
+++ b/src/cmd/linuxkit/pkglib/git.go
@@ -84,7 +84,21 @@ func (g git) isWorkTree(pkg string) (bool, error) {
 }
 
 func (g git) treeHash(pkg, commit string) (string, error) {
-	out, err := g.commandStdout(os.Stderr, "ls-tree", "--full-tree", commit, "--", pkg)
+	// we have to check if pkg is at the top level of the git tree,
+	// if that's the case we need to use tree hash from the commit itself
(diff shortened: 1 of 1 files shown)


deitch:
This looks good to me. My testing shows that it gives the same hash (i.e. identical behaviour) when in a subdir, and the correct hash when in root.

Would be nice to get some testing on it, but we don't have any for `git.go` anyways, so not going to complain.

Wait for CI to pass then we can merge it in (unless another maintainer objects, of course).

PR #2908, opened 2018-02-07
Extend kernel config test

### LINKED ISSUE #2890: Improve/extend kernel config test
We should add tests for KPTI, retpoline, and various other options we added the past months.

cji:
This commit adds tests for KPTI, retpoline, bpf_jit_always_on, and
disabled scsi proc fs configs.

Closes #2890

Signed-off-by: Craig Ingram <cingram@heroku.com>

The test.yml's will need to be updated with a new build of the linuxkit/test-kernel-config image, which (I think) I need someone else to build/push first. 

**- What I did**
I reviewed the checks added to the kernel configs for the past few months, in particular focusing on the Meltdown/Spectre mitigation checks, and added them to the `check-kernel-config.sh` script. 

**- How I did it**
Followed the pattern for other checks, and then ran the tests with the latest kernels using a local version of the `test-kernel-config` image to verify everything passed.

**- How to verify it**
* I had to build a local version of the test-kernel-config image (docker build in the `test/pkg/kernel-config` directory)
* modified the test.yml's for each of the linuxkit.kernel.config tests to use my local image 
* went into each kernel config test case, built the test.yml and ran it to confirm the tests still passed.

Kernel configuration check tests for KPTI, Retpoline, BPF JIT Always On enabled, and SCSI proc FS disabled

screen shot 2018-02-07 at 4 47 46 pm
Olive my goldendoodle ❤️

### DIFF
--- a/test/pkg/kernel-config/check-kernel-config.sh
+++ b/test/pkg/kernel-config/check-kernel-config.sh
@@ -41,6 +41,11 @@ echo $UNZIPPED_CONFIG | grep -q CONFIG_PANIC_ON_OOPS=y || fail "CONFIG_PANIC_ON_
 echo $UNZIPPED_CONFIG | grep -q CONFIG_SYN_COOKIES=y || fail "CONFIG_SYN_COOKIES=y"
 echo $UNZIPPED_CONFIG | grep -q CONFIG_LEGACY_VSYSCALL_NONE=y || fail "CONFIG_LEGACY_VSYSCALL_NONE=y"
 echo $UNZIPPED_CONFIG | grep -q CONFIG_RANDOMIZE_BASE=y || fail "CONFIG_RANDOMIZE_BASE=y"
+echo $UNZIPPED_CONFIG | grep -q CONFIG_PAGE_TABLE_ISOLATION=y || fail "CONFIG_PAGE_TABLE_ISOLATION=y"
+echo $UNZIPPED_CONFIG | grep -q CONFIG_RETPOLINE=y || fail "CONFIG_RETPOLINE=y"
+echo $UNZIPPED_CONFIG | grep -q CONFIG_GENERIC_CPU_VULNERABILITIES=y || fail "CONFIG_GENERIC_CPU_VULNERABILITIES=y"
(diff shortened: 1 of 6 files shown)


test/pkg/kernel-config/check-kernel-config.sh
```diff
@@ -41,6 +41,11 @@ echo $UNZIPPED_CONFIG | grep -q CONFIG_PANIC_ON_OOPS=y || fail "CONFIG_PANIC_ON_
 echo $UNZIPPED_CONFIG | grep -q CONFIG_SYN_COOKIES=y || fail "CONFIG_SYN_COOKIES=y"
 echo $UNZIPPED_CONFIG | grep -q CONFIG_LEGACY_VSYSCALL_NONE=y || fail "CONFIG_LEGACY_VSYSCALL_NONE=y"
 echo $UNZIPPED_CONFIG | grep -q CONFIG_RANDOMIZE_BASE=y || fail "CONFIG_RANDOMIZE_BASE=y"
+echo $UNZIPPED_CONFIG | grep -q CONFIG_PAGE_TABLE_ISOLATION=y || fail "CONFIG_PAGE_TABLE_ISOLATION=y"
```
rn: I think these should be conditional on the kernel versions

cji:
@rn no problem, I can make those changes here.

Would you prefer the conditional checks continue to only look at major/minor versions or to include patch releases as well? For example, RETPOLINE is in major/minor version 4.15 but not in 4.14 until 4.14.14.

rn:
Actually, come to think of it a bit more, we should check this unconditionally as these have been backported to all major kernels and one shouldn't really run a kernel without these days.

Let me build/push/sign the updated package so that you can add the new hashes to this PR. It'll take an hour or so as my arm64 server is currently busy with some other stuff

rn:
OK, Images pushed. Can you, in a separate commit change the kernel-config hashes. The simplest way to do this is from the top-level:
```
./scripts/update-component-sha.sh linuxkit/test-kernel-config:80bb1cd57205e47b5976f7fa4fb036384d012a1d linuxkit/test-kernel-config:f658811da20f069f71a863ba9469b4a2ae8910e1
```

That should update the YAML file in the test to pick up the updated tests. Once pushed, CI should run with these new tests.

thanks

PR #2909, opened 2018-02-08
Update kernels to 4.15.2/4.14.18

### LINKED ISSUE #2906: Improve kernel patching
When applying patches we should use the proper `patch` utility and use the following argument `-t -F0 -N -p1 -u` to error on any issue and disable fuzzing. The main kernel patches should all apply cleanly as they are derived from a branch which gets rebased for every release.

See https://github.com/linuxkit/linuxkit/pull/2897#discussion_r166590726

rn:
These kernels include significant changes to spectre mitigation but may require a micro-code update which has not yet been released.

Also improve the application of kernel patches to disable fuzzing (resolves #2906)

x86 kernels already build, arm64 kernels still building

baby-penguin

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -21,6 +21,7 @@ RUN apk add \
     mpc1-dev \
     mpfr-dev \
     ncurses-dev \
+    patch \
     sed \
     squashfs-tools \
     tar \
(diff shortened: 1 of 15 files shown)


rn:
arm64 kernel build finished too

PR #2915, opened 2018-02-16
This commit adds the functionality to run the qemu container detached

scotty-c:
**- What I did**
This PR adds the functionality to allow the user to run the qemu container detached via a flag `-detached`  under `linuxkit run qemu`. I have added this flag as we are building tooling around Linuxkit and want to have the option for the container to not have a tty
**- How I did it**
I added the detached to the qemuconfig struct that gets passed to down to the docker args

**- How to verify it**
I have run all the tests included in the readme and have been able to run the container in the background successfully

This PR adds the functionality to allow the user to run the qemu container detached via a flag `-detached`  under `linuxkit run qemu`

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -366,6 +366,11 @@ func runQemuLocal(config QemuConfig) error {
 		}
 	}
 
+	// Detached mode is only supported in a container.
+	if config.Detached == true {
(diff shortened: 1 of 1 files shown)


rn:
Thanks. How does this interact with the `qemuCmd.Run()` later? Does that start the qemu container in the background and return?

scotty-c:
@rn that is correct. Also, I added a fail to the func ` runQemuLocal` so the user cant accidentally pass `-detached` when using the local `qemu` binary

PR #2923, opened 2018-02-21
Make 4.14.x the default and reduce number of kernels build

### LINKED ISSUE #2918: Switch default kernel to 4.14
Currently the default LinuxKit kernel is the latest from the 4.9 series. 4.14 has been out for a while and receives some critical security updates earlier (easier back port) and also has better Meltdown mitigation. We should switch the default kernel to 4.14

### LINKED ISSUE #2921: Per arch selection of kernel builds
We currently compile 8 kernels for both x86 and arm64. We should add a simple mechanism to select per arch which kernels to compile.

rn:
- Switch to 4.14.x as the default kernel. 4.14.x has been out for long enough and receives critical security fixes quicker (and better fixes too). (resolves #2918)
- Remove 4.9.x-dbg and 4.14.x-rt-dbg kernels from the default build. We already build too many kernels and interested users can easily build their own by specifying `DEBUG=-dbg` on the `make` command-line.
- Split the kernels we compile by architecture. We may soon get another arch and we don't really want to build all kernels for all architectures. (resolves #2921)
- Remove the 4.4.x kernel from the arm64 build.
- While at it also correct the kmod tests.

pengu-fly


kernel/Makefile
```diff
@@ -214,12 +214,22 @@ endef
 # Build Targets
 # Debug targets only for latest stable and LTS stable
 #
+ifeq ($(ARCH), x86_64)
```
ijc: Just to double check (because I trip over this sometimes), is the whitespace here significant, i.e. the comparison is with `" x86_64"` including the space and unless `$(ARCH)` has the same space it wont match? I may be mis-remembering this though. I guess you will find out when building...

Also, would it be worth keeping the common (i.e. default) kernel(s) common and only splitting out some of the optional extras? Might be clearer?

rn: I copy and pasted the conditional from further up in the Makefile. I've now tidied it up (it did work before and now as well).

I've also followed your suggested in keeping the common kernels common. I suspect once s390x support lands more of the kernels will be moved to arch specific clauses, but that's fine

rn:
The `wireguard` test failed with:
```
[+ 20m 35s] [STDERR  ] 2018-02-21T13:45:22.780139685Z:   Create OCI config for nginx:alpine
[+ 20m 36s] [STDERR  ] 2018-02-21T13:45:24.121964919Z: Pull image: docker.io/library/nginx:alpine@sha256:84a0df9d3a0a67a4d62e8fb77833681f68c282a7f092269c74dd9f7e376f95db
[+ 20m 37s] [STDERR  ] 2018-02-21T13:45:25.183944340Z: time="2018-02-21T13:45:25Z" level=fatal msg="Failed to extract root filesystem for nginx:alpine: Failed to docker create image docker.io/library/nginx:alpine@sha256:84a0df9d3a0a67a4d62e8fb77833681f68c282a7f092269c74dd9f7e376f95db: Error: No such image: docker.io/library/nginx:alpine@sha256:84a0df9d3a0a67a4d62e8fb77833681f68c282a7f092269c74dd9f7e376f95db"
[+ 20m 37s] [STDERR  ] 2018-02-21T13:45:25.198316349Z: + clean_up
[+ 20m 37s] [STDERR  ] 2018-02-21T13:45:25.198368791Z: + rm -f wireguard*
[+ 20m 37s] [FAIL    ] linuxkit.build.examples.wireguard 20.17s
```

I've hit rebuild, but looks like the docker pull nginx:alpine manifest list is broken (which reminds me that this was reported elsewhere today as well)

justincormack:
you could add a tag into the test...

rn:
I just did...I pushed an update

I think the `nginx:alpine` is currently being converted into a proper multi-arch. Previously only amd64 was available. That's why in one test, I pull `nginx:1.13.7`, which is debian based, as that was avail for all archs.

PR #2927, opened 2018-02-26
Add s390x support for Linuxkit binary and tools

alicefr:
This PR adds the basic support for s390 architecture on linuxkit
command.

**- What I did**
Add basic support for s390 systems:
- s390-specific flags to run qemu on s390 architecture

**- How I did it**
Update qemu flag in `src/cmd/linuxkit/`.

### DIFF
--- a/scripts/push-manifest.sh
+++ b/scripts/push-manifest.sh
@@ -45,7 +45,7 @@ esac
 # Push manifest list
 OUT=$(manifest-tool $MT_ARGS push from-args \
                     --ignore-missing \
-                    --platforms linux/amd64,linux/arm64 \
+                    --platforms linux/amd64,linux/arm64,linux/s390x \
                     --template "$TARGET"-ARCH \
                     --target "$TARGET")
 
--- a/src/cmd/linuxkit/pkglib/manifest_push_script.go
+++ b/src/cmd/linuxkit/pkglib/manifest_push_script.go
@@ -48,7 +48,7 @@ esac
 # Push manifest list
 OUT=$(manifest-tool $MT_ARGS push from-args \
                     --ignore-missing \
-                    --platforms linux/amd64,linux/arm64 \
+                    --platforms linux/amd64,linux/arm64,linux/s390x \
                     --template "$TARGET"-ARCH \
                     --target "$TARGET")
 
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -67,8 +67,12 @@ func init() {
 		defaultArch = "aarch64"
 	case "amd64":
 		defaultArch = "x86_64"
+	case "s390x":
+		defaultArch = "s390x"
 	}
 	switch {
+	case runtime.GOARCH == "s390x":
+		defaultAccel = "kvm"
 	case haveKVM():
 		defaultAccel = "kvm:tcg"
 	case runtime.GOOS == "darwin":
@@ -156,7 +160,7 @@ func runQemu(args []string) {
 
 	// VM configuration
 	accel := flags.String("accel", defaultAccel, "Choose acceleration mode. Use 'tcg' to disable it.")
-	arch := flags.String("arch", defaultArch, "Type of architecture to use, e.g. x86_64, aarch64")
+	arch := flags.String("arch", defaultArch, "Type of architecture to use, e.g. x86_64, aarch64, s390x")
 	cpus := flags.String("cpus", "1", "Number of CPUs")
 	mem := flags.String("mem", "1024", "Amount of memory in MB")
 
@@ -511,6 +515,8 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
 	// goArch is the GOARCH equivalent of config.Arch
 	var goArch string
 	switch config.Arch {
+	case "s390x":
+		goArch = "s390x"
 	case "aarch64":
 		goArch = "arm64"
 	case "x86_64":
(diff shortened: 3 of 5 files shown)


deitch:
I really like the idea of getting more platforms. Thank you indeed. 

How would we make this part of our normal build? Pretty much every package nowadays is built on aarch64 and amd64. We have a packet arm64 server which we use. Build on one arch, creates manifest too, build on the other. Would we need an s390 server available to all maintainers or contributors to do so? Or is the intention just to make lkt tool itself available, but then how do we build anything without the right arch specific packages?

rn:
@deitch we'll get access to a s390 system (in fact Justin and I already have). Need to work out signing etc as well

src/cmd/linuxkit/run_qemu.go
```diff
@@ -67,8 +67,12 @@ func init() {
 		defaultArch = "aarch64"
 	case "amd64":
 		defaultArch = "x86_64"
+	case "s390x":
+		defaultArch = "s390x"
 	}
 	switch {
+	case runtime.GOARCH == "s390x":
+		defaultAccel = "kvm"
```
ijc: Does the `hackKVM()` case not work for s390x? That case handles fallback to tcg if for some reason kvm is not usable by the current user.

rn:
@alicefr thanks. could split this up? Changes to the `./tools` directory require signing to be set up. Also, `tools/alpine/versions.s390x` is autogenerated and a maintainer will have to do that, so please exclude it from your commit.

I would suggest, to initially just do a PR for stuff under `./src`. Once merged (and once I have the signing set up) have a separate PR with `./tools` changes (without `versions.s390x`). I can then create a `linuxkit/alpine` package for s390. From there things should become easier

src/cmd/linuxkit/run_qemu.go
alicefr: on s390x, TCG does not provide good machine support, so virtualization with KVM is essentially the only way to execute images on s390x

alicefr:
Ok, I'm going to remove the files in `tools`

src/cmd/linuxkit/run_qemu.go
ijc: Is https://wiki.qemu.org/Documentation/Platforms/S390X out of date then, it seems to suggest tcg works for modern distros with current (2.12+) versions of QEMU.

It might not be the best available option, and we expect that, but when it is the only option then `linuxkit run` tool (which is a dev not production tool) usually allows it (e.g. for existing architectures it does).

rn: I too would prefer if I could run s390x images on another architecture, no matter how slow...
Or is this a matter of virtio/emulated HW support not being avail?

rn:
I think you also need to update `src/cmd/linuxkit/pkglib/manifest_push_script.go`

ijc:
> I think you also need to update `src/cmd/linuxkit/pkglib/manifest_push_script.go`

That is generated from `scripts/push-manifest.sh` so please update that and then `go generate github.com/linuxkit/linuxkit/src/cmd/linuxkit/pkglib/` should update the `pkglib` one (we could really do with a) a header on the generated file and b) build and CI integration ... oh well)

src/cmd/linuxkit/run_qemu.go
alicefr: We only work and test KVM, but it might happen to work by chance on other platforms. Therefore, in the PR for s390 only kvm is enabled.

rn: I would still prefer if we don't actively disable it for now, so that we can at least easily test. If we find that it doesn't work at all, we can disable it later

ijc: I agree and I think this just means removing the various s390x special cases currently added here around `accel` and `kvm` vs `kvm:tcg`. I'd say there's no obligation for @alicefr to test the result in a non-kvm environment though.

alicefr: ok, would it be reasonable to set as default accelerator `kvm`, and then overwrite it with `-accel`?

rn: @alicefr yes, i think that would be fine.

alicefr: Great!:)

scripts/push-manifest.sh
```diff
@@ -45,7 +45,7 @@ esac
 # Push manifest list
 OUT=$(manifest-tool $MT_ARGS push from-args \
                     --ignore-missing \
-                    --platforms linux/amd64,linux/arm64 \
+                    --platforms linux/amd64,linux/arm64,linux/s390 \
```
justincormack: shouldnt that be `s390x`?

src/cmd/linuxkit/pkglib/manifest_push_script.go
```diff
@@ -48,7 +48,7 @@ esac
 # Push manifest list
 OUT=$(manifest-tool $MT_ARGS push from-args \
                     --ignore-missing \
-                    --platforms linux/amd64,linux/arm64 \
+                    --platforms linux/amd64,linux/arm64,linux/s390 \
```
justincormack: and here?

justincormack:
Just a question about manifest list arch, but otherwise LGTM

src/cmd/linuxkit/pkglib/manifest_push_script.go
ijc: Is it really taste and not what are the `manifest-tool` requirements? (or perhaps it is hub, or the eventual Docker engine which will pull the image, or OCI requirements?)

https://github.com/estesp/manifest-tool seems to call it `s390x` for example.

alicefr: ok, you are right, it depends on the tag. Before multi-arch, the docker images for z platforms were tagged with `s390x`. Anyway, I already pushed with `s390x`

justincormack: I think notionally we support 31 bit s390 images running on s390x, although I don't think anyone has ever tried running one...

PR #2933, opened 2018-02-27
Configure guest with 9P mount and vsock port when using existing VPNKit instance

ekcasey:
Signed-off-by: Emily Casey <ecasey@pivotal.io>

**- What I did**
I added support for port forwarding when running hyperkit with an existing vpnkit instance

**- How I did it**
I set Sockets9P and VSockPorts on the hyperkit instance when starting with an existing vpnkit process

**- How to verify it**
Start vpnkit kit. `linuxkit run hyperkit` with the `-neworking vpnkit[,eth-socket-path[,port-socket-path]]` flag. In the guest use `vpnkit-expose-port` to configure a forwarded port. From the host send traffic to the configured host port and observe that it is forwarded to the guest port.

Supports port forwarding when running hyperkit with an existing vpnkit image

image

### DIFF
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -245,6 +245,10 @@ func runHyperKit(args []string) {
 			if len(netMode) > 2 {
 				vpnkitPortSocket = netMode[2]
 			}
+			// The guest will use this 9P mount to configure which ports to forward
+			h.Sockets9P = []hyperkit.Socket9P{{Path: vpnkitPortSocket, Tag: "port"}}
+			// VSOCK port 62373 is used to pass traffic from host->guest
+			h.VSockPorts = append(h.VSockPorts, 62373)
 		} else {
 			// Start new VPNKit instance
 			h.VPNKitSock = filepath.Join(*state, "vpnkit_eth.sock")


justincormack:
Is the use case here that you start vpnkit first then attach to it, but there is still only one VM per vpnkit? In which case this looks fine. I think this will be confusing if eg there is already a VM on the bridge as I dont think we have a way to configure which machine ports are forwarded from yet, although I may be wrong @djs55 is the person to ask!

djs55:
I think @justincormack is right -- the port forwarding isn't multi-VM capable at the moment. `vpnkit` has an argument `--vsock-path <path to hyperkit connect socket>`. When `vpnkit` accepts connections on the host it `connect`s to that specific `hyperkit`. The `vpnkit-expose-port` program reads the destination (container) IP and port, but if this is a container on a different VM then it won't work.

So I think this is ok, provided it's used carefully with one VM per `vpnkit`, with the `--vsock-path` set correctly.

I guess to make it multi-VM capable we could model it on swarm services -- run `vpnkit-expose-port` on every VM binding the same `AF_VSOCK` port. When first exposing a port we could figure out the VM the request is coming from and modify the tunnelling protocol to forward to the VM first and then the container?

justincormack:
@ekcasey going to merge this but still interested in how you are using it re conversation above...

ekcasey:
Yeah, we are only using one vm per `vpnkit` instance and we are carefult to start the `vpnkit` with the correct `--vsock-path`. We are using a running vpnkit rather than allowing linuxkit to start vpnkit so that we can pass proxy settings to vpnkit.

justincormack:
@ekcasey cool, thanks. We should fix multi host at some point anyway though...

PR #2934, opened 2018-02-27
Add s390x support for linuxkit docker images

alicefr:
Add support for s390 architecture for linuxkit/alpine and the
other docker images in tools and pkg.

**- What I did**
I updated `tools` to be able to build and push linuxkit/alpine docker image for s390 architecture

### DIFF
--- a/tools/alpine/packages.aarch64
+++ b/tools/alpine/packages.aarch64
@@ -1,1 +1,3 @@
+gummiboot
+libunwind-dev
 qemu-system-aarch64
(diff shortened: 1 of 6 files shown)


tools/alpine/packages.aarch64
```diff
@@ -1 +1,3 @@
+libunwind-dev
+gummiboot
```
justincormack: can we keep these sorted

rn:
OK, so the next step is for me to build/update the alpine base image, push and sign it on all platforms. I still need to set up a few more things on the s390 system...

PR #2939, opened 2018-02-28
Add s390 support for linuxkit kernel

alicefr:
Update building process to add s390 support.

The patch serial-forbid-8250-on-s390.patch has been added to disable
8250 serial for s390.

The patch is available upstream https://patchwork.kernel.org/patch/10106437/
but it is not backported.

**- What I did**
Add s390 support in `kernel`.

**- How I did it**
I add the kernel configs for s390 and update the make and dockerfiles.
The patch serial-forbid-8250-on-s390.patch adds a trivial fix to avoid the usage of 8250 serial on s390.


alicefr:
In order to generate the kernel configs see [PR](https://github.com/linuxkit/linuxkit/pull/2940).

rn:
Thanks. Overall this looks good to me, except you need to restrict the kernel build targets to 4.14/4.15 for s390x around here: https://github.com/linuxkit/linuxkit/blob/master/kernel/Makefile#L214 because otherwise there will be targets defined for 4.9 and 4.4 as well, which won't compile.

Also, this needs a rebase to incorporate the changes from https://github.com/linuxkit/linuxkit/pull/2943. Most importantly, this will then pick up a muti-arch `linuxkit/alpine` which has s390x support.

alicefr:
@rn thanks! Do you want to build for s390 also `-rt` and `-dbg` or should I also exclude those from the building?

rn:
definitely leave out `-rt`. `-dbg` depends on your use case and how fast the compile is. Less kernels in general is better as there are too many releases per week

PR #2940, opened 2018-02-28
Script to generate s390 kernel configs

alicefr:
The script generates the kernel configs for s390 based on the
kernel configs for x86 and arm.

The configs are generated inside a docker container from the image
linuxkit/kconfig.

**- What I did** 
Add script `create_kernel_conf.sh` to generate s390 kernel configs.


rn:
@alicefr thanks for the script. It is very useful for reference but I think now that we have the config files in the repository we don't need the script anymore so I'm closing this PR.

PR #2942, opened 2018-02-28
Add logging to /var/log for onboot containers

### LINKED ISSUE #1688: onboot logs not showing up anywhere
**Description**

Output from `onboot` containers do not appear to have persistent logs or to console.

When running a linuxkit image, the `onboot` containers should have their output both persistent and to console. There does appear to be a plan for more advanced logging based on https://github.com/linuxkit/linuxkit/blob/d422c70ac6da78da3d32237735d86eb9001df384/projects/logging/README.md but in the meantime, the output is not in `/var/log` (where `services` logs are) or on console.

Simplest way is to take the sample `linuxkit.yml` and add an `onboot` service that just echos:

```yml
  - name: swap
    image: "alpine:3.5"
    readonly: true
    command: ["/bin/echo","I AM HERE"]
```

Look for "I AM HERE" in the console, or even an indication that a particular container failed.

deitch:

Fixes #1688 

**- What I did**
Changed the stdout/stderr of `onboot` `runc` commands to go to files, and when the command is done, dump the output to `os.Stdout` and `os.Stderr`

**- How I did it**
Modified `pkg/init` so that it writes output to `/var/log/<type>.<name>.out.log` (or `.err.log`). When the process is complete, dump to `os.Stdout` and `os.Stderr`.

As discussed in the Slack channel, this would be *much* better using a `MultiWriter` that streams to stdout/stderr and files simultaneously. However, that causes `runc` to hang due to a known bug; issue linked in comments and [here](https://github.com/opencontainers/runc/issues/1721#issuecomment-366315563) .

Once that `runc` issues is fixed, and we can use that version, this can be cleaned up. In the meantime, at least we get output on console and in files.

**- How to verify it**
Run it.

onboot and onshutdown containers now write to logfiles in /var/log in addition to console.

Once we have approval of an additional maintainer or two on here, I will push out, update the hashes in samples, etc.

**Notes for feedback:** 

1. I don't love the logging style at the end of the change, with separate errors for `Seek` and `Copy`. I can leave it, but if anyone has a better idea...
2. The messages just get dumped to console, no formatting or anything. Do we want to improve it?

### DIFF
--- a/pkg/init/cmd/service/runc.go
+++ b/pkg/init/cmd/service/runc.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strconv"
 
@@ -13,9 +15,23 @@ import (
 
 const (
 	runcBinary = "/usr/bin/runc"
+	logDirBase = "/run/log/"
+	varLogDir  = "/var/log"
 )
 
-func runcInit(rootPath string) int {
+func dumpFile(w io.Writer, filePath string) error {
+	f, err := os.OpenFile(filePath, os.O_RDONLY, 0644)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = io.Copy(w, f)
+
+	return err
+}
+
+func runcInit(rootPath, serviceType string) int {
 	// do nothing if the path does not exist
 	if _, err := os.Stat(rootPath); err != nil && os.IsNotExist(err) {
 		return 0
@@ -39,6 +55,13 @@ func runcInit(rootPath string) int {
 
 	status := 0
 
+	logDir := path.Join(logDirBase, serviceType)
+	varLogLink := path.Join(varLogDir, serviceType)
+
+	if err := os.MkdirAll(logDir, 0755); err != nil {
+		log.Fatalf("Cannot create log directory %s: %v", logDir, err)
+	}
+
 	for _, file := range files {
 		name := file.Name()
 		path := filepath.Join(rootPath, name)
@@ -52,8 +75,32 @@ func runcInit(rootPath string) int {
 		}
 		pidfile := filepath.Join(tmpdir, name)
 		cmd := exec.Command(runcBinary, "create", "--bundle", path, "--pid-file", pidfile, name)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+
+		// stream stdout and stderr to respective files
+		// ideally we want to use io.MultiWriter here, sending one stream to stdout/stderr, another to the files
+		// however, this hangs if we do, due to a runc bug, see https://github.com/opencontainers/runc/issues/1721#issuecomment-366315563
+		// once that is fixed, this can be cleaned up
+		stdoutFile := filepath.Join(logDir, serviceType+"."+name+".out.log")
+		stdout, err := os.OpenFile(stdoutFile, os.O_WRONLY|os.O_CREATE, 0644)
+		if err != nil {
+			log.Printf("Error opening stdout log file: %v", err)
+			status = 1
+			continue
+		}
+		defer stdout.Close()
+
+		stderrFile := filepath.Join(logDir, serviceType+"."+name+".err.log")
+		stderr, err := os.OpenFile(stderrFile, os.O_WRONLY|os.O_CREATE, 0644)
+		if err != nil {
+			log.Printf("Error opening stderr log file: %v", err)
+			status = 1
+			continue
+		}
+		defer stderr.Close()
+
+		cmd.Stdout = stdout
+		cmd.Stderr = stderr
+
 		if err := cmd.Run(); err != nil {
 			log.Printf("Error creating %s: %v", name, err)
 			status = 1
@@ -91,8 +138,9 @@ func runcInit(rootPath string) int {
 		}()
 
 		cmd = exec.Command(runcBinary, "start", name)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		cmd.Stdout = stdout
+		cmd.Stderr = stderr
+
 		if err := cmd.Run(); err != nil {
 			log.Printf("Error starting %s: %v", name, err)
 			status = 1
@@ -103,9 +151,24 @@ func runcInit(rootPath string) int {
 
 		cleanup(path)
 		_ = os.Remove(pidfile)
+
+		// dump the log file outputs to os.Stdout/os.Stderr
+		if err = dumpFile(os.Stdout, stdoutFile); err != nil {
+			log.Printf("Error writing stdout of onboot service %s to console: %v", name, err)
+		}
+		if err = dumpFile(os.Stderr, stderrFile); err != nil {
+			log.Printf("Error writing stderr of onboot service %s to console: %v", name, err)
+		}
 	}
 
 	_ = os.RemoveAll(tmpdir)
+
+	// make sure the link exists from /var/log/onboot -> /run/log/onboot
+	if err := os.MkdirAll(varLogDir, 0755); err != nil {
+		log.Printf("Error creating secondary log directory %s: %v", varLogDir, err)
+	} else if err := os.Symlink(logDir, varLogLink); err != nil && !os.IsExist(err) {
+		log.Printf("Error creating symlink from %s to %s: %v", varLogLink, logDir, err)
+	}
 
 	return status
 }
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.9.81
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:d899eee3560a40aa3b4bdd67b3bb82703714b2b9
+  - linuxkit/init:b212cfeb4bb6330e0a7547d8010fe2e8489b677a
   - linuxkit/runc:7c39a68490a12cde830e1922f171c451fb08e731
   - linuxkit/containerd:37e397ebfc6bd5d8e18695b121166ffd0cbfd9f0
   - linuxkit/ca-certificates:v0.2
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.9.81
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:d899eee3560a40aa3b4bdd67b3bb82703714b2b9
+  - linuxkit/init:b212cfeb4bb6330e0a7547d8010fe2e8489b677a
   - linuxkit/runc:7c39a68490a12cde830e1922f171c451fb08e731
   - linuxkit/containerd:37e397ebfc6bd5d8e18695b121166ffd0cbfd9f0
   - linuxkit/ca-certificates:v0.2
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.9.81
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
-  - linuxkit/init:d899eee3560a40aa3b4bdd67b3bb82703714b2b9
+  - linuxkit/init:b212cfeb4bb6330e0a7547d8010fe2e8489b677a
   - linuxkit/runc:7c39a68490a12cde830e1922f171c451fb08e731
   - linuxkit/containerd:37e397ebfc6bd5d8e18695b121166ffd0cbfd9f0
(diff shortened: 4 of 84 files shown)


pkg/init/cmd/service/runc.go
```diff
@@ -39,6 +41,10 @@ func runcInit(rootPath string) int {
 
 	status := 0
 
+	if err := os.MkdirAll(logDir, 0755); err != nil {
+		log.Fatalf("Cannot create log directory %s: %v", logDir, err)
```
justincormack: generally we try not to fatally error in init as it is hard to debug. Maybe just print an error and continue on?

justincormack: Although I guess a lot will go wrong below...

deitch: Crap. Stupid copy-paste error from my tests. Sorry.

deitch: I take it back. I was thinking of a different part of the file. We use `log.Fatalf` in a number of places in this file, if we think it cannot work.

```go
	// get files; note ReadDir already sorts them
	files, err := ioutil.ReadDir(rootPath)
	if err != nil {
		log.Fatalf("Cannot read files in %s: %v", rootPath, err)
	}

	tmpdir, err := ioutil.TempDir("", filepath.Base(rootPath))
	if err != nil {
		log.Fatalf("Cannot create temporary directory: %v", err)
	}

	// need to set ourselves as a child subreaper or we cannot wait for runc as reparents to init
	if err := sys.SetSubreaper(1); err != nil {
		log.Fatalf("Cannot set as subreaper: %v", err)
	}

	status := 0
```

We fatal out if we cannot:

* read onboot config files
* create temp dir
* set subreaper


I think being unable to write logs is equivalent. Leave it?

justincormack: well if we cant read the config files we cant try to run them. If we can not insanely just fallback to stdout it might be nice, in case one of your onboots leaves `/var` in an unwriteable state temporarily. But not too bothered really, maybe it is better if you fix up the errors.

deitch: Yeah, I lean towards, "if your onboot screwed up `/var`, I would rather you fail cleanly than get into a half-there not-really-known state."

pkg/init/cmd/service/runc.go
```diff
@@ -52,8 +58,32 @@ func runcInit(rootPath string) int {
...
+		stdout, err := os.OpenFile(stdoutFile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+		if err != nil {
+			log.Printf("Error opening stdout log file: %v", err)
+			status = 1
+			continue
+		}
+		defer stdout.Close()
+
+		stderrFile := filepath.Join(logDir, serviceType+"."+name+".err.log")
+		stderr, err := os.OpenFile(stderrFile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
```
justincormack: I don't know that we should append logs, as we cant rotate them. Normally they will be on tmpfs anyway so it won't matter.

Actually wondering if we should write the onboot logs to `/tmp` anyway, as often `/var` will be overmounted so these won't be any use.

deitch: > I don't know that we should append logs, as we cant rotate them

Do you mean we should remove `|os.O_APPEND`?

> wondering if we should write the onboot logs to /tmp anyway, as often /var will be overmounted so these won't be any use

services logs are in `/var/log`; do we not want them in the same place? If there is a concern that a `mount` container in `onboot` might overwrite it, then be explicit about ordering?

justincormack: Yes but `onboot` is often there to mount a new `/var`, that is all done by the time services start, but for `onboot` part way through the containers it may get overmounted and then you lose all the earlier logs.

(you can just use `os.Create` I think if you dont want to append)

deitch: Ok. I’ll switch it to remove `os.Apoend`

Logs in `/tmp` seem strange. Is there another place?

justincormack:
Well I think the only writeable places are /var, /tmp and /var/tmp... we could use /var/log and assume that you will bind mount somewhere else if you care....

deitch:
Or we could put it in `/tmp/log` and at the end of the `init` process, move all log files to `/var/log`. That way they are not trounced by anything in `onboot`, but are in the right place in the running system?

ijc:
Is `/run` not available? Some distros use `/run/initramfs` for logs from the ramfs, so using `/run/onboot` or similar here wouldn't be without a sort-of precedent.

justincormack:
Oh, yes `/run` is available too.

deitch:
I definitely like `/run/onboot` better than `/tmp`. I still am concerned about having a split in logs with services in `/var/log` and onboot in `/run/onboot`. 

How about we do the following:

1. Put onboot logs in `/run/logs/onboot/`
2. At the end of init, symlink from `/var/logs/onboot` to `/run/logs/onboot`

I think that might be the best of both worlds?

justincormack:
Sgtm

deitch:
Something like this? I don't love the `if-else` tree there, but couldn't think of a cleaner way to do it.

pkg/init/cmd/service/runc.go
```diff
@@ -103,9 +136,35 @@ func runcInit(rootPath string) int {
...
 
+	// make sure the link exists from /var/log/onboot -> /run/log/onboot
+	if err := os.MkdirAll(varLogDir, 0755); err != nil {
+		log.Printf("Error creating secondary log directory %s: %v", varLogDir, err)
+	} else if _, err := os.Stat(varLogLink); os.IsNotExist(err) {
+		// create the link
+		if err := os.Symlink(logDir, varLogLink); err != nil {
+			log.Printf("Error creating symlink from %s to %s: %v", varLogLink, logDir, err)
+		}
+	} else {
```
justincormack: dont bother with the `else` case error message

deitch: OK

pkg/init/cmd/service/runc.go
```diff
@@ -103,9 +136,33 @@ func runcInit(rootPath string) int {
...
 	}
 
 	_ = os.RemoveAll(tmpdir)
 
+	// make sure the link exists from /var/log/onboot -> /run/log/onboot
+	if err := os.MkdirAll(varLogDir, 0755); err != nil {
+		log.Printf("Error creating secondary log directory %s: %v", varLogDir, err)
+	} else if _, err := os.Stat(varLogLink); os.IsNotExist(err) {
+		// create the link
+		if err := os.Symlink(logDir, varLogLink); err != nil {
```
ijc: Not a big deal, but the normal in Go would be to just call `os.Symlink` and squash an `os.IsExist` result that than doing the stat.

pkg/init/cmd/service/runc.go
```diff
@@ -103,9 +136,33 @@ func runcInit(rootPath string) int {
 
 		cleanup(path)
 		_ = os.Remove(pidfile)
+
+		// dump the log file outputs to os.Stdout/os.Stderr
+		if _, err = stdout.Seek(0, 0); err != nil {
```
ijc: At first I thought this `Seek` was actually doing something subtle/clever, like testing if `stdout` is a file or something else (a tty etc), but now I realise you just want to rewind it to dump it (the error message was a big clue and probably mitigates the need for a comment).

I've a vague unease about reusing the fd and in particular the requirement it adds to pass a readable stdout/err to the process (it could be `O_WRONLY` otherwise I think). I don't think that unease is particularly well founded given the process will have exited by this point so there's no buffering etc to worry about and I don't think there is actually anything inherently wrong with stdout/err being readable (although I've not consulted any of the specs).

ijc:
SGTM (one tiny comment/nit but not worth losing sleep over), but I'll leave the final say to @justincormack.

I don't think the `if-else` trees are that bad. You could avoid them with a helper function or two (especially since there is two instances of the same pattern in a couple of cases) but I don't think it's necessarily worth it in this case.

pkg/init/cmd/service/runc.go
deitch: Oh ok. That makes sense. I will do that.

> Not a big deal,

No, it isn't but clean code is better.

deitch: Actually, I don't see how that would help. I would need to handle 2 error conditions: `os.IsExist` and anything else. So my code would be:

```go
	if err := os.MkdirAll(varLogDir, 0755); err != nil {
		log.Printf("Error creating secondary log directory %s: %v", varLogDir, err)
	} else {
                _, err := os.Symlink(logDir, varLogLink)
                switch {
                   case os.IsExist(err):
                        // do nothing, maybe report it
			log.Printf("Cannot create symlink from %s to %s, as file already exists", varLogLink, logDir)
                    default:
			log.Printf("Error creating symlink from %s to %s: %v", varLogLink, logDir, err)
		}
	}
```

Is that any better?

pkg/init/cmd/service/runc.go
deitch: > doing something subtle/clever, like testing if stdout is a file or something else 

I think `runc` does something a little too clever like that, hence all of these headaches here. I am trying to avoid it. :-)

> I've a vague unease about reusing the fd and in particular the requirement it adds to pass a readable stdout/err to the process

I can see that. I cannot really close the fd since I have a `defer f.Close()` earlier, which needs to stay. Maybe it is cleaner to use a separate fd and dump that. I don't mind.

deitch:
> I don't think it's necessarily worth it in this case.

OK, then I will avoid the extra work. :-)

pkg/init/cmd/service/runc.go
deitch: No, I got it. It is cleaner what you suggested. 

```go
	// make sure the link exists from /var/log/onboot -> /run/log/onboot
	if err := os.MkdirAll(varLogDir, 0755); err != nil {
		log.Printf("Error creating secondary log directory %s: %v", varLogDir, err)
	} else if err := os.Symlink(logDir, varLogLink); err != nil && !os.IsExist(err) {
			log.Printf("Error creating symlink from %s to %s: %v", varLogLink, logDir, err)
		}
	}
```

deitch:
Updated. Please take a look @ijc and @justincormack

pkg/init/cmd/service/runc.go
```diff
@@ -52,8 +75,32 @@ func runcInit(rootPath string) int {
...
 		cmd := exec.Command(runcBinary, "create", "--bundle", path, "--pid-file", pidfile, name)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+
+		// stream stdout and stderr to respective files
+		// ideally we want to use io.MultiWriter here, sending one stream to stdout/stderr, another to the files
+		// however, this hangs if we do, due to a runc bug, see https://github.com/opencontainers/runc/issues/1721#issuecomment-366315563
+		// once that is fixed, this can be cleaned up
+		stdoutFile := filepath.Join(logDir, serviceType+"."+name+".out.log")
+		stdout, err := os.OpenFile(stdoutFile, os.O_RDWR|os.O_CREATE, 0644)
```
justincormack: thats the same as `os.Create(stdoutFile)` as the default umask is `022`.

pkg/init/cmd/service/runc.go
ijc: Yep, that's what I was envisaging (apart from the indentation)

deitch: Took a try or two, but we got it. :-)

If you are good with the rest @ijc , please hit the approve button. Since you and Justin both commented, want approval from both (he already did).

pkg/init/cmd/service/runc.go
ijc: With the change to use `dumpFile` I think this can be `O_WRONLY` now, can't it? (same for stderr)

pkg/init/cmd/service/runc.go
```diff
@@ -103,9 +151,26 @@ func runcInit(rootPath string) int {
 
 		cleanup(path)
 		_ = os.Remove(pidfile)
+
+		// dump the log file outputs to os.Stdout/os.Stderr
+		err = dumpFile(os.Stdout, stdoutFile)
+		if err != nil {
```
ijc: You can use the `if err := .... ; err != nil {` style here (and below)

pkg/init/cmd/service/runc.go
deitch: Yes, it can actually. Worth doing.

pkg/init/cmd/service/runc.go
deitch: Yeah, good point there too. Done.

deitch:
Updated per @ijc's cleanup requests

deitch:
Thanks for the approvals. Doing the following:

1. `lkt pkg push`
2. Update hashes in separate commit
3. Wait for CI
4. Merge

PR #2943, opened 2018-02-28
Update tools/alpine (wireguard tools and s390x)

rn:
This was actually more complicated than it ought to have been.
- The `wireguard-tools` from edge became incompatible with stable. So we now build our own `wireguard-tools` package.
- The latest Docker for Mac edge renamed the macOS key chain helper, so needed to adjust `push-manifest.sh`
- Some additional minor tweaks were required to work around the non-exisiting s390x files.

However, with that fixed, there now is a signed x86_64/arm64/s390x alpine base image on hub.

I've updated the kernel Dockerfiles already, so that it gets picked up on the next build.

endangered-cover


alicefr:
@rn Do you re-build `wireguard-tools` because of s390 package? I was able to compile the `linuxkit/alpine` without problems with `wireguard-tools` from the edge. I actually also tried the wireguard.yml and it was working on s390.

tools/alpine/push-manifest.sh
```diff
@@ -46,7 +46,7 @@ EOF
 # we need them for notary on all platforms.
 case $(uname -s) in
     Darwin)
-        CRED=$(echo "https://index.docker.io/v1/" | /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain.bin get)
+        CRED=$(echo "https://index.docker.io/v1/" | /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain get)
```
justincormack: is it worth supporting both names for a while?

rn:
@alicefr I rebuild it because the edge wireguard-util package now depends on a different musl version than is in stable. So this is not an issue specific to s390. Turns out building the `.apk` is pretty straight forward

tools/alpine/push-manifest.sh
rn: Probably. Pushed  an update

tools/alpine/push-manifest.sh
```diff
@@ -46,7 +46,12 @@ EOF
 # we need them for notary on all platforms.
 case $(uname -s) in
     Darwin)
-        CRED=$(echo "https://index.docker.io/v1/" | /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain.bin get)
+        # DfM recently (2018-03-27) change the mac key chain utility binary name. Handle both for now
+        if [ -f /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain.bin ]; then
+            CRED=$(echo "https://index.docker.io/v1/" | /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain.bin get)
+        else
+            CRED=$(echo "https://index.docker.io/v1/" | /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain get)
```
ijc: Same change is needed in `script/push-manifest.sh`, I think. After that you'd want to run `go generate github.com/linuxkit/linuxkit/src/cmd/linuxkit/pkglib` too.

Your comment suggests you think this change might be reverted? If that is likely then fine, if it is unlikely then a comment like `# Prior to 2018-03-27 D4M used a .bin suffix on the keychain utility binary name` would be more appropriate.

rn:
@ijc I update the comment (your verbiage was clearer) and also added the change to the `./scripts/push-manifest.sh` and re-generated the source file

PR #2951, opened 2018-03-07
Update moby tool

justincormack:
Also update its notary vendor as that was part of update.

This also adds support for qcow EFI images cc @garthy


justincormack:
Note this does suppress all the certificate expiry errors! Great for demos!

rn:
the build error is for `linuxkit build pkg/tss` and i can't build it locally either (same error).

Specifically:
```
$ git clone https://git.code.sf.net/p/trousers/trousers
Cloning into 'trousers'...
remote: Counting objects: 12960, done.
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
fatal: protocol error: bad pack header
```

deitch:
Something clearly changed. I will take a look.

deitch:
No, it looks like it is just a network issue, or possibly git on `git.code.sf.net`. The first time I ran it it failed on `trousers/trousers`. Second time it succeeded, but failed on `trousers/tpm-tools`. The third time it went all the way through.

We could cache it all, but I think it would be a waste since once it does run, we cache the output as `linuxkit/tss` image.

 I suggested we leave it.

rn:
Hit rebuild in case this is temporary.

The CI is trying to rebuild the `tss` package because the git content hash changed (due to readme update in https://github.com/linuxkit/linuxkit/pull/2948 without the updated package having been pushed)

rn:
OK, also build/pushed the `tss` package with the new git tree hash and this CI sumpbling block should be fixed

deitch:
> the git content hash changed due to README update 

I didn’t think of that. Would it matter? Unless we update the dependent hashes elsewhere in tests and examples, shouldn’t it just stick with the old ones?

deitch:
Hit wrong button by accident; reopened.

rn:
@deitch `make ci-pr` executes `$(MAKE) -C pkg build` which will call `linuxkit pkg build` for each package, which in turn will do a `pull` of the package with the current git tree hash and if that fails will attempt to build the package. That was what was failing. Has nothing to do with the examples. they still pick up the old package

deitch:
> which will call linuxkit pkg build for each package, which in turn will do a pull of the package with the current git tree hash and if that fails will attempt to build the package

Got it. So if there is a commit to just a doc (like the `README.md` case referenced above), and we didn't do a `lkt pkg push`, then no image will be on the hub, then it should attempt to build the package, which would succeed.

Ah, but the build sometimes fails because of the `curl` issue. Got it.

I will just push out the latest tree image, then, so we don't risk it failing or slow down CI.

I think we should update the docs for `packages.md` so it is clear that the latest package pushed out should match the tree to keep CI moving.

deitch:
> I will just push out the latest tree image, then, so we don't risk it failing or slow down CI

Ah, @rn , you did already. Thanks.

PR #2955, opened 2018-03-09
Tidy kernel build, tweak s390 config, and update kernels

rn:
- Make it more explicit which kernels are build for which arch. While this introduces some duplication, i prefer it over trying to figure it out from ifeq/ifneq combinations.
- Remove un-supported device drivers from the s390x kernel config file
- Update kernels to 4.14.24/4.9.86/4.4.120

angry-penguin


rn:
x86 and s390 builds are done. arm64 is still going

PR #2961, opened 2018-03-14
Get architecture name from Docker image

arm64b:
Since the `build` process will run within the docker daemon, so it's
safer to get the Arch name based on the images in which the build
process is running while not depend on the host kernel.

**- What I did**
Change the arch name fetch method
**- How I did it**
Using the `apk --print-arch` provided by the docker image instead of the host
**- How to verify it**
`make build_` kernel

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -30,7 +30,7 @@ RUN apk add \
     zlib-dev
 
 # libunwind-dev pkg is missed from arm64 now, below statement will be removed if the pkg is available.
-RUN [ $(uname -m) == x86_64 ] && apk add libunwind-dev || true
+RUN [ $(apk --print-arch) == x86_64 ] && apk add libunwind-dev || true
 
 ARG KERNEL_VERSION
 ARG KERNEL_SERIES
@@ -84,7 +84,7 @@ RUN set -e && \
     fi
 
 # Kernel config
-RUN case $(uname -m) in \
+RUN case $(apk --print-arch) in \
     x86_64) \
         KERNEL_DEF_CONF=/linux/arch/x86/configs/x86_64_defconfig; \
         ;; \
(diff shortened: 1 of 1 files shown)


deitch:
When would `uname -m` be insufficient?

arm64b:
>When would uname -m be insufficient?

Hello @deitch , the justification can refer to https://github.com/moby/moby/pull/36572  😄

deitch:
[This comment](https://github.com/moby/moby/pull/36572/files#r174105503) here? 

> we should avoid uname -m;

and

> I'd suggest something based on the arch of the image, not the kernel that happens to be running it. 

OK. Doesn't explain why though, when it fails?

OK, I will back off. :-)

arm64b:
Hmm, IMO, it's safer and reasonable to abstract the arch name from the image we're running, since all we do will be depended on that execution environment. I guess it will fail if a compatible docker image running on a host which arch is different with the one we get from image, suppose an amd64 system is running a 32-bit busybox image. firework will blown with wrong package installed 🉑

ijc:
32on64 shouldn't be a problem on a properly configured environment:
```
$ linux32 uname -m
i686
$ uname -m
x86_64
```

In any case just fixing the `Dockerfile` is likely to be insufficient, you would also need to fix up `linuxkit pkg build` (which uses `runtime.GOARCH` on the host) or the appropriate `Makefile` (e.g. kernel and alpine which run `uname -m` on the actual host) since we use that as part of the image name, and once you do that you may as well use `--env BUILD_ARCH=...` when invoking the container/build to make sure they agree.

So even if decide we think such biarch build compatibility is important (and right now LinuxKit has no pairs of biarch architectures in the supported set) this PR seems to be at best unnecessary and at worst a step in the wrong direction.
So I think this PR is taking us in the wrong direction,

ijc:
BTW my logic here is specific to LinuxKit, the logic may well be different for the case in https://github.com/moby/moby/pull/36572, whatever that is, so the above shouldn't be taken as relevant to the discussion there...

arm64b:
is there any difference between Moby and LinuxKit in terms of the build pipeline? I think we're not talking about the usage scenario, or the image build process will be shaped by the specific usage mode?

ijc:
LinuxKit essentially wraps the moby pipeline (today at least, in the future it could change e.g. to use buildkit directly). See the `linuxkit pkg` stuff (and the equivalent `Makefiles` for the stuff which isn't ported yet) and the push manifest scripting, both of which are relevant here.

deitch:
I still don't understand what failure scenario this is trying to address.

arm64b:
> LinuxKit essentially wraps the moby pipeline (today at least, in the future it could change e.g. to use buildkit directly). See the linuxkit pkg stuff (and the equivalent Makefiles for the stuff which isn't ported yet) and the push manifest scripting, both of which are relevant here.

Ah, as summary our LinuxKit will touch more stuffs to make it happens... 😵

arm64b:
I think Makefile is not relevant with this PR, if we look at the PR, it just abstract the arch from the image the build process running inside. A standalone building block for that for the assemble line,haha

ijc:
The existing `uname -m` is equally always correct, just changing this one place makes no sense (and is solving a wider problem which we don't currently have).

rn:
I agree with @ijc. This seems to address an issue we don't have and likely won't have (no intention of supporting say 64 and 32 bit for a given arch). Just changing in one place is not sufficient in any case and I much rather keep using `uname -m` until we need to adjust (which as said above looks unlikely).

Cross compile would be nice at some point, but that would also require far more wide ranging changes.

rn:
closing this PR based on the discussion above

Issue #2967, opened 2018-03-23
Building `pkg/rngd` with latest alpine base fails on x86


rn:
With `linuxkit/alpine:96ad1eb5ec262b4cd0eef574cdc0b225b502d9ee`, currently the latest alpine base image, building on x86 we get an error:
```
linuxkit pkg build .
Building "linuxkit/rngd:245d6e560556875e0b3e3e7ecfdd86f33e6be52f"
Pull (1 of 1): linuxkit/rngd:245d6e560556875e0b3e3e7ecfdd86f33e6be52f@sha256:f3837e2f7bbb4187b389d1b1b1b0621fe2da9977f485883e9b20ae71acf71b14
sha256:f3837e2f7bbb4187b389d1b1b1b0621fe2da9977f485883e9b20ae71acf71b14: Pulling from linuxkit/rngd
no matching manifest for linux/amd64 in the manifest list entries
No image pulled, continuing with build
Sending build context to Docker daemon  4.291MB
Step 1/13 : FROM linuxkit/alpine@sha256:187a64e2608be2c836486fe0772b4849720534e83021d0b06447538e0c5838b4 AS mirror
 ---> 7b9c7ac9c061
Step 2/13 : RUN apk add --no-cache go gcc musl-dev linux-headers
 ---> Using cache
 ---> ebea81500a00
Step 3/13 : ENV GOPATH=/go PATH=$PATH:/go/bin
 ---> Using cache
 ---> 3858b30d7e81
Step 4/13 : COPY cmd/rngd/ /go/src/rngd/
 ---> Using cache
 ---> 49003a184a05
Step 5/13 : RUN REQUIRE_CGO=1 go-compile.sh /go/src/rngd
 ---> Running in a5490145b803
gofmt...
govet...
golint...
ineffassign...
go test...
go build rngd: invalid flag in #cgo CFLAGS: -mrdrnd
The command '/bin/sh -c REQUIRE_CGO=1 go-compile.sh /go/src/rngd' returned a non-zero code: 1
exit status 1
```

the `gcc --target-help` lists `-mrdrnd` as an option
```
  -mrdrnd                     Support RDRND built-in functions and code generation.
  -mrdseed                    Support RDSEED instruction.
```

rn:
ah, https://github.com/golang/go/issues/23672 limits the flags one can use with cgo

Issue #2949, opened 2018-03-06
finish initial s390x support


rn:
This is mostly a reminder to self, what is left for s390x integration
- Update `./tools` and `./pkg` to new alpine base
- build/push/sign packages
- test build

/cc @alicefr

alicefr:
Cool! Let me know if you have any issues

rn:
@alicefr am I right in assuming that currently s390x only boots via kernel+initrd? ie none of the `tools/mkimage-*` bits are needed?

alicefr:
Yes, I tried only `kernel+initrd`. If we noticed that another format is needed, we can added later on I guess.

ijc:
@alicefr the question is really which ones s390x as a platform supports. Can you boot from an ISO or a filesystem with a suitable bootloader embedded? Which bootloader would that be in each case (e.g. on x86 we use syslinux or isolinux or grub depending on the context, on arm64 it is grub for all the ones which are supported)

rn:
My question was mostly about which of the tools we should enabled for s390x. I think for now I won't enable any.

alicefr:
@rn. I'm not really a bootloader expert I guess there are 2 points to solve. First, which bootloader s390 support and how the kernel is booted on s390 KVM. I will ask my colleagues. I am afraid that it may be not the standard way as on x86. Second, which tools are already present in alpine for s390. As far as I know, there are not *syslinux* and *isolinux* on alpine for s390. It might be a subject for an extension.

ijc:
`syslinux` and `isolinux` are very x86 specific and unlikely to be used on any other platform.

My expectation is that s390x will either use grub2 (if it is supported on s390x, I've no idea, but it supports a bunch of architectures) or some other custom/platform-specific bootloader (a bit of googling suggests that `qemu-system-s390x` is often used along with something called `zipl`).

alicefr:
They are both but we principally use *zipl*

alicefr:
We support only *zipl*. You may find also *grub2* from SUSE.

Issue #2968, opened 2018-03-24
linuxkit run on s390x fails with kernel crash/and or KVM error


rn:
with https://github.com/linuxkit/linuxkit/pull/2969 all packages are available for s390x, but booting the minimal image fails:

```
$ cd examples
$ linuxkit build minimal.yml
$ sudo linuxkit run minimal
sudo: unable to resolve host lxdkci07
ioctl(KVM_CREATE_VM) failed: 22 Invalid argument
Host kernel setup problem detected. Please verify:
- for kernels supporting the switch_amode or user_mode parameters, whether
  user space is running in primary address space
- for kernels supporting the vm.allocate_pgste sysctl, whether it is enabled
qemu-system-s390x: failed to initialize KVM: Invalid argument
FATA[0000] exit status 1
```

Running without KVM acceleration results in a kernel crash:
```
$ linuxkit run qemu -accel tcg minimal
sudo: unable to resolve host lxdkci07
Uncompressing Linux... Ok, booting the kernel.
[    0.649107] Linux version 4.14.29-linuxkit (root@b4881791b088) (gcc version 6.4.0 (Alpine 6.4.0)) #1 SMP Fri Mar 23 10:46:26 UTC 2018
[    0.650205] setup: Linux is running under KVM in 64-bit mode
[    0.651892] setup: The maximum memory size is 1024MB
[    0.653282] cpu: 1 configured CPUs, 0 standby CPUs
[    0.707333] Write protected kernel read-only data: 13652k
[    0.739138] Zone ranges:
[    0.740478]   DMA      [mem 0x0000000000000000-0x000000007fffffff]
[    0.740789]   Normal   empty
[    0.740808] Movable zone start for each node
[    0.740843] Early memory node ranges
[    0.740953]   node   0: [mem 0x0000000000000000-0x000000003fffffff]
[    0.741167] Initmem setup node 0 [mem 0x0000000000000000-0x000000003fffffff]
[    0.760711] random: fast init done
[    0.762445] percpu: Embedded 23 pages/cpu @000000003fdd2000 s57344 r8192 d28672 u94208
[    0.764144] Built 1 zonelists, mobility grouping on.  Total pages: 258048
[    0.764383] Kernel command line: console=tty0 console=ttyS0 console=ttyAMA0
[    0.771562] PID hash table entries: 4096 (order: 3, 32768 bytes)
[    0.771984] Dentry cache hash table entries: 131072 (order: 8, 1048576 bytes)
[...]

PR #2969, opened 2018-03-24
Complete initial s390x support

rn:
- Update `moby/tool`
- Update all package to the latest `linuxkit/alpine` (which has s390x support)
- Update firmware package to use a newer kernel package with s390x support
- Disable `cadvisor` for s390x. There was a compile error
- Fix x86 `rngd` compilation (need to add some CFLAGS to the whitelist) resolves https://github.com/linuxkit/linuxkit/issues/2967
- Update all YAMLs to latest version

resolves https://github.com/linuxkit/linuxkit/issues/2949

Note, there is another issue with s390x: https://github.com/linuxkit/linuxkit/issues/2968

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 
 PREFIX?=/usr/local/
 
-RTF_COMMIT=f2409214ca3b719567aa15bb7b363f24876a4d98
+RTF_COMMIT=171155c375706f2616f0b9c96afe2240e15d1de1
 RTF_CMD=github.com/linuxkit/rtf/cmd
 RTF_VERSION=0.0
 bin/rtf: tmp_rtf_bin.tar | bin
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,23 +2,23 @@ kernel:
   image: linuxkit/kernel:4.14.29
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:b212cfeb4bb6330e0a7547d8010fe2e8489b677a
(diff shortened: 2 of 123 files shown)


deitch:
@rn are we compiling on an actual `s390x` (or at least VM)? Or are we cross-compiling?

* Compiling: can we add the location and requirement to `docs/PACKAGES.md`? 
* Cross: how did you get it to work? I have had issues running s390x binaries using `binfmt_misc` with `qemu`?

deitch:
When I say:

> issues running s390x binaries

it could be related to the issue you referenced, to possibly not. I can give a simple replica, if needed.

rn:
I'm compiling on a s390x system (VM). 

What do you mean by adding location/requirements to `docs/packages.md`? It doesn't have details for these on other system either. Justin or I can give you access to a system once back from PTO. It's just a standard Debian VM with docker installed.

deitch:
> I'm compiling on a s390x system (VM).

Ah ok. I was having issues with running a basic hello world compiled for s390x under qemu, was hoping this meant you had figured it out and I could leverage your knowledge. The binary ran fine on a VM, but not on an `amd64` with qemu.

> What do you mean by adding location/requirements to docs/packages.md? It doesn't have details for these on other system either

No, you are right, but it does say that we support most packages on `amd64` and `arm64`. If we are going to start supporting `s390x` (which is great), we probably should say so there.

> Justin or I can give you access to a system once back from PTO

Thanks. We probably should have a file with a list of archs to hostnames for compilation.

rn:
ping @ijc @justincormack

PR #2982, opened 2018-04-03
Bump to containerd v1.0.3

ijc:
https://github.com/containerd/containerd/releases/tag/v1.0.3


justincormack:
Can we update for s390x too?

ijc:
Yes, I'd forgtten about the new arch. @rn is sorting me out with access to a machine.

ijc:
I built `tools/alpine`, `pkg/containerd`,  `pkg/init` and `test/pkg/containerd` for s390x and appended the versions.s390x commit here. I think that's all which needed building here.

PR #2983, opened 2018-04-03
Use NOTARY_AUTH instead of the expect gymnastics 

rn:
`notary` 0.6.0 allows passing credentials via the `NOTARY_AUTH` env variable. Previously we had jump through hoops via an expect script to feed the credentials to `notary`.

image


justincormack:
Much nicer!

PR #2993, opened 2018-04-06
pkg: build: do not infer a release if not pushing

ijc:
If you happen to be exactly on a tag then:

    $ linuxkit pkg build --dev pkg/init
    Building "ijc/init:dev"
    Cannot release "v0.3" if not pushing

Do not try and infer a release if not pushing so this is possible again.

The subsequent check for `bo.release != "" && !bo.push` remains since the
caller could have used `WithRelease` but not `WithPush`. Our CLI never does
this, but a hypothetical other user of the library might.

### DIFF
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -82,7 +82,7 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 		return fmt.Errorf("Unknown arch %q", arch)
 	}
 
-	if p.git != nil && bo.release == "" {
+	if p.git != nil && bo.push && bo.release == "" {
 		r, err := p.git.commitTag("HEAD")
 		if err != nil {
 			return err


ijc:
You aren't relying on `linuxkit pkg push` doing the tagging? If it's not being used it would certainly simplify things to remove it.

rn:
I am, I do a `linuxkit pkg push -release v0.3 <dir>`

ijc:
So we could safely drop the bit which detects that you are on a git tag and automatically does the release on push?

rn:
I think we can do that for now. What I do for relasese is that I use `linuxkit pkg forcepush -release` once I updated the alpine base. Only when all the package builds worked, do I update the tags in the YAML files (typically the last step) and only once CI passes and it got merged do I tag.

If we were to build packages in CI I could simplify the workflow and hope that the package build succeeds etc, but we are quite a bit off from being there across all three arches as well.

So, yes, we could remove that logic for now if it is too complex/brittle

ijc:
Lets merge this fix for now and remove if it causes any future issues.

PR #3007, opened 2018-04-18
containerd v1.1.0-rc.2

ijc:
https://github.com/containerd/containerd/releases/tag/v1.1.0-rc.2

Mostly fairly standard but the newer containerd requires go 1.10 which is available in alpine as a separate go1.10 package so add that to the mirror and use it for containerd + runc + init (runc just for consistency, init uses containerd libraries so it requires 1.10 too).

When we bump to alpine 3.8 we will need to remember to undo this part to go back to the default go package.

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -33,9 +33,9 @@ kernel:
   image: linuxkit/kernel:4.9.91
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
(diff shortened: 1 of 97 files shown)


ijc:
Looks like I need to rebase+rebuild already! I'm on it.

ijc:
Rebased. I missed out on the merge of #3006 but I don't think it is necessary to rebase again to get that (and rebasing this PR is fiddly + time consuming so I don't want to do so unnecessarily)

PR #3013, opened 2018-04-23
move pkg config from Dockerfile to build.yml

w9n:
**- What I did**

convert config json from Dockerfile label to build.yml

fixes #2999

This will break docker build compatibility

**- How I did it**

http://convertjson.com/json-to-yaml.htm

**- How to verify it**

CI

**- Description for the changelog**

move pkg config from Dockerfile to build.yml

**- A picture of a cute animal (not mandatory but encouraged)**

cleaning monkey

PR #3014, opened 2018-04-23
golint: fix hosting issue

w9n:
Signed-off-by: Robin Winkelewski <w9ncontact@gmail.com>

fixes #3013 CI error

**- What I did**

fix issue of https://golang.org/x/lint being not available

**- How to verify it**

CI

golint: fix hosting issue by directly git cloning from github

monkey

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,21 +3,19 @@ jobs:
   build:
     working_directory: /go/src/github.com/linuxkit/linuxkit
     docker:
-      - image: circleci/golang:1.9-stretch
+      - image: linuxkit/alpine:f3cd219615428b2bd943411723eb28875275fae7
     steps:
       - checkout
       - run: mkdir -p ./bin
+      - run:
+          name: Dependencies
+          command: apk add go make git musl-dev
       - run:
           name: Versions
           command: |
             set -x
             go version
             cat /etc/os-release
-      - run:
-          name: Dependencies
-          command: |
-            go get -u github.com/golang/lint/golint
-            go get -u github.com/gordonklaus/ineffassign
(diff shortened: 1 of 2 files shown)


w9n:
hm https://github.com/golang/lint/issues/397

rn:
I hit the same issue yesterday but it went away. I'm not sure if this is temporary flakiness of golang.org...

In any case a better fix might be to use the alpine base image in CI which has the tools we need already baked in

rn:
there is go in the `linuxkit/alpine`, that's what I meant by "alpine base image".

w9n:
hm i tried this https://circleci.com/gh/linuxkit/linuxkit/800 on linuxkit/alpine:f3cd219615428b2bd943411723eb28875275fae7

but of course you are right its in the packages list: https://github.com/linuxkit/linuxkit/blob/master/tools/alpine/packages

.circleci/config.yml
```diff
@@ -16,7 +16,9 @@ jobs:
       - run:
           name: Dependencies
           command: |
-            go get -u github.com/golang/lint/golint
+            mkdir -p $GOPATH/src/golang.org/x
+            git clone https://github.com/golang/lint.git $GOPATH/src/golang.org/x/lint
```
justincormack: you can do a depth=1 clone here

ijc:
The immediate issue has been fixed at the golang end: https://github.com/golang/lint/issues/397#issuecomment-383997188

Not sure if we still want this -- using the alpine image for the CI base image is an interesting(ish) idea.

w9n:
Giving up with the current linker error, it works locally for me :/

PR #3016, opened 2018-04-25
Containerd v1.1.0

ijc:
Containerd release: https://github.com/containerd/containerd/releases/tag/v1.1.0

All mechanical/scripted.

Rebuilding the alpine base image bumped wireguardtools from 0.0.20180413-r3 to 0.0.20180420-r3. I think that's fine -- no need to coordinate with kernel updates these days, but cc @zx2c4 @rn in case that's wrong.

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -33,9 +33,9 @@ kernel:
   image: linuxkit/kernel:4.9.91
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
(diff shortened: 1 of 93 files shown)


ijc:
I checked with the containerd repo and they didn't bump runc between rc.2 and this GA release.

zx2c4:
> Rebuilding the alpine base image bumped wireguardtools from 0.0.20180413-r3 to 0.0.20180420-r3. I think that's fine -- no need to coordinate with kernel updates these days, but cc @zx2c4 @rn in case that's wrong.

Yea, no need to worry any more about that kind of thing. We've come far!

PR #3017, opened 2018-04-25
update-component-sha.sh cleanups and fixes.

ijc:
See the individual commits for details (and smaller, more readable diffs). The none-cleanup changes are:

* support update of image with tagged release
* don't update a file in the vendor dir.

### DIFF
--- a/scripts/update-component-sha.sh
+++ b/scripts/update-component-sha.sh
@@ -56,7 +56,7 @@ case "$1" in
         old=$2
         new=$3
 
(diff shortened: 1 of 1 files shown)


ijc:
I only went so far as to check that `sed -E` was documented in `sed(1)` on mac. It's part of POSIX and our usage here is fairly simple and something I've used on Mac in the past, so I think it should be ok.

But I'll check for sure!

ijc:
Yes, a roundtrip for runc through v0.4 back to acba8886e4b1318457c711700f695a02fef9493d and another via a different hash back to the start works on mac too.

w9n:
LGTM, nice

PR #1675, opened 2017-04-20
set the CONFIG_IP_SCTP=y to enable SCTP support in the kernel

razaborg:
**- What I did**

I just recompiled the linuxkit moby kernel to get a SCTP support directly from the kernel.

I personally needed it in a docker for mac usecase to get SCTP support in my containers

**- How I did it**

I edited the kernel/kernel_config file to set the **CONFIG_IP_SCTP** option to **y** and remake the kernel.

For the usecase of docker for mac (I don't know if it's the only usecase of this kernel), I replaced the */Applications/Docker.app/Contents/Resources/moby/vmlinuz64* file (on the host) by the newly compiled bzImage kernel, and restarted docker daemon.

**- How to verify it**

```
host>docker run -it debian
container>cat /proc/net/protocols 
protocol  size sockets  memory press maxhdr  slab module     cl co di ac io in de sh ss gs se re sp bi br ha uh gp em
SCTPv6    1416      1       0   no       0   yes  kernel      y  y  y  y  y  y  y  y  y  y  y  y  n  y  y  y  y  y  y
SCTP      1256      0       0   no       0   yes  kernel      y  y  y  y  y  y  y  y  y  y  y  y  n  y  y  y  y  y  y
```

**- Description for the changelog**

In the *kernel/kernel_config* file :
```
before: # CONFIG_IP_SCTP is not set
after: CONFIG_IP_SCTP=y
```

PR #3025, opened 2018-05-01
Enable SCTP protocol as a default kernel config

acetcom:
I'd just like to try to this pull request again. -- @razaborg had initiated the same issue on (#1675)
I believe that it is really helpful to most telco people even though the changes are very simple.

**- What I did**

I just recompiled the linuxkit moby kernel to get a SCTP support directly from the kernel.
I personally needed it in a docker for mac usecase to get SCTP support in my containers

**- How I did it**
I edited the kernel/kernel_config file to set the CONFIG_IP_SCTP option to y and remake the kernel.

**- How to verify it**

```
host>docker run -it debian
container>cat /proc/net/protocols 
protocol  size sockets  memory press maxhdr  slab module     cl co di ac io in de sh ss gs se re sp bi br ha uh gp em
SCTPv6    1416      1       0   no       0   yes  kernel      y  y  y  y  y  y  y  y  y  y  y  y  n  y  y  y  y  y  y
SCTP      1256      0       0   no       0   yes  kernel      y  y  y  y  y  y  y  y  y  y  y  y  n  y  y  y  y  y  y
```

In the kernel/kernel_config file :
```
before: # CONFIG_IP_SCTP is not set
after: CONFIG_IP_SCTP=y
```


justincormack:
It may make more sense to have it as a module I think rather than built in (it should autoload).

acetcom:
Basically, I understand your thought. ^^;

However, please check the SCTP expose at the recent [moby issue #9689](https://github.com/moby/moby/issues/9689).
If SCTP is not enabled in hyperkit VM, many people will have difficulty in Docker For Mac.

Thanks for quick response.

justincormack:
Enabling it as a module rather than built in should not affect this, it should autoload I believe if you try to use it.

acetcom:
Ah! I have misunderstanding. Do you want to change from CONFIG_IP_SCTP=y to CONFIG_IP_SCTP=m?

If so, I'll change the pull request with git commit --amend

Thanks!

acetcom:
I've changed the CONFIG_IP_SCTP=m for kernel module.
Let me know if this modification has wrong.

Thanks a lot!

justincormack:
LGTM.

acetcom:
I've installed D4M with the following version.

In About Menu
- Version 18.05.0-ce-mac67 (25042)
- Channel: edge
- 0c7f1aa2fa

I've checked with a command `cat /proc/net/protocols`.
Unfortunately, SCTP does not work properly in the container.

After doing like the following
```
docker run --rm -it --privileged --pid = host debian nsenter -t 1 -m -u -n -i sh
ls /lib/modules/4.9.93-linuxkit-aufs/kernel/net/
```
But, there is no sctp kernel module.

I've checked [Release Note for Edge Version of D4M](https://docs.docker.com/docker-for-mac/edge-release-notes/). It is said that linuxkit v0.4.0 is applied.

Can you help me with how to solve this problem?

Thanks in advance.

justincormack:
Can you open an issue on https://github.com/docker/for-mac/issues as the issue is with Docker for Mac.

Note that d4mac has its own kernel, adding aufs support for backwards compatibility, so although it uses LinuxKit 0.4 it does not necessarily have the same kernel config.

PR #3030, opened 2018-05-04
Reboot should reboot instead of powering off.

kmjohansen:
When busybox's reboot processing occurs in init, it runs all SHUTDOWN
actions that are defined in inittab.  Once those are complete, it will
trigger either a halt, poweroff, or reboot, depending upon what signal
is received.  The mechanism that's used to shell out through inittab
does not allow us to pass through exactly which invocation was
requested.

Due to the way that rc.shutdown works, it invokes the poweroff action
for any and all SHUTDOWN callbacks, whether they're a reboot, poweroff,
or halt.  Instead of handling the reboot(2) syscall in rc.shutdown,
return after killing and unmounting and let busybox's init process
decide which reboot(2) action to use.

Signed-off-by: Krister Johansen <krister.johansen@oracle.com>

**- What I did**

Since the `::shutdown` action in inittab can be invoked for multiple different types of shutdowns, use a dummy argument to `rc.shutdown` that does not trigger a reboot or poweroff.  Let init decide once the unmount and kill are complete.

**- How I did it**

Modified how `::shutdown` is calling `rc.shutdown` in /etc/inittab.

**- How to verify it**

I verified this by invoking `reboot`, `poweroff`, and `halt` and validating that each triggered the desired action.

Ensure that reboot triggers a reboot instead of a poweroff.

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -33,7 +33,7 @@ kernel:
   image: linuxkit/kernel:4.9.91
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
-  - linuxkit/init:a14b7ef6115e6d56fe5bf7c40517b9e190dd4dfb
+  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
 services:
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.48
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:a14b7ef6115e6d56fe5bf7c40517b9e190dd4dfb
+  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
   - linuxkit/ca-certificates:v0.4
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.48
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:a14b7ef6115e6d56fe5bf7c40517b9e190dd4dfb
+  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
   - linuxkit/ca-certificates:v0.4
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.48
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
-  - linuxkit/init:a14b7ef6115e6d56fe5bf7c40517b9e190dd4dfb
+  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
   - linuxkit/ca-certificates:v0.4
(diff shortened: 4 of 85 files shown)


kmjohansen:
An earlier version of this PR attempted to remove the `unix.Reboot()` call from rc.shutdown, however, it's still needed to handle the ctl-alt-del case.  This updated version simply uses an argument to `rc.shutdown` that won't trigger any `reboot(2)` actions from `rc.shutdown` and instead defers the decision to `init`.

kmjohansen:
Yes, thanks for suggesting that.  I'm sorry I didn't leave a comment in the first go round.  I've updated the review with something that hopefully makes sense.

justincormack:
CI failure unrelated (out of space). Tested this and seems to work fine.

kmjohansen:
The updated change is just a rebase against master.  This commit has been approved, however, I do not have permissions to merge it myself.  Keeping this fresh for now.

ijc:
(I forgot to hit "Comment" on this an hour or two ago and have since seen Justin's comment about the same further up, but anyway)
CI said:
```
[+ 74m 38s] Error processing tar file(exit status 1): write /usr/lib/go/pkg/tool/linux_amd64/test2json: no space left on device
[+ 74m 38s] exit status 1
[+ 74m 38s] Makefile:13: recipe for target 'build' failed
[+ 74m 38s] make[1]: Leaving directory '/tmp/build/pkg'
[+ 74m 38s] make[1]: *** [build] Error 1
[+ 74m 38s] Makefile:118: recipe for target 'ci-pr' failed
[+ 74m 38s] make: *** [ci-pr] Error 2[2018-05-23 19:12.00] Fetching results
[2018-05-23 19:12.00] Running "scp" "-r" "-i" "/run/secrets/builder-ssh" "-o" "StrictHostKeyChecking=no" "root@35.233.82.208:/tmp/build/test/_results" "."
```
I've kicked it, hopefully we don't need to track down someone to free up some diskspace for us...

justincormack:
I don't understand what is running out of space, and @talex5 couldn't find anything that appeared to be either, its very strange...

ijc:
Didn't we see something like this once before, ages ago. It was something like pkg had changed in such a way that `pkg build` was actually building some things which had huge disk requirements to build (which would be cleaned up on failure, hiding it from post-mortem investigations). If the images are built and pushed to hub then the `pkg build` short circuits into a `pull` and you avoid the issue.

The logs here does seem to run out of space while doing a package build, I haven't spotted which package it is yet though.

justincormack:
Its int `init` package (the one this PR changes). Other PRs are passing...

rn:
It's a little odd that it fails on the `init` build. Space requirements aren't/shouldn't be that high. In any case, this can be "fixed" by building/pushing/signing the init package so that the CI is not doing it. I can do that after the kernel updates I'm currently running

ijc:
If someone builds/pushes init then I bet it will pass.

It seems to be doing a `make -J«something»` over `pkg/*` I wonder if that is provoking it to use more disk space then is available for `/var/lib/docker/`, `pkg/init` seems pretty simple so surprising if it blew the disk space all by itself (e.g. the apk add is just adding 117M).

ijc:
It's also pulling everything in `pkg/*` it isn't building, which will be taking space in `/var/lib/docker`, quite why building a single package then pushes it over the edge I'm unsure. Maybe it's right on the edge most of the time?

rn:
@kmjohansen I've build `linuxkit/init` for all arches. Could you in a separate commit change all `linuxkit/init:v0.4` to `linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4`? There is `./scripts/update-component-sha.sh` which helps you with that.
Thanks

kmjohansen:
@rn Absolutely.  I've pushed a new version of these commits which include a separate change to move the init package to the hash value that you specified.

rn:
Hmm, CI failed in a strange way...looks like maybe a timeout. I've hit rebuild

rn:
@kmjohansen apologies for the delay, I was away/busy. I think I now know what's going on in CI. Could you try and rebase your PR to master to include https://github.com/linuxkit/linuxkit/pull/3051?

kmjohansen:
@rn Absolutely; I've rebased and pushed an updated set of commits.

rn:
Hmm, failed again and I saw the same symptoms on https://github.com/linuxkit/linuxkit/pull/3058 but then it went away after hitting rebuild.

Hitting rebuild one more time

@kmjohansen apologies about this. It seems unrelated to your PR.

justincormack:
Can you rebase now #3065 is merged, this should resolve CI.

kmjohansen:
Absolutely; updated commits pushed.

rn:
Yeah, CI passed. Apologies this took so long. I didn't have time until recently to look into the failure in more details

PR #3044, opened 2018-05-18
Only support long RSA and ed25519 ssh keys

justincormack:
1. Support ed25519 ssh keys in example
2. Only generate ed25519 and 4096 bit RSA keys for sshd by default

dsa and ecdsa and short RSA keys are not recommended. You should
probably just switch to ed25519.

This only affects installs if you do not add your own keys.

hippo-milk-pink-orig

### DIFF
--- a/examples/sshd.yml
+++ b/examples/sshd.yml
@@ -28,6 +28,10 @@ files:
     source: ~/.ssh/id_rsa.pub
     mode: "0600"
     optional: true
+  - path: root/.ssh/authorized_keys
+    source: ~/.ssh/id_ed25519.pub
+    mode: "0600"
+    optional: true
 trust:
   org:
     - linuxkit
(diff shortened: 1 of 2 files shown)


examples/sshd.yml
```diff
@@ -28,6 +28,10 @@ files:
     source: ~/.ssh/id_rsa.pub
     mode: "0600"
     optional: true
+  - path: root/.ssh/authorized_keys
```
ijc: Does having two entries with identical `path` really DTRT in all possible combinations of `source` being available?

Perhaps we should have a special entry (akin to the `metadata` key) which specifies a list of public keys to harvest and construct an authorized_keys from? e.g. `authorized_keys: ["rsa", "ed25519"]` (or full paths, etc)

PR #3047, opened 2018-05-24
wireguard: upgrade to 0.0.20180524

zx2c4:
```
  * allowedips: set pointer to null before freeing
  * ncat-client-server: do not always call sudo and use env bash
  * qemu: bump default kernel for gcc 8.1
  * compat: work around qcom 4.9 backports
  
  The usual fixes.
  
  * tools: fix OpenBSD build
  * tools: always pass -v as first argument to install
  
  Portability changes.
  
  * wg-quick: darwin: rename namefile environment variable
  * wg-quick: darwin: do not remove routes when no real interface
  * wg-quick: freebsd: add new implementation
  * wg-quick: openbsd: add new implementation
  * wg-quick: support FreeBSD/Darwin search path
  * wg-quick: better bash completion for non-renaming OSes
  * wg-quick: allow enumeration of socket files
  
  We now support OpenBSD and FreeBSD, in addition to macOS.
```


zx2c4:
Next round is fine.

PR #3060, opened 2018-06-06
Stash kernel source and WireGuard source in the package

rn:
The main commit is:
```
    While we can re-create the kernel source code we don't have it
    handily available in one place. This commit stashes the kernel
    and the WireGuard source as /src/linux.tar.xz and
    /src/wireguard.tar.xz in the kernel package.

    This increases the size of the hub image by around 100MB.
```
In addition, I also added `-f` to the `curl` invocations which were missing it.

image


justincormack:
Yes, its a larger download but it seems like the best option.

PR #3063, opened 2018-06-08
Add a restart and stop command to service.

justinbarrick:
Signed-off-by: justinbarrick <jbarrick@cloudflare.com>

**- What I did**

Added restart and stop commands to the `service` command to restart and stop services.

**- How to verify it**

Run `service restart someservice` and `service stop someservice`.

Added restart and stop commands to the `service` command to restart and stop services.

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -33,7 +33,7 @@ kernel:
   image: linuxkit/kernel:4.9.91
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
-  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
+  - linuxkit/init:eca81db9b072eb431c057bab77797e169295fa1e
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
 services:
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.48
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:fbcbbe3ace0b3e9455300017dd07625293a0d4d4
+  - linuxkit/init:eca81db9b072eb431c057bab77797e169295fa1e
   - linuxkit/runc:v0.4
   - linuxkit/containerd:72df922e49a363a4ad53d88cd1a8a3ae4d41086f
   - linuxkit/ca-certificates:v0.4
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.48
   cmdline: "console=ttyS0"
 init:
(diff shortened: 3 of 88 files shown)


deitch:
CI is failing, probably should look at the above and figure out why.

deitch:
I think that adding `stop` and `restart` is a good thing. I have some suggestions on the structure though. 

1. Why the `switch args[0]` statement [here in main.go](https://github.com/linuxkit/linuxkit/pull/3063/files#diff-d97c24f85a236f43d00bbda743ed98e0R102) and then another [here in start.go](https://github.com/linuxkit/linuxkit/pull/3063/files#diff-ff45df16bbcd1fbb7b9b7370398c6b50R50), when they basically do the same thing? Might as well just have the single one in `main.go` call the relevant `func` directly.
2. We should rename `start.go` to `cmd.go` if it will have `start`, `stop`, and `restart`; or alternatively have separate files `start.go`, `stop.go` and `restart.go`, each with its own `func`. Either way, match the filename to what it does.

That is about it.

rn:
I'm looking into the CI failure. Seem the builders run out of diskspace. Might take a day or two to fix.

talex5:
I've rebuilt it now with a larger disk. The build required 11G in the end. However, it still failed:

```
[+ 79m 02s] gofmt...
[+ 79m 02s] start.go
[+ 79m 02s] The command '/bin/sh -c go-compile.sh /go/src/cmd/service' returned a non-zero code: 1
```

justinbarrick:
Thanks for the review everyone. Are you happy to merge this?

justinbarrick:
Thanks!

rn:
@justinbarrick I've build and pushed an updated `linuxkit/init` package with this change for x86, arm64, and s390x. Could you add a commit to this PR to update the hashes in the YAML files?

There is script, so something like this from the top level directory should do the right thing:
```
./scripts/update-component-sha.sh --image linuxkit/init:9c91a7f714d84ef5f7aac8ec31536db9001d9930
```

Once you pushed the new commit, CI should pick it up and hopefully it will pass

justinbarrick:
I've got a merge conflict, but the button doesn't work for me since I don't have write access to Linuxkit. Do one of you want to press that or should I resolve it in my repo?

rn:
Ah, its a conflict with the power change in https://github.com/linuxkit/linuxkit/pull/3030

Could you:
- drop the hash update commit
- rebase to master
- force push

I will then rebuild the init package and give you a new hash

can you then add a commit with the update of the YAMLs.

Sorry, but the way the hashes work we need to go though this again, but shouldn't take long

Thanks

justinbarrick:
Done!

rn:
Thanks for the patience. I've rebuild and pushed the `init` package for all arches. Could you now push a new commit again updating the `init` package to:
```
linuxkit/init:eca81db9b072eb431c057bab77797e169295fa1e
```

This now should not have merge conflicts and also the right meta-data

deitch:
CI is green. I will merge it in. Thanks for all of it @justinbarrick

justinbarrick:
Awesome, thanks so much!

PR #3065, opened 2018-06-09
tests: Fix wireguard test shutdown

rn:
With PR #3030 the behaviour of poweroff/halt is changed. This
test relies on on-shutdown containers to be executed to display
the test result (service containers have their stdout redirected).
Use 'poweroff' (note, no '-f') to ensure that:
- the machine actually powers off
- the on-shutdown container is executed

Note, there are subtle differences between 'poweroff' and 'halt'
between hypervisors. With HyperKit, 'halt' actually works, but with
qemu/kvm, with 'halt' the process does not exit.

Signed-off-by: Rolf Neugebauer <rolf.neugebauer@gmail.com>

image

### DIFF
--- a/test/cases/040_packages/023_wireguard/check.sh
+++ b/test/cases/040_packages/023_wireguard/check.sh
@@ -20,8 +20,8 @@ fi
 # Nginx may not be up immediately as service startup is async
 for s in $(seq 1 10)
 do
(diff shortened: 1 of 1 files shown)


rn:
FWIW I've tested this patch on top of #3030 on Linux with qemu both on KVM and inside a container and both seem to work

PR #3071, opened 2018-06-19
Workaround DNS issue with qemu-user-static on alpine by using Debian in the first stage #2710

### LINKED ISSUE #2710: qemu-user-static executables from alpine are broken, use debian instead
**Description**
Run an emulated image, for instance ppc64le/debian:9 or ppc64le/fedora:26 . Try to use any program using dns, for instance curl or yum. Program does not find any host or fails.

As far as I could check, the qemu-user-*-static executables from alpine are broken.
Replacing them with debian ones fixes the issue.

Fastest to see that qemu from alpine is broken can be done by fedora images:

`docker run --rm  -it ppc64le/fedora:26 /usr/libexec/system-python /usr/bin/yum update`
Error: Failed to synchronize cache for repo 'updates'

Doublecheck that alpine-linux qemu- is the culprit:

`docker -v $(pwd):/ws debian:9 bash -c "apt-get update && apt-get -y install qemu-user-static && cp /usr/bin/qemu-ppc64le-static /ws/qemu-debian`
`docker -v $(pwd):/ws alpine sh -c "apk update && apk add qemu-ppc64le &&/usr/bin/qemu-ppc64le /ws/qemu-alpine`

`docker run --rm -v $(pwd)/qemu-alpine:/qemu -it ppc64le/fedora:26 /qemu /usr/libexec/system-python /usr/bin/yum update`
Error: Failed to synchronize cache for repo 'updates'

However debians qemu works:
`docker run --rm -v $(pwd)/qemu-debian:/qemu -it ppc64le/fedora:26 /qemu /usr/libexec/system-python /usr/bin/yum update`

Another one which fails 

`docker run  --rm -it ppc64le/debian:9  bash -c "apt-get update && apt-get install -y curl && curl https://www.docker.com "`
Return: curl: (6) Could not resolve host: www.docker.com

And btw: aarch64 is broken the same way.

justinbarrick:
Fixes #2710 

DNS lookups fail in qemu-user when it is built on Alpine: https://bugs.alpinelinux.org/issues/8131

Until this is resolved, we fetch the binaries from Debian and use those instead. The final stage
of the Dockerfile is still based on scratch.

We can revert this once the Alpine issue is fixed.

Signed-off-by: Justin Barrick <jbarrick@cloudflare.com>

### DIFF
--- a/pkg/binfmt/Dockerfile
+++ b/pkg/binfmt/Dockerfile
@@ -1,8 +1,9 @@
-FROM linuxkit/alpine:1b05307ae8152e3d38f79e297b0632697a30c65c AS qemu
-RUN apk add \
-    qemu-aarch64 \
-    qemu-arm \
-    qemu-ppc64le
+# Use Debian stretch until https://bugs.alpinelinux.org/issues/8131 is resolved.
+FROM debian@sha256:de3eac83cd481c04c5d6c7344cd7327625a1d8b2540e82a8231b5675cef0ae5f AS qemu
+RUN apt-get update && apt-get install -y qemu-user-static && \
+    mv /usr/bin/qemu-aarch64-static /usr/bin/qemu-aarch64 && \
+    mv /usr/bin/qemu-arm-static /usr/bin/qemu-arm && \
+    mv /usr/bin/qemu-ppc64le-static /usr/bin/qemu-ppc64le
 
 FROM linuxkit/alpine:1b05307ae8152e3d38f79e297b0632697a30c65c AS mirror
 


justincormack:
Sorry, just having some issues testing this as there seems to be some weirdness sharing the `binfmt` filesystem.

The reason I want to do some testing is that there are patches in the Alpine qemu binaries for signal handling in Go, which weren't in upstream previously.

pkg/binfmt/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM linuxkit/alpine:1b05307ae8152e3d38f79e297b0632697a30c65c AS qemu
-RUN apk add \
-    qemu-aarch64 \
-    qemu-arm \
-    qemu-ppc64le
+# Use Debian stretch until https://bugs.alpinelinux.org/issues/8131 is resolved.
+FROM debian@sha256:de3eac83cd481c04c5d6c7344cd7327625a1d8b2540e82a8231b5675cef0ae5f AS qemu
+RUN apt-get update && apt-get install -y qemu-user-static && \
+    mv /usr/bin/qemu-aarch64-static /usr/bin/qemu-aarch64 && \
+    mv /usr/bin/qemu-arm-static /usr/bin/qemu-arm4 && \
```
justincormack: there is a typo here, should not be "arm4" just "arm"

justincormack:
Ok, it seems to run at least some Go binaries (eg I compiled and run linuxkit for arm64) so I think its ok. However, the way this is being run is not really working, when I bind mount the `binfmt` filesystem it is empty, so I am going to do a PR to change how it works slightly.

justinbarrick:
@justincormack from what I could tell, the bind mount shows empty from inside of the container, but if you look outside of the container in /proc you can still find it and it's working on my VMs as well. So I think that's normal but I could be missing something.

justincormack:
I was having trouble getting it to work at all, I have a small patch to mount the binfmt filesystem inside the container instead that seems better...

justinbarrick:
Sounds good, should I close this then?

justincormack:
No leave this open for now. I may just cherry pick your commit thougg in
which case can close later.

justincormack:
cherry picked in #3075 so closing

PR #3075, opened 2018-06-22
Binfmt fixes

### LINKED ISSUE #2710: qemu-user-static executables from alpine are broken, use debian instead

justincormack:
Rather than bind mounting binfmt_misc filesystem, mount a copy.
    
For some reason, bind mounting does not always seem to work,
sometimes the filesystem is empty. Mounting a fresh copy seems
a better solution, and simplifies things. The container does
need `CAP_SYS_ADMIN` but only on boot.

also cherry pick #3071 to use Debian for qemu as Alpine currently broken for networking.

fix #2710

### DIFF
--- a/examples/docker-for-mac.yml
+++ b/examples/docker-for-mac.yml
@@ -17,7 +17,7 @@ onboot:
   - name: sysfs
     image: linuxkit/sysfs:v0.4
   - name: binfmt
-    image: linuxkit/binfmt:v0.4
+    image: linuxkit/binfmt:0f73915a3fe365615a10a811f5b83303f48949d5
   # Format and mount the disk image in /var/lib/docker
   - name: format
     image: linuxkit/format:v0.4
--- a/pkg/binfmt/Dockerfile
+++ b/pkg/binfmt/Dockerfile
@@ -1,8 +1,9 @@
-FROM linuxkit/alpine:1b05307ae8152e3d38f79e297b0632697a30c65c AS qemu
-RUN apk add \
-    qemu-aarch64 \
-    qemu-arm \
-    qemu-ppc64le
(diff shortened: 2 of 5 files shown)


justincormack:
Updated uses, pushed package.

PR #3078, opened 2018-06-26
Fix dhcpcd hooks config

justincormack:
These were being added to the incorrect directory.

panda lying down


justincormack:
Added packages and yaml updates.

justincormack:
Ah I realised why the config file was not in `/etc`, reverted that part.

PR #3080, opened 2018-06-26
Use apk --print-arch to discover architecture

### LINKED ISSUE #3079: APK Architecture not always reported correctly
**Description**

When trying to build tools/alpine on other architectures (namely armhf, aka armv7l), `uname -m` reports a different architecture (`armv7l`) than that expected by alpine's `apk`.

Run `make` in `tools/alpine` on a raspberry pi 2 (supported by docker, not yet by linuxkit)

Build failed; unable to find /mirrors/armv7l/

Build success

Although the problematic architecture isn't yet supported by linuxkit, the fix appears to work on an x86_64 architecture and should work on all others.

alanraison:
fixes #3079 

**- What I did**
Changed references to `uname -m` in the context of an apk command to `apk --print-arch`

**- How to verify it**
Build tools/alpine with `make` on different architectures

When dealing with apk, `uname -m` doesn't always match the architecture
name that apk uses. Instead `apk --print-arch` is used.

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -9,8 +9,8 @@ COPY packages* /tmp/
 
 # mirror packages
 RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \
-   mkdir -p /mirror/$(uname -m) && \
-   apk fetch --recursive -o /mirror/$(uname -m) $(apk info; cat /tmp/packages)
+   mkdir -p /mirror/$(apk --print-arch) && \
+   apk fetch --recursive -o /mirror/$(apk --print-arch) $(apk info; cat /tmp/packages)
 
 # It's tricky to mix edge/testing packages which sometimes leads to dependency conflicts.
 # wireguard-tools currently is only in edge, so here we build our own package using the
@@ -23,7 +23,7 @@ RUN apk add alpine-sdk libmnl-dev curl && \
     cd /wireguard && \
     curl -fsSLo APKBUILD https://git.alpinelinux.org/cgit/aports/plain/testing/wireguard-tools/APKBUILD && \
     su -c "abuild-keygen -a -n && abuild -r" builder && \
-    cp /home/builder/packages/$(uname -m)/wireguard-tools-[0-9]*.apk /mirror/$(uname -m)
+    cp /home/builder/packages/$(apk --print-arch)/wireguard-tools-[0-9]*.apk /mirror/$(apk --print-arch)
 
 # install abuild for signing
 RUN apk add --no-cache abuild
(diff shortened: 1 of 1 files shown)


tools/alpine/Dockerfile
```diff
@@ -9,8 +9,8 @@ COPY packages* /tmp/
 
 # mirror packages
 RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \
```
alanraison: FYI, this was intentionally left as `uname -m` to hide the apk-specific details; hence the packages files in this folder would still be named after `uname -m`. Since these relate to apk packages, perhaps this should also refer to the apk architecture.

deitch:
I thought we had a discussion about `uname -m` vs `print-arch` somewhere...

Ah yes, here it was #2961 . That was specifically around kernel builds, whereas this is building the base `alpine`, where it might be more relevant.

I think the primary objections were raised by @ijc ?

ijc:
The rationale for the change in #2961 seemed to be around supporting (cross?) biarch, for which it was not a complete fix (plenty of other places would have needed changing), nor was it clear that a fix was needed at all, for _that_ case.

The rationale here seems much clearer &mdash; specifically  `--rewrite-arch` certainly does need an apk arch name and if uname and apk disagree for arm32 we have to take the apk name. AIUI the paths within the mirror mush also use apk names or apk won't find them. Also unlike the kernel this package is pretty closely tied to alpine.

I'm assuming this doesn't require updates to all the downstream packages since they rely on `/etc/apk/repositories` and `apk` colluding to DTRT, very few (if any) of them should ever be hardcoding a path under `/mirror`.

In the issue it says `Build failed; unable to find /mirrors/armv7l/`, but I'd have expected you to trip over something along the lines of "unknown arch armv7l used with --rewrite-arch` first, what happened there? Did it actually build an empty repo for an unknown arch or something?

This PR goes part of the way to supporting the new arch but surely at least `tools/alpine/push-manifest.sh` needs an update before an armv7 image can actually be built and pushed. I guess there will be other bits across the tree to come too (kernel, `scripts/push-manifest.sh`, the linuxkit build tooling, probably some subset of pkg). Do you have a full working end to end build of armhf from which you cherry-picked this first standalone fix or are you still chipping away at it?

However before we go too far down the path of getting this PR in we should also step back and ask ourselves what we as a project need in order to support a new arch. In particular I think we need a maintainer accessible build machine in order to do the builds which get pushed to hub (which only maintainers can do, and which all maintainers have to be able to do for various boring reasons). I'm not sure what else is needed, @rn was more closely involved in the addition of s390x recently so he may have a better idea.

justincormack:
@ijc I think its fine to fix the tooling for more architectures even if we are not currently planning to have official builds. We don't currently have a machine that can do arm32 as people keep making 64 bit only arm servers.

alanraison:
This is my very first step into trying to get a port of linuxkit for armv7l (which may not even get anywhere). However, I thought that the issue I encountered (sorry I forget the exact error message, it took me a while to debug) seemed to be with a mismatch between apk and uname, that it was worth fixing for all architectures, since I guess it's just good fortune that it hasn't been spotted before.

As you say, there are plenty of other questions to ask before a full port can be considered.

deitch:
@ijc's comments convinced me too:

> The rationale here seems much clearer — specifically --rewrite-arch certainly does need an apk arch name and if uname and apk disagree for arm32 we have to take the apk name

deitch:
Do we need to update the hashes for the `FROM` lines in the `pkg` Dockerfiles? E.g. 

```dockerfile
FROM linuxkit/alpine:1b05307ae8152e3d38f79e297b0632697a30c65c AS mirror
```

And did we ever put in tooling for that? I am not sure if [update-component-sha.sh](https://github.com/linuxkit/linuxkit/blob/master/scripts/update-component-sha.sh) will handle that correctly, but I think it might.

deitch:
@rn wrote:

> I'm also fine with adding additional architecture support to the tooling (like push-manifest.sh and the linuxkit tool) later, to allow others building LinuxKit images for other architectures.
> 
> However, adding a new supported architecture, ie something the LinuxKit project would provide images for, update the kernels etc is something we would have to consider very carefully and it would require access to suitable build infrastructure etc.

We probably should have something in the docs about:

* how to build for a new architecture (requirements, process, etc.)
* requirements for a fully supported architecture

alanraison:
Ah yes, I remember the error now; the build didn't even reach --rewrite-arch; it failed on line 26:

```
cp /home/builder/packages/$(uname -m)/wireguard-tools-[0-9]*.apk /mirror/$(uname -m)
```

Since /home/builder/packages/armv7l didn't exist.

rn:
@deitch no need to update all the `FROM` hashes. We haven't done this in the past, e.g. for containerd updates. We let the `linuxkit/alpine` versions drift in the different packages and then resync for releases. I'm also planning to do the update to alpine 3.8 soon, which should pick these changes up.

PR #3085, opened 2018-07-01
Support pluggable logging systems

djs55:
**- What I did**

I re-integrated `projects/logging` (written by @MagnusS ) into current master to allow optional pluggable logging systems.

**- How I did it**

By default everything works as before: `onboot` and `services` log to files in `/var/log`.

However if the socket `/var/run/external-logging.sock` exists then the `service` command will

- connect the `runc` `stdout` and `stderr` to a socketpair, where the other end has been sent to the `/var/run/external-logging.sock`
- create FIFOs for the services, send the paths to `containerd`, open the other ends and send them to the `/var/run/external-logging.sock`
 
The package `memlogd` provides the `/var/run/external-logging.sock` and writes logs to an in-memory circular buffer. Later on in the boot a service can connect to `memlogd` and stream the logs somewhere else, for example over the network.

**- How to verify it**

```
$ cd examples
$ linuxkit build -disable-content-trust logging.yml
$ linuxkit run logging
```
and then in the `getty`:
```
/proc/1/root/usr/bin/logread -F
2018-07-01T13:37:55Z memlogd memlogd started
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: waiting for carrier
2018-07-01T13:37:55Z 001-dhcpcd.err eth0: could not detect a useable init system
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: carrier acquired
2018-07-01T13:37:55Z 001-dhcpcd.out DUID 00:01:00:01:22:cb:95:b3:02:50:00:00:00:0b
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: IAID 00:00:00:0b
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: adding address fe80::3c9c:5e2e:f13c:9016
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: soliciting an IPv6 router
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: soliciting a DHCP lease
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: offered 192.168.65.13 from 192.168.65.1 `vpnkit'
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: leased 192.168.65.13 for 7200 seconds
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: adding route to 192.168.65.0/24
2018-07-01T13:37:55Z 001-dhcpcd.out eth0: adding default route via 192.168.65.1
2018-07-01T13:37:55Z 001-dhcpcd.out exiting due to oneshot
2018-07-01T13:37:55Z 001-dhcpcd.out dhcpcd exited
2018-07-01T13:37:55Z getty.out [INFO  tini (444)] Spawned child process '/usr/bin/rungetty.sh' with pid '461'
2018-07-01T13:37:55Z getty.out getty: cmdline has console=tty0 but does not exist in /etc/securetty; will not be able to log in as root on this tty tty0.
2018-07-01T13:37:55Z write-to-the-logs.out hello Sun Jul 1 13:37:55 UTC 2018
2018-07-01T13:37:56Z write-to-the-logs.out hello Sun Jul 1 13:37:56 UTC 2018
2018-07-01T13:37:57Z write-to-the-logs.out hello Sun Jul 1 13:37:57 UTC 2018
2018-07-01T13:37:58Z write-to-the-logs.out hello Sun Jul 1 13:37:58 UTC 2018
2018-07-01T13:37:59Z write-to-the-logs.out hello Sun Jul 1 13:37:59 UTC 2018
2018-07-01T13:38:00Z write-to-the-logs.out hello Sun Jul 1 13:38:00 UTC 2018
2018-07-01T13:38:01Z write-to-the-logs.out hello Sun Jul 1 13:38:01 UTC 2018
```
and then edit the `logging.yml` to remove `memlogd` from the `init` section. The VM will now log to files as before.

Support optional pluggable logging systems

cute animals


justincormack:
Sounds good, I am away this week so may not have a chance to review, hopefully someone else can.

docs/logging.md
```diff
@@ -0,0 +1,88 @@
...
+- `/var/run/external-logging.sock`: used for receiving standard out and standard error
+  file descriptors
+- `/var/run/memlogdq.sock`: used for reading the buffered log
+
+When `memlogd` receives a file descriptor it will read from the file
+descriptor and timestamp and append the content to the in-memory log until
+the file is closed.
+
+Since `memlogd` logs to a circular buffer, when the buffer fills then log
+messages will be dropped. Therefore it is recommended that a `service` be
```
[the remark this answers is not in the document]
djs55: A good question -- I had a more thorough look at the `memlogd` code. It uses a fixed length `container/ring` initialised here: https://github.com/djs55/linuxkit/blob/cb4817bd419baedd4077bd349d7f989c1942d8d0/pkg/memlogd/cmd/memlogd/main.go#L112 It maintains a pointer into the ring and appends by overwriting and advancing the pointer here: https://github.com/djs55/linuxkit/blob/cb4817bd419baedd4077bd349d7f989c1942d8d0/pkg/memlogd/cmd/memlogd/main.go#L120 so I think it only drops when it cycles round to the beginning and hence drops old messages.

I think there is a memory usage issue. Although the input messages are in a fixed-size ring, there is a per-reader [bytes.Buffer](https://github.com/djs55/linuxkit/blob/cb4817bd419baedd4077bd349d7f989c1942d8d0/pkg/memlogd/cmd/memlogd/main.go#L47) which has no bounds. If the reader is slower than the writer then the data will fill up in memory. Perhaps this should be a ring too?

I think another nice-to-have in the `memlogd` code would be to keep a log message counter, so clients could detect when they've dropped messages (and log "dropped n messages") since it's useful to know when a log message was definitely not written as opposed to maybe written and then dropped.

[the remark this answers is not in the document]
ijc: [`bytes.NewBuffer`](https://golang.org/pkg/bytes/#NewBuffer) can be given a `[]byte` with a fixed capacity which should bound it I think.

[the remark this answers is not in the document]
ijc: My mistake, that just sets the initial size, it can still grow...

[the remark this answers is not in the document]
rn: bounding the buffers now would be good. Adding a counter later is fine. I'm not sure I understand your design regarding the counter though. I would think the client sending the counter as part of the message and `memlogd` keeping an internal counter allows `memlogd` to detect dropped messages and log it accordingly, ie `memlogd` detects dropped messages instead of the client (as you described above).

pkg/init/cmd/service/cmd.go
```diff
@@ -150,6 +151,35 @@ func stop(ctx context.Context, service, sock, basePath string) (string, uint32,
...
+		go func() {
+			fd, err := syscall.Open(path, syscall.O_RDONLY, 0)
+			if err != nil {
+				// Should never happen: we just created the fifo
+				log.Fatalf("failed to open fifo %s: %s", path, err)
+			}
+			defer syscall.Close(fd)
+			if err := sendToLogger(name, fd); err != nil {
+				// Should never happen: logging is enabled
+				log.Fatalf("failed to send fifo %s to logger: %s", path, err)
```
rn: Is `Fatalf` the right thing here? Maybe default to files?

pkg/init/cmd/service/cmd.go
```diff
@@ -150,6 +151,35 @@ func stop(ctx context.Context, service, sock, basePath string) (string, uint32,
 	return id, pid, "", nil
 }
 
+func newLogDestination(name string) string {
+	path := filepath.Join("/var/log", name+".log")
+
+	if loggingEnabled() {
+		if err := syscall.Mkfifo(path, 0600); err != nil {
```
rn: If I read this code right we either have `/var/log/<name>.log` being a text file or a FIFO. It might be better to have the FIFO in in `/var/run` to keep `/var/log` for text files (optionally rotated later)

pkg/init/cmd/service/runc.go
```diff
@@ -75,27 +75,41 @@ func runcInit(rootPath, serviceType string) int {
 		}
 		pidfile := filepath.Join(tmpdir, name)
 		cmd := exec.Command(runcBinary, "create", "--bundle", path, "--pid-file", pidfile, name)
+		var stdout, stderr *os.File
 
 		// stream stdout and stderr to respective files
```
rn: while fifos are files too, maybe this comment can be made more concise

djs55:
I've pushed a rebase with

- `if loggingEnabled ()` removed in favour of an interface and a pair of structs in logging.go. A function `GetLogger()` returns the most appropriate implementation.
- the unbounded size `bytes.Buffer` and `sync.Cond` in `memlogd` are replaced with a bounded size channel using `select` for non-blocking writes. This ensures that a slow client won't cause excessive buffering in the server and is a bit simpler to understand (no more `Lock` and `Unlock` to protect the buffer between goroutines)
- some test cases to `memlogd` to check some simple things (writers don't block, the ring is finite and slow clients don't cause the servers to buffer excessively)
- a `kmsg`-style log reading format
```
  <timestamp>,<origin>;<body>
```
- hopefully clearer docs (-- let me know what you think).
- the FIFOs moved to `/var/run`-- they were previously in `/var/log` which was confusing.

docs/logging.md
```diff
@@ -0,0 +1,93 @@
...
+   all the early `onboot` logs)
+2. a log writing `service` that starts later and can download and process
+   the logs from `memlogd`
+
+To use this new logging system, you should add the `memlogd` container to
+the `init` block in the LinuxKit yml. On boot `memlogd` will be started
+from `init.d` and it will listen on a Unix domain socket:
+
+```
+/var/run/external-logging.sock
```
ijc: Maybe involve the word `linuxkit` in the name of the socket, just in case? `linuxkit-external-logging.sock`?

docs/logging.md
```diff
@@ -0,0 +1,93 @@
...
+package demonstrates how to do this.
+
+### Message format
+
+The format used to read logs is similar to [kmsg](https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg):
+```
+<timestamp>,<origin>;<body>
+```
+where `<timestamp>` is an RFC3339-formatted timestamp, `<origin>` is the name of the
+original sending process and `<body>` is the output.
```
ijc: What is `<origin>` is it the processes `argv[0]` at time of launch, `argv[0]` now or something else? In principal it could contain a `;` which would royally confuse the recipient. Might the `pid` be sufficient?

The examples below suggest it might actually be the name of the log file rather than the process, in which case just the text needs changing, assuming we agree that `;` is verbotten in a service name...

docs/logging.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+Current issues and limitations:
+
+- No docker logger plugin support yet - it could be nice to add support to
+  memlogd, so the docker container logs would also be gathered in one place
+- No syslog compatibility at the moment and `/dev/log` doesn’t exist. This
+  socket could be created to keep syslog compatibility, e.g. by using
+  https://github.com/mcuadros/go-syslog. Processes that require syslog should
+  then be able to log directly to memlogd.
```
ijc: Do we have any processes/services in `pkg/` (or elsewehre we care about) which might want syslog?

pkg/init/cmd/service/logging.go
```diff
@@ -0,0 +1,177 @@
...
+	Path(name string) string            // Path of the log file (or FIFO)
+	Open(name string) (*os.File, error) // Opens a log file directly
+	DumpAll()                           // Copies all the logs to the console
+}
+
+// GetLog returns the log destination we should use.
+func GetLog(logDir string) Log {
+	// is an external logging system enabled?
+	_, err := os.Stat(logWriteSocket)
+	if !os.IsNotExist(err) {
```
ijc: nit: `if _, err := ... ; !os.IsNotExist(err) { ....`

pkg/init/cmd/service/logging.go
```diff
@@ -0,0 +1,177 @@
...
+// Path returns the name of a log file path for the named service.
+func (f *fileLog) Path(name string) string {
+	// We just need this to exist. If we cannot write to the directory,
+	// we'll discard output instead.
+	file, err := f.Open(name)
+	if err != nil {
+		return "/dev/null"
+	}
+	_ = file.Close()
+	return filepath.Join(f.dir, name+".log")
```
ijc: I'm confused, `file` had a different name (and potentially path, depending on `cwd`) to what is being returned.

pkg/init/cmd/service/logging.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Open a log file for the named service.
+func (f *fileLog) Open(name string) (*os.File, error) {
+	return os.Create(filepath.Join(f.dir, name+".log"))
```
ijc: The result here is `RDWR` rather than `WRONLY` as before. 
I think you can make the return an `io.WriteCloser` here instead of `*os.File` too, which more accurately reflects what the caller wants/needs.

pkg/init/cmd/service/logging.go
```diff
@@ -0,0 +1,177 @@
...
+		if err := sendToLogger(name, fd); err != nil {
+			// Should never happen: logging is enabled
+			log.Printf("failed to send fifo %s to logger: %s", path, err)
+		}
+	}()
+	return path
+}
+
+// Open a log file for the named service.
+func (r *remoteLog) Open(name string) (*os.File, error) {
```
ijc: `io.WriteCloser` again?

docs/logging.md
djs55: It's actually the name of the log -- it's chosen by `pkg/init` when it sets up the logging, based on the service name (usually with a `.out` or `.err` suffix). I'll clarify the text and note that `;` is banned. Thinking about it I could also reject `;' in `memlogd` just in case.

docs/logging.md
djs55: I don't know of any myself-- everything I've seen is ok with `stdout`/`stderr` logging. I guess the main advantage of `syslog` is that it's a bit more structured, since the client gets to choose a facility and level, where here with `stdout` and `stderr` it's all just lines.

justincormack:
I think we dont have anything logging to syslog. The container and systemd
style is stderr logging so its most common.

docs/logging.md
ijc: > I don't know of any myself

Cool. What I was really meaning was "if we know of any we should list them", if there are none that's even better ;-)

docs/logging.md
ijc: We should possibly also disallow `;` in service names in the first place, if we don't already

pkg/init/cmd/service/logging.go
djs55: This just confused me as well (!) It's confusing because `f.Open(name)` looks like a call to `os.Open(path)` but is actually a call to
```golang
func (f *fileLog) Open(name string) (*os.File, error) {
	return os.Create(filepath.Join(f.dir, name+".log"))
}
```
so I think it is working by computing the path twice. I'll see if I can rename or otherwise refactor `Open` (and `Path`) to avoid them being confused with `os.Open` etc. Since this behaviour was copied from the previous code I re-checked what happens if you omit it -- unfortunately `containerd` fails to start the task if the path doesn't exist. I'll mention that in the comments.

ijc: `*os.File` does have a `Name` method but a) I'm not convinced that'll be the full path b) I think the return should be an `os.WriteCloser` anyway.

Maybe just have `fileLog.Open` return the name as well as the `os.WriteCloser` (and the err)?

ijc: Or since the `fileLog.Open` is also part of the public interface (meaning you'd have to change the other one too, where it might make sense) make an internal helper which `Open` and `Path` both call internally.

djs55: Sorry I just raced with your comment and updated with something similar-ish. I've made a common function
```golang
func (f *fileLog) localPath(n string) string {
	return filepath.Join(f.dir, n+".log")
}
```
and now `Path` and `Open` call this function and then directly call `os.*` functions (rather than one calling the other which was hard to read):
```golang
// Path returns the name of a log file path for the named service.
func (f *fileLog) Path(n string) string {
	path := f.localPath(n)
	// We just need this to exist, otherwise containerd will say:
	//
	//   ERRO[0000] failed to create task error="failed to start io pipe
	//     copy: containerd-shim: opening /var/log/... failed: open
	//     /var/log/...: no such file or directory: unknown"
	file, err := os.Create(path)
	if err != nil {
		// If we cannot write to the directory, we'll discard output instead.
		return "/dev/null"
	}
	_ = file.Close()
	return path
}

// Open a log file for the named service.
func (f *fileLog) Open(n string) (io.WriteCloser, error) {
	return os.OpenFile(f.localPath(n), os.O_WRONLY|os.O_CREATE, 0644)
}
```
-- does that seem reasonable?

rn:
Thanks, looking good. 

I've set up signing for `linuxkit/memlogd` and build and pushed both `linuxkit/init` and `linuxkit/memlogd`,

Could you, in a separate commit, update the YAMLs to use:
```
linuxkit/init:4fa16c9045d60732dadb04f5d7f5c18d83c6551e
```
and
```
linuxkit/memlogd:883f0d46e7d3ae2d787e8acb496da115a4707cbc
```

Can you update the *all* the YAMLs with the new `linuxkit/init` as that way the CI will test it. You can use `./scripts/update-component-sha.sh` for that.

rn:
weird:
```
[+ 31m 54s] [FAIL    ] linuxkit.kernel.config_4.4.x 57.31s
[+ 32m 51s] [FAIL    ] linuxkit.kernel.config_4.9.x 57.00s
[+ 33m 44s] [FAIL    ] linuxkit.kernel.config_4.14.x 52.38s
[+ 34m 44s] [FAIL    ] linuxkit.kernel.config_4.17.x 60.56s
[+ 39m 43s] [FAIL    ] linuxkit.packages.binfmt 53.53s
...
```

Looking at the logs of theses tests I don't see any output from onboot/serrvice containers on the console. @djs55 do you changes disable this somehow?

djs55:
Hm, it's supposed to behave as before i.e. write the logs to files first
and then copy them to the console afterwards (because multiwriter doesn't
work with runc because of an open bug). I'll take a look

djs55:
It's because of a slight semantic change. Consider the `binfmt` yaml:
```
onboot:
  - name: binfmt
    image: linuxkit/binfmt:0f73915a3fe365615a10a811f5b83303f48949d5
  - name: test
    image: alpine:3.7
    binds:
      - /check.sh:/check.sh
      - /proc/sys/fs/binfmt_misc:/binfmt_misc
    command: ["sh", "./check.sh"]
  - name: poweroff
    image: linuxkit/poweroff:d75d94d9603bde9ab95b914deed0d069d05d8b20
    command: ["/bin/sh", "/poweroff.sh", "10"]
```
Previously [the log files were dumped one by one](https://github.com/linuxkit/linuxkit/blob/2c8438f64ba737b88bd1f69b87c2790a4a9cfd3a/pkg/init/cmd/service/runc.go#L155) inside the `for` loop, but here the [log files are dumped at the end](https://github.com/djs55/linuxkit/blob/649cff9a44efb2daaa4dd6af47eba08c571ede23/pkg/init/cmd/service/runc.go#L155). This means the `poweroff` happens before the call to `DumpAll` so the logs are missing. I'll see if I can refactor it to put the old behaviour back tomorrow. I think I can easily change `DumpAll()` into `Dump(log string)` for the file case but I'll have to download the logs from `memlogd` and filter to avoid repetition to get the same behaviour there. Not a big amount of change. Sorry for the signing inconvenience!

rn:
I'm not entirely convinced. If I change the `10` in the `poweroff` to something like 30s I still don't see the output on the console.

rn:
Also, if you look at the wireguard test, it passes, but there it is the `service` container printing the `PASSED`:

```
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060409354Z: INFO[0000] loading plugin "io.containerd.grpc.v1.leases"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060423249Z: INFO[0000] loading plugin "io.containerd.grpc.v1.namespaces"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060434773Z: INFO[0000] loading plugin "io.containerd.grpc.v1.snapshots"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060446847Z: INFO[0000] loading plugin "io.containerd.grpc.v1.tasks"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060466910Z: INFO[0000] loading plugin "io.containerd.grpc.v1.version"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060478770Z: INFO[0000] loading plugin "io.containerd.grpc.v1.introspection"...  type=io.containerd.grpc.v1
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060490308Z: INFO[0000] serving...                                    address="/run/containerd/debug.sock"
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060506155Z: INFO[0000] serving...                                    address="/run/containerd/containerd.sock"
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060517973Z: INFO[0000] containerd successfully booted in 0.338532s  
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060529875Z: INFO[0001] shim containerd-shim started                  address="/containerd-shim/services.linuxkit/nginx/shim.sock" debug=false pid=438
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060542071Z: INFO[0004] shim containerd-shim started                  address="/containerd-shim/services.linuxkit/test/shim.sock" debug=false pid=479
[+ 72m 39s] [STDERR  ] 2018-07-06T18:22:59.060553957Z: wireguard test suite PASSED
```

I haven't double checked all the tests, but looks like the `onboot` container output is going AWOL but `service` container output is fine.

djs55:
@rn I've rebased with this patch: https://github.com/djs55/linuxkit/commit/c45427711b770cf793601dd9fb31c3e8a3ce03b7 and updated the `linuxkit/init` hash (which won't exist but I think it's got the correct value)

rn:
can you check locally? Picking a random failing test, in `test/cases/020_kernel/009_config_4.17.x` I've done `linuxkit build test.yml` followed by `linuxkit run qemu test` on a mac. make sure you have the `linuxkit/init` hash adjusted and have a local `linuxkit pkg build .` in `./pkg/init`

rn:
we probably ought to have a test for memlogd as well...but maybe that can wait and be rolled in with thttps://github.com/linuxkit/linuxkit/pull/3095

djs55:
@rn with the new hash  `linuxkit/init:6cc1442112980c889230b6449df09d5b48de6854` (and `-disable-content-trust`)
```
$ cd test/cases/020_kernel/009_config_4.17.x
$ linuxkit build -disable-content-trust test.yml
$ linuxkit run qemu test
...
Limits:
- /proc/sys/kernel/keys/root_maxkeys: 1000000

Kernel config test suite PASSED
[    5.695258] ACPI: Preparing to enter system sleep state S5
[    5.696474] reboot: Power down
```

So ... I think it's looking ok.

Regarding the poweroff delay, it looks like [poweroff.sh](https://github.com/linuxkit/linuxkit/blob/master/test/pkg/poweroff/poweroff.sh) calls `sleep` and then `poweroff -f` so I think the `poweroff -f` was killing the VM before `init` could call `DumpAll` since I mistakenly moved `DumpAll` to after the `for` loop. Now that `init` is calling `Dump` in the `for` loop it should write the test results to the console before the `poweroff.sh`

rn:
rebuild/pushed `linuxkit/init`for all arches...let's see what's happening

PR #3095, opened 2018-07-06
Support log rotation and logging of /dev/kmsg

### LINKED ISSUE #3072: Docker daemon log taking up too much disk space
**Description**

To start off with, I'd like to clarify that this concern is for the docker **_daemon_** logs, **not** for docker's container-level logs. The `log-driver` and `log-opts` docker daemon configurations seem to handle container-level logs just fine. 

With that out of the way, the issue I'm running into is that the `/var/log/docker.err.log` log file is growing too quickly and not being rotated on our production servers, making us nervous that it will eventually fill up enough disk space to cause issues on the server itself. Concretely, we've been running for 19 days, and, on one of the nodes, that log file has grown to `3.6GB`. As our VM's are expected to run for months on end, this is obviously a concerning trend. 

The `linuxkit` OS that we're using has very similar configuration to the [example docker configuration](https://github.com/linuxkit/linuxkit/blob/master/examples/docker.yml), with some additional configuration in the `/etc/docker/daemon.json` file, most notably that we set `debug` to `True`. That was clearly a mistake, as my tests show that, had we not used that option, our log files would substantially smaller.

However, even if we had left the log-level at it's default of `info`, it looks to me like the log file could still cause issues if the server is left running for too long. I did some rough calculations, and my guess is that the file could still grow to something like 10GB if the daemon is left running for 6 months or so. 

I have been trying to get a manual `logrotate` solution working to protect against this, but it seems like the docker daemon never reloads it's log file, which means that when `logrotate` does it's thing and creates a blank new `docker.err.log`, the docker daemon continues to write at whatever offset it left off at, and backfills the rest of the file with null bytes, taking up as much space as it was before. I've tried some solutions involving sending the `HUP` signal to the docker process without any success; it seems like the docker daemon doesn't handle that signal, or at least not in a way that reloads it's log file. 

All that to say, I have two things I'm wondering about:

1. Is there an accepted way to implement rotating of the docker daemon logs? It seems silly that I can't find any good information about this. Surely someone else has bumped into this issue before? Or is it just accepted that the `docker.err.log` file keeps on growing forever, and you'd better restart your server at some point? If that's the current answer, I'd love to have a linuxkit-based way of rotating that log file specifically. 
2. Do you have any suggestions on how to truncate the existing `docker.err.log` on a running server _without_ shutting down the server or docker daemon instance? We'd like to avoid having to update the OS image we're running just to prevent the docker daemon log from using up all our disk space. 

Thanks for your time!

1. Create a linuxkit OS ISO with a docker daemon service with `debug: True` in it's configuration file
1. Using that ISO, run a docker swarm cluster in production for several days

* `/var/log/docker.err.log` file grows rapidly, and is not rotated

* Log file is rotated automatically, conserving disk space
* Alternatively, the docker daemon responds to the `HUP` signal by reloading it's log file, and beginning to write from the beginning, so that we can use `logrotate` properly.

djs55:
**- What I did**

- added `pkg/logwrite` which can connect to `memlogd` and stream logs to files in `/var/log`, with log rotation (Fixes #3072)
- added `pkg/kmsg` which connects to `/dev/kmsg` and logs the kernel log

**- How I did it**

This builds on top of the infrastructure proposed in #3085 (note I rebased this PR on top of that branch but only the last few patches are different)

**- How to verify it**

First build and run the logging example:
```
cd examples
linuxkit build -disable-content-trust logging.yml
linuxkit run logging
```

Then execute `watch ls -l /var/log` in the terminal. After a while it should look like:
```
total 72
-rw-r--r--    1 root     root           405 Jul  6 16:59 getty.out
-rw-r--r--    1 root     root           969 Jul  6 16:59 kmsg.out
-rw-r--r--    1 root     root          1051 Jul  6 16:59 kmsg.out.0
-rw-r--r--    1 root     root          1048 Jul  6 16:59 kmsg.out.1
-rw-r--r--    1 root     root          1100 Jul  6 16:59 kmsg.out.2
-rw-r--r--    1 root     root          1120 Jul  6 16:59 kmsg.out.3
-rw-r--r--    1 root     root          1029 Jul  6 16:59 kmsg.out.4
-rw-r--r--    1 root     root          1077 Jul  6 16:59 kmsg.out.5
-rw-r--r--    1 root     root          1048 Jul  6 16:59 kmsg.out.6
-rw-r--r--    1 root     root          1027 Jul  6 16:59 kmsg.out.7
-rw-r--r--    1 root     root          1034 Jul  6 16:59 kmsg.out.8
-rw-r--r--    1 root     root          1076 Jul  6 16:59 kmsg.out.9
-rw-r--r--    1 root     root            45 Jul  6 16:59 memlogd
lrwxrwxrwx    1 root     root            15 Jul  6 16:59 onboot -> /run/log/onboot
-rw-r--r--    1 root     root            88 Jul  6 16:59 onboot.001-dhcpcd.err
-rw-r--r--    1 root     root          1003 Jul  6 16:59 onboot.001-dhcpcd.out
-rw-r--r--    1 root     root           693 Jul  6 17:00 write-to-the-logs.out
-rw-r--r--    1 root     root          1078 Jul  6 16:59 write-to-the-logs.out.0
-rw-r--r--    1 root     root          1078 Jul  6 16:59 write-to-the-logs.out.1
```

Add support for log rotation and logging `/dev/kmsg`.

Cute animals

### DIFF
--- a/examples/logging.yml
+++ b/examples/logging.yml
@@ -26,6 +26,8 @@ services:
     command: ["/bin/sh", "-c", "while /bin/true; do echo hello $(date); sleep 1; done" ]
   - name: write-and-rotate-logs
     image: linuxkit/logwrite:7859c102a963828fd9c5aa3837db9600483220c7
+  - name: kmsg
+    image: linuxkit/kmsg:3dfa0d5b4027ecc16089f27fbcffa15d6aa5438e
 trust:
   org:
     - linuxkit
--- a/examples/logging.yml
+++ b/examples/logging.yml
@@ -24,6 +24,8 @@ services:
   - name: write-to-the-logs
     image: alpine
     command: ["/bin/sh", "-c", "while /bin/true; do echo hello $(date); sleep 1; done" ]
+  - name: write-and-rotate-logs
+    image: linuxkit/logwrite:7859c102a963828fd9c5aa3837db9600483220c7
 trust:
   org:
     - linuxkit
--- a/pkg/logwrite/logwrite.go
+++ b/pkg/logwrite/logwrite.go
@@ -1,0 +1,203 @@
+package main
+
+// Write logs to files and perform rotation.
+
+import (
+	"bufio"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+// These must be kept in sync with memlogd:
+const (
+	logDump byte = iota
+	logFollow
+	logDumpFollow
+)
+
+const mb = 1024 * 1024
+
+// LogMessage is a message received from memlogd.
+type LogMessage struct {
+	Time    time.Time // time message was received by memlogd
+	Name    string    // name of the service that wrote the message
+	Message string    // body of the message
+}
+
+func (m *LogMessage) String() string {
+	return m.Time.Format(time.RFC3339) + " " + m.Name + " " + m.Message
+}
+
+// ParseLogMessage reconstructs a LogMessage from a line of text which looks like:
+// <timestamp>,<origin>;<body>
+func ParseLogMessage(line string) (*LogMessage, error) {
+	bits := strings.SplitN(line, ";", 2)
+	if len(bits) != 2 {
+		return nil, errors.New("Failed to parse log message: " + line)
+	}
+	bits2 := strings.Split(bits[0], ",")
+	if len(bits2) < 2 {
+		// There could be more parameters in future
+		return nil, errors.New("Failed to parse log message: " + line)
+	}
+	Time, err := time.Parse(time.RFC3339, bits2[0])
+	if err != nil {
+		return nil, err
+	}
+	return &LogMessage{
+		Time:    Time,
+		Name:    bits2[1],
+		Message: bits[1],
+	}, nil
+}
+
+// LogFile is where we write LogMessages to
+type LogFile struct {
+	File         *os.File // active file handle
+	Name         string   // filename of log file
+	Dir          string   // log file directory
+	BytesWritten int      // total number of bytes written so far
+}
+
+// NewLogFile creates a new LogFile.
+func NewLogFile(dir, name string) (*LogFile, error) {
+	// If the log exists already we want to append to it.
+	f, err := os.OpenFile(filepath.Join(dir, name), os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
+	if err != nil {
+		return nil, err
+	}
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+	return &LogFile{
+		File:         f,
+		Name:         name,
+		Dir:          dir,
+		BytesWritten: int(fi.Size()),
+	}, nil
+}
+
+// Write appends a message to the log file
+func (l *LogFile) Write(m *LogMessage) error {
+	s := m.String()
+	_, err := io.WriteString(l.File, s)
+	if err == nil {
+		l.BytesWritten += len(s)
+	}
+	return err
+}
+
+// Close a log file
+func (l *LogFile) Close() error {
+	return l.File.Close()
+}
+
+// Rotate closes the current log file, rotates the files and creates an empty log file.
+func (l *LogFile) Rotate(maxLogFiles int) error {
+	if err := l.File.Close(); err != nil {
+		return err
+	}
+	path := filepath.Join(l.Dir, l.Name)
+	for i := maxLogFiles - 1; i >= 0; i-- {
+		newerFile := fmt.Sprintf("%s.%d", path, i-1)
+		// special case: if index is 0 we omit the suffix i.e. we expect
+		// foo foo.1 foo.2 up to foo.<maxLogFiles-1>
+		if i == 0 {
+			newerFile = path
+		}
+		olderFile := fmt.Sprintf("%s.%d", path, i)
+		// overwrite the olderFile with the newerFile
+		err := os.Rename(newerFile, olderFile)
+		if os.IsNotExist(err) {
+			// the newerFile does not exist
+			continue
+		}
+		if err != nil {
+			return err
+		}
+	}
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	l.File = f
+	l.BytesWritten = 0
+	return nil
+}
+
+func main() {
+	socketPath := flag.String("socket", "/var/run/memlogdq.sock", "memlogd log query socket")
+	logDir := flag.String("log-dir", "/var/log", "Directory containing log files")
+	maxLogFiles := flag.Int("max-log-files", 10, "Maximum number of rotated log files before deletion")
+	maxLogSize := flag.Int("max-log-size", mb, "Maximum size of a log file before rotation")
+	flag.Parse()
+
+	addr := net.UnixAddr{
+		Name: *socketPath,
+		Net:  "unix",
+	}
+	conn, err := net.DialUnix("unix", nil, &addr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer conn.Close()
+
+	n, err := conn.Write([]byte{logDumpFollow})
+	if err != nil || n < 1 {
+		log.Fatalf("Failed to write request to memlogd socket: %v", err)
+	}
+
+	// map of service name to active log file
+	logs := make(map[string]*LogFile)
+
+	r := bufio.NewReader(conn)
+	for {
+		line, err := r.ReadString('\n')
+		if err == io.EOF {
+			return
+		}
+		if err != nil {
+			log.Fatalf("Failed to read from memlogd: %v", err)
+		}
+		msg, err := ParseLogMessage(line)
+		if err != nil {
+			log.Println(err)
+			continue
+		}
+		if strings.HasPrefix(msg.Name, "logwrite") {
+			// don't log our own output in a loop
+			continue
+		}
+
+		var logF *LogFile
+		var ok bool
+		if logF, ok = logs[msg.Name]; !ok {
+			logF, err = NewLogFile(*logDir, msg.Name)
+			if err != nil {
+				log.Printf("Failed to create log file %s: %v", msg.Name, err)
+				continue
+			}
+			logs[msg.Name] = logF
+		}
+		if err = logF.Write(msg); err != nil {
+			log.Printf("Failed to write to log file %s: %v", msg.Name, err)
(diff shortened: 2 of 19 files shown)


djs55:
@rn rebased (and made a minor update to a commit message and README)

pkg/kmsg/README.md
```diff
@@ -0,0 +1,2 @@
+A simple program to log the kernel log buffer
```
rn: Can you drop this. While this is a one liner, I prefer to have package documentation in `./docs`. Main reason is that if people change the docs it affects the git tree hash, requiring package rebuild for no code changes. There are some other packages which have `README.md`s but that seem the minority.

Or add a comment in `main.go` after `package main`

examples/logging.yml
```diff
@@ -24,6 +24,17 @@ services:
...
     image: alpine
     command: ["/bin/sh", "-c", "while /bin/true; do echo hello $(date); sleep 1; done" ]
+  - name: write-and-rotate-logs
+    image: linuxkit/logwrite:80b8619e0a2e9ffaa4ec314886b2606a839a094e
+    binds:
+      - /var/run:/var/run
+      - /var/log:/var/log
+  - name: kmsg
+    image: linuxkit/kmsg:862384637bd541d499b2f1356f71a1d304f79cb5
+    binds:
```
rn: can you add the binds and caps to the `build.yml` for `kmsg`. They are unlikely to change and keep the YAMLs cleaner. Look at `pkg/getty/build.yml` for examples.

examples/logging.yml
```diff
@@ -24,6 +24,17 @@ services:
   - name: write-to-the-logs
     image: alpine
     command: ["/bin/sh", "-c", "while /bin/true; do echo hello $(date); sleep 1; done" ]
+  - name: write-and-rotate-logs
+    image: linuxkit/logwrite:80b8619e0a2e9ffaa4ec314886b2606a839a094e
+    binds:
```
rn: You already specify the binds in `build.yml` so should not need them here.

pkg/logwrite/README.md
```diff
@@ -0,0 +1,5 @@
+A simple multi-file log writer
```
rn: See my comment above for `kmsg`

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	"time"
+)
+
+const (
+	logDump byte = iota
+	logFollow
+	logDumpFollow
+)
+
+// LogMessage is a single message sent to memlogd
```
rn: This comment seems imprecise. We *receive* a comma/semicolon separate string from memlogd and turn it into this struct.

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	Name    string    // name of the service that wrote the message
+	Message string    // body of the message
+}
+
+func (m *LogMessage) String() string {
+	return m.Time.Format(time.RFC3339) + " " + m.Name + " " + m.Message
+}
+
+// ParseLogMessage reconstructs a LogMessage from a line of text which
+// looks like:
```
rn: nit, keep this in one line (no line limit in go :)

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	s := m.String()
+	_, err := io.WriteString(l.File, s)
+	if err == nil {
+		l.BytesWritten += len(s)
+	}
+	return err
+}
+
+// Rotate closes the current log file, rotates the files and creates an
+// empty log file.
```
rn: Nit, comment on one line

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	l.File = f
+	l.BytesWritten = 0
+	return nil
+}
+
+func main() {
+	socketPath := flag.String("socket", "/var/run/memlogdq.sock", "memlogd log query socket")
+	logDir := flag.String("log-dir", "/var/log", "Directory containing log files")
+	maxLogFiles := flag.Int("max-log-files", 10, "Maximum number of rotated log files before deletion")
+	maxLogSize := flag.Int("max-log-size", 1048576, "Maximum size of a log file before rotation")
```
rn: Maybe express `1048576` as  an computation like `1024 * 1024` to make it clearer this 1 MB. Also say "Maximum size in Bytes..."

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	"log"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const (
+	logDump byte = iota
```
rn: Maybe add a comment that these need to be kept in  sync with `memlogd`

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	conn, err := net.DialUnix("unix", nil, &addr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer conn.Close()
+
+	n, err := conn.Write([]byte{logDumpFollow})
+	if err != nil || n < 1 {
+		log.Fatalf("Failed to write request to memlogd socket: %s", err)
```
rn: errors should be printed with `%v`

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+			// don't log our own output in a loop
+			continue
+		}
+
+		var logF *LogFile
+		var ok bool
+		if logF, ok = logs[msg.Name]; !ok {
+			logF, err = NewLogFile(*logDir, msg.Name)
+			if err != nil {
+				log.Fatalf("Failed to create log file %s: %s", msg.Name, err)
```
rn: This stops logging for all sources if *one* log file creation fails. Is this the failure mode we want?

pkg/logwrite/logwrite.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+			logs[msg.Name] = logF
+		}
+		if err = logF.Write(msg); err != nil {
+			log.Fatalf("Failed to write to log file %s: %s", msg.Name, err)
```
rn: Ditto. Maybe disable the log file by setting to `nil` on error?

rn:
@djs55 as mentioned I got the alpine:3.8 update going in https://github.com/linuxkit/linuxkit/pull/3098. I'd prefer holding off with this PR till that is merged as the update requires rebuilding everything for all arches. Rebasing this onto alpine 3.8 update should be painless.

rn:
#3098 is merged. Cold you also update the `linuxkit/alpine` hash to the latest?

djs55:
Thanks, will do tomorrow!

djs55:
@rn I've made all the updates and added 2 test cases. Let me know what you think!

djs55:
2 tests failed pulling `linuxkit/kmsg` and `linuxkit/logwrite` I think as expected.

test/cases/040_packages/030_logwrite/test.yml
```diff
@@ -0,0 +1,34 @@
...
+  - name: write-and-rotate-logs
+    image: linuxkit/logwrite:b43f735d365d39539013d1678d8cb87a8bceb1a5
+  - name: check-the-logs
+    image: alpine:3.7
+    binds:
+      - /check.sh:/check.sh
+      - /dev/console:/dev/console
+      - /var/log:/var/log
+    command: ["sh", "./check.sh"]
+  - name: poweroff
```
rn: Any reason you don't powerdown in `check.sh`? Otherwise the test always runs for 30 seconds even if it succeeds earleir

test/cases/040_packages/031_kmsg/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# SUMMARY: Check that the logwrite package works
```
rn: logwrite -> kmsg?

test/cases/040_packages/031_kmsg/test.sh
```diff
@@ -0,0 +1,24 @@
...
+# LABELS: amd64
```
rn: why only on `amd64`?

test/cases/040_packages/030_logwrite/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# SUMMARY: Check that the logwrite package works
+# LABELS: amd64
```
rn: Why only on `amd64`?

test/cases/040_packages/031_kmsg/test.sh
```diff
@@ -0,0 +1,24 @@
...
+# REPEAT:
+
+set -e
+
+# Source libraries. Uncomment if needed/defined
+#. "${RT_LIB}"
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+NAME=binfmt
```
rn: name shouldn't be `binfmt`

test/cases/040_packages/030_logwrite/test.sh
```diff
@@ -0,0 +1,24 @@
...
+# REPEAT:
+
+set -e
+
+# Source libraries. Uncomment if needed/defined
+#. "${RT_LIB}"
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+NAME=binfmt
```
rn: Name shouldn't be `binfmt`

test/cases/040_packages/031_kmsg/test.yml
```diff
@@ -0,0 +1,32 @@
...
+  - linuxkit/containerd:f2bc1bda1ab18146967fa1a149800aaf14bee81b
+  - linuxkit/ca-certificates:v0.4
+  - linuxkit/memlogd:883f0d46e7d3ae2d787e8acb496da115a4707cbc
+services:
+  - name: kmsg
+    image: linuxkit/kmsg:dd3fdc6eb26941f6a6c569807bc60176f8d32f48
+  - name: write-and-rotate-logs
+    image: linuxkit/logwrite:b43f735d365d39539013d1678d8cb87a8bceb1a5
+  - name: check-the-logs
+    image: alpine:3.7
```
rn: `alpine:3.7` -> `alpine:3.8`

rn:
```
linuxkit/logwrite:7859c102a963828fd9c5aa3837db9600483220c7
linuxkit/kmsg:3dfa0d5b4027ecc16089f27fbcffa15d6aa5438e
```
pushed and signed for all arches

djs55:
Thanks!

PR #3101, opened 2018-07-09
Enable building of bcc in kernel build

kmjohansen:
**- What I did**

This PR is an opening gambit for moving the bcc build out of projects and into the kernel build.

**- How I did it**

This leveraged the existing work done earlier by @rn.  The change collapses the dual stage build into a single stage that builds a bcc container after the kernel headers have been generated by an earlier stage of the kernel build.

**- How to verify it**

I've tested by building against all versions of the kernel supported by linuxkit.  Once built, I placed the bcc package into one of the init layers and then validated that I could run both Python and luaj scripts that are packaged from the bcc build.

Move bcc to the kernel build.


kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,103 @@
+ARG IMAGE
+FROM ${IMAGE} as ksrc
+
+FROM alpine:3.7 as build
+RUN apk update && apk upgrade -a && \
```
justincormack: We should use the `linuxkit/alpine` package here; we will need to add the extra packages that are currently missing.

projects/ebpf/ebpf.build/Dockerfile
```diff
@@ -37,8 +43,8 @@ RUN cat elfutils-$ELFUTILS_VERSION.tar.bz2 | tar xjf - && \
   patch -p1 < ../100-musl-compat.patch && \
   patch -p0 < ../decl.patch && \
   patch -p0 < ../intl.patch
-ENV BCC_COMMIT=d4fc95d92ec9bace9bd607dfd1833e9e06457486
+ENV BCC_COMMIT=6972806729da00ecda0235abac61d66c8fad7fad
```
rn: might be worth in a comment to say this is `v0.6.0` or better, use the release tag instead of the commit

projects/ebpf/ebpf.pkg/Makefile
```diff
@@ -1,9 +1,9 @@
-KERNEL_FILES=-C ../alpine/kernel/x86_64 kernel-headers.tar kernel-dev.tar kernel-modules.tar
+KERNEL_IMAGE=iad.ocir.io/odx-fn/linuxkit/kernel:4.14.48-ee964ca1fb8641370812be8ef4592b2708d3551b-amd64
```
rn: Any reason for changing this?

docs/kernel-bcc.md
```diff
@@ -0,0 +1,31 @@
...
+The `bcc` utility is a standard Linux tool to access performance
+counters, trace events and access various other kernel internals for
+performance analysis.
+
+The `bcc` utility needs to matched be with the kernel. For recent
+kernel build, LinuxKit provides a `linuxkit/kernel-bcc` package with
+a matching tag for each kernel under `linuxkit/kernel`.
+
+The preferred way of using the `linuxkit/kernel-bcc` package is to
+add it to the `init` section. This adds `/usr/bin/perf` to the
```
rn: `/usr/bin/perf`?

rn:
Thanks for picking this up. I've tried a couple of times a few months ago but had issues compiling it against musl and other bits. Once I had it compiling I also had issues running stuff. Have you tried some of the samples?

I made a few comments and also this needs to be build with `linuxkit/alpine` as Justin pointed out, but generally I'd be ok with carrying this forward. It's good to have these tools available!

kernel/Dockerfile.bcc
kmjohansen: I was unsure whether it was desirable to add larger packages like Python and Lua to the `linuxkit/alpine` base images.  However, I'm happy to do so if that's the preferred approach.

docs/kernel-bcc.md
kmjohansen: Thanks for catching.  That's a copy/pasta error.  Will correct.

kmjohansen:
Thanks for the review comments.

> Have you tried some of the samples?

I've been running the `hardirqs.py` and the `task_switch.lua` scripts as sanity checks but I haven't attempted to use any of my own scripts or complete a complicated kernel debugging task yet.  Were there specific scripts that had proved problematic in the past for you?

Unrelated: I noticed that some of the delta in the earlier commits in this change caused a bit of unnecessary confusion.  Is it preferable to collapse the change set into a single delta, or are multiple smaller commits the preferred style for this project?

kernel/Dockerfile.bcc
rn: In general we prefer repeatable builds. Adding python and lua to `tools/alpine` is fine. Maybe do this as a separate PR first as this allows me to build and push the image and then you can use that in a subsequent PR to add `bcc`

rn: Actually, even better, open an issue with the packages you need and I add them. Adding packages to `tools/alpine` is very manual and best done by a maintainer.

kmjohansen: Thanks, I've filed #3106 to cover the additional packages that weren't in the linuxkit alpine base image.

kmjohansen:
> Maybe it would be good to add a quick test under ./tests/cases/040_packages for one of the samples.

Sure, let me see if I can put something together.

kmjohansen:
Added a test case, but the test's image.yml depends upon the availability of the `kernel-bcc` package, which is not presently being built by default.  Is there a mechanism to trigger a specific portion of the kernel build so that the tests may consume it?  Presumably we'd need to do this if we were to ever write a test case for the `kernel-perf` image, too?

test/cases/040_packages/032_bcc/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# SUMMARY: Check that the logwrite package works
```
rn: copy and paste...

test/cases/040_packages/032_bcc/test.sh
```diff
@@ -0,0 +1,24 @@
...
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+NAME=bcc
+
+clean_up() {
+	rm -rf ${NAME}-*
+}
+trap clean_up EXIT
+
+# Test code goes here
+linuxkit build -disable-content-trust -format kernel+initrd -name "${NAME}" test.yml
```
rn: please don't disable content trust for tests. All packages should be signed once this is ready to get merged

test/cases/040_packages/032_bcc/test.yml
```diff
@@ -0,0 +1,29 @@
...
+  image: linuxkit/kernel:4.14.54
+  cmdline: "console=ttyS0 console=ttyAMA0"
+init:
+  - linuxkit/init:v0.5
+  - linuxkit/runc:v0.5
+  - linuxkit/containerd:v0.5
+  - linuxkit/ca-certificates:v0.5
+  - linuxkit/kernel-bcc:4.14.54
+services:
+# A service which generates logs of log messages
```
rn: copy and paste...

test/cases/040_packages/032_bcc/test.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: check-bcc
```
rn: this should be implementable as a `onboot` as most other tests. The `logwrite` test you copied this from need to run this as a service

test/cases/040_packages/032_bcc/check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+nsenter -t 1 -m -i -u -n -p --  /bin/ash -c "/usr/share/bcc/tools/softirqs 1 1"
+if [ "$?" -eq "0" ]; then
+	printf "bcc test suite PASSED\n" > /dev/console
+	/sbin/poweroff -f
```
rn: If you run this as a `onboot` you can add the `linuxkit/poweroff` package to handle poweroff.

kmjohansen:
Thanks for the additional comments on the tests.  I've re-written as `onboot` tests as you've requested.  I realize there's still a bit more to tidy up here, but I wanted to at least resolve the issues that you've raised so far.

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,102 @@
...
+    ln -s /usr/include/llvm5/llvm-c/ /usr/include/llvm-c; \
+    ln -s /usr/include/llvm5/llvm/ /usr/include/llvm
+
+WORKDIR /build
+COPY ./bcc.patches/ ./
+COPY Dockerfile.bcc /
+COPY ./bcc.patches/error.h /usr/include/
+COPY ./bcc.patches/cdefs.h /usr/include/sys/
+ENV ELFUTILS_VERSION=0.165
+RUN curl -sSL -O https://fedorahosted.org/releases/e/l/elfutils/0.165/elfutils-$ELFUTILS_VERSION.tar.bz2
```
rn: We should verify the downloaded file with some checksum, like a SHA256 similar to what we do with the wireguard code in `kernel/Dockerfile`

kernel/Makefile
```diff
@@ -192,6 +193,40 @@ push: push_perf_$(2)$(3)$(4)
...
+forcepush_bcc_$(2)$(3)$(4): forcebuild_bcc_$(2)$(3)$(4)
+	@if [ x"$(DIRTY)" != x ]; then echo "Your repository is not clean. Will not push image"; exit 1; fi
+	docker push $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4)-$(TAG)$(SUFFIX) && \
+	docker tag $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4)-$(TAG)$(SUFFIX) $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4)$(SUFFIX) && \
+	docker push $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4)$(SUFFIX) && \
+	$(PUSH_MANIFEST) $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4)-$(TAG) $(DOCKER_CONTENT_TRUST) && \
+	$(PUSH_MANIFEST) $(ORG)/$(IMAGE_BCC):$(1)$(3)$(4) $(DOCKER_CONTENT_TRUST)
+
+
+build: build_bcc_$(2)$(3)$(4)
```
rn: The compile fails on s390x. The compile also takes quite a while and our arm64 server.

 I'd be inclined to limit adding iovisor/bcc as a default option to x87 only and there only on the latest long term support kernel (currently 4.14.x) and the latest kernel (currently 4.17.x).

test/cases/040_packages/032_bcc/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# SUMMARY: Check that the bcc package works
+# LABELS:
```
rn: can you make this:
```
# LABEL: skip
```
for now. That will skip the test and I can enable it once the we have the BCC package build.

kmjohansen:
Ok, sorry it took me a little longer than usual to address these comments.  I did make the bcc build conditional on x86 and newer kernels, despite the fact that github's review software hasn't quite realized it.

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+  llvm \
+  llvm-dev \
+  llvm-static \
+  luajit-dev \
+  m4 \
+  python \
+  zlib-dev \
+  && true
+
+RUN ln -s /usr/lib/cmake/llvm5/ /usr/lib/cmake/llvm; \
```
rn: could you replace the `;` with `&&` to catch any errors?

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+    ln -s /usr/include/llvm5/llvm-c/ /usr/include/llvm-c; \
+    ln -s /usr/include/llvm5/llvm/ /usr/include/llvm
+
+WORKDIR /build
+COPY ./bcc.patches/ ./
+COPY Dockerfile.bcc /
```
rn: Why do you copy `Dockerfile.bcc` here?

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+COPY ./bcc.patches/error.h /usr/include/
```
rn: you already have `error.h` and `cdefs.h` in the build context. Why not copy them with `RUN cp` instead (or better move them with `mv`)?

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+COPY ./bcc.patches/cdefs.h /usr/include/sys/
+ENV ELFUTILS_VERSION=0.165
+ENV ELFUTILS_SHA256="a7fc9277192caaa5f30b47e8c0518dbcfd8c4a19c6493a63d511d804290ce972"
+RUN curl -sSL -O https://fedorahosted.org/releases/e/l/elfutils/0.165/elfutils-$ELFUTILS_VERSION.tar.bz2
+ 
+RUN echo "${ELFUTILS_SHA256}  /build/elfutils-$ELFUTILS_VERSION.tar.bz2" | sha256sum -c - && \
+  cat elfutils-$ELFUTILS_VERSION.tar.bz2 | tar xjf - && \
```
rn: Why `cat` if you can just `tar xjf` the file?

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+  make -C libelf && make -C libelf install
+RUN mkdir -p bcc/build && cd bcc/build && \
+  cmake .. -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_C_FLAGS="-I/build/usr/include" -DCMAKE_CXX_FLAGS="-I/build/usr/include" -DCMAKE_INSTALL_PREFIX=/usr -DLUAJIT_INCLUDE_DIR=/usr/include/luajit-2.1 && \
+  make && \
+  make install
+RUN mkdir -p /usr/local/share/lua/5.1/ && cd ljsyscall && cp -a *.lua syscall /usr/local/share/lua/5.1/
+
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/lib64
+
+FROM alpine:3.7 as mirror
```
rn: This should also use `linuxkit/alpine` as otherwise you will in unknown versions of the tools

kernel/Dockerfile.bcc
```diff
@@ -0,0 +1,105 @@
...
+COPY --from=build /build/usr/src /usr/src/
+COPY --from=build /build/usr/include /usr/include/
+COPY --from=build /usr/lib/libelf* /usr/lib/
+COPY --from=build /usr/lib/libstdc* /usr/lib/
+COPY --from=build /usr/lib/libintl* /usr/lib/
+COPY --from=build /usr/lib64 /usr/lib/
+COPY --from=build /usr/lib/python2.7/site-packages /usr/lib/python2.7/site-packages/
+COPY --from=build /usr/share/bcc /usr/share/bcc/
+COPY --from=build /usr/bin/bcc-lua /usr/bin/
+COPY --from=build /usr/local/share/lua /usr/local/share/lua/
```
rn: These many `COPY` commands create many layers in the final image. It might be better to build the toot file system in `/out` in the previous stage and then have a single `COPY` statement here.

justincormack:
I wonder if we should disable `CONFIG_GCC_PLUGIN_RANDSTRUCT` - it seems to potentially cause other issues eg making external modules difficult to build.

rn:
@justincormack external modules should compile/work fine with `CONFIG_GCC_PLUGIN_RANDSTRUCT`. See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/Kconfig#n529. We save the random seed in `kernel-dev.tar` and the external module build should pick this up.

PR #3102, opened 2018-07-10
Containerd v1.1.1

ijc:
Containerd release: https://github.com/containerd/containerd/releases/tag/v1.1.1


ijc:
Done. I made a mental note to do that yesterday and then promptly forgot.

PR #3109, opened 2018-07-11
Add '.log' as file extension to 'pkg/logwrite' files

rn:
Without this change, logfiles just use the log name provided by init, typically `<name>.err` and `name.out`. This PR adds `.log` as generic extension to all log files.

While at it, I also simplified the logwrite code a tiny bit and fixed up the test and example.

/cc @djs55 

olinguito

### DIFF
--- a/examples/logging.yml
+++ b/examples/logging.yml
@@ -22,7 +22,7 @@ services:
      - INSECURE=true
 # A service which generates log messages for testing
   - name: write-to-the-logs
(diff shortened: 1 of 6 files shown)


rn:
CI failed because I had a typo image name in the test YAMLs. I had that fixed locally but must have forgotten to push. Fixed now

djs55:
Looks good to me too! And the CI is green now...

Issue #3106, opened 2018-07-11
Additional packages requested for alpine base


kmjohansen:
During the review for #3101, @rn requested that I open an issue for the packages that are missing from `linuxkit/alpine` such that those could be used instead of an upstream alpine image.  The PR originally used the `alpine-3.7` base image, but it's preferable to use a `linuxkit/alpine` base instead.

After looking through the list of packages already in the `linuxkit/alpine` base image, it appears that the following are used in #3101 but not yet present in the linuxkit base:

- build-base
- clang
- clang-dev
- clang-static
- flex-dev
- fts-dev
- iperf
- libedit-dev
- llvm
- llvm-dev
- llvm-static
- luajit-dev
- python

rn:
These look reasonable to me. I probably won't have time to add them tomorrow but try to do so before the end of the week

rn:
@kmjohansen why is `iperf` needed?

kmjohansen:
@rn `iperf` is a test tool, and it's not strictly required.  I think we could omit it if need be.  (https://github.com/iovisor/bcc/blob/master/tests/python/CMakeLists.txt)

As far as Python goes, I can try building with Python 3 to see if it works.  A quick look at the bcc repo shows that they're building RH based releases against Python2, and Debian based releases against both Python 2 and 3.  I had mostly defaulted to Python 2 because building that way had worked for me in the past.

rn:
these were added as part of https://github.com/linuxkit/linuxkit/pull/3112
closing

PR #3112, opened 2018-07-12
Add packages needed for iovisor/bcc

rn:
This adds most of the packages needed to add the iovisor/bcc tools to LinuxKit as per #3106.

`python` and `iperf` were omitted as I don't think they are essential. Also `luajit-dev` is not avail on s390x so the build of bcc might need to take that into account (/cc @
kmjohansen).

The new package is: `linuxkit/alpine:cbf183a92cb4800c4129c2b9bb6e4d3eb3226330` and the size increased by about 140MB to about 600MB on x86.

image

### DIFF
--- a/tools/alpine/packages
+++ b/tools/alpine/packages
@@ -22,6 +22,9 @@ busybox-initscripts
 ca-certificates
 cdrkit
 cifs-utils
+clang
+clang-dev
+clang-static
 cmake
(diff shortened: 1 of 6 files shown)


justincormack:
Yeah there is no luajit port for s390x... are you sure python is not needed?

Issue #3111, opened 2018-07-11
Improve logging


rn:
Had a quick chat with @justincormack and @djs55 about further improvements to logging.

- Ideally all packages should log to stderr. This is kinda the default for go based programs and also seems to be encouraged by `systemd`.
- The init process should then pipe stderr for services to `<name>.log` (or register a logger with `memlogd` as `<name>` for stderr) and do the equiv for `onboot` containers as well.
- The init process should still log stdout to `<name>.out.log` (or register `<name>.out` with `memlogd`).

We should also make sure that scripts we execute as part of packages redirect to stderr.

ijc:
How broken would it be to launch these processes with fd 1 and 2 referencing the same thing? (ie. arranging for `dup(1, 2)` somewhere along the line)?

justincormack:
That might work. Most things are logging only to stderr, but some shell scripts for startup are not at present, so fixing is not a huge amount of work.

PR #3114, opened 2018-07-13
Write stderr to /var/log/<name>.log

djs55:
**- What I did**

- Ensured that `stderr` logs are written to `/var/log/<name>.log` (instead of `/var/log/<name>.err.log`). 
- Ensured that `pkg/kmsg` logs to `stderr`

See #3111

**- How I did it**

I renamed the `stderr` output to `<name>` (instead of `<name>.err`). The log file is therefore `<name>.log`.

**- How to verify it**

```
cd examples
linuxkit build -disable-content-trust logging.yml
linuxkit run logging
...
(ns: getty) linuxkit-025000000005:~# ls /var/log
getty.out.log              onboot.001-dhcpcd.log
kmsg.log                   write-to-the-logs.out.log
memlogd.log
```

With `memlogd` disabled (i.e. commented out)
```
(ns: getty) linuxkit-025000000005:~# ls /var/log
getty.log                      write-and-rotate-logs.log
getty.out.log                  write-and-rotate-logs.out.log
kmsg.log                       write-to-the-logs.log
kmsg.out.log                   write-to-the-logs.out.log
onboot
(ns: getty) linuxkit-025000000005:~# ls /var/log/onboot
onboot.000-sysctl.log      onboot.001-dhcpcd.log
onboot.000-sysctl.out.log  onboot.001-dhcpcd.out.log
```

Write `stderr` logs to `/var/log/<name>.log` instead of `/var/log/<name>.err.log`

Cute picture

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -33,7 +33,7 @@ kernel:
   image: linuxkit/kernel:4.9.91
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
-  - linuxkit/init:v0.5
+  - linuxkit/init:3f1e5c32d6ca9e83a7ea7ad9854da2032971a83f
   - linuxkit/runc:v0.5
   - linuxkit/containerd:v0.5
 services:
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:4.14.54
   cmdline: "console=ttyS0"
 init:
-  - linuxkit/init:v0.5
+  - linuxkit/init:3f1e5c32d6ca9e83a7ea7ad9854da2032971a83f
   - linuxkit/runc:v0.5
   - linuxkit/containerd:v0.5
   - linuxkit/ca-certificates:v0.5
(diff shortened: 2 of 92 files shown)


rn:
`init` and `kmsg` packages pushed

ijc:
```
(ns: getty) linuxkit-025000000005:~# ls /var/log
getty.out.log              onboot
kmsg.log                   onboot.001-dhcpcd.log
memlogd.log                write-to-the-logs.out.log
```

Having `onboot` and `onboot.*.log` there looks odd, maybe one of them is stale?

ijc:
That might have been from one of the previous batches of changes rather than here I guess?

rn:
The `onboot` is a symlink to `/run/log/onboot` and is created here: https://github.com/linuxkit/linuxkit/blob/master/pkg/init/cmd/service/runc.go#L162

Maybe we can make this conditional on the logger being registered?

rn:
Yeah, @ijc i vaguely recall that discussion regarding loosing onboot logs. I think we should only add the symlink if `init` writes to files. We don't need the symlink if `memlogd` is being used.

djs55:
OK, I've got a patch for the symlink -- I'll force push in a moment.

PR #3125, opened 2018-07-20
Enable 'linuxkit build' with private repositories and registries

rn:
This PR adds the ability to pull images from private repositories and registries during `linuxkit build`. It reuses the credentials provided with `docker login` and thus also utilises credential helpers such as the macOS keychain.

The actual code is quite small but required significant changes to the vendored packages. The first vendoring related commit updates all vendored packages to the same version as used by `docker/cli` version 18.06-ce. This required vendoring a whole bunch of new dependencies as well. The second vendoring related commit then adds the new packages needed to implement the authentication.

Also included is a minor tweak to use the `context` package (which wasn't available when this code was first written) and documentation updates.

resolves https://github.com/moby/tool/issues/202
resolves https://github.com/moby/tool/issues/104

stingrays


docs/yaml.md
```diff
@@ -10,6 +10,11 @@ The yaml configuration specifies the components used to build up an image . All
 are downloaded at build time to create an image. The image is self-contained and immutable,
 so it can be tested reliably for continuous delivery.
 
+Components are specified as Docker images which are pulled from registry during build if they
```
ijc: "from a registry" (or "the", but I think not)

rn:
great...more vendoring required, this time for windows only...

rn:
Update pushed

src/cmd/linuxkit/moby/docker.go
```diff
@@ -13,12 +14,23 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/reference"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/configfile"
+	distref "github.com/docker/distribution/reference"
```
ijc: AFAICT this is needed because the "github.com/containerd/containerd/reference" one above insists on the `docker.io` prefix being included, which we don't want to do in this context. That's a shame but I can't see a way to make this work with the containerd one as is.

rn: indeed. It's all a bit of a mess...

FWIW, the code which does `docker.io` -> `https://index...` is here: https://github.com/moby/moby/blob/master/registry/config.go#L405

Instead of faffing about with yet another type I replicated it below...which seems quite common

docs/yaml.md
rn: tahnsk. pushed a fix

justincormack:
Ugh what a lot of garbage vendoring, there is no way much of that stuff is actually needed!

rn:
Indeed. Even just the update of the vendored versions pulled in a bunch of crap, but that was necessary to pull in the vendored packages for the auth stuff which then pulled in even more crap...

PR #3126, opened 2018-07-20
memlogd/logwrite: use the same naming convention as init

djs55:
**- What I did**

Note there is a confusing name clash between the `pkg/logwrite` package which writes logs to disk and the command-line wrapper program `cmd/logwrite` (part of the `memlogd` package) modified here.

In e8786d73bb5665b4d548781949e318af6967be03 the logwrite package will automatically append .log to every log filename.

In 5201049f2ccb92d3933ab39f1a1a498c09dda669 the init package will send stderr of a service `s` to a log named `s` and the stdout to `s.out`. Therefore the files we create on disk are `s.log` and `s.out.log`.

This patch modifies the memlogd `logwrite` command-line wrapper to use the same convention.

**- How to verify it**

```
/ # ./pkg/memlogd/cmd/logwrite/logwrite -n hello ls /fooo
/ # ls -1 /var/log
...
hello.log
...
/ # cat /var/log/hello.log
2018-07-20T14:46:45Z hello ls: cannot access '/fooo': No such file or directory
```

The `logwrite` command-line wrapper now follows the same logging convention as the `init` package.

Kitten looking confused

### DIFF
--- a/examples/logging.yml
+++ b/examples/logging.yml
@@ -7,7 +7,7 @@ init:
   - linuxkit/runc:v0.5
   - linuxkit/containerd:0784cc754edb296b996c3510abbdf69686ef0f24
   - linuxkit/ca-certificates:v0.5
-  - linuxkit/memlogd:v0.5
+  - linuxkit/memlogd:ba4fcf55c35a3833546a1e39125bf0fa940327b0
 onboot:
   - name: sysctl
     image: linuxkit/sysctl:v0.5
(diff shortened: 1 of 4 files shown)


rn:
doh, sorry, I now rebuild the memlogd packages. Could you update to:
```
linuxkit/memlogd:ba4fcf55c35a3833546a1e39125bf0fa940327b0
```

Issue #2817, opened 2017-12-15
Passing config options for image build


rvs:
**Description**

Currently the kernel build section has a few hardcoded options for things like filer container, etc. It would be nice to have a configurable options that one can pass. The same mechanism could be used to potentially control things like GRUB, etc.

errordeveloper:
@ijc and I just discussed this in the context of Kubernetes packages, where we have `kubernetes_version` and `cni_version` already, and it would be convenient to make it easier to pass other options, e.g. repo URL for developers to test a version in a fork...

So a natural thing to do seems to map it to `docker build --build-args`, aka `ARG` keyword in Dockerfile, and set defaults in `build.yaml`. The main concern is how to allow a developer to easily build packages with different parameters without breaking the mapping between git hash in the image tag and the content of the image.

One option I can think would involve:

- adding `args` field to `build.yml` to set default values
- adding `-args` flag to `linuxkit pkg build` to override the defaults set in `build.yml`

Let's consider how this could work nicely.

Given we have `pkg/kubelet/build.yml` that looks like this:

```yaml
org: linuxkit
image: kubelet
network: true
arches:
  - amd64
config:
  build_args:
    kubernetes_version: v1.9.0
    kubernetes_repo_url: git@github.com:kubernetes/kubernetes
```

When `linuxkit pkg build ./pkg/kubelet` runs, it produces `linuxkit/kubelet:<git_hash>` as it does at present.

However, if I were to run the following command instead:
```
linuxkit pkg build  -args \
   kubernetes_version=myfeature,kubernetes_repo_url=git@github.com:errordeveloper/kubernetes \
   ./pkg/kubelet
```
It would produce `linuxkit/kubelet:<git_hash>-<args_hash>`, where `args_hash` is the hash of `-args` and is also stored as an image label.

If we need a more formal design doc, I'm happy to produce such a thing, but let's discuss it first :)

justincormack:
Yes, i discussed adding build args to build.yml with Roman and it seemed to
be a good solution, and doesnt break tree hash.

rvs:
+1 @errordeveloper and @justincormack

errordeveloper:
Should we have a more formal design doc for this?

rvs:
I can certainly help -- but someone else will have to take a stab at it first.

w9n:
If you do changes in `build.yml` the git subtree hash of the pkg dir changes so i dont think its necessary to introduce an args_hash.

[...]

PR #3127, opened 2018-07-22
Allow building with custom containerd repo

hairyhenderson:
**- What I did**

Trying to hack on containerd with LinuxKit I found that I could _almost_ do `CONTAINERD_REPO=... CONTAINERD_COMMIT=... make iid` (in `tools/alpine`) to build from a custom containerd repo. But not quite, so I made this possible 😉.

Note: There are a few similar places where it might be useful to be able to override repo/commit - not sure if it makes sense to also add those?

**- How I did it**

- replaced `git clone`'s reference to the containerd repo with a reference to `$CONTAINERD_REPO` (which was unreferenced before)
- changed `ENV` to `ARG` so the 2 variables could be set as `--build-arg`s
- added the appropriate `--build-arg` flags to the `docker build` command

**- How to verify it**

```console
$ cd tools/alpine; CONTAINERD_REPO=https://github.com/hairyhenderson/containerd.git CONTAINERD_COMMIT=master make iid
...
Step 21/39 : RUN mkdir -p $GOPATH/src/github.com/containerd &&   cd $GOPATH/src/github.com/containerd &&   git clone ${CONTAINERD_REPO} &&   cd $GOPATH/src/github.com/containerd/containerd &&   git checkout $CONTAINERD_COMMIT
 ---> Running in e4adacf8228c
Cloning into 'containerd'...
Already on 'master'
Your branch is up to date with 'origin/master'.
...
```

- Allow building with custom containerd repo

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -52,11 +52,11 @@ RUN go get -u github.com/LK4D4/vndr
 # checkout and compile containerd
 # Update `FROM` in `pkg/containerd/Dockerfile`, `pkg/init/Dockerfile` and
 # `test/pkg/containerd/Dockerfile` when changing this.
-ENV CONTAINERD_REPO=https://github.com/containerd/containerd.git
-ENV CONTAINERD_COMMIT=v1.1.2
+ARG CONTAINERD_REPO=https://github.com/containerd/containerd.git
+ARG CONTAINERD_COMMIT=v1.1.2
 RUN mkdir -p $GOPATH/src/github.com/containerd && \
   cd $GOPATH/src/github.com/containerd && \
-  git clone https://github.com/containerd/containerd.git && \
+  git clone ${CONTAINERD_REPO} && \
   cd $GOPATH/src/github.com/containerd/containerd && \
   git checkout $CONTAINERD_COMMIT
 RUN apk add --no-cache btrfs-progs-dev gcc libc-dev linux-headers make libseccomp-dev
--- a/tools/alpine/Makefile
+++ b/tools/alpine/Makefile
@@ -30,7 +30,11 @@ show-tag:
 	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions.$(ARCH)
 
 iid: Dockerfile Makefile $(DEPS)
-	docker build --no-cache --iidfile iid .
+	docker build --no-cache \
+		--iidfile iid \
+		--build-arg CONTAINERD_REPO \
+		--build-arg CONTAINERD_COMMIT \
+		.
 
 hash: Makefile iid
 	docker run --rm $(shell cat iid) sh -c 'echo Dockerfile /lib/apk/db/installed $$(find /mirror -name '*.apk' -type f) $$(find /go/bin -type f) | xargs cat | sha1sum' | sed 's/ .*//' | sed 's/$$/$(SUFFIX)/'> $@


rn:
I'm not sure we  want to support building packages build on something not version controlled. Maybe I don't quite understand your use case, but is there a reason you can't just edit the Docker file and then build normally? For development, maybe we could make an override for the hash.

hairyhenderson:
> is there a reason you can't just edit the Docker file and then build normally

not at all - this is really just a convenience thing. But it turns out the convenience is minimal without also being able to override the `linuxkit/alpine` tag...

> For development, maybe we could make an override for the hash.

Yeah, my first thought was:

```Dockerfile
ARG ALPINE_IMG=linuxkit/alpine:8a89682421abf0d886d777235a05f4a04a736df2
FROM $ALPINE_IMG as alpine
...
```

But `linuxkit pkg build` doesn't support the `--build-arg` flag, so that's pretty much pointless.

Oh well - clearly I didn't think this one through 😉

rn:
for `linuxkit pkg build` there is `-dev` which overrides the org and the tag. We could add something similar to the `tools/alpine` `Makefile` for local development. This would allow you to not having to override the `linuxkit/alpine` tag, for example. You just have to edit, say the `pkg/containerd/Dockerfile` once

ijc:
It's been a while since I needed/used it but FWIW here is my skanky script for making an overlay (add to your yaml `init` section after the normal one) "dev" containerd image from a normal containerd source checkout, which (IMHO) is most convenient when debugging/hacking on containerd itself:

```shell
#!/bin/bash
set -e

src=$GOPATH/src/github.com/containerd/containerd
lk=/home/ijc/development/Docker/LinuxKit/gocode/src/github.com/linuxkit/linuxkit

make -C ${src} binaries EXTRA_FLAGS="-buildmode pie" EXTRA_LDFLAGS="-extldflags \\\"-fno-PIC -static\\\""

root="/home/ijc/development/Docker/LinuxKit/make-containerd-dev.tmp"
rm -rf "$root"
mkdir -p "$root"

for bin in containerd containerd-shim ctr ; do
    bins="$bins ${src}/bin/$bin"
done

mkdir -p $root/usr/bin
for bin in $bins ; do
    cp "$bin" "$root/usr/bin/"
done

if true ; then
    # etc should be provided by the base containerd, which is needed for the service command, but sometimes its nice to override things.
    cp -r $lk/pkg/containerd/etc $root/etc
    etc=etc
fi

# service comes from init now.
mkdir -p $root/etc/init.d && ln -s /usr/bin/service $root/etc/init.d/020-containerd

tar --owner 0:0 -C $root -cf $root/rootfs.tar $etc usr

cat > $root/Dockerfile <<EOF
FROM scratch
ADD rootfs.tar /
ENTRYPOINT []
WORKDIR /
EOF

docker build -t ijc/containerd:dev "$root"
```

hairyhenderson:
Thanks @ijc - that's helpful! I'll pore over that and see what I can glean for my own uses.

I think ultimately it'd be useful for `linuxkit pkg build` to allow passing through arbitrary `docker build` arguments, or at the very least, `--build-arg`. Or maybe that's counter to the goals of the project?

rn:
we deliberately don't want to do that to allow building with something else in the future, e.g. containerd + build kit. So trying to keep it simple for now

hairyhenderson:
thanks @rn - makes sense!

ijc:
The other reason to avoid build args is because that introduces a way to have two builds with different contents but the same hash (since the hash is the git tree hash). It was proposed (in https://github.com/linuxkit/linuxkit/issues/2817#issuecomment-368735626) to support them but have them unconditionally mark the image `-dirty`, which probably makes sense.

Also containerd is a bit of a weird special case due to the fact we ended up putting the source into the alpine mirror, rather than cloning in `pkg/containerd`, which makes "just use this other one for testing" far more complicated than it would need to be for any other package since there's a chain of packages to build.

hairyhenderson:
> we ended up putting the source into the alpine mirror

I found that a bit odd TBH - is that just because it's referenced in both `pkg/containerd` and `pkg/init`? or is there another reason?

ijc:
> is that just because it's referenced in both `pkg/containerd` and `pkg/init`? 

Mostly (possibly entirely, I can't think of/remember another reason right now). There's also a use in `test/pkg/containerd/`.

PR #3136, opened 2018-07-25
Allow external directories for 'linuxkit pkg build'

### LINKED ISSUE #3108: Improve linuxkit pkg build to be able to use external sources
Currently, `linuxkit pkg build` assumes that all the sources for a package are in the package directory. The hash for the package is the git tree hash of the package directory. In some cases this leads to duplication of code, e.g., for go vendoring.

In a conversation with @ijc and @justincormack we discussed to potential extension to `linuxkit pkg build` to also consider external sources. Here is a more detailed proposal:

- Add a new `sources` entry to `build.yml` which takes a list of `<src>:<dst>` strings, where `<src>` and `<dst>` are directories and `<src>` should be under git source control.
- During build the <src> directory is copied to <dst> in the build context where it will be available to the `Dockerfile`.
- To compute the package hash `linuxkit pkg` will use the git tree hash of the package directory and XOR it with the git tree hash of all the `<src>` directories.

rn:
This adds adds a new `sources` field to the `build.yml` where one can specify additional source directories for the package build outside the package directory. This is useful for sharing files, such as vendored go code, between packages.

If `sources` are specified, the package hash is calculated as the `sha1` of the strings of the git tree hashes of all the externals sources and the package directory. Therefore, `sources` need to be under git revision control.

`sources` are specified as a `src:dst` pair, where `src` can be a relative path the package directory and `dst` is the destination with the docker build context.

The `sources` features is implemented by `tar`ing up (and rewriting the paths) of the sources and pipe them to `stdin` of the `docker` process used to build the package.

There should be no changes for packages not using the `sources` features.

resolves #3108

whales

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -37,6 +37,7 @@ A package source consists of a directory containing at least two files:
 - `image` _(string)_: *(mandatory)* The name of the image to build
 - `org` _(string)_: The hub/registry organisation to which this package belongs
 - `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
+- `extra-sources` _(list of strings)_: Additional sources for the package outside the package directory. The format is `src:dst`, where `src` can be relative to the package directory and `dst` is the destination in the build context. This is useful for sharing files, such as vendored go code, between packages.
 - `gitrepo` _(string)_: The git repository where the package source is kept.
 - `network` _(bool)_: Allow network access during the package build (default: no)
 - `disable-content-trust` _(bool)_: Disable Docker content trust for this package (default: no)
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -1,12 +1,18 @@
 package pkglib
 
 import (
+	"archive/tar"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
+	"path"
+	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/version"
+	log "github.com/sirupsen/logrus"
 )
 
 type buildOpts struct {
@@ -141,6 +147,8 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 		args = append(args, "--label=org.mobyproject.linuxkit.version="+version.Version)
 		args = append(args, "--label=org.mobyproject.linuxkit.revision="+version.GitCommit)
 
+		d.ctx = &buildCtx{sources: p.sources}
+
 		if err := d.build(p.Tag()+suffix, p.path, args...); err != nil {
 			return err
 		}
@@ -188,6 +196,62 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 	}
 
 	fmt.Printf("Build, push and release of %q complete, all done.\n", bo.release)
+
+	return nil
+}
+
+type buildCtx struct {
+	sources []pkgSource
+}
+
+// Copy iterates over the sources, tars up the content after rewriting the paths.
+// It assumes that sources is sane, ie is well formed and the first part is an absolute path
+// and that it exists. NewFromCLI() ensures that.
+func (c *buildCtx) Copy(w io.WriteCloser) error {
+	tw := tar.NewWriter(w)
+	defer func() {
+		tw.Close()
+		w.Close()
+	}()
+
+	for _, s := range c.sources {
+		log.Debugf("Adding to build context: %s -> %s", s.src, s.dst)
+
+		f := func(p string, i os.FileInfo, err error) error {
+			if err != nil {
+				return fmt.Errorf("ctx: Walk error on %s: %v", p, err)
+			}
+
+			h, err := tar.FileInfoHeader(i, "")
+			if err != nil {
+				return fmt.Errorf("ctx: Converting FileInfo for %s: %v", p, err)
+			}
+			h.Name = path.Join(s.dst, strings.TrimPrefix(p, s.src))
+			if err := tw.WriteHeader(h); err != nil {
+				return fmt.Errorf("ctx: Writing header for %s: %v", p, err)
+			}
+
+			if !i.Mode().IsRegular() {
+				return nil
+			}
+
+			f, err := os.Open(p)
+			if err != nil {
+				return fmt.Errorf("ctx: Open %s: %v", p, err)
+			}
+			defer f.Close()
+
+			_, err = io.Copy(tw, f)
+			if err != nil {
+				return fmt.Errorf("ctx: Writing %s: %v", p, err)
+			}
+			return nil
+		}
+
+		if err := filepath.Walk(s.src, f); err != nil {
+			return err
+		}
+	}
 
 	return nil
 }
--- a/src/cmd/linuxkit/pkglib/pkglib.go
+++ b/src/cmd/linuxkit/pkglib/pkglib.go
@@ -1,11 +1,13 @@
 package pkglib
 
 import (
+	"crypto/sha1"
 	"flag"
 	"fmt"
 	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -177,6 +179,37 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 		}
 	})
 
+	var srcHashes string
+	sources := []pkgSource{{src: pkgPath, dst: "/"}}
+
+	for _, source := range pi.ExtraSources {
+		tmp := strings.Split(source, ":")
+		if len(tmp) != 2 {
+			return Pkg{}, fmt.Errorf("Bad source format in %s", source)
+		}
+		srcPath := filepath.Clean(tmp[0]) // Should work with windows paths
+		dstPath := path.Clean(tmp[1])     // 'path' here because this should be a Unix path
+
+		if !filepath.IsAbs(srcPath) {
+			srcPath = filepath.Join(pkgPath, srcPath)
+		}
+
+		g, err := newGit(srcPath)
+		if err != nil {
+			return Pkg{}, err
+		}
+		if g == nil {
+			return Pkg{}, fmt.Errorf("Source %s not in a git repository", srcPath)
+		}
+		h, err := g.treeHash(srcPath, hashCommit)
+		if err != nil {
+			return Pkg{}, err
+		}
+
+		srcHashes += h
+		sources = append(sources, pkgSource{src: srcPath, dst: dstPath})
+	}
+
 	git, err := newGit(pkgPath)
 	if err != nil {
 		return Pkg{}, err
@@ -195,6 +228,11 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 				return Pkg{}, err
 			}
 
+			if srcHashes != "" {
+				hash += srcHashes
+				hash = fmt.Sprintf("%x", sha1.Sum([]byte(hash)))
+			}
+
 			if dirty {
 				hash += "-dirty"
 			}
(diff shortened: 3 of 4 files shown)


rn:
@justincormack thanks for the review. I'll wait till tomorrow for @ijc eyes. This is mostly his code I modified

docs/packages.md
```diff
@@ -37,6 +37,7 @@ A package source consists of a directory containing at least two files:
 - `image` _(string)_: *(mandatory)* The name of the image to build
 - `org` _(string)_: The hub/registry organisation to which this package belongs
 - `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
+- `sources` _(list of strings)_: Optionally specifies sources for the package build which are outside the package directory. The format is `src:dst`, where `src` can be a relative path to the package directory and `dst` is the destination in the build context. This is useful for sharing files, such as vendored go code, between packages.
```
ijc: I think `extra-sources` would be a more accurate name.

> where `src` can be a relative path to the package directory

"can" or "must"? I would also word it as "path relative to the".

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -191,3 +201,62 @@ func (p Pkg) Build(bos ...BuildOpt) error {
...
+			}
+
+			f, err := os.Open(p)
+			if err != nil {
+				log.Warnf("ctx: Open %s: %v", p, err)
+				return nil
+			}
+			defer f.Close()
+
+			h, err := tar.FileInfoHeader(i, "")
```
ijc: In `linuxkit build` we force `PAX` headers, IIRC for reasons to do with path length limitations. Should we do the same here?

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -191,3 +201,62 @@ func (p Pkg) Build(bos ...BuildOpt) error {
...
+	for _, source := range c.sources {
+		tmp := strings.Split(source, ":")
+		src := tmp[0]
+		dst := tmp[1]
+		log.Debugf("Adding to build context: %s -> %s", src, dst)
+
+		f := func(p string, i os.FileInfo, err error) error {
+			if err != nil {
+				log.Warnf("ctx: Walk error on %s: %v", p, err)
+				return nil
```
ijc: This function appears to warn and eat all the errors -- should it not return the error and fail the build?

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +172,43 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+		// If we have external sources ensure the package path is included first
+		sources = []string{pkgPath + ":/"}
+	}
+	for _, source := range pi.Sources {
+		tmp := strings.Split(source, ":")
+		if len(tmp) != 2 {
+			return Pkg{}, fmt.Errorf("Bad source format in %s", source)
+		}
+
+		// Make sure the srcPath
```
ijc: missing the end of the sentence?

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +172,43 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 		}
 	})
 
+	var srcHashes string
+	var sources []string
+	if len(pi.Sources) != 0 {
+		// If we have external sources ensure the package path is included first
+		sources = []string{pkgPath + ":/"}
+	}
```
ijc: If we made it so that sources always contains at least the pkgPath (and nothing else if `pi.Sources` is unused) then wouldn't that remove a bunch of conditionality elsewhere -- e.g. the building bit would then always pass the context on stdin even in the single source case.

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +172,43 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+	for _, source := range pi.Sources {
+		tmp := strings.Split(source, ":")
+		if len(tmp) != 2 {
+			return Pkg{}, fmt.Errorf("Bad source format in %s", source)
+		}
+
+		// Make sure the srcPath
+		srcPath := tmp[0]
+		if !filepath.IsAbs(srcPath) {
+			srcPath, err = filepath.Abs(filepath.Join(pkgPath, srcPath))
```
ijc: `filepath.Join(pkgPath, srcPath)` is already absolute (and `.Clean`), so `filepath.Abs` here does nothing.

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -191,3 +201,62 @@ func (p Pkg) Build(bos ...BuildOpt) error {
...
+			}
+
+			if i.IsDir() {
+				return nil
```
ijc: I guess the untar on the other end automatically creates missing dirs, but aren't you throwing away ownership and perms etc? Is there any harm in outputting the directories too?

I'm not entirely sure that the remaining code handles symlinks and the like ok, it seems to either collapse them to files or output a symlink header followed by some actual data, not sure what the result is.

src/cmd/linuxkit/pkglib/build.go
rn: There does not seem to be a way to return an error all the way back to the pipe and no way yo handle the error. I could change this to log.Fatal()` or similar to exit

src/cmd/linuxkit/pkglib/pkglib.go
rn: Almost certainly. I contemplated it but initially wanted to add this feature but leave the other code path untouched, but maybe now it's better to always use the build context

src/cmd/linuxkit/pkglib/build.go
rn: Let me try to refactor this a little bit. I got an error when hitting directories, but I think I now know a better way to handle this

src/cmd/linuxkit/pkglib/build.go
ijc: Returning an error in this function propagates to the error return of the `filepath.Walk`, you just need to handle it there instead of ignoring it. I think.

Maybe you need to propagate it further, like back the `if dr.ctx != nil {` block goroutine, an `errgroup` there would save you messing around with channels manually.

src/cmd/linuxkit/pkglib/pkglib.go
ijc: I think it is necessary to leave the behaviour the same, but changing the internals as needed to make the new stuff fit cleanly is the right thing to do.

src/cmd/linuxkit/pkglib/build.go
rn: turns out `dockerd` doesn't like it:
```
Sending build context to Docker daemon   16.9kB
Error response from daemon: Dockerfile parse error line 1: unknown instruction: /PAXHEADERS.00000000000000000000000000000013200000000000010217
exit status 1
```

ijc: Interesting!

I guess the tarfile is all relative to the root of the build context, so it won't matter if `cwd` is super long and we only worry about very deep paths within the actual context itself (which is less likely).

rn:
@ijc thanks for the review. I think I have addressed all the comments:
- `sources` -> `extra-sources`
- Always use the build context. Makes the code a little simpler
- Propagate errors from the build context `Copy()` function and then use an `errgroup`
- plus the minor bits and pieces

docs/packages.md
```diff
@@ -37,6 +37,7 @@ A package source consists of a directory containing at least two files:
 - `image` _(string)_: *(mandatory)* The name of the image to build
 - `org` _(string)_: The hub/registry organisation to which this package belongs
 - `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
+- `extra-sources` _(list of strings)_: Additional sources for the package outside the package directory. The format is `src:dst`, where `src` can be a relative to the package directory and `dst` is the destination in the build context. This is useful for sharing files, such as vendored go code, between packages.
```
ijc: "can be a relative" → "can be relative".

Is there really a usecase for absolute paths here?

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -191,3 +199,62 @@ func (p Pkg) Build(bos ...BuildOpt) error {
...
+		f := func(p string, i os.FileInfo, err error) error {
+			if err != nil {
+				return fmt.Errorf("ctx: Walk error on %s: %v", p, err)
+			}
+
+			h, err := tar.FileInfoHeader(i, "")
+			if err != nil {
+				return fmt.Errorf("ctx: Converting FileInfo for %s: %v", p, err)
+			}
+			h.Name = path.Join(dst, strings.TrimPrefix(p, src))
```
ijc: `p` at this point has, I think, been [`Clean`ed](https://golang.org/pkg/path/filepath/#Clean) relative to the original `src`.

If `src` was unclean, perhaps having `//` or "/../" in it or similar then `TrimPrefix` won't DTRT. It looks like [`filepath.Rel`](https://golang.org/pkg/path/filepath/#Rel) might do exactly what you want. It might be sufficient to `filepath.Clean` both `src` and `dst` yourself, but I think `filepath.Rel` is probably better.

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +172,36 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 		}
 	})
 
+	var srcHashes string
+	sources := []string{pkgPath + ":/"}
+
+	for _, source := range pi.ExtraSources {
+		tmp := strings.Split(source, ":")
```
ijc: You do this splitting and what have you in two places, you could consider making sources be a `[]struct Foo {src, dst string}` perhaps and accumulating it once here.

docs/packages.md
rn: it doesn't matter. In most cases it will be relative but it is not enforced to be an absolute path.

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +178,36 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+	var srcHashes string
+	sources := []pkgSource{pkgSource{src: pkgPath, dst: "/"}}
+
+	for _, source := range pi.ExtraSources {
+		tmp := strings.Split(source, ":")
+		if len(tmp) != 2 {
+			return Pkg{}, fmt.Errorf("Bad source format in %s", source)
+		}
+
+		srcPath := tmp[0]
```
ijc: An absolute `srcPath` is never cleaned AFAICT (a relative one gets it implicitly in `filepath.Join`).

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +178,36 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
 		}
 	})
 
+	var srcHashes string
+	sources := []pkgSource{pkgSource{src: pkgPath, dst: "/"}}
```
ijc: I think `gofmt` is going to want to make this  `[]pkgSource{{src: pkgPath, dst: "/"}}` which is more concise (the inner type is implicit)

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -169,6 +178,36 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+		if g == nil {
+			return Pkg{}, fmt.Errorf("Source %s not in a git repository", srcPath)
+		}
+		h, err := g.treeHash(srcPath, hashCommit)
+		if err != nil {
+			return Pkg{}, err
+		}
+
+		srcHashes += h
+		sources = append(sources, pkgSource{src: srcPath, dst: tmp[1]})
```
ijc: `dst` is also never `Clean`d (which may or may not matter).

rn:
Now cleaning paths and fix gofmt issue

PR #3149, opened 2018-08-02
update -rt to 4.14.59-rt37

TiejunChina:
Update rt stuff with the latest -rt release.


TiejunChina:
Any comments?

TiejunChina:
@rn NP. Rolf, looks you have already left from Docker, right?

PR #3150, opened 2018-08-02
Update the vendoring of virtsock

djs55:
In addition to bug fixes, this removes the special protocol used for `shutdown` needed by old Windows builds < 14393-- this is of particular interest to `vsudd` which is used to proxy arbitrary socket connections.

Update the version of `virtsock` used by `vsudd` to the latest version, removing the old workaround for lack of `shutdown` on Windows builds < 14393.

Sunbathing pig


djs55:
@rn thanks! I've pushed the yaml updates and the CI is green.

djs55:
@rn thanks again!

PR #3160, opened 2018-08-15
Improve docker container startup times on 4.14, 4.17, and 4.18

kmjohansen:
**- What I did**

This change ensures that docker container startup times are in the 390-450ms range.  Previously, I was seeing > 2000ms times on 4.14, and ~1260ms times on 4.17.  The timing was observed by using the example dind build, but adding a shell container so that it's possible to login and run docker commands in the dind container.  After ensuring that a hello-world image has been pulled, the test measured the total time for a 'docker run hello-world' to complete.

**- How I did it**

I used bcc to drill into this problem.  On 4.14, all of the pernet init functions happen serially and under a single lock.  This has been refactored since.  The predominant area where we're spending lots of time blocking in the docker run case is in network namespace creation.  Most of that time is in setup_net().  Having to call per-net initializers for all of the subsystems registered with networking is time consuming.  By switching to modules, we modprobe in just what we need and are able to save about 750ms on a docker run.

The second observation was that we were spending a large amount of time in the rcu_wait_gp and rcu_barrier code in paths that were also part of network namespace creation.  These times remained long even after module conversion.  By increasing the value of HZ, the periodic timer checks more frequently to see if CPUs have reached a quiescent rcu state, which can decrease the amount of time we spend waiting for grace periods and barriers.  This provides a substantial performance increase by removing more than 800ms of time spent waiting for docker run to complete on both 4.14 and 4.17. 

**- How to verify it**

From a docker/dind linuxkit image,  do a nsenter(1) to enter the docker namespaces.  Pull the hello-world image, and then do a 'time docker run hello-world'.  Before this change the times would be in the 2000ms range on 4.14 and in the 1260ms range on 4.17.  Afterwards, they're in the 390-450ms range.

Improve docker container startup times on 4.14 and 4.17.


rn:
Thanks for the analysis and the patches!

Do you have an idea if this also affects arm64 and s390x (I would suspect it does) so we probably should also add these changes to the respective kernel configs. You could check the configs with `make ARCH=arm64 oldconfig` and `make ARCH=s390 oldconfig`.

The other slight concern I have with modularising the network subsystem is that currently the LCOW build drops all the kernel modules as they are not needed. If we  modularising the network subsystem we'd have to pull in all the kernel module making the LCOW initrd quite a big bigger (6MB -> 22MB), but I guess that would be offset by the faster container start up times.

kmjohansen:
I didn't test on ARM or S390, mostly due to lack of access to hardware.  AIUI, HZ is valid for all architectures.

I actually suspect that there's a bug in one of the network subsystems that is exacerbating this problem, and converting everything to modules is a workaround until I (or somebody else) can track down the culprit.  Note that on 4.9 and 4.17, the container launch times are in the 1200ms range, so globally applying HZ, and only converting modules on 4.14 is one possibility.

OTOH, if switching everything to using modules is particularly obnoxious, it might be possible to see if the issues in 4.14 can be fixed by backporting the appropriate patches.  (Means that we'll have to figure out exactly what in the setup_net() path is taking all that time.  Although, with the HZ changed, it means we only need to recover 120ms or so to reach parity with the other releases.)

rn:
>  Although, with the HZ changed, it means we only need to recover 120ms or so to reach parity with the other releases.)

Could you clarify this statement? Are you saying that of we only increase HZ to 1000 but not covert to use network modules even for 4.14 we'd see a significant increase and converting to modules on top of the HZ change will "only" get us an extra 120ms faster?

kmjohansen:
Sorry I wasn't clear about this.  I attempted to call out the individual improvement of each patch against baseline, but may not have gotten all of that from the commit messages into the PR's description.

In the case of 4.14, just applying the HZ patch takes us from 2000ms -> 570ms, and just applying the modules patch takes us from 2000ms -> 1260ms.  If you combine both, then you can go from 2000ms -> 450ms.  On 4.17, we need just the HZ patch, and that takes us from 1260ms -> 390ms.

kmjohansen:
It may be worth testing the performance of arm64 and s390 to see how they fare compared to x86.  I had a colleague anecdotally report acceptable performance using 300Hz instead of 1000Hz.  I'm re-spinning a 4.17 kernel with that clock rate instead to see how it compares to the 1000Hz kernel.

kmjohansen:
After re-testing 4.17 with 300Hz, I'm seeing container start times in the 1200ms range.  I'm going to leave this at 1000Hz for now, absent a better suggestion.

kmjohansen:
Looking at just the per-net ops_init() calls that take more than 1ms, I get the following list:
```
__ip_vs_ftp_init 1.205770
clusterip_net_init 8.582198
frame_filter_net_init 23.559237
frame_nat_net_init 23.989305
__ip_vs_init 99.905918
```
So it's likely that the majority of the savings we're getting from modularization are the result of the ipvs and frame init calls being omitted.  If the mass modularization is not desirable, I can experiment with a more targeted change here.  I haven't had a chance to dig into why these are slower on 4.14 than other versions, though.

kmjohansen:
Confirmed that turning just those 5 subsystems into modules, and combining them with the HZ fix also gives the desired performance on 4.14, at least for x86-64.  Would that be an acceptable compromise solution, or is a different approach preferred here?

kmjohansen:
Swapped the kmod commit for one that only turns the slow subsystems into modules.  Let me know if this seems more tractable.

rn:
Thanks for looking more into this. In terms of picking which network subsystem to modularise, instead of picking very specific modules from a sub system (like `CONFIG_BRIDGE_EBT_T_NAT` or `CONFIG_IP_VS_FTP`) I think it would be cleaner to then turn the entire subsystem into modules.

So basically turn all of IPVS and EBT into modules. For `CONFIG_IP_NF_TARGET_CLUSTERIP` I think it's fine to just have that as a module but not all of `CONFIG_IP_NF`. I think this config would allow us to still run LCOW without modules.

It would also be good to apply the same changes to the arm64 and s390x config to keep them in sync.

@ijc @justincormack any objections to the changes, in particular the HZ change? It might increase the CPU consumption a little with HyperKit as there will be more interrupts...

ijc:
We have CONFIG_NO_HZ_IDLE which turns off the tick when idle which I think is the more important thing wrt CPU consumption, so fine by me in that respect.

Although, since RCU is the motivation for this change have we checked if any of the other RCU settings/modes might be a worthwhile improvement without switching to such a fast clock?

(this all has the slight whiff of working around a kernel bug, has this been reported upstream?)

kmjohansen:
@rn Sounds reasonable.  Would you like me to make the HZ changes to the older kernels too? (<= 4.9)  Once we get that sorted, I'll work up another version that contains the changes we've discussed.

@ijc I played around with NO_HZ_FULL, but didn't find that it made much of a difference in terms of the performance that we observed for this particular benchmark.  There are some other RCU Kconfig options available in expert mode, but they seem to mostly pertain to when RCU will agree to enable dyntick mode.

I haven't reported this as an issue upstream, mostly because I hadn't managed to convince myself that this was a kernel bug.  The documentation for the HZ values in Kconfig, mentions that a higher HZ rate leads to lower system latencies, though the connection between HZ and RCU isn't explicitly spelled out.  The documentation suggests that lower clock rates are better for throughput based workloads, but that ignores the case here, where we actually get better throughput and latency when RCU grace periods occur more frequently.

The symptom that I observed was that we had a bunch of docker threads waiting in the rcu code for long periods of time:
```
    finish_task_switch
    schedule
    schedule_timeout
    __wait_for_common
    wait_for_completion
    __wait_rcu_gp
    synchronize_sched
    dev_change_name
    do_setlink
    rtnl_setlink
    rtnetlink_rcv_msg
    netlink_rcv_skb
    rtnetlink_rcv
    netlink_unicast
    netlink_sendmsg
    sock_sendmsg
    SYSC_sendto
    sys_sendto
    do_syscall_64
    entry_SYSCALL_64_after_hwframe
    -                dockerd (7162)
        162558

    finish_task_switch
    schedule
    schedule_timeout
    __wait_for_common
    wait_for_completion
    _rcu_barrier
    rcu_barrier_sched
    rcu_barrier
    dev_change_net_namespace
    do_setlink
    rtnl_setlink
    rtnetlink_rcv_msg
    netlink_rcv_skb
    rtnetlink_rcv
    netlink_unicast
    netlink_sendmsg
    sock_sendmsg
    SYSC_sendto
    sys_sendto
    do_syscall_64
    entry_SYSCALL_64_after_hwframe
    -                dockerd (7162)
        214721
```
So in the above example, the rtnl_setlink being done as part of docker network setup is blocking for 162ms in a `__wait_rcu_gp()` and 214ms in a `rcu_barrier()`.  A RCU quiescent state must occur before a grace period can advance.  The `rcu_barrier()` code is checking to ensure that all queued RCU callbacks are completed, which also waits for a grace period if any were in fact queued.

The code that notes quiescent states and pokes the RCU core processing is invoked from `rcu_check_callbacks()`, which is called from `update_process_times()`.  Quiescent states may also be noted on context switch.  The update_process_times function is called by the scheduler tick handler and the periodic tick handler.  If we're invoking the periodic tick more frequently, then the kernel is actually generating quiescent states, and running the rcu core processing more frequently.  This is why we see better times. Though, it is a tradeoff.  The kernel is performing periodic work more frequently and taking more interrupts (at a factor of 10x), but it allows us to get a 3-5x improvement on container launch time.

kmjohansen:
Forgot to mention in my last post that there's some decent documentation about how/when RCU checks for grace periods based upon the tick mode enabled in the kernel.  That's here:

https://www.kernel.org/doc/Documentation/RCU/Design/Requirements/Requirements.html#Scheduling-Clock%20Interrupts%20and%20RCU

ijc:
> I played around with NO_HZ_FULL, but didn't find that it made much of a difference in terms of the performance that we observed for this particular benchmark. 

The purpose of `NO_HZ_*` is to mitigate the performance overheads (esp on virtualised platforms) from having a higher HZ value, rather than something you would expect to effect the benchmark here directly. The interesting measurement would be the amount of VCPU time used when running virtualised before and after this change (and on an idle vs active system).

> I haven't reported this as an issue upstream, mostly because I hadn't managed to convince myself that this was a kernel bug.

A performance issue of this sort (or other unreasonable overhead/obvious inefficiency etc) is certainly something should be reported, whether it is strictly "a bug" or not (IMHO).

kmjohansen:
Rebased against the tip of master, and made the requested changes to aarch64 and s390.  Added 4.18 to the mix, since it also has performance similar to that of 4.17.

rn:
no objections, so merging.

kmjohansen:
> Looking at just the per-net ops_init() calls that take more than 1ms, I get the following list:
> 
> ```
> __ip_vs_ftp_init 1.205770
> clusterip_net_init 8.582198
> frame_filter_net_init 23.559237
> frame_nat_net_init 23.989305
> __ip_vs_init 99.905918
> ```

A postscript on this, which I think was left partially unresolved at the time this PR landed.  I went back and looked through the code again as part of reviewing something else.  What I found was that these per-net init slowdowns actually have the same underlying root cause.  The heavy hitters in that chart all call `nf_register_net_hook()` -> `synchronize_net()` -> `synchronize_sched()` which puts us right into the slow path that we were hitting elsewhere.  The `synchronize_net` in `nf_register_net_hook` was removed by a fixed that landed in 4.15, which is why it wasn't seen on newer kernels.  (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8c873e2199700c2de7dbd5eedb9d90d5f109462b)

PR #3162, opened 2018-08-17
Update and fix compose project

deitch:
**- What I did**
Updated the compose project as follows:

* remove `rm` of pre-loaded tar files in static as they are in a read-only filesystem
* Updated docker version to `18.06`
* Bumped memory for static in Makefile's `run-static` target to handle the larger file
* Added `getty` to the compose examples
* Removed static list of images from `Makefile` and just looked for them dynamically in the compose file

**- How I did it**

Changed the `linuxprojects/compose` image, updated and pushed, then changed the `Makefile` and the two linuxkit ymls

**- How to verify it**
For both static and dynamic, build the `lkt` images, run them, check the output:

1. build: `make static` or `make dynamic`
2. run: `make run-static` or `make run-dynamic`
3. Wait for the console to come up
4. Check the logs in `/var/log/compose.log` and `/var/log/compose.out.log` - it takes a few moments for docker to start up, and then (in dynamic case) for it to download the images
5. Exec into docker container: `ctr -n services.linuxkit t exec --tty --exec-id 999 docker sh`
6. List the containers running: `docker ps`. Expect output as follows (with different timestamps and IDs, of course):

```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                NAMES
46bc1627bde0        traefik             "/traefik --web --do…"   19 seconds ago      Up 17 seconds       0.0.0.0:80->80/tcp   compose_proxy_1
9cc75e427ba7        nginx:alpine        "nginx -g 'daemon of…"   19 seconds ago      Up 17 seconds       80/tcp               compose_b_1
0af4179f1251        nginx:alpine        "nginx -g 'daemon of…"   19 seconds ago      Up 17 seconds       80/tcp               compose_a_1
```

Update compose example and fix startup errors

### DIFF
--- a/projects/compose/Makefile
+++ b/projects/compose/Makefile
@@ -3,9 +3,7 @@
 ORG?=linuxkitprojects
 IMAGE=compose
 DEPS=image/Dockerfile docker-compose.yml image/waitfordocker.sh image/load-images-and-compose.sh
-COMMON_IMAGES := \
-	nginx:alpine \
-	traefik
+COMMON_IMAGES := $(shell awk '/image:/ {print $$2}' docker-compose.yml | sort | uniq)
 
 
 HASH?=$(shell git ls-tree HEAD -- ./image | awk '{print $$3}')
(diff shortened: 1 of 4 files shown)


deitch:
Thanks @rn. I will merge in once ci is done (even if it is just a project).

PR #3163, opened 2018-08-17
Update kernels, Intel ucode and add 4.18.x kernels

rn:
- The 4.14.63 kernel has some important security updates against L1TF (CVE-2018-3615, CVE-2018-3620, CVE-2018-3646) and userspace-userspace SpectreRSB.
- Update the Intel ucode which quite possibly has more speculative execution fixes.
- Add support from 4.18.x kernels. The config was derived from 4.17.x but I additionally enabled:
    - XDP_SOCKETS
    - NFT_CONNLIMIT
    - IP_VS_MH
    - BPFILTER (as a module)
- The 4.18.x update also required a tweak to Docker.kconfig

As part of this update I also pushed the `kernel-bcc` package and updated the `bcc` test ti use packages which exist on hub. The test is not enabled by default. While at it I also tidied up `Dockerfile.bcc`.

The x86_64 and arm64 kernels are build and pushed. There is an issue with the s390x machine which has been reported TLS handshake timeouts when talking to the notary server (and hub), and issue I have not seen on the other arches. I'll try to build the s390x kernel later

elephant-bull


rn:
The `STACKPROTECTOR_STRONG` option has changed the name so updated the `test-kernel-config` accordingly

rn:
still getting errors on the s390x machine:
```
4.17.15-8025d130daad3418fbc8f39d0a461a368e745dd4-s390x: digest: sha256:ab0833b2a93354c17d088d086b9a33fa463d74f0fbd83fced75327b6c4a5c170 size: 530
ERRO[0069] Could not publish Repository since we could not update: net/http: TLS handshake timeout
failed to sign docker.io/linuxkit/kernel:4.17.15-8025d130daad3418fbc8f39d0a461a368e745dd4-s390x: net/http: TLS handshake timeout
```
This does not seem to be an issue with hub/notary server as doing this from several other machines works fine. It also seems to be temporal, as in this case the 4.18 kernel push/sign worked fine

kernel/Dockerfile.bcc
```diff
@@ -107,6 +117,6 @@ FROM scratch
 ENTRYPOINT []
 CMD []
 WORKDIR /
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/lib64
```
ijc: The dockerfile seems to copy stuff which was built for `lib64` to `lib` in the output, so is this really needed? Would be better to arrange for it not to be in any case.
(for a future change though if you prefer)

rn: this was actually in the previous stage, where i made no sense to me, so I just moved it to the final stage. Can probably be removed...Will do when I get around to switching the kernel build to `linuxkit pkg build`

rn:
The s390x trust issue is definitely temporal (but fatal). I sometimes see TLS handshake timeouts on the `linuxkit/kernel:4.18.1` portions, sometime on the `linuxkit/kernel-perf:4.18.1` portions, sometimes on the 4.17.15 kernel. This is also after clearing all the trust cache. It also seems to be specific to the s390 system as i had no issues with the arm64 and the x86 builds which were on completely different networks.

The issue with the s390x has been there since at least saturday. @alicefr do you have any idea what could cause TLS handshake errors such as this:
```
New signed multi-arch image: linuxkit/kernel:4.18.1-8025d130daad3418fbc8f39d0a461a368e745dd4

time="2018-08-20T12:31:14-04:00" level=error msg="Error trying v2 registry: read tcp 148.100.33.156:54968->52.22.201.61:443: read: connection reset by peer"
time="2018-08-20T12:31:14-04:00" level=warning msg="Couldn't find or access image reference \"linuxkit/kernel:4.18.1-arm64\". Skipping image."
time="2018-08-20T12:31:25-04:00" level=error msg="Error trying v2 registry: Get https://registry-1.docker.io/v2/linuxkit/kernel/manifests/4.18.1-s390x: net/http: TLS handshake timeout"
time="2018-08-20T12:31:25-04:00" level=warning msg="Couldn't find or access image reference \"linuxkit/kernel:4.18.1-s390x\". Skipping image."
Digest: sha256:03e4980e643cec0642886965560543e41cfeac365b02132a5ad5d099264eca0d 433
```

rn:
In the meantime, I'm merging this.

alicefr:
@rn The error is due to a network problem of the machines. Please, see my answer in slack

rn:
Just for the record. The networking issues were indeed resolved and I managed to build/push the s390x kernel images

Issue #3172, opened 2018-08-30
Enable network CAN flag (SocketCAN)


duraki:
As pointed by docker/for-mac#2580, `linuxkit` should enable CAN flag in kernel configuration to allow Docker containers to setup virtual CAN device. The namespace was presented by O. Hartkopp in [Linux Kernel 4.12](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/log/net/can?h=linux-4.12.y); update docs available on [SocketCAN April 2018](https://drive.google.com/viewerng/viewer?url=https://wiki.automotivelinux.org/_media/agl-distro/agl2018-socketcan.pdf).

This **issue affects** both Linux & MacOS. 

With `linux-kernel@4.12` update, we can now bridge cap for virtual CAN interface in Docker, instead of relying on hardware devices and tunnelling.

https://github.com/linuxkit/linuxkit/blob/262e6ed3ec9d1d6327df58eda02c8f20f3571b93/kernel/config-4.18.x-x86_64#L1451

This wouldn't be such a hassle, if D4M (Docker 4 Mac) wasn't private. We can't recompile `docker-for-mac.iso` properly as seen in linked issue. This needs to be set on [linuxkit/linuxkit](/linuxkit/linuxkit) side.

duraki:
Since it's my first time contributing to [linuxkit/linuxkit](/linuxkit/linuxkit), I'm allowed to ask a few questions:

* Shall we bring this topic to KSPP newsletter?
* Shall we flag on `check-kernel-config.sh` to verify?
* Shall it [officially] support landlock?

rn:
Can't you compile the require kernel modules externally and modprobe them? All the config options in `net/can` seem to be `tristate` suggesting this should be possible

duraki:
No. Tested under RedHat Linux and MacOS. Check [this comment](https://github.com/docker/for-mac/issues/2580#issuecomment-375726422) on why it's not possible. We need explicit allowance from linuxkit side.

brandonros:
Can this be closed?

duraki:
@brandonros not yet, give me some more time so I can coop with @hartkopp and make a final decision. Thank you. @rn: will provide additional information.

thomasantony:
@duraki Were you able to get this to work with D4M ?

PhilippKirchnerOSMA:
Are there any news on support for SocketCan in D4M?

PR #3173, opened 2018-08-31
Enable CAN networking subsystem

duraki:
**- What I did**

This PR should fix issue #3172 by enabling CAN network module for `clear-containers` within kernel flag config. This feature works for Linux Kernel 4.12 and above. 

**- How I did it**

By enabling kernel flag `CONFIG_CAN` to load this module. This namespace is introduced in [linux/linux-kernel@4.12](/torvalds/linux/tree/v4.12/drivers/net/can). After extension is loaded, we may bring up new `vcan` devices in container environment.

**- How to verify it**

Built against 4.12, or higher version supported within linuxkit kernel. Once built, we may either force or set bootable module. No patch for previous kernels, given the feature is proposed for upstream. Not implemented or tested in landlock since KSPP. Host machine using Kernel 4.12. **Not tested on MacOS nor equivalent image**.

See commit for details.

Manual (if avoiding mdev, or rebuilding):

```yml
onboot:
...
  - name: modprobe
    image: linuxkit/modprobe:<hash>
    command: ["modprobe", "vcan"]
```

Testing module:
```
$ modprobe vcan
$ sudo ip link add dev vcan0 type vcan
$ sudo ip link set up vcan0
```

Built passed for all [Linux] images, not tested is `docker-for-mac.yml` base.

* Enable CAN module.

### DIFF
--- a/projects/clear-containers/kernel/kernel_config
+++ b/projects/clear-containers/kernel/kernel_config
@@ -986,7 +986,7 @@ CONFIG_NET_FLOW_LIMIT=y
 #
 # CONFIG_NET_PKTGEN is not set
 # CONFIG_HAMRADIO is not set
-# CONFIG_CAN is not set
+CONFIG_CAN=y
 # CONFIG_IRDA is not set
 # CONFIG_BT is not set
 # CONFIG_AF_RXRPC is not set


duraki:
Paging @hartkopp for the confirmation on PR for CAN support in the namespace environment.

hartkopp:
The CAN support has been enabled even in RHEL7 (obviously due to multiple requests from RHEL6 users). IMO linuxkit and CAN fit perfectly together when I look into the linuxkit goals - which might even be interesting for embedded use cases. Thanks for the support!

Clovel:
> `CAN` networking seems a bit niche.

A lot of embedded engineers use the CAN protocol, and being able to test their programs with `vcan` is essential. macOS users do not have access to SocketCAN's `vcan`, so a Docker image (which uses LinuxKit I believe) could be one way to easily have access to such Linux features.

PR #3179, opened 2018-09-04
[WIP] Add TPM2 random number generator support

justincormack:
This should work on any platform with a tpm2 device or vtpm.

Still testing this, and need to rebuild packages.  cc @deitch 

pic-dump-217-20


deitch:
This looks good to me. I don't have an immediately accessible device with `tpm2` (wish I did) or I would test it out as well.

Do we want some sort of preference of one over the other if both exist? Based on this, it looks like it tries `/dev/random` (assuming it to be hwrng, which may or may not be accurate) and then also tries tpm2? My assumption is that we would do one or the other.

It may be fair to say, highly unlikely to have both, but that may be a bad assumption.

justincormack:
Closing, see https://github.com/linuxkit/linuxkit/pull/4170 now the kernel handles this just fine.

PR #3184, opened 2018-09-10
Fix path separator on Windows

mat007:
Sorry for opening a PR per fix, It's just I'm so glad I'm making progress every time. 😋 

**- What I did**

Fixed more file path handling on Windows.

**- How I did it**

On Windows `os.PathSeparator` is `\` but here it's all unix file paths being manipulated.

And not really related but `byte('/')` is really just `'/'`.

**- How to verify it**

Before
```
../bin/linuxkit build  -format iso-efi -name docker-for-mac -dir build image.yml image-mac.yml
…
Create outputs:
  build\docker-for-mac-efi.iso
[31mFATA[0m[0096] Error writing outputs: Error writing iso-efi output: docker run linuxkit/mkimage-iso-efi:248673cd3784eb3664d0f80b5bd31829465ca784 failed: exit status 1 output:

```
(errors right away with no output, no logs, no clue …)

Now
```
../bin/linuxkit build  -format iso-efi -name docker-for-mac -dir build image.yml image-mac.yml
…
Create outputs:
  build\docker-for-mac-efi.iso
mv -fv build/docker-for-mac-efi.iso build/docker-for-mac.iso
…
```

### DIFF
--- a/src/cmd/linuxkit/moby/build.go
+++ b/src/cmd/linuxkit/moby/build.go
@@ -454,7 +454,7 @@ func filesystem(m Moby, tw *tar.Writer, idMap map[string]uint32) error {
 			return errors.New("Did not specify path for file")
 		}
 		// tar archives should not have absolute paths
-		if f.Path[0] == os.PathSeparator {
+		if f.Path[0] == '/' {
 			f.Path = f.Path[1:]
 		}
 		mode := int64(0600)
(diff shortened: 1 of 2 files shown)


ijc:
This could/should have used `path.IsAbs` in a bunch of places I think (which also removes a panic if the string is empty).

mat007:
Good point!
Something is still amiss somewhere on Windows which I am investigating, I'll add a revision with that cleanup.

ijc:
Cool ta!.

BTW I think there is a similar panic in the suffix check in `prefix[len(prefix)-1]` when `len(prefix) == 0`. Might be worth encapsulating in a `HasTrailingSlash` or somesuch.

PR #3186, opened 2018-09-16
Update kernels to 4.18.8/4.14.70/4.9.127/4.4.156

rn:
And also 4.18.7/4.14.69/4.9.126/4.4.155

dog


rn:
Pulling `linuxkit/kernel:4.14.70` seems to fail for x86. When building the kernel yesterday I had some issues with notary again...so force pushing the kernel now

rn:
yes, the `-amd64` image is on hub but not in the manifest list. I think this is when the trust update timed out or something (I can't recall the exact error message). Anyway, just `forcepush_4.14.x` should fix this and then I hit rebuild on the CI

PR #3197, opened 2018-10-08
GCP PUSH: Fix typo and add error message when project is missing

pyaillet:
Signed-off-by: Pierre-Yves Aillet <pyaillet@gmail.com>

**- What I did**
Fix error message for gcp push when project is not specified
Before:
FATA[0004] Error creating Google Compute Image: googleapi: got HTTP response code 404 with body: Not Found

**- How I did it**
Check for project var being empty and exiting with an error message

**- How to verify it**
Launching linuxkit push gcp without project being set should output an error:
```shell
$ linuxkit push gcp nginx-os.img.tar.gz
FATA[0000] Please specify the project to use
```

GCP push - Notify when project is not set

image

### DIFF
--- a/src/cmd/linuxkit/gcp.go
+++ b/src/cmd/linuxkit/gcp.go
@@ -63,7 +63,7 @@ func NewGCPClient(keys, projectName string) (*GCPClient, error) {
 			projectName: projectName,
 		}
 	} else {
-		log.Debugf("Using Application Default crednetials")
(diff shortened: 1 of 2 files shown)


justincormack:
Hmm, seems like a CI issue, I tried restarting but no effect

rn:
I've hit rebuild a few times to no avail

pyaillet:
A better solution is proposed by ebriand in #3202

PR #3203, opened 2018-10-15
Fix golint on circleci and GCP testing

justincormack:

### DIFF
--- a/test/hack/test.yml
+++ b/test/hack/test.yml
@@ -15,7 +15,7 @@ onboot:
     image: linuxkit/test-kernel-config:1aaef970b5f70791d74d6f980ad38af4035948f8
   - name: poweroff
     image: linuxkit/poweroff:afe8f7dd0d47a7991c54519b0f09124cb8c4e300
-    command: ["/bin/sh", "/poweroff.sh", "3"]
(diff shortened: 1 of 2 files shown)


test/hack/test.yml
```diff
@@ -13,6 +13,9 @@ onboot:
     command: ["/sbin/dhcpcd", "--nobackground", "-f", "/dhcpcd.conf", "-1"]
   - name: check-kernel-config
     image: linuxkit/test-kernel-config:1aaef970b5f70791d74d6f980ad38af4035948f8
+  - name: sleep
+    image: linuxkit/poweroff:afe8f7dd0d47a7991c54519b0f09124cb8c4e300
+    command: ["/bin/sleep", "120"]
```
rn: The `poweroff` package also takes an numerical argument for sleep. Maybe this needs bumping from 3s to something a bit more

justincormack: ah yes I forgot that; lets see if this has any effect anyway and will change that instead.

justincormack: ok that did fix it, will adjust....

justincormack:
When I did `linuxkit run gcp` by hand the console connected quite late, so I guess they just cut some boot time somehow...

PR #3205, opened 2018-10-17
Missing s390x in the switch for qemu Dockerfile

alicefr:
Signed-off-by: Alice Frosi <afrosi@de.ibm.com>


justincormack:
I think this is only used for the `linuxkit run qemu ...` functionality when qemu is not installed on the host. There is a discussion (somewhat unresolved) in #3178 about whether to just remove this and insist on qemu being installed on the host instead, which would make this unnecessary...

alicefr:
thanks @justincormack ! If you plan to remove the `-containerized` option than this PR isn't necessary. Now the `linuxkit\qemu` on s390x image doesn't work because it doesn't have qemu-system-s390x inside

PR #3212, opened 2018-10-29
containerd v1.2.0

ijc:
This is mostly the usual mechanical bump (no code changes needed in LinuxKit) except I switched to using credentials helpers since the last bump I did so I had to update some of the scripting/tooling to cope with that on Linux (see the first commit).

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -18,7 +18,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:v0.6
   - name: sshd
-    image: linuxkit/sshd:v0.6
+    image: linuxkit/sshd:c4bc89cf0d66733c923ab9cb46198b599eb99320
     binds:
      - /run/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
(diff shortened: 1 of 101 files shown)


ijc:
I've not run it at all, CI should cover the basics but this being a new minor version I guess there is a chance that e.g. `ctr` interfaces have changed in ways which people might trip over.

ijc:
CI:
```
[FAIL    ] linuxkit.packages.getty-containerd 90.49s
```
Not exactly a huge shock. I'll take a look.

ijc:
I've kicked CI to have another go while I dig into it, on the off chance it was just a flake.

ijc:
Failed for me locally too, so not a flake. Trying to obtain any useful logs.

ijc:
Looks like I need to rebase so I'll sort out the minor nit as I go.

ijc:
Rebased and force pushed, will merge with the existing approvals once CI is green.

I dropped the containerd maintainers a line re the new need to bind `/var/lib/containerd` just in case it wasn't expected. We shouldn't block this PR if it wasn't though and should instead revert the relevant commit whenever we pickup a fixed v1.2.x.

rn:
SGTM

justincormack:
Yes, curious as to why that change is needed...

PR #3213, opened 2018-10-29
build: Don't create empty kernel/cmdline files

rn:
If the YAML does not specify a kernel, kernel commandline
or any containers, don't create empty files. Note, an
initrd file is still created if the kernel image contains
CPU ucode.

This only applies to kernel+initrd and tar-kernel-initrd
output formats.

humminbird


justincormack:
Just wondering if any of the formats that use these currently require an empty `cmdline` file? eg looks like https://github.com/linuxkit/linuxkit/blob/master/tools/mkimage-qcow2-efi/make-efi will error (as it has `set -e`) if the cmdline file is missing. Should be easy enough to fix to accept an empty file.

rn:
`qcow2-efi` works without a cmdline in the YAML. This commit does not remove the empty file from the `tar` format which is used as the base for all the other formats.

PR #3214, opened 2018-10-29
Update crosvm

rn:
- Fix crosvm documentation
- Update crosvm build to rust 1.30
- Update crosvm commit to latest

I've done some light testing and the non-jail/non-seccomp version seems to boot

waterant


rn:
@justincormack I pushed two more commits with an update of rust and latest commit of crosvm. Could you re-approve as this is a bit more than then initial commit.

Issue #2836, opened 2018-01-05
Investigate eudev


rn:
We seem to get some reports where cold/hot plug may not work all the time. A simple test case seems to be using a kernel package build from the debian kernel:
```
./scripts/kernels/debian.sh foobar/kernel-debian 4.14.0-2 4.14.7-1
```
and then use `foobar/kernel-debian:4.14.7-1` in a YAML file and run with hyperkit (or qemu).

The debian kernels have all of virtio drivers configured as modules and `mdev` only seem to modprobe the virtio PCI driver but not, for example the `virtio_net` driver. It's unclear, why this does not happen, as, in theory, the bus driver gets cold plugged, and then the devices, such as the network device driver, should get hotplugged. `eudev` may handle this better. 

With `eudev` we may also get https://github.com/linuxkit/linuxkit/issues/2418 for free.

PR #3217, opened 2018-11-01
Allowing modprobe modules to load firmware if needed

rvs:
**- What I did**

Allowed for modprobe modules to load firmware files.

**- How I did it**

Made /lib/firmware available in the modprobe container and also added capabilities all. This last bit I'm really not sure of, but try as hard as I may the CAP_SYS capabilities don't seem to allow for kernel to load firmware files from the filesystem. I think it should be fine for this pkg, but would appreciate any guidance on how to make it more granular.

**- How to verify it**

Modprobe any WiFi driver that requires a firmware

Allowed for modprobe modules to load firmware files.

=!=!=
    - 
   \/

### DIFF
--- a/pkg/modprobe/build.yml
+++ b/pkg/modprobe/build.yml
@@ -2,6 +2,7 @@ image: modprobe
 config:
   binds:
     - /lib/modules:/lib/modules
+    - /lib/firmware:/lib/firmware
     - /sys:/sys
   capabilities:
-    - CAP_SYS_MODULE
+    - all


justincormack:
`all` literally just adds the full list of CAP_SYS caps, so some subset of them must also work... I would suspect you might need CAP_SYS_ADMIN as well?

ijc:
I think I must misunderstand how firmware loading works, perhaps some one can enlighten me.

My understanding was that firmware loading was an asynchronous process (triggered by drivers while they are probing) which ends up triggering a userspace event which udev (or perhaps mdev) then satisfies by loading the firmware (by dumping it to some file under `/sys`).

I don't think that `modprobe` ever does any modprobing (I don't see anything like that in `modprobe` applet in the `busybox` source, that all seems to be in the `mdev` applet).

It's also my understanding that the userspace events are processed in the host/root namespace, not within any particular container, even if the device probe/driver bind that triggered the userspace firmware loading event was caused by modprobe loading some new module from inside a container.

The `pkg/init/usermode-helper.c` helper executes both `modprobe` and `mdev` directly in the root namespace, it is not forwarding to a container AFAICT.

What am I missing? How/where does firmware get loaded within any `modprobe` container?

justincormack:
You could try just putting the full list in and removing them one by one. It seems weird though, I would try `CAP_MKNOD` and maybe `CAP_DAC_OVERRIDE` as plausible ones...

rvs:
@ijc here's a few things you're missing:
  * while it is true that by default modprobe container doesn't do modprobe it actually has the binary to do so and a capability CAP_SYS_MODULE to allow it. Hence the typical usage is to specify something like command: ["modprobe", "-a", "XXX" ] in your linuxkit.yml file. The problem, of course, is how that interacts with the case when the firmware is required for the driver being modrpobed. Which brings us to...
  * ...how firmware gets loaded (note this changed at around 2016 https://lwn.net/Articles/676101/): modern linux kernels actually default to read firmware from the filesystem (instead of going through udev and userspace). The question then becomes what namespace does it look under. This gets determined by whether the firmware loading event got triggered from within a container (or even chroot for that matter or not). 

You're right that in linuxkit all of this a bit manual since we're not using anything like mdev (although as @rn pointed out we do have this issue pending https://github.com/linuxkit/linuxkit/issues/2836). However, I'd argue that for as long as we use modprobe package for manually orchestrating something that mdev would've done -- we might as well empower it to do its job.

Of course, to @justincormack point -- nobody's suggesting that we add capabilities all -- that's still need to get clarified. The rest, however, I would argue stands.

Does it make it clear(er)?

rn:
> You're right that in linuxkit all of this a bit manual since we're not using anything like mdev

We are actually using `mdev` in the root namespace but it sometimes fails to load modules. This seems particularly the case for USB attached devices. On the RPi3, for example, the smsc95xx modules needs to be loaded manually.

PR #3225, opened 2018-11-14
Add experimental support for building uncompressed kernels

### LINKED ISSUE #3215: And an option to extract a uncompress linux kernel image
For some backends (e.g. crosvm) it might be useful to boot an uncompressed kernel images. So this proposal is to have a `-vmlinux` option to `linuxkit build` to extract the raw kernel image during the build (basically implement the equivalent of [extract-vmlinux](https://raw.githubusercontent.com/torvalds/linux/master/scripts/extract-vmlinux) in go).

rn:
This restructures the `linuxkit build` subcommand to:
- stash the kernel image in an internal buffer
- add a new `-vmlinux` build option to decompress the kernel
- decompress the kernel into an internal buffer

It also updates `crosvm` to the latest commit and updates the instructions to use the `-vmlinux` build option.

The resulting vmlinux image is binary identical to using the `extract-vmlinux` script from the Lnux kernel `./script` directory. The resulting kernel boots fine on `crosvm` but there is no interactive console (same as with manually extracting the kernel).

The support is experimental in the sense that it assume the fist gzip header found in the vmlinux image is the kernel image and that we do not check the resulting vminux image.

resolves #3215

peacock-mantis-shrimp

### DIFF
--- a/contrib/crosvm/README.md
+++ b/contrib/crosvm/README.md
@@ -26,35 +26,29 @@ You may also have to create an empty directory `/var/empty`.
 
 You can build a LinuxKit image suitable for `crosvm` with the
 `kernel+squashfs` build format. For example, using `minimal.yml` from
-the `./examples` directory, run:
+the `./examples` directory, run (but also see the known issues):
 
 ```sh
-linuxkit build -format kernel+squashfs minimal.yml
+linuxkit build -format kernel+squashfs -decompress-kernel minimal.yml
 ```
 
-The generated kernel file (`minimal-kernel`) needs to be converted as
-`crosvm` does not grok `bzImage`s. You can convert the LinuxKit kernel
-image with
-[extract-vmlinux](https://raw.githubusercontent.com/torvalds/linux/master/scripts/extract-vmlinux):
-
-```sh
-extract-vmlinux minimal-kernel > minimal-vmlinux
-```
+The `-vmlinux` switch is needed since `crosvm` does not grok
+compressed linux kernel images.
 
 Then you can run `crosvm`:
 ```sh
-./crosvm run --seccomp-policy-dir=./seccomp/x86_64 \
+crosvm run --disable-sandbox \
     --root ./minimal-squashfs.img \
     --mem 2048 \
-    --multiprocess \
     --socket ./linuxkit-socket \
-    minimal-vmlinux
+    minimal-kernel
 ```
 
 ## Known issues
 
 - With 4.14.x, a `BUG_ON()` is hit in `drivers/base/driver.c`. 4.9.x
   kernels seem to work.
+- With the latest version, I don't seem to get a interactive console.
 - Networking does not yet work, so don't include a `onboot` `dhcpd` service.
 - `poweroff` from the command line does not work (crosvm does not seem
   to support ACPI). So to stop a VM you can use the control socket
--- a/src/cmd/linuxkit/build.go
+++ b/src/cmd/linuxkit/build.go
@@ -49,6 +49,7 @@ func build(args []string) {
 	buildSize := buildCmd.String("size", "1024M", "Size for output image, if supported and fixed size")
 	buildPull := buildCmd.Bool("pull", false, "Always pull images")
 	buildDisableTrust := buildCmd.Bool("disable-content-trust", false, "Skip image trust verification specified in trust section of config (default false)")
+	buildDecompressKernel := buildCmd.Bool("decompress-kernel", false, "Decompress the Linux kernel (default false)")
 	buildCmd.Var(&buildFormats, "format", "Formats to create [ "+strings.Join(outputTypes, " ")+" ]")
 
 	if err := buildCmd.Parse(args); err != nil {
@@ -203,7 +204,7 @@ func build(args []string) {
 	if moby.Streamable(buildFormats[0]) {
 		tp = buildFormats[0]
 	}
-	err = moby.Build(m, w, *buildPull, tp)
+	err = moby.Build(m, w, *buildPull, tp, *buildDecompressKernel)
 	if err != nil {
 		log.Fatalf("%v", err)
 	}
--- a/src/cmd/linuxkit/moby/build.go
+++ b/src/cmd/linuxkit/moby/build.go
@@ -257,6 +257,7 @@ func Build(m Moby, w io.Writer, pull bool, tp string) error {
 type kernelFilter struct {
 	tw          *tar.Writer
 	buffer      *bytes.Buffer
+	hdr         *tar.Header
 	cmdline     string
 	kernel      string
 	tar         string
@@ -288,6 +289,19 @@ func (k *kernelFilter) finishTar() error {
 	if k.buffer == nil {
 		return nil
 	}
+
+	if k.hdr != nil {
+		if err := k.tw.WriteHeader(k.hdr); err != nil {
+			return err
+		}
+		if _, err := k.tw.Write(k.buffer.Bytes()); err != nil {
+			return err
+		}
+		k.hdr = nil
+		k.buffer = nil
+		return nil
+	}
+
 	tr := tar.NewReader(k.buffer)
 	err := tarAppend(k.tw, tr)
 	k.buffer = nil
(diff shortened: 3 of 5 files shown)


contrib/crosvm/README.md
```diff
@@ -26,35 +26,29 @@ You may also have to create an empty directory `/var/empty`.
 
 You can build a LinuxKit image suitable for `crosvm` with the
 `kernel+squashfs` build format. For example, using `minimal.yml` from
-the `./examples` directory, run:
+the `./examples` directory, run (but also see the known issues):
 
 ```sh
-linuxkit build -format kernel+squashfs minimal.yml
+linuxkit build -format kernel+squashfs -vminux minimal.yml
```
justincormack: typo, `vminux`

src/cmd/linuxkit/build.go
```diff
@@ -49,6 +49,7 @@ func build(args []string) {
 	buildSize := buildCmd.String("size", "1024M", "Size for output image, if supported and fixed size")
 	buildPull := buildCmd.Bool("pull", false, "Always pull images")
 	buildDisableTrust := buildCmd.Bool("disable-content-trust", false, "Skip image trust verification specified in trust section of config (default false)")
+	buildVMLinux := buildCmd.Bool("vmlinux", false, "Decompress the Linux kernel (default false)")
```
ijc: The name of this option is a bit opaque for what it does. I think it might be implying "vmlinux not vmlinuz" or some such, but that's a bit of an obscure internal detail for an end user.

`-decompress-kernel` perhaps?

Same sort of thing should be used for the variable names too I reckon.

src/cmd/linuxkit/moby/build.go
```diff
@@ -427,6 +452,35 @@ func tarAppend(iw *tar.Writer, tr *tar.Reader) error {
 	return nil
 }
 
+// Attempt to decompress a Linux kernel image
+// This only looks for the first gzip signature in the image and decompresses it.
+// There are no sanity check that the result is an ELF binary. YMMV
```
ijc: The (x86) kernel has fields in its header which should allow you to precisely locate the payload. Here is my ancient C hack to do it:
```c
#include <stdio.h>
#include <fcntl.h>
#include <stdint.h>
#include <unistd.h>

#include <inttypes.h>

#include <err.h>

#include <sys/types.h>
#include <sys/mman.h>
#include <sys/stat.h>

int main(int argc, char **argv)
{
	int fd;
	struct stat sb;
	void *p;
	uint8_t *hdr;
	int setup_sectors;
	uint32_t compressed_payload_offset;
	uint32_t compressed_payload_length;

	if (argc != 2)
		errx(1, "usage: bzexplode <bzImage>");

	fd = open(argv[1], O_RDONLY);
	if (fd < 0)
		err(1, "open");

	if (fstat(fd, &sb) < 0)
		err(1, "fstat");

	p = mmap(0, sb.st_size, PROT_READ, MAP_SHARED, fd, 0);
	if (p == MAP_FAILED)
		err(1, "mmap");

	hdr = p;
	setup_sectors = hdr[0x1f1];
	compressed_payload_offset = *(uint32_t*)&hdr[0x248];

	fprintf(stderr, "setup sectors %d\n", setup_sectors);

	compressed_payload_offset += (setup_sectors+1) * 512;

	//compressed_payload_length = *(uint32_t*)(p + compressed_payload_offset - 4);
	compressed_payload_length = *(uint32_t*)&hdr[0x24c];

	fprintf(stderr, "compressed_payload_offset %"PRIx32" (abs)\n",
		compressed_payload_offset);
	fprintf(stderr, "compressed_payload_length %"PRIx32"\n",
		compressed_payload_length);

	write(1,
	      p + compressed_payload_offset,
	      compressed_payload_length);
	return 0;
}

```

rn:
@ijc I renamed the flag and restructured the code to find the gzip header. It now tests offset 0 (for arm64) and the offset based on information in a bzImage header (also verifies the gzip signature).

I tested the result with `crossvm` and the decompressed x86 kernel boots fine there. Haven't checked arm64 but it is just a plain gzip'ed kernel so should be fine

src/cmd/linuxkit/moby/build.go
```diff
@@ -427,6 +453,53 @@ func tarAppend(iw *tar.Writer, tr *tar.Reader) error {
 	return nil
 }
 
+// Attempt to decompress a Linux kernel image
+// arm64 kernel images are plain gzip
+// x86 kernel images are in bzImage format, where the offset for the gzip'ed image is described in the header
```
ijc: FWIW it can actually be one of several possible compression algs. Fine to just support one for now though.

e.g. on Debian:
```
$ grep CONFIG_KERNEL /boot/config-4.17.0-1-amd64
# CONFIG_KERNEL_GZIP is not set
# CONFIG_KERNEL_BZIP2 is not set
# CONFIG_KERNEL_LZMA is not set
CONFIG_KERNEL_XZ=y
# CONFIG_KERNEL_LZO is not set
# CONFIG_KERNEL_LZ4 is not set
```

src/cmd/linuxkit/moby/build.go
```diff
@@ -427,6 +453,53 @@ func tarAppend(iw *tar.Writer, tr *tar.Reader) error {
...
+	s := src.Bytes()
+
+	if bytes.HasPrefix(s, []byte(gzipMagic)) {
+		log.Debugf("Found gzip signature at offset: 0")
+	} else {
+		// Not a plain gzip image, assume a bzImage and try to locate the compressed image
+		const setupSectorsIdx = 0x1f1
+		const sectorSize = 512
+		const payloadIdx = 0x248
+		const payloadLengthIdx = 0x24c
```
ijc: Perhaps link to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/x86/boot.txt for these?

src/cmd/linuxkit/moby/build.go
```diff
@@ -427,6 +453,53 @@ func tarAppend(iw *tar.Writer, tr *tar.Reader) error {
 	return nil
 }
 
+// Attempt to decompress a Linux kernel image
+// arm64 kernel images are plain gzip
+// x86 kernel images are in bzImage format, where the offset for the gzip'ed image is described in the header
+// s390x kernels images are not compressed.
+func decompressKernel(src *bytes.Buffer) (*bytes.Buffer, error) {
+	const gzipMagic = "\037\213\010" // Note, this is for gzip with max decoding used by the kernel
```
ijc: Perhaps rather than this, which encodes a specific set of options you could just try and then handle `gzip.ErrHeader`. i.e. let the gzip library probe the header.

src/cmd/linuxkit/moby/build.go
ijc: Also, I reckon you should test 0x1fe and 0x202:
```
01FE/2	ALL	boot_flag	0xAA55 magic number
0200/2	2.00+	jump		Jump instruction
0202/4	2.00+	header		Magic signature "HdrS"
```
for the `AA55` and `HdrS` magic numbers and if not find conclude this isn't an x86 bzImage at all.

Otherwise the behaviour of this code on uncompressed or non-x86 kernels is a little surprising I think.

You might optionally want to test the version header >= 2.08 which indicates the presence of the fields you consult.

AIUI we do want to support cross building images (or at least not make it harder to do in the future), so we can't use `runtime.GOARCH` to make decisions, right?

justincormack: Yes, there is demand for cross build at some point.

rn: > Otherwise the behaviour of this code on uncompressed or non-x86 kernels is a little surprising I think.

I don't understand this comment. If the user asks to decompress the kernel and we can't decompress it (or if it is not compressed), I think it's fair to error out. This should work for the arm64 LinuxKit kernels and x86 bzImage kernels. For s390x I don't think they are compressed, so don't specify `--decompress-kernel`...

> AIUI we do want to support cross building images (or at least not make it harder to do in the future), so we can't use runtime.GOARCH to make decisions, right?

I explicitly avoid this so the code should work on support kernel formats irrespective of the architecture we run `linuxkit build` on.

rn:
restructured the code:
- make it easier to add additional kernel image formats
- more sanity checks on bzImage images

The code still errors if a uncompressed image is used and `--decompress-kernel` is used, which I think is ok.

Also fixed up the `crosvm` README with the updated `linuxkit build` commandline which I forgot in the prefix revision.

src/cmd/linuxkit/moby/build.go
ijc: > I don't understand this comment. If the user asks to decompress the kernel and we can't decompress it (or if it is not compressed), I think it's fair to error out

Right, but the error would have been something about no gzip header in the middle of a file, found via reading some offset which wasn't actually an offset just random bytes. This looks better now though.

I think the only remaining issue is that won't it panic on an array/slice overrun if the file is too short to have the header bytes in it, or if the file is corrupt such that payload offset points beyond the limits of the file?

rn: I'm fine with panicking if the kernel image is not large enough or corrupted, but added some check now.

Issue #3201, opened 2018-10-10
Create a bpftrace package


rn:
[bpftrace](https://github.com/iovisor/bpftrace) was recently released. It creates a single binary (and some sample scripts) for tracing similar to dtrace.

It would be nice to have a package containing the binary and sample so that it can be included, for example in the init section.

Need to investigate, though, if it requires matching kernel header for compilation or use.

simar7:
@rn:  Is the idea of making bpftrace a core package under ```pkg/``` or something else?

(I'm fairly new to this project but looking to pick up some low hanging items to play with)

rn:
yes, that would be the idea.

simar7:
Cool! I've started to hack on this and here's the progress so far: https://github.com/linuxkit/linuxkit/pull/3230

simar7:
This can be closed now as https://github.com/linuxkit/linuxkit/pull/3230 is merged.

PR #3230, opened 2018-11-17
pkg/bpftrace: Add bpftrace as a core pkg

simar7:
**- What I did**
This PR adds [bpftrace](https://github.com/iovisor/bpftrace) as a package to linuxkit. It includes all the necessary components and directory structures in order to build a bpftrace inside of a docker container.

Addresses: https://github.com/linuxkit/linuxkit/issues/3201

**- How I did it**
I started off by looking at existing packages for a reference model of how to structure a package in linuxkit. I also read the https://github.com/linuxkit/linuxkit/blob/master/docs/packages.md doc.

**- How to verify it**
You can build it locally to test it out using the following command:
```linuxkit pkg build -org=<orgname> -disable-content-trust <path/to/bpftrace>```

This commit adds bpftrace to linuxkit packages and addresses the issue: https://github.com/linuxkit/linuxkit/issues/3201

image

Signed-off-by: Simarpreet Singh <simar@linux.com>

### DIFF
--- a/pkg/bpftrace/Dockerfile
+++ b/pkg/bpftrace/Dockerfile
@@ -1,0 +1,37 @@
+FROM linuxkit/alpine:ace75d0ec6978762d445083d6c8c6336bdb658ed AS build
+RUN apk add --update \
+    bison \
+    build-base \
+    clang-dev \
+    clang-static \
+    cmake \
+    elfutils-dev \
+    flex-dev \
+    git \
+    linux-headers \
+    llvm5-dev \
+    llvm5-static \
+    zlib-dev
+
+# Put LLVM directories where CMake expects them to be
+RUN ln -s /usr/lib/cmake/llvm5 /usr/lib/cmake/llvm
+RUN ln -s /usr/include/llvm5/llvm /usr/include/llvm
+RUN ln -s /usr/include/llvm5/llvm-c /usr/include/llvm-c
+
+ENV GITREPO=https://github.com/iovisor/bpftrace.git
+ENV COMMIT=8f7f8214d7dd7bc25b7740a3c0e9a580a89e0244
+ENV STATIC_LINKING=on
+ENV RUN_TESTS=0
+
+RUN git clone ${GITREPO} /bpftrace \
+    && cd /bpftrace/ \ 
+    && mkdir -p /bpftrace/build \
+    && git checkout ${COMMIT} \
+    && chmod +x /bpftrace/docker/build.sh \
+    && /bpftrace/docker/build.sh /bpftrace/build Release -j$(nproc)
+
+FROM scratch
+ENTRYPOINT []
+CMD []
+WORKDIR /
+COPY --from=build /bpftrace/build/src/bpftrace /usr/bin/bpftrace
--- a/pkg/bpftrace/build.yml
+++ b/pkg/bpftrace/build.yml
@@ -1,0 +1,4 @@
+image: bpftrace
+network: true
+arches: 
+  - amd64
--- a/test/cases/040_packages/002_bpftrace/test.yml
+++ b/test/cases/040_packages/002_bpftrace/test.yml
@@ -1,0 +1,29 @@
+kernel:
+  image: linuxkit/kernel:4.14.86
+  cmdline: "console=ttyS0 console=ttyAMA0"
+init:
+  - linuxkit/init:v0.5
+  - linuxkit/runc:v0.5
+  - linuxkit/bpftrace:67b5c3b4f13a4acf22eab8f422091e1efc06852d
+onboot:
+  - name: bpftrace-test
+    image: alpine:3.8
+    binds:
+      - /check.sh:/check.sh
+    command: ["sh", "./check.sh"]
+    pid: host
+    net: host
+    ipc: host
+    capabilities:
+      - all
+  - name: poweroff
+    image: linuxkit/poweroff:2687ed712690766b8156a77ac600005b532d2c7d
+    command: ["/bin/sh", "/poweroff.sh", "10"]
+files:
+  - path: check.sh
+    source: ./check.sh
+trust:
+  org:
+    - linuxkit
+  image:
+    - alpine:3.8
--- a/test/cases/040_packages/002_bpftrace/check.sh
+++ b/test/cases/040_packages/002_bpftrace/check.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+nsenter -t1 -m -- /usr/bin/bpftrace -l
+if [ "$?" -ne "0" ]; then
+	printf "bpftrace test suite FAILED\n" >&1
+	exit 1
+fi;
+
+printf "bpftrace test suite PASSED\n" >&1
--- a/test/cases/040_packages/002_bpftrace/test.sh
+++ b/test/cases/040_packages/002_bpftrace/test.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+# SUMMARY: Check that the bpftrace package works
+# REPEAT:
+
+set -e
+
+# Source libraries. Uncomment if needed/defined
+#. "${RT_LIB}"
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+NAME=bpftrace
+
+clean_up() {
+	rm -rf ${NAME}-*
+}
+trap clean_up EXIT
+
+# Test code goes here
+linuxkit build -format kernel+initrd -name "${NAME}" test.yml
+RESULT="$(linuxkit run ${NAME})"
+echo "${RESULT}"
+echo "${RESULT}" | grep -q "suite PASSED"
+
+exit 0

### CODE REVIEW

File: pkg/bpftrace/build-release.sh [hunk 1]
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
justincormack: can you use /bin/sh - this does not use any bash features...

pkg/bpftrace/build.sh
```diff
@@ -0,0 +1,14 @@
+set -e
```
justincormack: this is missing the hashbang...

justincormack:
I don't think we should vendor the bpftrace source here, can we download from a specific git commit instead?

rn:
Thanks for importing, this. But as @justincormack said, could you please follow a pattern similar to the other packages which have a `Dockerfile` which check out a specific revision of the source code and then compile it. If you look at the `bpftrace` `Dockerfile` is basically just installs the too chain, which should already be present in the `linuxkit/alpine` package, and then calls `build.sh`

pkg/bpftrace/build.sh
simar7: Thanks. Fixed in the next commit.

pkg/bpftrace/build-release.sh
simar7: Thanks. Fixed in the next commit.

simar7:
@rn and @justincormack : Good point on the vendoring out the bpftrace files. I've modified the Dockerfile to checkout from source.

pkg/bpftrace/build-release.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
rn: this should no longer be needed

pkg/bpftrace/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
rn: this should no longer be needed

pkg/bpftrace/build.yml
```diff
@@ -0,0 +1,13 @@
+image: bpftrace
+network: true
+arches: 
+  - amd64
+gitrepo: github.com/iovisor/bpftrace
```
rn: this is not needed.

pkg/bpftrace/build.yml
```diff
@@ -0,0 +1,13 @@
...
+config:
+  pid: host
+  binds:
+    - /proc:/host/proc
+    - /sys:/host/sys
+    - /:/rootfs
```
rn: why is this here?

pkg/bpftrace/build.yml
simar7: I can remove this. It was part of the packaging readme https://github.com/linuxkit/linuxkit/blob/master/docs/packages.md#package-source so I added.

rn: I think this refers to the location of the `build.yml` and defaults to the linuxkit repository so can be omited

pkg/bpftrace/build-release.sh
simar7: Removed.

pkg/bpftrace/build.yml
simar7: Gotcha. Fixed in the next commit.

pkg/bpftrace/build.yml
simar7: copy/pasta. Removed.

pkg/bpftrace/build.sh
simar7: Removed.

simar7:
Squashed into a single commit and force pushed with all the recommendations.

pkg/bpftrace/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+ENV COMMIT=8f7f8214d7dd7bc25b7740a3c0e9a580a89e0244
+ENV STATIC_LINKING=on
+ENV RUN_TESTS=0
+
+RUN git clone ${GITREPO} /bpftrace \
+    && cd /bpftrace/ \ 
+    && mkdir -p /bpftrace/build \
+    && git checkout ${COMMIT}
+
+ENTRYPOINT ["/bin/sh", "/bpftrace/docker/build.sh", "/bpftrace/build", "Release" ]
```
tjfontaine: Should this really be an `ENTRYPOINT` or another command in the previous `RUN`? I think it's the latter.

And then if it is in fact a fully static binary, then the final stanzas can be

```
FROM scratch
WORKDIR
ENTRYPOINT[]
## Assuming the real value of the path to bpftrace
COPY --from=build /bpftrace/build/bpftrace /bpftrace
CMD ["/bpftrace"]
```

and the first can be `FROM linuxkit/alpine:ace75d0ec6978762d445083d6c8c6336bdb658ed AS build`

simar7: Great suggestion. I'll try playing around with this.

simar7: I've implemented this in the next commit.

simar7:
Although this now runs as expected, I'm curious if this means we'd need a custom kernel with appropriate debug capabilities. For instance this is an example run that should've listed some probes.

```docker run -it bpftrace -l

No such file or directory: /sys/kernel/debug/tracing/available_filter_functions
```

test/cases/040_packages/002_bpftrace/test.yml
```diff
@@ -0,0 +1,29 @@
+kernel:
+  image: linuxkit/kernel:4.14.81
+  cmdline: "console=ttyS0 console=ttyAMA0"
+init:
+  - linuxkit/init:v0.5
+  - linuxkit/runc:v0.5
+  - wombat/bpftrace:1a2cc3766bf1f12090a3f8ea584d141f0ec860c0
```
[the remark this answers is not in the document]
simar7: Any thoughts on this cc @justincormack @rn @tjfontaine

rn:
@simar7 this looks good to me. Thanks for setting this up.

To add a new package, one of the maintainers has to add this on hub and set up signing on the notary server. A maintainer also has to build and push the new package for the CI to pass.

Unfortunately, I hadn't had time earlier this week and am travelling next week. I see if I can do the first bit tomorrow morning

pkg/bpftrace/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ENV COMMIT=8f7f8214d7dd7bc25b7740a3c0e9a580a89e0244
+ENV STATIC_LINKING=on
+ENV RUN_TESTS=0
+
+RUN git clone ${GITREPO} /bpftrace \
+    && cd /bpftrace/ \ 
+    && mkdir -p /bpftrace/build \
+    && git checkout ${COMMIT} \
+    && chmod +x /bpftrace/docker/build.sh \
+    && /bpftrace/docker/build.sh /bpftrace/build Release
```
tjfontaine: While testing this, I noticed that there's no parallelism done automatically, maybe add a `-j$(nproc)` or similar here? The `build.sh` seems to support arbitrary `make` arguments subsequent to `Release`

tjfontaine:
This pkg seems to work, I haven't tried running the linuxkit test though

```
(ns: getty) linuxkit-3efcfc3734a0:~# /proc/1/root/bpftrace -e 'tracepoint:raw_syscalls:sys_enter { @[comm] = count(); }'
chdir(/lib/modules/4.14.81-linuxkit/build): No such file or directory
definitions.h:1:10: fatal error: 'linux/types.h' file not found
Attaching 1 probe...


@[bpftrace]: 6
@[tini]: 30
@[containerd-shim]: 148
@[containerd]: 799


(ns: getty) linuxkit-3efcfc3734a0:~# /proc/1/root/bpftrace -l | wc -l
38542
```

tjfontaine:
Ok, I took a few laps testing this today, and have some suggestions --

```diff
diff --git a/pkg/bpftrace/Dockerfile b/pkg/bpftrace/Dockerfile
index 207345449..88c409cd1 100644
--- a/pkg/bpftrace/Dockerfile
+++ b/pkg/bpftrace/Dockerfile
@@ -27,11 +27,15 @@ RUN git clone ${GITREPO} /bpftrace \
     && cd /bpftrace/ \ 
     && mkdir -p /bpftrace/build \
     && git checkout ${COMMIT} \
-    && chmod +x /bpftrace/docker/build.sh \
-    && /bpftrace/docker/build.sh /bpftrace/build Release
+    && cd /bpftrace/build \
+    && export MAKEFLAGS=-j$(nproc) \
+    && cmake -DCMAKE_BUILD_TYPE="Release" -DSTATIC_LINKING:BOOL=ON -DRUN_TESTS=0 -DCMAKE_INSTALL_PREFIX=/usr ../ \
+    && make -j$(nproc) \
+    && DESTDIR=/out make install \
+    && chmod +x /out/usr/share/bpftrace/tools/*.bt
 
 FROM scratch
 WORKDIR /
-COPY --from=build /bpftrace/build/src/bpftrace /bpftrace
-ENTRYPOINT [ "/bpftrace" ]
-CMD ["/bpftrace", "-h"]
\ No newline at end of file
+COPY --from=build /out/ /
+ENTRYPOINT [ "/usr/bin/bpftrace" ]
+CMD ["/usr/bin/bpftrace", "-h"]
```

Essentially this just implements the `build.sh` inline to be able to pass more flags to `cmake` than what is currently available [there may be a more elegant `cmake` approach for this]. The key here, is we're setting the prefix to `/usr` so the tools look in more accurate platform locations for things, like `/usr/include` etc etc, also it uses `make install DESTDIR=/out` which allows us to ship the tools that are shipped with `bpftrace` as well.

As a result, you can either `nsenter -t1 -a -- bpftrace` or `export PATH=$PATH:/proc/1/root/usr/bin` though the latter is slightly "dangerous", assuming it is on your `$PATH` though, you can use the tools by either copying them into `getty`/`ssh` from the root namespace, or using them in place `/proc/1/root/usr/share/bpftrace/tools/pidpersec.bt`

also, using an image like

```yml
kernel:
  image: linuxkit/kernel:4.14.81
  cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
init:
  - linuxkit/init:c563953a2277eb73a89d89f70e4b6dcdcfebc2d1
  - linuxkit/runc:83d0edb4552b1a5df1f0976f05f442829eac38fe
  - linuxkit/containerd:326b096cd5fbab0f864e52721d036cade67599d6
  - linuxkit/ca-certificates:v0.6
  - linuxkit/bpftrace:1a2cc3766bf1f12090a3f8ea584d141f0ec860c0-dirty
  - linuxkit/kernel:4.14.81
onboot:
  - name: sysctl
    image: linuxkit/sysctl:v0.6
  - name: dhcpcd
    image: linuxkit/dhcpcd:v0.6
services:
  - name: getty
    image: linuxkit/getty:2eb742cd7a68e14cf50577c02f30147bc406e478
    env:
     - INSECURE=true
  - name: rngd
    image: linuxkit/rngd:v0.6
trust:
  org:
    - linuxkit
```

You're able to `tar xf /proc/1/root/kernel-headers.tar` such that you can avoid `bpftrace`'s invocation of `clang` from failing, and it can be in sync with the kernel version you're running.

These are just some suggestions, I'll file some tickets with `bpftrace` directly regarding some of the pitfalls of their implementation, but that's outside the scope of this PR.

simar7:
@tjfontaine those are some good suggestions but I feel as an initial implementation of this PR it's best to keep it small to start and later on we can improve upon this. Thoughts?

tjfontaine:
I'm fine with whatever the maintainers decide, I was just testing it out :)

simar7:
hey @justinbatcf and @rn sorry to nudge but if you folks could see if this PR makes the cut for a first thin slice implementation, then we could merge it in.

pkg/bpftrace/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apk add --update \
+    bison \
+    build-base \
+    clang-dev \
+    clang-static \
+    cmake \
+    elfutils-dev \
+    flex-dev \
+    git \
+    linux-headers \
```
rn: This compiles it against the headers from Alpine Linux but not the LinuxKit kernels. Is that a problem?

pkg/bpftrace/Dockerfile
rn: Adding some parallelism here would be good. It seems to take some time to compile...

pkg/bpftrace/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    && cd /bpftrace/ \ 
+    && mkdir -p /bpftrace/build \
+    && git checkout ${COMMIT} \
+    && chmod +x /bpftrace/docker/build.sh \
+    && /bpftrace/docker/build.sh /bpftrace/build Release
+
+FROM scratch
+WORKDIR /
+COPY --from=build /bpftrace/build/src/bpftrace /bpftrace
+ENTRYPOINT [ "/bpftrace" ]
```
rn: I would remove this and the next line. The `bpftrace` is unlikely to be used as a stand alone package running in a container. to be consistent with other package file make this:
```
FROM scratch
ENTRYPOINT []
CMD []
WORKDIR /
COPY --from=build /bpftrace/build/src/bpftrace /usr/bin/bpftrace
```
or similar.

pkg/bpftrace/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN git clone ${GITREPO} /bpftrace \
+    && cd /bpftrace/ \ 
+    && mkdir -p /bpftrace/build \
+    && git checkout ${COMMIT} \
+    && chmod +x /bpftrace/docker/build.sh \
+    && /bpftrace/docker/build.sh /bpftrace/build Release
+
+FROM scratch
+WORKDIR /
+COPY --from=build /bpftrace/build/src/bpftrace /bpftrace
```
rn: Please copy this to a standard location, like `/usr/bin`

pkg/bpftrace/build.yml
```diff
@@ -0,0 +1,8 @@
+image: bpftrace
+network: true
+arches: 
+  - amd64
+config:
```
rn: This can be dropped if if the main consumption model is to drop the binary into the root filesystem

test/cases/040_packages/002_bpftrace/test.yml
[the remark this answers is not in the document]
rn: This should be `linuxkit/bpftrace:<hash>`. To avoid the chicken and egg, I will compile and push the package once and restart the CI.

pkg/bpftrace/Dockerfile
simar7: Hm – not sure about that. I seem to think this will be fine as I was able to run the compiled packaged but is there any other way I could figure that out?

pkg/bpftrace/Dockerfile
simar7: ACK. Makes sense.

pkg/bpftrace/Dockerfile
simar7: ACK.

simar7:
Thanks for looking into this @rn. I've fixed the PR with the suggestions made.

Issue #3245, opened 2018-12-16
Mounting an encrypted filesystem


chriswue:
I spent several days on building a linuxkit image which allows me to mount an encrypted file system backed by a loop device (essentially making this work on linuxkit: http://goohackle.com/how-to-create-a-portable-encrypted-file_system-on-a-loop-file/). 

Ended up introducing two utility containers: one to run `losetup` and one to run `cryptsetup`.

Is there any interest in upstreaming this? If yes then I'll tidy it up and add documentation etc.

justincormack:
No, I haven't seen an example, and there is definitely interest, so this would be helpful, yes.

chriswue:
@justincormack: Currently the utility containers I have are shell scripts - looking over the repo I guess you guys would prefer the container entrypoint to be in go?

justincormack:
Shell scripts are better than nothing! `losetup` is only a line or so of Go code to do `mount` but haven't looked at the internals of `cryptsetup` it could be a whole lot of work...

PR #3249, opened 2018-12-20
Adding support to mount an encrypted filesystem 

chriswue:
**- What I did**
The PR has 3 commits:
1. A losetup utility container allowing to setup a loop device backed by a regular file, including test and docs
2. A dm-crypt container allowing to map an device with dm-crypt including tests and docs
3. A couple of examples

Obviously I have no access to the linuxkit hub so a contributor would have to built+push the containers (all .yml files assume that this will happen as they use linuxkit/... images)

Fixes: https://github.com/linuxkit/linuxkit/issues/3245

**- How I did it**
Tried to follow https://github.com/linuxkit/linuxkit/blob/master/docs/packages.md and practices and structure of the other existing packages.

**- How to verify it**
Build it locally with: 
```
linuxkit pkg build pkg/losetup
linuxkit pkg build pkg/dm-crypt

linuxkit build -format kernel+initrd examples/dm-crypt.yml
linuxkit run -disk size=20M dm-crypt
```

Adding support and examples for loop device and dm-crypt support

sorry no time to sift through the copyright implications

### DIFF
--- a/pkg/dm-crypt/Dockerfile
+++ b/pkg/dm-crypt/Dockerfile
@@ -1,0 +1,19 @@
+FROM linuxkit/alpine:3683c9a66cd4da40bd7d6c7da599b2dcd738b559 AS mirror
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out  \
+    alpine-baselayout \
+    cryptsetup \
+    e2fsprogs
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY crypto.sh /usr/bin/crypto
+RUN chmod +x /usr/bin/crypto
+
+CMD ["/usr/bin/crypto"]
--- a/pkg/dm-crypt/crypto.sh
+++ b/pkg/dm-crypt/crypto.sh
@@ -1,0 +1,84 @@
+#!/bin/sh
+
+set -e
+
+help()
+{
+    echo "Usage: $0 [options] <dm_name> <device>"
+    echo
+    echo "Options:"
+    echo "    -l|--luks       Use LUKS extension"
+    echo "    -k|--key-file   Name of the key file, default: key"
+    echo "    <dm_name>       Name of the device mapper file, the encrypted device will become available under /dev/mapper/<dm_name>"
+    echo "    <device>        The encrypted device (e.g. /dev/sda1, /dev/loop0, etc)"
+    echo
+}
+
+luks=false
+key_file="key"
+
+O=`getopt -l key-file:luks,help -- k:lh "$@"` || exit 1
+eval set -- "$O"
+while true; do
+    case "$1" in
+        -l|--luks)      luks=true; shift;;
+        -k|--key-file)  key_file=$2; shift 2;;
+        -h|--help)      help; exit 0;;
+        --)             shift; break;;
+        *)              echo "Unknown option $1"; help; exit 1;;
+        esac
+done
+
+if [ -z "$1" ]; then
+    echo "Missing argument <dm_name>"
+    help
+    exit 1
+fi
+
+if [ -z "$2" ]; then
+    echo "Missing argument <device>"
+    help
+    exit 1
+fi
+
+dm_name=$1
+device=$2
+dmdev_name="/dev/mapper/$dm_name"
+cipher="aes-cbc-essiv:sha256"
+
+case "$key_file" in
+    /*) ;;
+    *)  key_file="/etc/dm-crypt/$key_file" ;;
+esac
+
+if [ ! -f "$key_file" ]; then
+    echo "Couldn't find encryption keyfile $key_file!"
+    exit 1
+fi
+
+if [ ! -d "/run/cryptsetup" ]; then
+    echo "Creating cryptsetup lock directory"
+    mkdir /run/cryptsetup
+fi
+
+if [ $luks = true ]; then
+    echo "Creating dm-crypt LUKS mapping for $device under $dmdev_name"
+    if ! cryptsetup isLuks $device; then
+        echo "Device $device doesn't seem to have a valid LUKS setup so one will be created."
+        cryptsetup --key-file "$key_file" --cipher "$cipher" luksFormat "$device"
+    fi
+    cryptsetup --key-file "$key_file" luksOpen "$device" "$dm_name"
+else
+    echo "Creating dm-crypt mapping for $device under $dmdev_name"
+    cryptsetup --key-file "$key_file" --cipher "$cipher" create "$dm_name" "$device"
+fi
+
+o=`blkid $dmdev_name`
+if [ -z "$o" ]; then
+	echo "Device $dmdev_name doesn't seem to contain a filesystem, creating one."
+    # dd will write the device until it's full and then return with an error because "no space left"
+    dd if=/dev/zero of="$dmdev_name" || true
+    mkfs.ext4 "$dmdev_name"
+else
+	echo "Device $dmdev_name seems to contain filesystem: $o"
+fi
--- a/docs/encrypted-disk.md
+++ b/docs/encrypted-disk.md
@@ -1,0 +1,86 @@
+# Device encryption with dm-crypt
+
+In the packages section you can find an image to setup dm-crypt encrypted devices in [linuxkit](https://github.com/linuxkit/linuxkit)-generated images.
+
+## Usage
+
+The setup is a one time step during boot:
+
+```yaml
+onboot:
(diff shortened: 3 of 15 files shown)


deitch:
Overall, I like this idea. I had planned at some point to make it able to boot off an encrypted filesystem, and then eventually store/retrieve keys from tpm, but haven't found the time yet.

I have a few small comments (inline).

Separately, CI is failing, but that looks like trust issues, mainly because the images don't exist and weren't pushed out.

Like to get some comments from the other maintainers, in addition to these.

Nice work!

pkg/losetup/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY loopy.sh /loopy.sh
```
deitch: Same comment as with `crypto.sh`. I think this should just be 

```Dockerfile
COPY loopy.sh loopy
```

This requires changing it everywhere else in this PR, but search-and-replace should make it easy.

deitch: BTW, I love that you called it "loopy". :-)

deitch:
I also think some of the docs should go in [docs/](https://github.com/linuxkit/linuxkit/tree/master/docs) rather than in the specific `pkg/`. Someone who already knows about crypt or losetup will look there, but someone looking to answer, "how can I do a loopback device" or "how can I use an encrypted filesystem" will start by going to [docs/](https://github.com/linuxkit/linuxkit/tree/master/docs)

pkg/dm-crypt/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY crypto.sh /crypto.sh
```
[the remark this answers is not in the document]
chriswue: good point, should be fixed now

pkg/losetup/Dockerfile
chriswue: was inspired by "mountie" :)
fixed as well

chriswue:
@deitch: I expanded the dm-crypt documentation a little bit and moved it into docs/

Thanks for the review

pkg/dm-crypt/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY crypto.sh /crypto
```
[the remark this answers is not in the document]
deitch: @rn that was my request. This is in `sh` now, but will likely convert to pure binary (along with everything else we do). Keeping it named just `crypto` made it consistent as "just an executable". Same for `loopy`.

pkg/losetup/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY loopy.sh /loopy
```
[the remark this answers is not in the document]
deitch: Ditto

pkg/dm-crypt/Dockerfile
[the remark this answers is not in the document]
chriswue: Ok, so I'll move it into `/usr/bin` but keep the name as is. I tried to avoid the term `mount` since it's not actually mounting anything - it's mapping a device.

pkg/losetup/loopy.sh
```diff
@@ -0,0 +1,57 @@
...
+
+container_file=$1
+
+if [ ! -b "$loop_device" ]; then
+    echo "Loop device $loop_device doesn't exist! Did you forget to bind-mount '/dev'?"
+    exit 2
+fi
+
+if [ ! -f "$container_file" ]; then
+    if [ $create = true ]; then
```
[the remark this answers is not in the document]
chriswue: Well, there didn't seem to a nice way to create a file of arbitrary size in linuxkit. I'm using the loop device as a portable storage location hence creating it if it doesn't exist made things easier. I suppose one could make another package to create files of arbitrary size if they don't exists but not sure if that's starting to get a bit too granular.

deitch:
I am good with this, subject to @rn signing off as well. @rn, if you are ok with it, I am happy to build the packages on arm64/amd64 and push them out so we can see ci pass (or fail).

pkg/dm-crypt/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY crypto.sh /usr/bin/crypto
```
rn: This still has the `crypto.sh` -> `crypto` rename which neither @deitch nor me liked. Could you fix this?

pkg/losetup/README.md
```diff
@@ -0,0 +1,27 @@
+# LinuxKit losetup
```
rn: Could you move this documentation to `./docs` as also @deitch suggested? One reason to have it in `./docs` is that if it stays here and someone fixes, say a typo, then the content hash of the package changes and we'd have to rebuild, although the package itself did not change.

pkg/losetup/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Remove apk residuals
+RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=mirror /out/ /
+
+COPY loopy.sh /usr/bin/loopy
```
rn: this also has the `loopy.sh` -> `loopy` rename

pkg/dm-crypt/crypto.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "Creating dm-crypt mapping for $device under $dmdev_name"
+    cryptsetup --key-file "$key_file" --cipher "$cipher" create "$dm_name" "$device"
+fi
+
+o=`blkid $dmdev_name`
+if [ -z "$o" ]; then
+	echo "Device $dmdev_name doesn't seem to contain a filesystem, creating one."
+    # dd will write the device until it's full and then return with an error because "no space left"
+    dd if=/dev/zero of="$dmdev_name" || true
+    mkfs.ext4 "$dmdev_name"
```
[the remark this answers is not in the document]
rn: any reason this is not formated with the `format` package?

pkg/losetup/Dockerfile
chriswue: same as per https://github.com/linuxkit/linuxkit/pull/3249#discussion_r243295417

pkg/dm-crypt/Dockerfile
chriswue: Well, it was actually @deitch who suggested to do it like this, see https://github.com/linuxkit/linuxkit/pull/3249#discussion_r243295417

pkg/dm-crypt/crypto.sh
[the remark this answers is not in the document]
chriswue: That's what I tried first but since the dm-crypt device has no partition table `format` will try to create a partition and `sfdisk` chokes since it thinks the device is already in use. Unless there is an option I missed for `format` which makes it generate the fs only?

pkg/dm-crypt/Dockerfile
deitch: Guilty as charged. I just wanted it to be `crypto` in the final package, not `crypto.sh`. @rn wanted it to be consistent from source to final package. I don't mind it being `crypto.sh` if it is hidden, but it is exposed when you call it, hence the desire for `crypto` (without the `.sh` extension). Maybe we should just keep it `crypto` I both src and pkg?

pkg/losetup/Dockerfile
deitch: Yes again. If @rn is ok with it, let's just name it `loopy` in the src and be consistent. I would say that is what we do elsewhere, but I think almost all of the `sh` code is replaced by `go`-sourced binaries by now. :-)

rn:
Thanks for the response. let's leave the `.sh` rename out. Could you just move the doc from ` pkg/losetup/README.md` either to `./docs` or incorporate it into crypto setup doc?

chriswue:
PR updated with correct SHA

PR #3254, opened 2018-12-29
Add runtime.envFromFiles

errordeveloper:


pkg/init/cmd/service/prepare.go
```diff
@@ -195,6 +196,28 @@ func prepareFilesystem(path string, runtime Runtime) error {
 	return nil
 }
 
+// prepareSpec appends spec based on runtime config;
+// NOTE these changes are runtime-only, we don't update config.json after this
```
errordeveloper: To be honest I was a little surprised by this, maybe it's something that we should fix. It seems like a plausible thing for one to look at the file when debugging it.

src/cmd/linuxkit/moby/config.go
```diff
@@ -110,12 +110,13 @@ type ImageConfig struct {
...
-	Interfaces *[]Interface   `yaml:"interfaces,omitempty,omitempty" json:"interfaces,omitempty"`
-	BindNS     Namespaces     `yaml:"bindNS,omitempty" json:"bindNS,omitempty"`
-	Namespace  *string        `yaml:"namespace,omitempty" json:"namespace,omitempty"`
+	Cgroups     *[]string          `yaml:"cgroups,omitempty" json:"cgroups,omitempty"`
+	Mounts      *[]specs.Mount     `yaml:"mounts,omitempty" json:"mounts,omitempty"`
+	Mkdir       *[]string          `yaml:"mkdir,omitempty" json:"mkdir,omitempty"`
+	Interfaces  *[]Interface       `yaml:"interfaces,omitempty,omitempty" json:"interfaces,omitempty"`
+	BindNS      Namespaces         `yaml:"bindNS,omitempty" json:"bindNS,omitempty"`
+	Namespace   *string            `yaml:"namespace,omitempty" json:"namespace,omitempty"`
+	EnvFromFile *map[string]string `yaml:"envFromFile,omitempty" json:"envFromFile,omitempty"`
```
errordeveloper: It wasn't very clear that I was supposed to do this, and having a reference to a map also felt a little odd.

src/cmd/linuxkit/moby/config.go
```diff
@@ -590,6 +591,7 @@ func assignRuntime(v1, v2 *Runtime) Runtime {
 	runtimeMkdir := assignStrings(v1.Mkdir, v2.Mkdir)
 	runtimeInterfaces := assignRuntimeInterfaceArray(v1.Interfaces, v2.Interfaces)
 	runtimeNamespace := assignString(v1.Namespace, v2.Namespace)
+	runtimeEnvFromFile := assignMaps(v1.EnvFromFile, v2.EnvFromFile)
```
errordeveloper: I didn't realise I had to do this had to spend some time debugging why my change didn't work. Perhaps we need to comment structs with references to other copies of the same struct definition and this function also.

src/cmd/linuxkit/moby/schema.go
```diff
@@ -249,7 +249,8 @@ var schema = string(`
         "mkdir": {"$ref": "#/definitions/strings"},
         "interfaces": {"$ref": "#/definitions/interfaces"},
         "bindNS": {"$ref": "#/definitions/namespaces"},
-        "namespace": {"type": "string"}
+        "namespace": {"type": "string"},
+        "envFromFile": {"$ref": "#/definitions/mapstring"}
```
errordeveloper: This wasn't immediately obvious either.

errordeveloper:
Just a few thoughts on my contributor experience, hopefully useful :)

src/cmd/linuxkit/moby/config.go
errordeveloper: Also, I'm not sure if this has to be a map - `env` is a list, but this has different meaning really. I'm not feeling too strongly about it and happy to change.

errordeveloper: Perhaps we should change it ot `envFromFiles` actually? ...unless someone has ideas for even better name?

pkg/init/cmd/service/prepare.go
justincormack: The filesystem is read only usually, so it cannot just be updated.

src/cmd/linuxkit/moby/config.go
justincormack: From memory, they are pointers so they get totally omitted from the generated JSON file if not set, rather than being set to an empty list or map, so the JSON is less verbose. Could add a comment to this effect.

justincormack: yeah `envFromFiles` probably better. I think a map is best, we try to avoid microformats...

src/cmd/linuxkit/moby/config.go
justincormack: It is kind of ugly and verbose, maybe we could replace all this with `reflect`?

src/cmd/linuxkit/moby/config.go
errordeveloper: Sure, happy to make it plural and go with a map.

src/cmd/linuxkit/moby/config.go
errordeveloper: Not too sure. Perhaps much of these concerns can be solved by comments that someone updating either of the structs will read. It's okay that this is not pretty, just would be nice to know all parts of the code that need updating. I don't know if this is just the case with `Runtime`, or there are other structs that are handled in the same way like this.

pkg/init/cmd/service/prepare.go
errordeveloper: Ah, obviously... :)

src/cmd/linuxkit/moby/config.go
justincormack: I think this is the only case, could probably be fixed by exporting the types from a library instead I guess (I can't remember why it is like this and I don't think it has to be).

errordeveloper:
@justincormack I've updated this. Let me know if you want me to open another PR with https://github.com/errordeveloper/linuxkit/commit/3141cee90cf7be7b451121e39dffbfaf4439d50d.

errordeveloper:
I've rebased this.

errordeveloper:
@justincormack ping

errordeveloper:
@justincormack ping

errordeveloper:
@justincormack would you mind to take a look at this once again please?

Issue #2813, opened 2017-12-15
Should builds be reproducible?


dprotaso:
**Description**

When building ISOs using the `linuxkit build` command multiple builds with the same configuration will result in different artifacts.

**Steps to reproduce the issue:**

In the examples folder
```
$ linuxkit build -format iso-efi sshd.yml
$ mv sshd-efi.iso sshd-efi-first.iso
$ linuxkit build -format iso-efi sshd.yml
$ diffoscope sshd-efi.iso sshd-efi-first.iso
...
```

**Describe the results you received:**

Within the ISO the OCI config.json output will vary across builds due to ordering of nested array elements.

**Describe the results you expected:**

The ISOS to be identical

**Additional information you deem important (e.g. issue happens only occasionally):**

xtuc:
Linuxkit is meant for "Immutable infrastructure", builds must be deterministic.

The ordering change is probably due to Golang (iteration over a list is guaranty to be random). I think it's fine since the resulting container is still the same?

ijc:
Builds are currently not expected to be bit for bit identical, although they _should_ be functionally identical (modulo things like `files` entries which reference external files).

Personally I think it would be good to strive for bit for bit reproducibility not just in the image build but in package build too (the former is likely to be much easier to achieve), but so fart as I know nobody is working on either right now.

So depending on how you meant "Should" the answer is either "yes" or "no"  ;-)

> The ordering change is probably due to Golang (iteration over a list is guaranty to be random). 

Hopefully you meant "map" and not "list" (or array or slice)?

Can we fix that by sorting some elements or is this something embedded in `encoding/json` and so inaccessible to us?

xtuc:
@ijc yes I meant maps. As far as I know the `encoding/json` serialization also uses Golang's map iteration.

deitch:
They are functionally identical, but it would be good to get it bit-for-bit identical such that `diff` shows identical files.

@ijc do we have any metadata on/in the image that might include a timestamp? Actually, I am sure we do. Some of the image builds for filesystems or partitioning will use date/time stamps as defaults for various in-filesystem bits. I do not recall which off-hand, but I do recall it is there.

If/when we get the tar or kernel+initrd output to be bit-for-bit identical, someone can open an issue to make the built image files identical and assign it to me, and I will track it down. 

Hopefully, but that point, the native build-image-in-go will be complete and merged, and will make it much easier.

justincormack:
[...]

PR #3255, opened 2018-12-29
Initial support for reproducible builds

rn:
This PR adds initial support for reproducible `linuxkit build`s (wip #2813).

- Files created during the build have their timestamp fixed
- Slices/Arrays are sorted before being marshalled to JSON

Currently, the `tar`, `tar-kernel-initrd`, `docker`, and `kernel+initrd` output formats support reproducible builds and tests for `tar` and `kernel+initrd` are added with this PR.

ladybugs

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ LinuxKit, a toolkit for building custom minimal, immutable Linux distributions.
 - Completely stateless, but persistent storage can be attached
 - Easy tooling, with easy iteration
 - Built with containers, for running containers
+- Designed to create [reproducible builds](./docs/reproducible-builds.md) [WIP]
 - Designed for building and running clustered applications, including but not limited to container orchestration such as Docker or Kubernetes
 - Designed from the experience of building Docker Editions, but redesigned as a general-purpose toolkit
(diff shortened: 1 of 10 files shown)


deitch:
This is _rather_ nice!

Is there concern that the modtime now is just the timestamp that lkt open sourced? Might we want to have a cmdline option to set the date-time as an input, with a default for that time, and maybe an option that says, "now"?

E.g. 

```
$ lkt build --modtime now ...
$ lkt build --modtime 20181229T175500Z ...
```

Or `-timestamp` or some other flag name, but that idea? And if not specified, just uses the date/time lkt was open-sourced?

rn:
Yeah, was considering making the timestamp configurable. But rather than via the commandline, we should probably support [this proposal](https://reproducible-builds.org/docs/source-date-epoch/) and make it configurable via the environment variable `SOURCE_DATE_EPOCH`

deitch:
If that’s where it’s going as a standard, I guess. I like command line options as they are explicit. But not going to be a stickler about it. :-)

justincormack:
LGTM as a first pass. I think setting all times to the git commit time of the linuxkit build itself might also be a reasonable option in future, but a static time is fine for now.

PR #3261, opened 2019-01-07
Feature/logging compression

teeg82:
**- What I did**
 - Modified `pkg/logwrite` to include a boolean option to compress log files as they rotate
    - The `-compress` option defaults to false, to preserve the existing behaviour.
 - Updated logging documentation to mention compression option
 - This feature was born from this discussion (https://github.com/linuxkit/linuxkit/issues/3072#issuecomment-450986701)

**- How I did it**
This is a modification of the original logging infrastructure created by @djs55 in #3095

- Added new function to handle compression using `compress/gzip` package
- Compression occurs immediately prior to rotation if enabled.
  - For clarity, compressed files will be appended with `.gz`
      - File lookup during rotation will attempt to locate files expecting this extension when compression is enabled. 
      - Note: With this implementation, this means that if compression is enabled after a period of time, the existing uncompressed logs will be ignored forever. The same occurs for the compressed logs if disabling compression after a period of time. Basically, toggling back and forth may create some left over files. This doesn't seem like it should be a big deal, but I'm happy to put some more work into it to make it more robust if needed.

**- How to verify it**

1. Create a logging.yml file with the following:

   ```
    kernel:
      image: linuxkit/kernel:4.14.91
    init:
      - linuxkit/init:c563953a2277eb73a89d89f70e4b6dcdcfebc2d1
      - linuxkit/runc:83d0edb4552b1a5df1f0976f05f442829eac38fe
      - linuxkit/containerd:326b096cd5fbab0f864e52721d036cade67599d6
      - linuxkit/ca-certificates:v0.6
      - linuxkit/memlogd:v0.6
    onboot:
      - name: sysctl
        image: linuxkit/sysctl:v0.6
      - name: dhcpcd
        image: linuxkit/dhcpcd:v0.6
    services:
    # Inside the getty type `/proc/1/root/usr/bin/logread -F` to follow the log
      - name: getty
        image: linuxkit/getty:2eb742cd7a68e14cf50577c02f30147bc406e478
        env:
         - INSECURE=true
    # A service which generates log messages for testing
      - name: write-to-the-logs
        image: alpine:3.8
        command: ["/bin/sh", "-c", "while /bin/true; do echo hello $(date); sleep 1; done" ]
      - name: logwrite
        image: linuxkit/logwrite
        command: ["/usr/bin/logwrite", "-max-log-size", "20480", "-compress", "true"]
      - name: kmsg
        image: linuxkit/kmsg:v0.6
    trust:
      org:
        - linuxkit
        - library
1. `linuxkit build -disable-content-trust logging.yml`
1. `linuxkit run logging`
1. In getty, run `watch ls -l /var/log`
1. You should expect to see a file called `write-to-the-logs.log` filling up to about 20kb, before rotating. After several minutes, you should expect to see at least 1 rotated compressed file a fraction of the size of the original, marked with `.gz`.

Example output:

```
Every 2s: ls -l                                             2019-01-07 18:57:47

total 68
-rw-r--r--    1 root     root           405 Jan  7 18:29 getty.out.log
-rw-r--r--    1 root     root          3352 Jan  7 18:29 kmsg.log
-rw-r--r--    1 root     root          4492 Jan  7 18:29 kmsg.log.0.gz
-rw-r--r--    1 root     root          4820 Jan  7 18:29 kmsg.log.1.gz
-rw-r--r--    1 root     root            45 Jan  7 18:29 memlogd.log
-rw-r--r--    1 root     root          1047 Jan  7 18:29 onboot.001-dhcpcd.log
-rw-r--r--    1 root     root         11935 Jan  7 18:57 write-to-the-logs.out.log
-rw-r--r--    1 root     root          1471 Jan  7 18:55 write-to-the-logs.out.log.0.gz
-rw-r--r--    1 root     root          1521 Jan  7 18:50 write-to-the-logs.out.log.1.gz
-rw-r--r--    1 root     root          1473 Jan  7 18:45 write-to-the-logs.out.log.2.gz
-rw-r--r--    1 root     root          1536 Jan  7 18:41 write-to-the-logs.out.log.3.gz
-rw-r--r--    1 root     root          1472 Jan  7 18:36 write-to-the-logs.out.log.4.gz
-rw-r--r--    1 root     root          1519 Jan  7 18:32 write-to-the-logs.out.log.5.gz
```

<!--
Write a short (one line) summary that describes the changes in this
pull request for inclusion in the changelog:
-->
Adding option to allow log files to compress after rotation

![java1](https://user-images.githubusercontent.com/5384181/50787069-5950e400-1283-11e9-8b30-1c0d290ec344.jpg)


teeg82:
@GordonTheTurtle I believe I've fixed commit signature

justincormack:
Hey, doesn't look like your commits are signed. It might be easier if you squash them into one first.

rn:
This seems to have fallen through the cracks. @djs55 care to review this as this mostly touches code you added?

@teeg82 once this is reviewed, can you add a commit which updates the YAML files using the `linuxkit/logwrite` package with the output of `linuxkit pkg show-tag .` in the `./pkg/logwrite` directory (using `./scripts/update-component-sha.sh`. This will cause CI to pick up the new package and it will fail as it has not been signed/pushed yet. I will do that and then kick CI again

teeg82:
Will do

pkg/logwrite/logwrite.go
```diff
@@ -100,19 +109,71 @@ func (l *LogFile) Close() error {
...
+
+	if err != nil {
+		return err
+	}
+
+	// Create a GZIP writer
+	var outputBuffer bytes.Buffer
+	writer := gzip.NewWriter(&outputBuffer)
+	writer.Write(rawBytes)
+	writer.Close()
```
djs55: I would recommend (if possible)
- use the simpler `os.Create` and `os.Open` functions rather than `os.OpenFile`;
- compress in a streaming fashion to avoid excessive buffering; which requires
- compressing to a new file (which you can rename afterwards if you want).

Maybe something like:
```go
input, err := os.Open(l.Path)
if err != nil {
    return err
}
defer input.Close()

output, err := os.Create(l.Path + gzipFileExtension)
if err != nil {
    return err
}
defer output.Close()

gzipWriter := gzip.NewWriter(output)
if _, err := io.Copy(gzipWriter, input); err != nil {
    return err
}
if err := gzipWriter.Close(); err != nil {
    return err
}
// delete the original?
if err := os.Remove(l.Path); err != nil {
    return err
}
// now we only have l.Path + gzipFileExtension
return nil
```

teeg82:
Thanks @djs55, I will look into implementing that suggestion shortly. Its been a hell of a week. That suggestion sounds reasonable though. Worst case, Monday.

pkg/logwrite/logwrite.go
teeg82: Thanks @djs55, I will look into implementing that suggestion shortly. Its been a hell of a week. That suggestion sounds reasonable though. Worst case, Monday.

teeg82: apologies, I haven't forgotten about this, just been extremely busy. Hopefully should have a bit of time soon.

PR #3262, opened 2019-01-08
Containerd v1.2.2

ijc:
Picks up v.1.2.1 through to 1.2.2.
https://github.com/containerd/containerd/releases/tag/v1.2.1
https://github.com/containerd/containerd/releases/tag/v1.2.2

Slight tweak needed to our rebuild of the wireguard userspace since the upstream/edge apk we rebuild gained another file at some point.


pkg/runc/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN \
   make \
   && true
 ENV GOPATH=/go PATH=$PATH:/go/bin
-ENV RUNC_COMMIT=58592df56734acf62e574865fe40b9e53e967910
```
rn: shouldn't this also bump the `linuxkit/alpine` hash?

ijc: I'm  a couple of mins away from pushing the bump, so we may as well wait for the CI again.

rn: did you also rebuild the `runc` package and bumped the hash? I don't trust the GH ordering of commits...

ijc: Yes, I bumped the hash in the same commit (the only new one, which is now the head of the branch) as I changed the `FROM` and I rebuilt manually on all three arches.

PR #3269, opened 2019-01-16
Switch the default kernel to 4.19.x

### LINKED ISSUE #3268: Switch to 4.19 kernel as the default LinuxKit kernel
4.19 is the latest LTS and it had a good burn in time of 14 dot releases. Switch to using it as the new default kernel.

rn:
4.19.x is the new LTS kernel and has been out for a while. Switch
all examples and tests to using it instead of the 4.14.x kernel.
resolves #3268 

pink-dolphin

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:4.14.92
+  image: linuxkit/kernel:4.19.14
   cmdline: "console=ttyS0"
 init:
   - linuxkit/init:e0dced91adbfba34c53cd673760e011f410638ff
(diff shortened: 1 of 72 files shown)


thaJeztah:
Do we know if https://github.com/torvalds/linux/commit/7056d3a37d2c6aaaab10c13e8e69adc67ec1fc65 has been back ported to 4.19? See https://github.com/containerd/cgroups/issues/74

justincormack:
It is in 4.19.15 https://cdn.kernel.org/pub/linux/kernel/v4.x/ChangeLog-4.19.15

justincormack:
So not in this PR but in the next kernel update....

rn:
> So not in this PR but in the next kernel update....

which is coming up soon (later today)

rn:
there isn't really a stable version of hyperkit. I think @ijc tags hyperkit whenever docker desktop updates the version used. Maybe it's time for the Docker Desktop folks to update. /cc @djs55

ijc:
In the end that only led me to tag v0.20180621, which is not new enough for the fix in #239.

PR #3273, opened 2019-01-19
Use latest stable release of ZFS

### LINKED ISSUE #2950: zfs does not build on 4.14
Probably needs a later zfs version.

errordeveloper:
This change makes `make -C kernel build_zfs_4.14.x` succeed.

Closes #2950.

### DIFF
--- a/kernel/Dockerfile.zfs
+++ b/kernel/Dockerfile.zfs
@@ -23,10 +23,10 @@ COPY --from=ksrc /kernel.tar /
 RUN tar xf kernel.tar
 
 # Note: ZFS and SPL commits must match. It's unclear how much the user
-# space tools must match the kernel module version. The current zfs
-# package on Alpine is 0.6.5.9. We pick 0.6.5.10 because it has
-# support for 4.12 based kernels.
-ENV VERSION=0.6.5.10
+# space tools must match the kernel module version.
+# package on Alpine is 0.6.5.9. We pick the version that compiles with
+# latest kernel we support.
+ENV VERSION=0.7.12
 
 ENV SPL_REPO=https://github.com/zfsonlinux/spl.git
 ENV SPL_COMMIT=spl-${VERSION}


justincormack:
Does this work on 4.19? Given we switched default kernel to 4.19 maybe makes sense to switch (although having a 4.14 version checked in may also be useful)

errordeveloper:
I've tested `make -C kernel build_zfs_4.9.x` just now, and that worked. Will try 4.19.x.

errordeveloper:
By the way, they have [`0.8.0-rc3`](https://github.com/zfsonlinux/zfs/tree/zfs-0.8.0-rc3)... Not too sure as it's an RC, so I'd rather stick to latest 0.7.x.

errordeveloper:
@justincormack whatever is the outcome of this, should we enable ZFS in CI, and possibly push the images also?

errordeveloper:
I've not been able to boot the 4.19 one (getting meaningless error from hyperkit), but 4.14 appears to load fine (see below).

```
(ns: getty) linuxkit-66511097e4f7:~# nsenter -t 1 -a 
linuxkit-66511097e4f7:/# modprobe zfs
[  171.206712] spl: loading out-of-tree module taints kernel.
[  171.206712] spl: loading out-of-tree module taints kernel.
[  171.211434] SPL: Loaded module v0.7.12-1
[  171.211434] SPL: Loaded module v0.7.12-1
[  171.215384] icp: module license 'CDDL' taints kernel.
[  171.215384] icp: module license 'CDDL' taints kernel.
[  171.217719] Disabling lock debugging due to kernel taint
[  171.217719] Disabling lock debugging due to kernel taint
[  173.596814] ZFS: Loaded module v0.7.12-1, ZFS pool version 5000, ZFS filesystem version 5
[  173.596814] ZFS: Loaded module v0.7.12-1, ZFS pool version 5000, ZFS filesystem version 5
linuxkit-66511097e4f7:/# grep zfs /proc/filesystems 
nodev	zfs
linuxkit-66511097e4f7:/# uname -a
Linux linuxkit-66511097e4f7 4.14.94-linuxkit #1 SMP Sat Jan 19 10:50:24 UTC 2019 x86_64 Linux
linuxkit-66511097e4f7:/# 
```

rn:
Last time I looked at ZFS one issue was that the version of the tools which came with Alpine did not quite match the version of the kernel module and there was some doubt if they'd work.

We currently do not build official images for ZFS and therefore don't have it in the CI because none of us fancied taking it up with the lawyers (I still don't) so we just provide the mechanism to build it yourself.

errordeveloper:
The userspace compatibility is something we should confirm, I shall play
around with it. The question of shipping it is understood, but I also
wonder if running buidls against it in CI is quite doable, which would let
us observe breakages early.

rn:
> (getting meaningless error from hyperkit)

What type of errors? Might be related to https://github.com/moby/hyperkit/issues/226 which was fixed a while back. I don't know if this has been pulled into the hyperkit shipped with Docker Desktop. Maybe try the qemu backend, where latest versions should also use the hypervisor framework for acceleration

errordeveloper:
@rn yes, I think I got that same error, but with the staircase effect :)

Should we merge and open an issue about userspace?

PR #3278, opened 2019-01-27
Add rsync in sshd package

errordeveloper:
The `rsync` command has to be available on both ends.

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -18,7 +18,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:v0.6
   - name: sshd
-    image: linuxkit/sshd:c4bc89cf0d66733c923ab9cb46198b599eb99320
+    image: linuxkit/sshd:a0ac44ab1cbfd14e368ed139c6a4e128f2a33513
     binds:
      - /run/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -15,7 +15,7 @@ services:
   - name: dhcpcd
     image: linuxkit/dhcpcd:v0.6
   - name: sshd
-    image: linuxkit/sshd:c4bc89cf0d66733c923ab9cb46198b599eb99320
+    image: linuxkit/sshd:a0ac44ab1cbfd14e368ed139c6a4e128f2a33513
 files:
   - path: root/.ssh/authorized_keys
     source: ~/.ssh/id_rsa.pub
--- a/examples/gcp.yml
+++ b/examples/gcp.yml
@@ -22,7 +22,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:v0.6
   - name: sshd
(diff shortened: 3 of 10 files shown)


errordeveloper:
I am not sure why CI gives me this error:
```
Failed to create OCI spec for linuxkit/sshd:a0ac44ab1cbfd14e368ed139c6a4e128f2a33513: Trusted pull for linuxkit/sshd:a0ac44ab1cbfd14e368ed139c6a4e128f2a33513 failed: No valid trust data for a0ac44ab1cbfd14e368ed139c6a4e128f2a33513
```

From previous experience, I recalled that I need to bump all uses of a package in the same PR/commit that introduces the changes to the package itself - does CI now assume that changes are made separately?

deitch:
> I'm not sure we want the sshd base layer to be the union of anything anyone might ever want in that context. e.g. we currently IIRC don't include the packages necessary to make scp "just work" (or perhaps I'm thinking of sftp, either way).

Tough call. I tend to agree with this sentiment, but then why do we have the wireguard package installed (see [here](https://github.com/linuxkit/linuxkit/blob/master/pkg/sshd/Dockerfile#L13))? 

I think this might be easier if there were a simple way for a consumer of `lkt` to just use the standard `linuxkit/sshd` package, but yet have the desired tools made available without having to extend the base. Mounted in? Standardly available dir? Unlike most `pkg/` parts, `sshd` (it keeps auto-correcting to `sushi`) is a CLI, so people will want things in there that they find useful, and so we are likely to get more and more of these, and more and more discussion of what is "standard" and "base" and "commonly used". We definitely _have_ an answer with "extend it in a `Dockerfile` with what you want," but I am wondering if we could have an easier one?

errordeveloper:
> Tough call. I tend to agree with this sentiment, but then why do we have the wireguard package installed (see here)?

Yes, inderd. I had second thoughts, but after finding wireguard, I figured rsync might be reasonable to add.

deitch:
I just looked at the history; wireguard has been in since July 2017, when it was added to everything else. I don't recall the PR that added it, or the discussions around it, offhand. I do recall other discussions around adding more "standard" packages to getty and sshd, and usually they were held back to keep it minimal.

I would like to see none of it in there, and one of:

* a way to mount in desired `apk` packages (or binaries) from another location, so they can be kept clean
* a way to add them in "on the fly", i.e. during `lkt build`

Either of these essentially comes down to some higher-order form of composition. We already have composition of layers and packages via `Dockerfile` to build OCI images, and we have composition of those images using a `linuxkit.yml` to build OS images via `lkt build`. There is some middle layer here, where lkt can know to do some additional composition.

I am just not sure what that is, and if it is worth it...

errordeveloper:
Making it easier to fork a package without copying much of the config would be great. For example, sshd package has many binds, and wouldn't want to make a copy of those, I would rather add one or two on top. Perhaps we could add 'from' field in pkg config?

deitch:
We have spoken in the past about some of the issues with appending to binds rather than overriding. Lots to do here for configurability...

rn:
I'm with @ijc on leaning towards not adding `rsync` to the `sshd` package. It seems a bit niche.

PR #3280, opened 2019-01-30
Extend CDROM provider to support any common type of disk/partition with ISO format

errordeveloper:


rn:
The diff in GH and git does not look right. Can you do a `git mv provider_cdrom.go provider_iso.go` and then make the changes. That way one actually sees the differences. Or even split this up into two commits. One which does the rename and one which adds the new functionality? That way one can see the actual changes.

It would also be good if the commit message explains more of the rationale for this change.

ijc:
`git mv` is no different to `git rm && git add` in terms of what is stored in the repo, there is no specific metadata recorded about such things. The "move" and "copy" display is entirely a presentation layer thing. IOW `git mv` wouldn't help here -- it's just that the old and new files differ enough that GH didn't guess there was a move involved while rendering the diff.

For the CLI you can control the heuristics a bit with the `-M` and `-C` options to `git show`, I suppose that isn't possible in the GH webui though.

Making it two commits would indeed be one way to avoid the problem though. and +1 to better commit messages...

errordeveloper:
Sounds good, will do! I have indeed use 'git mv'.

errordeveloper:
I've split the commits as advised.

rn:
Thanks for the update. the diff is not much easier to read.

I'm not sure I like this approach. This seems to mount/unmount all disk type devices instead of stopping when one suitable device is found.

errordeveloper:
> the diff is not much easier to read.

I thought https://github.com/linuxkit/linuxkit/pull/3280/commits/5c3bde1f775ac17feb658db0abb93531050141ec fairly trivial...

> I'm not sure I like this approach. This seems to mount/unmount all disk type devices instead of stopping when one suitable device is found.

That's just what it did before, only limited to all `/dev/sr*` devices. If we believe that behaviour should change, I'm more than happy to implement that in this PR, or perhaps separate PR would make sense.
Perhaps the current behaviour was somewhat satisfactory, considering that systems with many optical drives aren't very common. I agree that trying to mount each and every drive is not a good idea, perhaps pre-filtering on filesystem type would be a good idea? We could even match on a well-known ISO volume ID, e.g. something like 'LinuxKit metadata'.

rn:
Sorry, I meant that the diff is *now* much easier to read.

But the code is mounting every disk in the system. And only later, during probe, it goes through the providers and stops on the first suitable on. This doesn't look nice. Why not mount only during probe and stop mounting once a suitable device is found. I dn't think we want to make it more complicated with filesystem probing in whatnot, just stop mounting once you found a suitable device.

errordeveloper:
Sure, that makes sense. Happy to adjust the behaviour that way.

errordeveloper:
@rn please take a look at 247bdaf.

errordeveloper:
There is a challenge with USB devices, I ended up having to do this:

```yaml
onboot:
  - name: modprobe-usb-storage
    image: linuxkit/modprobe:v0.6
    command: ["modprobe", "usb-storage"]
  - name: wait-usb-storage
    image: linuxkit/alpine:3683c9a66cd4da40bd7d6c7da599b2dcd738b559
    binds:
      - /dev:/dev
    command:
      - sh
      - -c
      - until test -b /dev/sdc ; do sleep 1 ; done
  - name: metadata
    image: errordeveloper/metadata:9250ccfc8a4ba879989281b5f782bae2551521b9
```

I might find a better way, but this is what I have to do at the moment.

errordeveloper:
@rn would you mind to take another look at this please?

PR #3300, opened 2019-03-07
Removed fixed size arg from mkimage-gcp script.

danrzs:
**- What I did**
Removed size arg from the tools/mkimage-gcp/make-gcp.sh script virt-make-fs command as this fixed size was failing with larger images.

**- How I did it**
delete key

**- How to verify it**
We have a large file included in our 'files:' directive, so this is where we found this issue. When making a large (>1GB) image with the old make-gcp script you'd get a "failed: exit code 1". We no longer get this error with the included change.

Removed size arg from mkimage-gcp script.

elephants

### DIFF
--- a/tools/mkimage-gcp/make-gcp
+++ b/tools/mkimage-gcp/make-gcp
@@ -42,7 +42,7 @@ cd ..
 
 tar cf files.tar -C files .
 
-virt-make-fs --size=1G --type=ext4 --partition files.tar disk.raw
+virt-make-fs --type=ext4 --partition files.tar disk.raw
 
 guestfish -a disk.raw -m /dev/sda1 <<EOF
   upload /usr/lib/SYSLINUX/mbr.bin /mbr.bin


justincormack:
I think the reason this is there is that GCP said (maybe it has changed?) that 1G was the minimum allowed size.

danrzs:
Interesting. I've made a smaller image that seems to have been bumped to 1GB by google, or maybe 1GB is just the smallest size it reports.

If you're worried about that being an issue, we can make it '--size=+1GB'* . A tad inefficient, but guarantees minimum size.

\* - http://libguestfs.org/virt-make-fs.1.html - "For example: --size=+200M means enough space for the input files, and (approximately) an extra 200 MB free space."

thehhugg:
I wish I'd looked at open pull requests sooner, because we were having the exact same issue with the vmdk output. I'd been using bash magic to evaluate the actual size of the file tar, but your approach seems saner.

rn:
will submit a PR shortly which fixes up use of this new package

PR #3323, opened 2019-04-05
Update kernels to 5.0.6/4.19.33/4.14.110/4.9.167

rn:
Note the somewhat gross hack in the second commit. The kernel compile for 5.06 started failing with the error mentioned in the commit message. I couldn't find a conclusive culprit, but fixing up the libelf header files seemed like the best way forward.

seabright-chicken

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:4.19.32
+  image: linuxkit/kernel:4.19.33
   cmdline: "console=ttyS0"
 init:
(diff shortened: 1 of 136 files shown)


justincormack:
I think we should have installed `elfutils-dev` which corresponds to what is recommended elsewhere, not `libelf-dev` that we have installed and seems to be a different elf library...

rn:
thanks, will try that later today

rn:
@justincormack your suggestion worked. Updated the PR and rebuild the kernels

rn:
The `kmod` build tests failed due to incompatible `libelf` (it installed `libelf-dev`. A new commit fixes this

PR #3327, opened 2019-04-14
Update containerd to v1.2.6

rn:
Also bump runc to the recommended version.

weasel-riding-woodpecker

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -56,7 +56,7 @@ RUN go get -u github.com/LK4D4/vndr
 # Update `FROM` in `pkg/containerd/Dockerfile`, `pkg/init/Dockerfile` and
 # `test/pkg/containerd/Dockerfile` when changing this.
 ENV CONTAINERD_REPO=https://github.com/containerd/containerd.git
-ENV CONTAINERD_COMMIT=v1.2.5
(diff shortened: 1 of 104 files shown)


tools/alpine/Dockerfile
```diff
@@ -56,7 +56,7 @@ RUN go get -u github.com/LK4D4/vndr
 # Update `FROM` in `pkg/containerd/Dockerfile`, `pkg/init/Dockerfile` and
 # `test/pkg/containerd/Dockerfile` when changing this.
 ENV CONTAINERD_REPO=https://github.com/containerd/containerd.git
-ENV CONTAINERD_COMMIT=v1.2.5
+ENV CONTAINERD_COMMIT=v1.2.6
```
thaJeztah: Does bumping this version also bump runc to the same version as is used in the containerd's vendor.conf? (just double checking)

rn: no, but the subsequent commit (https://github.com/linuxkit/linuxkit/pull/3327/commits/fd67969697c3be7b8cea2268de235f792e7a5f2a) bumps `linuxkit/runc` to the recommended version

thaJeztah: Ah, thanks!

thaJeztah: LOL, shouldn't be looking at large diffs from my phone

Issue #3324, opened 2019-04-08
Unclear documentation about booting on Raspberry Pi


dapopescu:
Hi,

I built an example minimal.tar for rpi3 as described in the documentation, but it is unclear to me what are the next steps. The documentation says extract the archive on a formatted SD card. I did so, but the image did not boot. Could you please elaborate on this part?

Thank you

rn:
do you have the serial console setup? It is on the GPIO pins and there are adapters to USB, which I used

dapopescu:
I have set it up today, I do not see any boot sequence, just two random
characters in the beginning.

How should the SD card be formatted? I currently have two partitions (boot
fat32 and a second ext4 one), and I copied the contents of the minimal.tar
archive on the first partition.

rn:
I have the first partition as fat32 as well (IIRC) and just untarred the tar file on that partition. It has been a while since I did this (10+ months ago) so i'm not sure if this bit rotted. When I developed this i used both serial console and a monitor. Have you tested with a known working SD with a standard distro and checked that serial and monitor output works?

dapopescu:
I tried on the same SD card Ubuntu 18.04 for Raspberry Pi and it worked
(monitor, keyboard, serial console), so this is not the issue.

nakato:
What do your RPi LED's do?

I just tried on my RPi-3 B+ and saw the power LED turn on, go off, then ACT blink 8 times indicating that the firmware needs to be updated.

dapopescu:
When I connect the monitor, I get the rainbow square, which I read means
that the kernel is not booting.

dapopescu:
Yes, I also see the ACT blink 8 times. How do I update the firmware then?

dapopescu:
So the LEDs do the following:
1. The power led (red) turns on, then goes off
2. The ACT led (green) blinks 8 times, four times slower, and then four
times quicker.

nakato:
The firmware is available from https://github.com/raspberrypi/firmware

dapopescu:
Thank you, it booted after I copied the contents of the firmware/boot folder.

PR #3335, opened 2019-04-23
Update RPi firmware

nakato:
**- What I did**
Update commit hash to include updated RPi firmware

**- How to verify it**
Attempt to boot a RPi and expect the power LED to remain on, and ACT to not flash with a debug code.  If successful, expect uboot to log to both serial consol and HDMI.

Vendor RPi firmware blob license
Update RPi3 firmware to fix startup on newer kit

---

This is a potential fix for #3324, however the fix will be when the new image is updated in `src/cmd/linuxkit/moby/output.go`.  I assume this must merge before an image is produced that can provide an updated hash for `src/cmd/linuxkit/moby/output.go`?


rn:
@nakato thanks for the patch (and @dprotaso for testing it).

@nakato could you add another commit to this PR to update the hash in src/cmd/linuxkit/moby/output.go? You can get the hash with the `linuxkit` utility:
```
linuxkit pkg show-tag tools/mkimage-rpi3
```
 Once this PR is updated, I will build and sign the packge and a `linuxkit` binary from build from the merged PR should pick up the new firmware.

nakato:
I have updated output.go, let me know if you'd prefer that rebased into `ea66def`

dprotaso:
fyi I think you tagged the wrong dave (@dprotaso) ;)

PR #3341, opened 2019-05-02
Fix issue wherein multiple files named kernel in the tar break various mkimage utils

### LINKED ISSUE #3340: bug in `mkimage-raw-efi` choking on kernel file and directory
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

Unable to build a `raw-efi` image.

1. Reproducible with the `linux.yml` example.
2. Build the example with a format of `tar`
3. Attempt to convert the tar ball to a `raw-efi` image.

```
 2019-05-01 16:35:01 ⌚  C02WW2ZCHTDD in ~/Code/teamcity-agent-kubevirt
± |master U:4 ?:5 ✗| → cat vm/teamcity-agent-vm.tar | docker run -i -e DEBUG=true linuxkit/mkimage-raw-efi:2d1a0e6e67c967bd4d5a859eb9044571f6844577
+ IMGFILE=//disk.img
+ exec
+ ESP_FILE=//boot.img
+ uname -m
+ ARCH=x86_64
+ BOOTFILE=BOOTX64.EFI
+ LINUX_ENTRY=linuxefi
+ INITRD_ENTRY=initrdefi
+ mkdir -p /tmp/efi
+ cd /tmp/efi
+ '[' -t 0 ]
+ bsdtar xzf -
+ find . -name '*.img'
+ INITRD=
+ find . -name kernel
+ KERNEL='./lib/modules/4.19.37-linuxkit/kernel
./boot/kernel
./containers/services/teamcity-agent/lower/etc/kernel'
+ find . -name cmdline
+ CMDLINE_FILE=./boot/cmdline
+ cat ./boot/cmdline
+ CMDLINE='console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0 pti=off'
+ cat /proc/sys/kernel/random/uuid
+ PARTUUID=17afb7ef-d237-4d6f-8099-d33cf7ccf1e6
+ cp /usr/local/share/BOOTX64.EFI .
+ mkdir -p EFI/BOOT
+ cat
+ stat -c '%s' './lib/modules/4.19.37-linuxkit/kernel
./boot/kernel
./containers/services/teamcity-agent/lower/etc/kernel'
stat: can't stat './lib/modules/4.19.37-linuxkit/kernel
./boot/kernel
./containers/services/teamcity-agent/lower/etc/kernel': No such file or directory
+ KERNEL_FILE_SIZE=
```

No error and an `img` file.

This will also effect:

```
mkimage-dynamic-vhd
mkimage-gcp
mkimage-qcow2-efi
mkimage-raw-bios
mkimage-raw-efi
mkimage-vhd
mkimage-vmdk
```

cc @deitch

deitch:

fixes #3340 

**- What I did**

Changes the 7 `mkimage-*` utilities that call `KERNEL=$(find -name kernel)` or similar to `KERNEL=$(find -type f -name kernel)`

**- How I did it**

With `vi` (because I still like `vi`)

**- How to verify it**

Build and deploy. And CI

Fix bug wherein dir named `kernel` in image breaks builder

Notes:

* This is not ready. I did not updated hashes in the `output.go` file yet because:
* I want feedback from other maintainers on if this is sufficient or if they have other ideas. Once we are agreed, I will updated the hashes.


deitch:
From the open issue:

> The easy fix would be find . -type f -name kernel, but I am wondering if we shouldn't be concerned there could be other files somewhere in the image named kernel? Should we be a bit smarter about it at this point and perhaps do find /boot -type f -name kernel? Maybe something else?

Thoughts? @dave-tucker @ijc @justincormack @rn @ijc ?

deitch:
> Is the find really still needed, do we not always publish the kernel (and initrd + cmdline) in the same place within the images these days? IOW perhaps the find is a relic of a time when things were more in flux and we can just do away with them altogether now?
> Even if not, are the references to bzImage obsolete now? They are only in a couple of the scripts which is suspicious.

I would be more than happy to change this to `KERNEL=/boot/kernel` and be done with it. I _think_ we put it all in one place now, but not 100% sure.

Others know?

rn:
as per https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile#L118 the kernel is always called `kernel`. This did not used to be the case but seem to have been the case since 2017-07-11 (or commit `e0a8a1d4df`). I'd say, if you run kernels that old you have other issues, and would be fine to break that "backwards compatibility"

justincormack:
Yes, we can drop support for any other name now; there was a transitional period when it was bzImage but that was ages ago...

deitch:
And will it always be `/boot/kernel`, or can it be elsewhere?

ijc:
I believe from Rolf's link that in the linuxkit kernel container images it is always `/kernel` today (that snippet installs to `/out` but then the final image is `FROM scratch` with a `COPY --from=kernel-build /out/* /`).

deitch:
I think it does `/kernel` in the `linuxkit/kernel` image, but that is copied to `/boot/kernel` in the final tar. 

I’ll update this shortly, do the hashes as well, and let CI run it

deitch:
CI failed because of the notary issue. See Slack channel discussion. Will repush and restart CI when it is fixed.

deitch:
I got it going. Restarted CI. Awaiting clearing CI and review from either @rn or @justincormack (as @ijc said their opinion is more relevant here, and I changed it since then).

deitch:
This ended up being confusing, but I think I got it. All of the above use `/boot/kernel` for `tar` output, but `/kernel` for the `tar-initrd` output. We `tar-initrd` for all of the above, so need to switch it.

tools/mkimage-dynamic-vhd/make-dynamic-vhd
```diff
@@ -17,7 +17,7 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name kernel -or -name '*bzImage')"
+KERNEL="./kernel"
 CMDLINE="$*"
 
 [ "$KERNEL" = "./kernel" ] || mv "$KERNEL" vmlinuz64
```
rn: This conditional should not be necessary anymore

tools/mkimage-gcp/make-gcp
```diff
@@ -17,7 +19,7 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name kernel -or -name bzImage)"
+KERNEL="./kernel"
 CMDLINE="$*"
 
 [ "$KERNEL" = "./kernel" ] || mv "$KERNEL" kernel
```
rn: ditto

tools/mkimage-vhd/make-vhd
```diff
@@ -17,7 +17,7 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name kernel -or -name '*bzImage')"
+KERNEL="./kernel"
 CMDLINE="$*"
 
 [ "$KERNEL" = "./kernel" ] || mv "$KERNEL" vmlinuz64
```
rn: ditto

tools/mkimage-vmdk/make-vmdk
```diff
@@ -17,7 +17,7 @@ cd files
 [ -t 0 ] || tar xf -
 
 INITRD="$(find . -name '*.img')"
-KERNEL="$(find . -name kernel -or -name '*bzImage')"
+KERNEL="./kernel"
 CMDLINE="$*"
 
 [ "$KERNEL" = "./kernel" ] || mv "$KERNEL" vmlinuz64
```
rn: ditto

deitch:
At least CI is nice and clean now. :-)

tools/mkimage-dynamic-vhd/make-dynamic-vhd
deitch: Yeah, I saw that, but didn't want to go through the "change it all, `lkt pkg push` it all, restart the 60-90 min CI again" cycle. :-) 

Will do it.

deitch:
OK, removed all as requested by @rn . Please review again.

Any idea how to get CI to kick off again? I did a force-push, but that didn't seem to cause it to run again.

rn:
hmm, looks like datakit ci fell off. I manually kicked it

deitch:
Thanks

PR #3345, opened 2019-05-07
rngd exits after being stopped and resumed

kmjohansen:
Linux has documented but somewhat unusual behavior around
SIGSTOP/SIGCONT and certain syscalls, of which epoll_wait(2) is one.  In
this particular case, rngd exited unexpectedly after getting ptrace'd
mid-epoll_wait.  Fix this by handling EINTR from this syscall, and
continuing to add entropy and wait.

Signed-off-by: Krister Johansen <krister.johansen@oracle.com>

**- What I did**

Changed rngd so that it will no longer exit 1 upon being ptrace'd.

**- How I did it**

This change requires that the error handling around the `epoll_wait(2)` in rngd be modified so that it can cope with receiving an EINTR (or any temporary error) and continue instead of exiting.

**- How to verify it**

I validated this by running `strace(1)` against `rngd`.  Prior to the fix it would exit moments after the tracer attached.  With this fix, it's possible to observe the behavior of `rngd` with strace, and to have it attach and detach without issue.

Rngd should gracefully handle SIGSTOP/SIGCONT

920x920

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -16,7 +16,7 @@ onboot:
     image: linuxkit/metadata:v0.7
 services:
   - name: rngd
-    image: linuxkit/rngd:v0.7
+    image: linuxkit/rngd:02c555b50cd1887aa628836662d2eec54c0d7e81
   - name: sshd
     image: linuxkit/sshd:v0.7
     binds:
--- a/examples/azure.yml
+++ b/examples/azure.yml
@@ -11,7 +11,7 @@ onboot:
     image: linuxkit/sysctl:v0.7
 services:
   - name: rngd
-    image: linuxkit/rngd:v0.7
+    image: linuxkit/rngd:02c555b50cd1887aa628836662d2eec54c0d7e81
   - name: dhcpcd
     image: linuxkit/dhcpcd:v0.7
   - name: sshd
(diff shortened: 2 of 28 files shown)


kmjohansen:
I seem to have gotten unlucky with the build system a second time around.  Can I trouble you to give it another poke, @justincormack?

kmjohansen:
@justincormack It looks like the build is green again.  Is this good to merge?

kmjohansen:
@justincormack another nudge on this one.  I've had it in production for a month without any issues on my side.  Does this look good to merge from yours?

justincormack:
yeah its fine, someone needs to build the new image I haven't had time been travelling constantly, maybe @rn or @deitch can...

rn:
I've build/pushed/signed the image for all three arches. In a separate commit, could you update the use of the rngd package in the YAML files to:
```
linuxkit/rngd:02c555b50cd1887aa628836662d2eec54c0d7e81
```

`./scripts/update-component-sha.sh` will help you with that

kmjohansen:
@rn Thanks; I've rebased and pushed another commit that updates the image id on `rngd`.  It _looks_ like circleci is unhappy about the `gcexportdata`, which doesn't seem to be in any of the code that I have access to.  Are you able to give CI another poke?

rn:
looks like some circleCI flakinesss. I had to kick it twice before it went green (got a different error the first time)

PR #3349, opened 2019-05-15
Update GCP metadata provider

ptone:
Updates the userdata key to the more common and documented `user-data` instead of `userdata`.

https://cloud.google.com/container-optimized-os/docs/how-to/create-configure-instance#using_cloud-init

Checks to see if the /run/config/ssh directory exists; this allows new ssh keys to be fetched on reboot.

Signed-off-by: Preston Holmes <preston@ptone.com>

### DIFF
--- a/pkg/metadata/provider_gcp.go
+++ b/pkg/metadata/provider_gcp.go
@@ -53,7 +53,7 @@ func (p *ProviderGCP) Extract() ([]byte, error) {
 	}
 
 	// Generic userdata
-	userData, err := gcpGet(instance + "attributes/userdata")
+	userData, err := gcpGet(instance + "attributes/user-data")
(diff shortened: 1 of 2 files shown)


justincormack:
You need to change it at https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/gcp.go#L269 as well.

rn:
I've build/push/signed the updated metadata package. In a separate commit, could you update the YAMLs to use the new version. Something like:

```
./scripts/update-component-sha.sh --image linuxkit/metadata:1b80de92a781d6884c26c8375a18698c3eb65111
```

should do the trick

rn:
will merge and then update the packages shortly

PR #3351, opened 2019-05-19
Update referenced Raspberry Pi firmware

riconnon:
Update Raspberry Pi firmware used in mkimage-rpi3 to the latest stable
version to support newer hardware models such as the 3B+

Signed-off-by: Richard Connon <richard@connon.me.uk>

**- What I did**
Updated the version of the raspberry pi firmware used by mkimage-rpi3

**- How I did it**
Modified the commit sha referenced in the Dockerfile

**- How to verify it**
Build an image using mkimage-rpi3 and test it on newer hardware such as a raspberry pi 3B+

Update referenced Raspberry Pi firmware to 1.20190517

IMG_20160427_090752
13.jpg)

### DIFF
--- a/src/cmd/linuxkit/moby/output.go
+++ b/src/cmd/linuxkit/moby/output.go
@@ -27,7 +27,7 @@ var (
 		"vhd":         "linuxkit/mkimage-vhd:4cc60c4f46b07e11c64ba618e46b81fa0096c91f",
 		"dynamic-vhd": "linuxkit/mkimage-dynamic-vhd:99b9009ed54a793020d3ce8322a42e0cc06da71a",
 		"vmdk":        "linuxkit/mkimage-vmdk:b55ea46297a16d8a4448ce7f5a2df987a9602b27",
-		"rpi3":        "linuxkit/mkimage-rpi3:cb8427df175b61dd6b635e76f0e88bf261e30d8b",
+		"rpi3":        "linuxkit/mkimage-rpi3:9f2d993daa83152c5d52af16ebb7fefa1e69f28e",
 	}
 )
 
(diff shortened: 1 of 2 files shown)


riconnon:
It looks like the firmware had already been updated since v0.7 (which I was using) and the version in master will already support the 3B+
This still bumps to a later release tag on the firmware repo which seems a valid thing to do, but my original reasoning is no longer relevant.

src/cmd/linuxkit/moby/output.go
```diff
@@ -27,7 +27,7 @@ var (
 		"vhd":         "linuxkit/mkimage-vhd:4cc60c4f46b07e11c64ba618e46b81fa0096c91f",
 		"dynamic-vhd": "linuxkit/mkimage-dynamic-vhd:99b9009ed54a793020d3ce8322a42e0cc06da71a",
 		"vmdk":        "linuxkit/mkimage-vmdk:b55ea46297a16d8a4448ce7f5a2df987a9602b27",
-		"rpi3":        "linuxkit/mkimage-rpi3:cb8427df175b61dd6b635e76f0e88bf261e30d8b",
+		"rpi3":        "linuxkit/mkimage-rpi3:9e2dc61823ec82ac4a76b0a3c30038894c5d7a54",
```
rn: this hash does not match what I get when I build the package:
```
% linuxkit pkg show-tag tools/mkimage-rpi3
linuxkit/mkimage-rpi3:9f2d993daa83152c5d52af16ebb7fefa1e69f28e
```
I've update the package and pushed it, but the tag needs updating.

riconnon: I've updated this now.
I think that I initially confused the tree hash you're using for this with the docker image hash.

PR #3357, opened 2019-05-26
Support for dynamic selection of Raspberry Pi DTB

riconnon:
**- What I did**
Support selecting devicetree for raspberry pi 3B vs 3B+ dynamically

**- How I did it**
Updated UBoot to a version which can detect the raspberry pi 3B+ correctly.
Install both device tree binaries.
Select the correct binary using the fdtfile variable populated by UBoot.

**- How to verify it**
Boot a raspberry pi 3B+ and run
`cat /proc/device-tree/model`
Should output correct model information.

Support selecting devicetree for raspberry pi 3B vs 3B+ dynamically

IMG_20160408_211321

### DIFF
--- a/src/cmd/linuxkit/moby/output.go
+++ b/src/cmd/linuxkit/moby/output.go
@@ -27,7 +27,7 @@ var (
 		"vhd":         "linuxkit/mkimage-vhd:4cc60c4f46b07e11c64ba618e46b81fa0096c91f",
 		"dynamic-vhd": "linuxkit/mkimage-dynamic-vhd:99b9009ed54a793020d3ce8322a42e0cc06da71a",
 		"vmdk":        "linuxkit/mkimage-vmdk:b55ea46297a16d8a4448ce7f5a2df987a9602b27",
-		"rpi3":        "linuxkit/mkimage-rpi3:9f2d993daa83152c5d52af16ebb7fefa1e69f28e",
+		"rpi3":        "linuxkit/mkimage-rpi3:9dd4f7735e19e495c2b0a856a52af15141816534",
 	}
 )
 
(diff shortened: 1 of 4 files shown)


riconnon:
May have fired the PR too quickly here. Looks like the DTB detection is working fine but the new u-boot can't load the same kernel image.

riconnon:
Ah, so the kernel was being loaded from the SD card to `kernel_addr_r` which is incorrect. I'm not sure why this was working in the older u-boot but this is definitely waht `loadaddr` is for. `kernel_addr_r` is the location that the kernel executes from after being decompressed. I've updated the PR with my working version now.

rn:
@riconnon thanks for this work. I currently don't have an easy way to test this. Did you test on 3b and 3b+? I cobbled together the RPi config a week or so before a demo at DockerCon a few years back and hacked it around till it worked. So it's great to see someone cleaning it up

riconnon:
@rn unfortunately I don't have a 3b to test on, only a 3b+. I'll try and get hold of one.
Best to wait and be sure this is fully working on both platforms before merging.

I have a few other bits I may try to push back at some point too, such as a contrib Dockerfile for a linuxkit kernel package from the upstream raspberry pi kernel tree.

riconnon:
@rn is there a particular reason for the use of u-boot at all as opposed to loading the images directly using the raspberry pi bootloader?

rn:
@riconnon no particular reason apart from that I couldn't get the native method to work reliably and I made quicker progress with uBoot and it seems to provide more flexibility wrt to boot options and has/had better documentation.

riconnon:
@rn I've now successfully tested this using both a 3B and a 3B+ using linuxkit/kernel:4.19.47 kernel package

rn:
@riconnon thanks for checking and getting back. Could you also, in a separate commit update the use the the rpi image in `src/cmd/linuxkit/moby/output.go` to `linuxkit/mkimage-rpi3:9dd4f7735e19e495c2b0a856a52af15141816534`?

I've build/push/signed the new image already.

Thanks

riconnon:
@rn done as requested.

PR #3360, opened 2019-05-29
Extract method that launch runc service

guillaumerose:
**- What I did**

Simplify a long method by extracting code inside the for-loop.

My goal is to open a follow-up PR that will log the time required to execute onboot/shutdown services. It will be useful in Docker Desktop for getting a better knowledge on "Why the boot is slow?"

**- How I did it**

Extract the method, use `fmt.Errorf` to carry the information.

**- How to verify it**

aaaproduction_0_0

### DIFF
--- a/pkg/init/cmd/service/runc.go
+++ b/pkg/init/cmd/service/runc.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


guillaumerose:
Should I do something else ? Like add a unit test ?

PR #3363, opened 2019-06-02
Fixed urls used to download kernel images from kernel.org in kernel/Dockerfile.kconfig, used by make kconfig

lpottier:
**- What I did**

`make kconfig` was broken, I fixed the issues by updating 
the urls used to download linux kernel images from kernel.org

**- How I did it**

Updated kernel/Dockerfile.kconfig

**- How to verify it**

run `make kconfig`


rn:
in what way was it broken? I use it pretty often, including yesterday.

lpottier:
Hi,
As the directory "sources" does not exist curl failed to write the downloaded kernels into it. Add a "RUN mkdir sources" before solve the problem, or use the "--create-dirs" option of curl.
I tested it on MacOS 10.14.5 with curl 7.54.0 and on Linux 3.10 with curl 7.29.0, on both systems I had this issue.

rn:
I see. Could you squash the commits into one and sign-off the commit?

lpottier:
Done, sorry for the delay.

PR #3370, opened 2019-06-07
Create swap file only if absent

guillaumerose:
**- What I did**

Speed up swap onboot container.

**- How I did it**

`swapon || (dd && mkswap)`  instead of `dd && mkswap && swapon`. 

**- How to verify it**

Without this PR, container log always show dd output. With the PR, it appears only once.

pkg/swap: create swap file only if absent.

### DIFF
--- a/examples/docker-for-mac.yml
+++ b/examples/docker-for-mac.yml
@@ -26,7 +26,7 @@ onboot:
     command: ["/usr/bin/mountie", "/var/lib"]
   # make a swap file on the mounted disk
   - name: swap
-    image: linuxkit/swap:v0.7
+    image: linuxkit/swap:f8b0434963b4c9114db473aab3b9409bccd4f85b
     command: ["/swap.sh", "--path", "/var/lib/swap", "--size", "1024M"]
   # mount-vpnkit mounts the 9p share used by vpnkit to coordinate port forwarding
   - name: mount-vpnkit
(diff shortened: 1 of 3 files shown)


rn:
The rest of the file uses tabs for indentation. Please use them in your changes too to leave it consistent.

Otherwise the changes look ok.

To speed things up a little, in a separate commit, please change the uses of the `linuxkit/swap` package in all the YAML files with the new tag (output of `linuxkit pkg show-tag .` in the `./pkg/swap` directory). `scripts/update-component-sha.sh` will help with that. The CI will then fail but one of the maintainers can then build and push the new package and re-kick CI

guillaumerose:
I updated the PR. Thanks for the feedback !

djs55:
This looks good to me too (although a shame a picture of a cute animal was omitted)

guillaumerose:
Just added a panda eating all my disk space :)

guillaumerose:
While opening the PR in Desktop, I just realize that if the size change, the disk is not regenerated. I will fix this, sorry :/

guillaumerose:
The missing check is `$(stat --printf="%s" $path) == $(disksize_to_count 1 $size)`

guillaumerose:
I added the extra check and tested reset/change swap size/restart multiples times on Desktop. It works fine. Sorry for this.

djs55:
Thanks!

PR #3371, opened 2019-06-08
Skip disk resize for dos partition if no free space is available

guillaumerose:
**- What I did**

LastLBA doesn't exist for DOS partition. This PR verifies that `total size > partition size + 2048` before doing a resize. It saves 5s boot time on my Mac running Docker Desktop.

**- How to verify it**

Logs are different if resize is skipped.

extend: skip resize dos partition if no free space is available


pkg/extend/extend.go
```diff
@@ -83,10 +84,24 @@ func extend(d, fsType string) error {
...
+	if f.PartitionTable.Label == "dos" {
+		out, err := exec.Command("blockdev", "--getsz", d).Output()
+		if err != nil {
+			return fmt.Errorf("Unable to get total size of %s: %v", d, err)
+		}
+		totalSize, err := strconv.Atoi(strings.TrimSpace(string(out)))
+		if err != nil {
+			return fmt.Errorf("Unable to convert total size from string to int: %v", err)
+		}
+		if totalSize == partition.Size+2048 {
```
guillaumerose: 2048 is also partition.Start. I can refactor to have a function that returns the total size of the disk (gpt and dos) and have a single check.

dave-tucker: Why the magic number vs. using `partition.Start` in this case?

guillaumerose: Yes it's a mistake, I will fix it.

pkg/extend/extend.go
```diff
@@ -87,6 +88,20 @@ func extend(d, fsType string) error {
 		log.Printf("No free space on device to extend partition")
 		return nil
 	}
+	if f.PartitionTable.Label == "dos" {
+		out, err := exec.Command("blockdev", "--getsz", d).Output()
```
djs55: It might be easier to use this https://github.com/moby/moby/blob/master/pkg/devicemapper/ioctl.go
```golang
func ioctlBlkGetSize64(fd uintptr) (int64, error) {
	var size int64
	if _, _, err := unix.Syscall(unix.SYS_IOCTL, fd, BlkGetSize64, uintptr(unsafe.Pointer(&size))); err != 0 {
		return 0, err
	}
	return size, nil
}
```

guillaumerose: I guess I should also changes sfdisk json structure to use int64 ? https://github.com/linuxkit/linuxkit/blob/master/pkg/extend/extend.go#L43

guillaumerose: @justincormack should I make this change ? I can open a separate follow-up PR to change the two calls to blockdev by ioctl.

guillaumerose:
PTAL @justincormack 

Thanks :)

pkg/extend/extend.go
justincormack: Yes to both of those please.

justincormack: It is probably easier to do it in one PR.

guillaumerose:
I removed calls to blockdev. Now it uses ioctl directly. I tested the package successfully on desktop.

Only issue is when I shrink the disk but it was already present before.

guillaumerose:
@justincormack please review :)

guillaumerose:
It's merged in Docker Desktop and will be released in the next edge version.

guillaumerose:
It's in Desktop edge for a month now. We didn't see issues related to this change.

guillaumerose:
It's in edge for 2 months now. I guess it has proven its stability.

guillaumerose:
up!

guillaumerose:
Thanks Dave!

rn:
Pushed the image and kicked CI

PR #3377, opened 2019-06-17
Update Scaleway Provider with new Go SDK

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

**- What I did**
I updated the Scaleway provider to use the new Golang SDK

**- How I did it**
Replace old SDK with new one

**- How to verify it**
Try to build, push and run Linuxkit in Scalway

Update Scaleway Provider with new Go SDK

cute turtle


Sh4d1:
That's why one of the dependencies (`sshkeys`) require a `blowfish` which is in `crypto`. I tries do add `blowfish` to the vendor.conf but it said it was a duplicate of `golang.org/x/crypto`. Maybe there is a trick I don't know about (no expert with vndr :sweat_smile:)

Sh4d1:
@rn didn't find a solution, seems like the way to do it :thinking_face:

PR #3385, opened 2019-06-27
kernel: Update wireguard to 0.0.20190601

rn:
This skips 0.0.20190531

Changelog for 0.0.20190601

== Changes ==

  * compat: don't call xgetbv on cpus with no XSAVE

  There was an issue with the backport compat layer in yesterday's snapshot,
  causing issues on certain (mostly Atom) Intel chips on kernels older than
  4.2, due to the use of xgetbv without checking cpu flags for xsave support.
  This manifested itself simply at module load time. Indeed it's somewhat tricky
  to support 33 different kernel versions (3.10+), plus weird distro
  frankenkernels.

Changelog for 0.0.20190531

== Changes ==

  * tools: add wincompat layer to wg(8)

  Consistent with a lot of the Windows work we've been doing this last cycle,
  wg(8) now supports the WireGuard for Windows app by talking through a named
  pipe. You can compile this as `PLATFORM=windows make -C src/tools` with mingw.
  Because programming things for Windows is pretty ugly, we've done this via a
  separate standalone wincompat layer, so that we don't pollute our pretty *nix
  utility.

  * compat: udp_tunnel: force cast sk_data_ready

  This is a hack to work around broken Android kernel wrapper scripts.

  * wg-quick: freebsd: workaround SIOCGIFSTATUS race in FreeBSD kernel

  FreeBSD had a number of kernel race conditions, some of which we can vaguely
  work around. These are in the process of being fixed upstream, but probably
  people won't update for a while.

  * wg-quick: make darwin and freebsd path search strict like linux

  Correctness.

  * socket: set ignore_df=1 on xmit

  This was intended from early on but didn't work on IPv6 without the ignore_df
  flag. It allows sending fragments over IPv6.

  * qemu: use newer iproute2 and kernel
  * qemu: build iproute2 with libmnl support
  * qemu: do not check for alignment with ubsan

  The QEMU build system has been improved to compile newer versions. Linking
  against libmnl gives us better error messages. As well, enabling the alignment
  check on x86 UBSAN isn't realistic.

  * wg-quick: look up existing routes properly
  * wg-quick: specify protocol to ip(8), because of inconsistencies

  The route inclusion check was wrong prior, and Linux 5.1 made it break
  entirely. This makes a better invocation of `ip route show match`.

  * netlink: use new strict length types in policy for 5.2
  * kbuild: account for recent upstream changes
  * zinc: arm64: use cpu_get_elf_hwcap accessor for 5.2

  The usual churn of changes required for the upcoming 5.2.

  * timers: add jitter on ack failure reinitiation

  Correctness tweak in the timer system.

  * blake2s,chacha: latency tweak
  * blake2s: shorten ssse3 loop

  In every odd-numbered round, instead of operating over the state
      x00 x01 x02 x03
      x05 x06 x07 x04
      x10 x11 x08 x09
      x15 x12 x13 x14
  we operate over the rotated state
      x03 x00 x01 x02
      x04 x05 x06 x07
      x09 x10 x11 x08
      x14 x15 x12 x13
  The advantage here is that this requires no changes to the 'x04 x05 x06 x07'
  row, which is in the critical path. This results in a noticeable latency
  improvement of roughly R cycles, for R diagonal rounds in the primitive. As
  well, the blake2s AVX implementation is now SSSE3 and considerably shorter.

  * tools: allow setting WG_ENDPOINT_RESOLUTION_RETRIES

  System integrators can now specify things like
  WG_ENDPOINT_RESOLUTION_RETRIES=infinity when building wg(8)-based init
  scripts and services, or 0, or any other integer.

image


rn:
ping @justincormack @deitch

PR #3386, opened 2019-06-28
Bump containerd to v1.2.7

rn:
This also fixes the wireguard-tools build

secretary-bird


rn:
ping @justincormack @deitch

deitch:
Saw this as I was going to sleep, figured I would hit it in the am.... but then @justincormack is just faster on the draw! Fastest PR approval in the West! :-)

PR #3400, opened 2019-08-09
Metadata: add support for loading from a file

simonferquel:
**- What I did**
This adds a new configuration provider that just reads a file.
This is needed for Docker Desktop, where we will run a LinuxKit distro in an isolated namespace within WSL 2.
In this scenario, the config will be accessible trough the WSL2 built-in 9p mount of the Windows filesystem.

**- How I did it**
If no provider match the first command line argument, and there is an actual file at this location, it directly read the json payload from it

**- How to verify it**
- Create a LinuxKit image with a static config file in it
- Add an onboot metadata service passing the path to this file as an argument
- Check that /run/config is correctly populated

Machine configuration can now be read directly from a file

animal

### DIFF
--- a/pkg/metadata/main.go
+++ b/pkg/metadata/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"strings"
 	"syscall"
 )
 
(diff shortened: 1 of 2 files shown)


justincormack:
I think I would rather the provider was specified as `file=/var/metadata` rather than just a fallback to a file, not sure what anyone else thinks.

rn:
I agree, just fallback is not nice.

simonferquel:
Will do it once I am back from pto

simonferquel:
@rn @justincormack I just came back from PTO and did the change. PTAL.

rn:
@simonferquel I've build and pushed the updated metadata package. In a separate commit, could you please update the YAML files to use it.

Something like:
```
./scripts/update-component-sha.sh --image linuxkit/metadata:60fe3f6dbe61dbe4ed2fa97bf0acf6891994ed93
```

rn:
This now has merge conflicts. When I tried to resolve them, I also noticed that this is now two commits with the second commit fixing up the first commit. This should really just be one commit, could you please squash.

simonferquel:
@rn I'll do it yes, sorry I've let this pr go stale :/

simonferquel:
I have rebased and squashed. Let me know if I steel need to add a commit with a new image version.

PR #3404, opened 2019-09-03
Add docker-standard ftp_proxy + socks-standard all_proxy env vars

deitch:
Also accepts upper-case `FTP_PROXY` or `ALL_PROXY`

**- What I did**

We already support `http_proxy`/`HTTP_PROXY`, `https_proxy`/`HTTPS_PROXY`, `no_proxy`/`NO_PROXY` env vars to add build args. This does:

1. Adds support for `ftp_proxy` docker standard build-arg
2. Adds support for non-docker standard but socks standard `all_proxy`/`ALL_PROXY`
3. Adds documentation about using proxies when building
4. Fixes bug in building args wherein if build args are present, has them twice and overrides existing args.

**- How I did it**

Changed `src/cmd/linuxkit/pkglib/docker.go` and `docs/packages.md`

**- How to verify it**

Run `lkt pkg build` with the env vars set and `-v` to see the output

Fixes bug in package building; adds support for more proxy variables.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -151,3 +151,27 @@ linuxkit pkg build -org=wombat -disable-content-trust -hash=foo push
 
 and this will create `wombat/<image>:foo-<arch>` and
 `wombat/<image>:foo` for use in your YAML files.
+
+### Proxies
+
+If you are building packages from behind a proxy, `linuxkit pkg build` respects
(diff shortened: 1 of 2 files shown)


justincormack:
Docker supports ftp proxy? Hmm, inclined to not support that...

justincormack:
Ah, it doesn't do anything...

Issue #3052, opened 2018-05-31
duplicates of file paths with "linuxkit build -f docker ..." 


w9n:
**Description**

Its currently not possible to export a tar with linuxkit build and import it directly into docker. As discussed on Slack untar and retar fixes this.

**Steps to reproduce the issue:**
linuxkit build -format docker sshd.yml
docker import sshd.docker

**Describe the results you received:**
An imported image

**Describe the results you expected:**
Error processing tar file(duplicates of file paths not supported):

**Additional information you deem important (e.g. issue happens only occasionally):**

tar -tf sshd.docker | sort | uniq -c | grep -v '^ *1 '
      6 containers
      2 containers/onboot
      4 containers/services
      4 dev/
      4 dev/pts/
      4 dev/shm/
      4 etc/
      4 etc/hosts
      3 etc/init.d/
      4 etc/mtab
      4 etc/resolv.conf
      2 etc/ssl/
      2 etc/ssl/certs/
      4 proc/
      4 sys/
      3 usr/
      3 usr/bin/

deitch:
Only came across this now, thanks to @rmb938 pointing me here. He kindly is working to remove duplicates via #3411 .

I would point out, though, that there should be an open issue on docker (or moby/moby), if there isn't already. It is perfectly legitimate for a tar archive to have duplicate files. The consuming system (`tar -x`, docker, whatever) is supposed to know what to do with them, normally overwrite earlier ones in the stream. If docker doesn't that is a bug in docker.

`linuxkit build ` _is_ putting out a legitimate tar; it is docker that doesn't know how to consume it.

PR #3406, opened 2019-09-08
add the ability to extend linuxkit images

### LINKED ISSUE #3405: Ability to add additonal "layers"
I have a use case where I have a "base" yaml configuration file to build a LinuxKit image. I then would like to take that base image and add things to it. i.e change configuration, add additional packages ect... Somewhat similar to docker images where I can expand upon and add more layers to existing images.

Currently to get this effect I need a full yaml definition, add the additional things and rebuild the whole thing. 

It would be nice to be able to have something like this:

base.yml
```
kernel:
  image: linuxkit/kernel:4.19.52
  cmdline: "console=tty0 console=tty1 console=ttyS0 kubeadm=init"
init:
  - linuxkit/init:1d8e0532ca588c5ad0d9ca6038349a70bb7ac626
  - linuxkit/runc:c1f0db27e71d948f3134b31ce76276f843849b0a
  - linuxkit/containerd:751de142273e1b5d2d247d2832d654ab92e907bc
  - linuxkit/ca-certificates:v0.7
  - linuxkit/memlogd:v0.7
onboot:
  - name: dhcpcd1
    image: linuxkit/dhcpcd:v0.7
services:
  - name: getty
    image: linuxkit/getty:v0.7
  - name: sshd
    image: linuxkit/sshd:v0.7
```

Then I could build the base into a tar or docker image or other format.

And have an "app.yaml" for example with the following contents:
``
from: base.tar
services:
  - name: ntpd
    image: linuxkit/openntpd:v0.7
files:
  - path: /run/config/ssh/authorized_keys
    source: .ssh/id_rsa.pub
    mode: "0600"
    optional: true
``

Then build the app and it would download or extract or something the base image and add the additional things that where specified.

Thoughts?

rmb938:
closes #3405 

**- What I did**

Added a `from` variable to the yaml schema. This allows the user to specify a docker image (atm it's a tar image due to https://github.com/linuxkit/linuxkit/issues/3052) to extend upon.

This allows the user to "layer" images by adding additional packages, files, ect... similar to adding layers to a docker image.

**- How I did it**

Modified the build process to use an existing image instead of a kernel if given. This will copy the kernel and cmdline from the existing image and layer everything else on top.

There may be some bugs and unexpected behaviours but after some quick tests it seems like everything is still working as expected.

**- How to verify it**

Build an image i.e `examples/sshd.yml` then build another image that uses the first i.e `from-sshd.yml`. Then run the image and the additional ntpd service should be running.

Add the ability to extend and layer on top of images.


rn:
@rmb938 apologies for the very delayed reply. If I read this PR and #3405 correctly, could this not be achieved by using something like:
```
% cat openntp.yml
services:
  - name: ntpd
    image: linuxkit/openntpd:v0.7
% linuxkit build sshd.yml openntp.yml
```
This will merge the two yml files when building the resulting image.

rmb938:
@rn No problem.

Yes this could be done by just providing multiple yamls. However doing that doesn't indicate any form of dependency. This PR also has the advantage of not needing to constantly built the "base" image over and over. 

Being that linuxkit images can be built into docker images it made sense (at least to me) to use those as a base layer as they can be pushed to docker repo's and don't have to constantly be built.

It is similar to building a docker image. If I want to use alpine as the base image, I don't build alpine every time I want to build my image, I just download it once.

rn:
@rmb938 I'm not sold on this approach and reluctant to add more syntax to the YAML file. Could you elaborate a bit more about your use case? We do support multiple YAML files, the `init` section already takes images which are expanded directly into the rootfs as well. So there are multiple ways of achieving something similar to what you propose.

rmb938:
The use case was to build upon or layer on top of images built via linuxkit, similar to docker images or OS images provided by cloud providers.

So for example for docker images; When I build a docker image I don't need the dockerfile for whatever base image I am building on top on. All I need is my own dockerfile that starts with `FROM alpine` and the build command will download that image and start layering on top of it.

Another example would be cloud images; When creating images in a cloud like GCP or AWS I can start from their prebuilt images, i.e centos. Boot my VM from that image, install the applications I need then convert that VM back to an image. I don't need the kickstart or upstart file that they used to create the base image I chose, just the resulting image.

My goal was to provide the ability to give someone a pre-built linuxkit image (via a docker image, #3411 fixes the ability to output into a docker image) that they can add more things to. The consumer of the image should not need to be bothered with knowing how it was built (via the yaml definition) or be bothered with building it themselves.


If that can be accomplished by just putting the image into the `init` section then I guess we don't need this PR, just #3411 to fix the docker image output.

deitch:
I think I am beginning to understand the use case. There is something to extensibility that has appeal, similar to the docker example. I can layer on top of existing images (i.e. apply diffs) to get another image without needing to rebuild the entire thing from scratch.

The other advantage is reproducibility. If I do `lkt build a.yml b.yml`, there is no single yml that captures the entire thing. On the other hand, if I do `lkt build combined.yml` which references an immutable image built by `a.yml` (or even just the files, although better with hashed references), then I always can reproduce it. To be fair, I think that is a separate discussion, though.

@rmb938 it might help to include sample use cases in the docs in the PR, along with the description/use case from above?

That having been said, I would be concerned about the `from` being from a tar file, as that reduces reproducibility. `from` an OCI image where you can restrict it right down to a hash, e.g. `from: docker.io/linuxkit/base@sha256:abcd1245ff`? That gets interesting.

rmb938:
@deitch There is an example use case in the PR `examples/from-sshd.yml` however yes I can add some documentation for it.

I also do agree that the `from` should reference a docker image. If you look at the code there are a few commented out lines that I put in that can use a docker image. 

However for that to be a possibility #3411 **must be merged** and I need eyes on that PR first. That PR fixes the docker image output that is needed by this PR.

If I can get more eyes on #3411 as the original responder stopped replying that would be great.

deitch:
> There is an example use case in the PR examples/from-sshd.yml

Ah, now I get it. I had thought that was to be used for chaining, like @rn described. This is the entire file? Interesting...

> I can add some documentation for it.

Sounds good.

> more eyes on #3411

I will take a look.

rmb938:
Nope no chaining needed, you would just run it like so `linuxkit build examples/from-sshd.yml`, linuxkit would then download the `from` docker image (eventually once the above mentioned PR is merged), stuff it's contents into the rootfs then it would continue to add the additional services, files, ect... to the image.

However being that @rn mentioned not changing the syntax and just using the existing `init` list that would probably be better. I'll undo my code changes and keep this PR all documentation and examples, of course once the other PR is merged.

Issue #3376, opened 2019-06-14
linuxkit kernels >= v4.14.106 have broken Hyper-V AF_VSOCK transport


djs55:
**Description**

As far as I can tell there is an upstream bug introduced by [vsock/virtio: fix kernel panic after device hot-unplug](https://github.com/linuxkit/linux/commit/6a78f5dccfff5c0b64b6956e6f6beadd669e0108#diff-c7455644fff4aabad361b451eec8b66e) which first appeared in v4.14.106. It seems to break Hyper-V `AF_VSOCK` only when support is compiled statically; switching to a dynamically loaded module seems to fix it. Other transports seem unaffected.

When running a broken kernel. calling `connect` on an `AF_VSOCK` returns `ENODEV` when running under Hyper-V.

**Steps to reproduce the issue:**

Reproduction examples and bisection data: https://github.com/djs55/bug-linuxkit-vsock-module

**Additional information you deem important (e.g. issue happens only occasionally):**

Issue happens every time (as you would expect).

I guess there are 2 things we could do:
1. as a workaround we could switch to loading `AF_VSOCK` as a module
2. report this bug upstream (although it'll take me a while to set up my git -> email -> lkml -> flame workflow)

Ideally someone could sanity-check this first to make sure I'm not completely mad :)

(/cc @rn for your interest / amusement)

rn:
Have you reported upstream? You can get the email addresses by running some script in the linux source tree (not at hand at the moment) which you point at a file and it tells you who to email

justincormack:
Has this been reported upstream yet? I am ok with changing VSOCK to a module though in general, it is somewhat specialist.

rn:
Closing as this is not LinuxKit specific and the LinuxKit kernel with #3407 moved vsock to be a kernel module. However the real issue seems to be upstream and should be reported and fixed there

PR #3407, opened 2019-09-10
Switch to vsock as a module

djs55:
**- What I did**

Modified the LinuxKit 4.14 kernel configuration so that `AF_VSOCK` is loaded as a module. This works around an upstream bug described in #3376 where `connect` returns `ENODEV` when `AF_VSOCK` is compiled statically.

This modified configuration is currently shipping in Docker Desktop. Ideally if this change is merged then Desktop could switch back to unmodified LinuxKit kernels.

**- How I did it**

I tweaked the 4.14 kernel config in the usual manner.

**- How to verify it**

Using `AF_VSOCK` on Hyper-V will work correctly afterwards. There's a test program on #3376.

- Load `AF_VSOCK` as a Linux 4.14 kernel module to work around an upstream bug (see linuxkit/linuxkit#3376)

a small but colourful bug


rn:
Is this limited to 4.14 or also for 4.19 and 5.x?

djs55:
I don't know -- I've not had a chance to try later kernels. I guess I can fire some PRs against the Desktop CI and see what happens :)

rn:
I prefer having them as similar as possible. I'll update the other kernel configs

djs55:
Thanks!

PR #3411, opened 2019-09-14
WIP: remove duplicate files from output so docker can import images

### LINKED ISSUE #3052: duplicates of file paths with "linuxkit build -f docker ..."

rmb938:
fixes #3052

**- What I did**

remove duplicate files from output so docker can import images

**- How I did it**

Write tar output to a tempfile, then read it back in and write the contents to the actual output while removing duplicates.

**- How to verify it**

```
linuxkit build -format docker sshd.yml
docker import sshd.docker
```

remove duplicate files from output so docker can import images


justincormack:
Hmm, I don't like this solution much. I spent quite a bit of time making many of the output formats streamable, so they could be piped into other commands etc. I was thinking that you would just maintain a list of files, and deduplicate at write time, so it could still be streaming (but with a memory overhead for the list to de-dup).

rmb938:
Well this still works with streaming, it doesn't break that.

What did you have in mind that was different? As far as I could tell there is no point were I could maintain a list of files without re-reading. All images are processed here [ImageTar](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/moby/image.go#L99) that uses the docker client API to stuff the contents of the image into a Reader. That just writes it directly to the tar writer while excluding or replacing some files.

I guess I could put all the logic there, however I would need the same logic when files are added outside of images. :woman_shrugging: I went this way at first but it seemed a bit gross passing around a pointer to a list.

rmb938:
@justincormack Could I get some input on this?

deitch:
I think the principle of deduplicating files is a good one. I suspect the objection from @justincormack has to do with the tmpfile in the middle.

Is there any reason that this cannot be done in the stream processing? It would require more intimate knowledge of the tar stream, but that isn't too bad.

At a deeper level, why is dedup necessary (as opposed to desired)?

rmb938:
@deitch See the issue #3052 docker images with duplicate files cannot be imported into docker. This is why dedup is needed as without it the linuxkit docker output is completely useless.

Would you recommend another place in the linuxkit code to put this? Because as of right now, without this fix being part of linuxkit this command `linuxkit build -format docker sshd.yml` is useless because the output cannot be used.

deitch:
OK, that context helps. I see two potential results out of this.

* there should be an open issue on docker (or moby/moby), if there isn't already. It is perfectly legitimate for a tar archive (that seems redundant, as "tar" = "Tape ARchive") to have duplicate files. The consuming system (`tar -x`, docker, whatever) is supposed to know what to do with them, normally overwrite earlier ones in the stream. If docker doesn't that is a bug in docker. I will mention it on #3052 .
* it may be worth deduplicating here as you recommended, if it helps. We do need a reasonable way of doing it. I think your location makes sense, i.e. in `moby.Build`. A tmpfile is an interesting choice, but if it can be done with direct streaming, it might be better as @justincormack pointed out. The list of files would need to be in-memory either way, so might as well avoid the tmpfile.

rmb938:
Yup I agree that this is more of a docker issue. There is no issue about it in any official docker repo. I did some additional research and it seems when other projects hit this issue every one of them did the debup within their project instead of requesting a fix in docker. So there may be some reason for docker not making this change that I am not finding.

Per your second bullet, maybe I am just miss understanding the term "direct streaming" can you explain what you mean by that? I did manage to get rid of the tempfile just now, I don't know what I did before when I couldn't get it to work.

deitch:
This approach writes it all to memory, which means you need the whole thing in memory for it to work. Some linuxkit images get big. I think what Justin was after was working with the stream. Reading it on the fly as the files go by and dropping duplicates. That means only one file, at most, is in memory (plus the file list). I don't know if that is easily doable or not. 

I also don't know if that is what he had in mind. Let's give him a chance to respond.

rmb938:
Ah I see, I wasn't thinking about the memory part of it.

deitch:
Yes, I think that is how I would do it. Each stage writes its files to the [tar.Writer](https://golang.org/pkg/archive/tar/#Writer) which is represented by the var `iw`. I think I would replace that with something that wraps it but maintains a list of all files and filters duplicates on the fly. So you have the `tar.Writer`, and it is wrapped in some custom struct, and that is what gets passed to all of the various functions here. Those write the same way they did, but the writer here filters out duplicates as they come along. Each time one of those functions writes a file to the `tar.Writer`, it calls `WriteHeader()` and then `Write()` for the file data. Pretty easy to filter on that. 

The only risk here is that, according to how proper tar programs work, if there are multiple copies of a file, the latter one overrides the earlier, since it just reads the archive blindly and writes the files out to the filesystem as they come along. But we can live with it for now in my view.

rmb938:
Oh yea wrapping `tar.Writer` would make it much nicer. I can look into doing that after the holidays. Thanks for all the input :+1:

DavidingPlus:
Shit, I'm facing the same problem too.

deitch:
This old issue bit you? How did you replicate it? Feel free to create a new PR with it up to date.

PR #3417, opened 2019-09-23
Fix label argument mkfs.btrfs

hyperized:
**- What I did**
I changed the `-l` flag to `-L` because in the current version (`v.0.7`) of `LinuxKit/format` the `mkfs.btrfs` `-l` flag means 'leafsize'. The `-L` flag means label

**- How I did it**
n/a

**- How to verify it**
Verified the fix on a local build.

Fix label argument mkfs.btrfs

Sloth

### DIFF
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -15,7 +15,7 @@ onboot:
   - name: sysfs
     image: linuxkit/sysfs:v0.7
   - name: format
-    image: linuxkit/format:v0.7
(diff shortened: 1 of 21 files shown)


rn:
@hyperized I've build and pushed the update format package. Could you update the YAMLs using the format package with something like this (on the top level) and push as a separate commit to this PR:
```
./scripts/update-component-sha.sh --image linuxkit/format:65b9e0a76d0b9fb8ac5c5f3bc8d3131109290f56
```

hyperized:
@rn took me a second, between using the wrong branch and messing up the sign-off, but there it is :).

PR #3423, opened 2019-10-02
go tool vet no longer supported

justincormack:
Just use go vet. The compat was removed in Go 1.13, but
go vet works in all supported versions now.

puppt


justincormack:
Ah, actually behaviour is different, will fix/upgrade Go.

justincormack:
Fixed up should now work with all recent Go versions and will go vet correctly cross platform.

justincormack:
Ah, its not working due to the skanky vendor script, which means go vet doesn't work. Tempted to just remove this...

rn:
This was done as part of https://github.com/linuxkit/linuxkit/pull/3502 and https://github.com/linuxkit/linuxkit/pull/3500.

Closing this one

PR #3425, opened 2019-10-03
Add Hetzner provider in the metadata package

zimme:
The title of the PR is pretty self-explanatory 😜

I copied the AWS provider file and removed a few lines that get metadata that's not available on Hetzner and updated the function `handleSSH` to parse the JSON array of ssh keys Hetzner provides and for each key write it to the authorized_keys file on a new line.

I'm not a Go developer so I might have missed something obvious, etc. so please do a proper review before merging 😄

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -101,6 +101,14 @@ hostname and populate the `/run/config/ssh/authorized_keys` from metadata.
 AWS userdata is extracted from `http://169.254.169.254/latest/user-data` and
 and made available in `/run/config/userdata`.
 
+## Hetzner
+
+Hetzner metadata is reached via the following URL
+(`http://169.254.169.254/latest/meta-data/`) and currently we extract the
+hostname and populate the `/run/config/ssh/authorized_keys` from metadata.
+
+Hetzner userdata is extracted from `http://169.254.169.254/latest/user-data` and
+and made available in `/run/config/userdata`.
 
 ## HyperKit
 
--- a/examples/hetzner.yml
+++ b/examples/hetzner.yml
@@ -1,0 +1,39 @@
+kernel:
+  image: linuxkit/kernel:4.19.71
+  cmdline: console=ttyS1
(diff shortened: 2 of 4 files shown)


deitch:
@zimme this is great, thank you. I am more than happy to have Hetzner on here. A few comments/ requests:

* The provider itself looks pretty good
* Can you update the [docs/metadata.md](https://github.com/linuxkit/linuxkit/blob/master/docs/metadata.md) file to include it, so people know support is there?
* You might want to put in an `examples/hetzner.yml` file so people can adopt it immediately. See the `gcp.yml` or `aws.yml` or `packet.yml` examples. You will have to use the actual hash for `linuxkit/metadata:<tag>`, since the most recent release (`v0.7`) doesn't include the Hetzner support, as you are just adding it now.
* Do you want to add support for `linuxkit push` and `linuxkit run` for Hetzner? It definitely should be a separate PR, but if you see the docs (or run `linuxkit push help` and `linuxkit run help`, you will see support for other cloud providers. All of the support is in [src/cmd/linuxkit](https://github.com/linuxkit/linuxkit/tree/master/src/cmd/linuxkit)

Other than that, quite good to go.

zimme:
I've added Hetzner to `docs/metadata.md`, however I see that Scaleway and some other providers seems to be missing from there too but I don't know if those providers does anything else than what AWS and Hetzner is doing so I don't feel I can add those sections to the document.

I don't actually use linuxkit directly but are using the `metadata` package via `k3os` so I don't know what a good config would look like as an example.

deitch:
Yeah, I know others are missing, but as long as we are adding this, let's do it correctly (as you did).

deitch:
> I don't actually use linuxkit directly but are using the metadata package via k3os so I don't know what a good config would look like as an example.

Just copy the [packet example](https://github.com/linuxkit/linuxkit/blob/master/examples/packet.yml) and make the tiny change.

zimme:
Example added

PR #3018, opened 2018-04-25
Add Scaleway provider to linuxkit

Sh4d1:
Fixes #2487

**- What I did**
I added support for scaleway provider. Right now I only did the push command. 
Some modifications are still needed, but I wanted some feedback.
**- How I did it**
I used to Scaleway Go API to implement the provider.
The tricky part was the process of the image creation since Scaleway relies on snapshot to make images. 
Basically I did:
  * Create a temporary instance with a second volume attached
  * Copy the iso file to the instance via ssh
  * Write the iso to the second volume using dd command
  * Terminate the instance
  * Snapshot the second volume
  * Create the image from snapshot

**- How to verify it**
Pull and try :grin: 

**- Description for the changelog**

Add scaleway support for LinuxKit

**- A picture of a cute animal (not mandatory but encouraged)**
cute cat

PR #3429, opened 2019-10-14
Fix the Scaleway provider in the metadata package

zimme:
Currently the Scaleway provider in the `metadata` package isn't working as expected, at least for me in k3os, which is using that package.

This will update the provider to use the endpoint `.../user_data/cloud-init` where the cloud-init data is served if it was provided. This is also the endpoint the python cloud-init scaleway provider is using from what I could tell. If no cloud-init data was provided this request will return a 404 and check for that in this fix so that we don't get an error if there is no cloud-init data.

This will also remove some code that stripped away the `ssh-rsa`/`ssh-dsa` part in the SSH keys, as no other provider code does that and a complete ssh key should include that part, as least to my knowledge.

### DIFF
--- a/pkg/metadata/provider_scaleway.go
+++ b/pkg/metadata/provider_scaleway.go
@@ -210,11 +210,11 @@ func scalewayGetUserdata() ([]byte, error) {
 		return nil, errors.New("not able to found a free port below 1024")
 	}
 	defer conn.Close()
-	fmt.Fprintf(conn, "GET /user_data HTTP/1.0\r\n\r\n")
+	fmt.Fprintf(conn, "GET /user_data/cloud-init HTTP/1.0\r\n\r\n")
 
 	reader := bufio.NewReader(conn)
 	resp, err := http.ReadResponse(reader, nil)
-	if err != nil {
+	if err != nil || resp.StatusCode == 404 {
 		return nil, err
 	}
 	defer resp.Body.Close()
@@ -245,10 +245,7 @@ func (p *ProviderScaleway) handleSSH(metadata []byte) error {
 		}
 
 		line := string(bytes.Trim(sshKey, "'"))
-		parts := strings.SplitN(line, " ", 2)
-		if len(parts) == 2 {
-			rootKeys = rootKeys + parts[1] + "\n"
-		}
+		rootKeys = rootKeys + line + "\n"
 	}
 
 	if err := os.Mkdir(path.Join(ConfigPath, SSH), 0755); err != nil {


zimme:
@Sh4d1 Care to have a quick look as you made the initial implementation?

Sh4d1:
@zimme I'll take a look tomorrow!
Quicky though I think Scaleway puts the SSH keys in `/user_data` and not in `/user_data/cloud-init` (I'll verify tomorrow as well). In that case maybe it'll be nice to support both cases.

zimme:
@Sh4d1 This doesn't change the way the ssh keys are fetched. It just removes some code that stripped away the first part of the ssh keys.

The update to the `user_data/cloud-init` endpoint is for fetching cloud-init data that one provides to the server. The current implementation just fetches `user_data´ which is an index route that only have information if there are any other endpoints under `user_data`.

But have a look when you have time :+1: Thanks 😄

Sh4d1:
@zimme yeah sorry tired 😅 (But still you can put user data other than cloud init, so supporting both might be good?)

But yeah I'll take a closer look with fresh eyes. Also wanted to clean some code in this package as `sendBootSignal` is no longer needed.

(Also, just curious, where are you using this package in k3os?)

zimme:
k3os is using linuxkit's metadata package for it's cloud-init/cloud-config stuff. However currently they are using a fork of linuxkit as they need some tweaks to create some stuff.

Not working at rancher, so I'm just getting to know the project. I'm just playing around with it and wanted to get it up and running properly on Scaleway and Hetzner so I potentally can setup k3os, k3s, rio and open-faas as our cluster and deployment stuff at the consultancy firm (VNTRS) I work at that helps startups get off the ground =)

Also, That other stuff you said sounds like a good idea too =)

pkg/metadata/provider_scaleway.go
```diff
@@ -220,7 +220,7 @@ func scalewayGetUserdata() ([]byte, error) {
 	defer resp.Body.Close()
 
 	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
+	if err != nil || resp.StatusCode == 404 {
```
Sh4d1: That's weird. First the `Extract` method (the one you are using I guess) does not return any errors for failed userdata. 
Second, if I understood correctly, you wanted to treat 404 as no error, but here you return an error. I would check the error code here https://github.com/linuxkit/linuxkit/pull/3429/files#diff-1b03b5d41b65159de69dd1b6ea7a31daR217 and return no error if 404

pkg/metadata/provider_scaleway.go
```diff
@@ -210,7 +210,7 @@ func scalewayGetUserdata() ([]byte, error) {
 		return nil, errors.New("not able to found a free port below 1024")
 	}
 	defer conn.Close()
-	fmt.Fprintf(conn, "GET /user_data HTTP/1.0\r\n\r\n")
+	fmt.Fprintf(conn, "GET /user_data/cloud-init HTTP/1.0\r\n\r\n")
```
Sh4d1: As said in the coversation, it would be nice to support cloud-init + user data. One thing though, when looking at other providers, they only return the user data keys, and note the value. (Which would not be the case with cloud-init). @rn or @justincormack any thoughts on this? Shoud the `Extract` method returns only the user data keys or also the values?

pkg/metadata/provider_scaleway.go
zimme: I'll move this check up to the previous error check, it was put here because I first thought the ststuscode was on the body. I'll also fix the check as this is not working as expected as you said.

pkg/metadata/provider_scaleway.go
zimme: I looked around and I didn't see that you could add any other userdata than cloud-init stuff and that's why I did this and it's based on the following code https://git.launchpad.net/cloud-init/tree/cloudinit/sources/DataSourceScaleway.py?id=d4d11c78e5e78999356fd0c3d124b5a298735b65

zimme: I do see that they also query vendor data which we could also do

pkg/metadata/provider_scaleway.go
zimme: I moved the error check [here](https://github.com/linuxkit/linuxkit/pull/3429/files#diff-1b03b5d41b65159de69dd1b6ea7a31daR217), as you said.

However, my intension was to actually make 404 responses return an error as letting it through will return `"Invalid Key"` as the userdata because that's the body of the response in the case of a 404.

This should be ok though, as the `Extract` method catches all errors and treats them as non errors [here](https://github.com/linuxkit/linuxkit/pull/3429/files#diff-1b03b5d41b65159de69dd1b6ea7a31daR143-R150)

pkg/metadata/provider_scaleway.go
zimme: Ok, so I look into the user data stuff and from what I can tell you can provide any user data when creating an instance via the api, the question is then how do we merge this properly. From what I can tell most other providers in this package is just returning the userdata as is, and in most if not all cases that's just whatever the user provides and in most cases that's a cloud-config/cloud-init data.

k3os is depending on that kind of behaviour to just read text from userdata and if it's a script, add it to the run_cmd and if it's yaml use it as a cloud-config.

nustiueudinastea:
Hi @Sh4d1 @zimme , is this PR mergeable? If it is, it would be awesome to merge it. I am hitting the same problem that @zimme mentioned, where the SSH keys get the key type stripped, which leads to the inability to ssh into linuxkit instances. Thanks so much!

zimme:
From my side it's good to go, tested the changes out on scaleway and they worked as far as I could tell.

Sh4d1:
Sorry haven't had time to try it :( I'll memo it for tomorrow and I'll get back :)

Sh4d1:
LGTM :tada: 
(Still having doubts about userdata/cloud-init, but I'm not using it, and I guess I'm th eonly user of LK on Scaleway :sweat_smile: )

justincormack:
The linked doc says that the source port for metadata requests needs to be <1024 to force root clients. We don't seem to implement that. I am kind of surprised it works at all given that...

nustiueudinastea:
I just built the metadata package based on this PR and managed to deploy a machine on Scaleway and everything works fine as far as I can tell.

justincormack:
I think the comments must be wrong, as it would never work as the default ephemeral port range is 32768-60999, so as it does work it must be ok...

Sh4d1:
@justincormack the port below 1024 is implemented for scaleway here https://github.com/linuxkit/linuxkit/blob/master/pkg/metadata/provider_scaleway.go#L197 :)

justincormack:
Oh somehow I missed that!

Sh4d1:
@justincormack https://github.com/linuxkit/linuxkit/pull/3018#discussion_r189224468 :sweat_smile:

rn:
Will merge and then update the package shortly

PR #3431, opened 2019-10-16
Fix the Hetzner provider in the metadata package

zimme:
There were some mistakes made in the initial code where writes didn't work, this PR fixes that.

I've tested these changes properly on Hetzner using the `metadata` package in a custom build of k3os.

### DIFF
--- a/pkg/metadata/provider_hetzner.go
+++ b/pkg/metadata/provider_hetzner.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


zimme:
I updated the code to truncate the `authorized_keys` file instead of appending as most other providers do via `ioutil.writeFile`.

rn:
Will merge and rebuild/update the metadata package shortly

Issue #3437, opened 2019-11-05
Linuxkit support


hyperized:
A bit of a meta issue here, but:

How is this project supported?
The only thing I can find is that currently two people are assigned to the project.
(Maybe: Can you document this?)

We're really liking the way Linuxkit is working and we're looking to use Linuxkit as a building block for our definition driven infrastructure going forward.

However I do notice there's a bit of a drop off in activity since the end of 2018 and it seems that keeping this project up to date is not getting a unified effort.

Ideally I'd like to be able to run Kubernetes on this, but currently there's a lot of hoops to jump through to get 1.16 properly supported.

There are a few PR's open that sort of overlap, but they have been quite silent for a bit:

- https://github.com/linuxkit/linuxkit/pull/3426
- https://github.com/linuxkit/linuxkit/pull/3388
- https://github.com/linuxkit/linuxkit/pull/3423

Is there any way we can help with this? How?
And going forward, will this project receive continuous support from Docker?

justincormack:
Hi. The current situation is that there are three active maintainers, but none of us are working full time on the project at present. Avi is available for consulting work. Docker are using the project actively, as are many other organizations, but mostly people are using as is, using their own containers but not making significant core modifications that we are aware of. There are a bunch of projects that need a significant investment of time to do, which would make things way more usable, in particular reworking CI, integrating the filesystem work, and integrating containerd in the build process.

Those three PRs are things that just need finishing off, and someone can contribute them. The current system for building images is annoying though, and currently only maintainers can do it, part of the CI rework would be building containers in CI so they can be tested by anyone and then built in CI, rather than the current manual process, so that would make contribution way easier.

We are happy to add new maintainers if people want to contribute.

deitch:
I can only second what Justin said. There are some improvements that we as a team would like to get done, and then there are community/customer-driven improvements we would like to see. We would be happy both to have more contributors, and if it is a business-specific need, we can look at a consulting arrangement as well.

hyperized:
Thank you for the reponses, it's given us a clearer picture as to what to expect in terms of support.

We're currently not interested in consultancy, but we'll know where to look for when we do :).

Might it be an idea to map out these goals as milestones (under Projects) and then link the issues to them so it's easier for a non-maintainer like myself to see what needs attention?

I would definately be interested in contributing and I think a decently working CI/CD setup would be a requirement. Would the new Github Actions be something for this project?

justincormack:
[...]

PR #3442, opened 2019-11-26
Actions

dave-tucker:
**- What I did**

Per the discussion in #3437, having a working CI system is important to attract contributors to the project and reduce the burden on maintainers. I have spent some time looking at consolidating our CI system sprawl (CircleCI, LinuxKitCI etc..) and replacing it with GitHub Actions.

This PR introduced a single worflow for CI.
It:
- Builds binaries and runs unit tests on all supported platforms
- Runs all integration tests (in parallel)

In future this can be extended to include anything and everything that is on the CI wishlist.

A few points of note:

1. Tests are run using containerized qemu (no nested virt)
2. Overall CI time is still 1hr, but that is mainly spent running the packages tests
3. The GCP test that was running on LinuxKit CI has been ported to RTF but isn't running just yet due to security concerns around secrets
4. I'm not sure what the security model is around changes to workflow/actions. I'll clarify before merge.

**- How I did it**

Wrote lots of YAML and tested on my LinuxKit fork.

**- How to verify it**
You can see the latest green run on my fork: https://github.com/dave-tucker/linuxkit/commit/71a95bff490e7d4505767dbf337b769c58e46f16/checks?check_suite_id=329481528

image


test/Makefile
```diff
@@ -42,8 +42,8 @@ ltp: $(LINUXKIT) test-ltp.img.tar.gz
 	$(call check_test_log, test-ltp.log)
 ### ------
 
-test: gcp-hack
+test:
 	@rtf -l build -v run -x
 
-test-pr: gcp-hack
```
deitch: With `gcp-hack` no longer a dependency, can we get rid of the `gcp-hack` structure? It is [here](https://github.com/linuxkit/linuxkit/blob/71a95bff490e7d4505767dbf337b769c58e46f16/test/Makefile#L21-L30)

src/cmd/linuxkit/run_qemu.go
```diff
@@ -452,8 +453,13 @@ func runQemuContainer(config QemuConfig) error {
...
 	}
 
+	currentUser, err := user.Current()
+	if err != nil {
+		return err
+	}
+	userString := fmt.Sprintf("%s:%s", currentUser.Uid, currentUser.Gid)
 	dockerArgs := append([]string{"run", "--interactive", "--rm", "-w", cwd}, binds...)
-	dockerArgsImg := append([]string{"run", "--rm", "-w", cwd}, binds...)
+	dockerArgsImg := append([]string{"run", "--user", userString, "--rm", "-w", cwd}, binds...)
```
deitch: I assume this is necessary when running on GitHub actions. Are there any knockdown effects? This does get called anytime someone calls `lkt run ` with qemu, does it not?

dave-tucker:
Additional note to reviewers. I've got a help request in to GitHub support to ask:

1) What happens if contributors open PRs against the workflow?
2) Whether self-hosted runners that are only used for `push` events (maintainers have already reviewed code) would prevent forks from gaining access to run untrusted code.

I'll update once I hear back.

test/Makefile
dave-tucker: Not quite yet. It's still being used by LinuxKitCI and deleting it now would make this PR red 😉 
If we merge this, I can come in and clean up afterwards.
Not only can `gcp-hack` disappear but so can the `ci`, `ci-pr` `ci-tag` targets in the top-level file.

I also look at moving the LTP tests to RTF outside of this PR so things will be super clean

deitch: Ah, I should have realized that. Of course it makes sense.

src/cmd/linuxkit/run_qemu.go
dave-tucker: It's not just for actions (see commit message for more info).
TL;DR on Linux a disk created using containerized `qemu-img` will be owned by `root:root`.

Adding `sudo` to a test seemed like a bad idea and this was the best alternative.

If someone with a mac (without qemu installed) could run this from  the `tests` directory it should confirm whether or not this breaks anything.
```
rtf -vvv -l build run -x linuxkit.package.extend.ext4
```

deitch: I would do it, but definitely have `qemu` installed. It will not help.

justincormack: I think the containerized qemu is a mistake, and we should remove support for it.

dave-tucker: @justincormack I'll take a look at removing it + then we can rebase this PR on top of that.

dave-tucker:
Added `[DO NOT MERGE]` here since this now:

a) Depends on #3443 
b) Needs a little more testing since I added an action for `rtf` and attempted to use caching of the binary to shave a few minutes off the overall build time.

dave-tucker:
Rebased since the QEMU patch has merged.
Ready for review.
Still waiting to hear back from GitHub support though 😢

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
...
+
+    - name: Restore LinuxKit From Cache
+      uses: actions/cache@v1
+      with:
+        path: lkt
+        key: linuxkit-amd64-linux-${{hashFiles('src/**')}}
+  
+    - name: Symlink Linuxkit
+      run: |
+        sudo ln -s `pwd`/lkt/linuxkit-amd64-linux /bin/linuxkit
```
justincormack: I think `/usr/local/bin/linuxkit` is better - adding stuff in `/bin` is a bit weird...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
...
+        path: bin
+        key: rtf-${{hashFiles('Makefile')}}
+    
+    - name: Build RTF
+      if: steps.cache-rtf.outputs.cache-hit != 'true' 
+      run: make bin/rtf
+    
+    - name: Symlink RTF
+      run: |
+        sudo ln -s `pwd`/bin/rtf /bin/rtf
```
justincormack: same here

dave-tucker:
Comments addressed @justincormack!
I think this in a state where it can be merged and safely replace CircleCI and LinuxKitCI.

The only delta would be the `lkt run gcp` test that LinuxKitCI does, but that can be added back in the next piece of work I have planned.

That would be to add workflows that cover:
- Building and signing images
- Running the LTP tests
- Running integration tests for all `lkt run` platforms etc...

The approach I have in mind is to keep the logic for ☝️ in `linuxkit/actions`.
The workflows would be in `linuxkit/private-workflows` - although I need to look at how best to handle triggering.
This will keep the most sensitive parts of CI/CD ring fenced off to avoid leaking secrets or providing broad access to self-hosted runners.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
+name: LinuxKit CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build & Test
+    strategy:
+      matrix:
+        arch:
+          - amd64-linux
```
deitch: I do wish that there were a way to source this from some other file, so that the "official supported matrix" would be available as a core part of the repo at the top level, and not buried deep under `.github/...`. Is that doable? Or not really?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
...
+          - amd64-windows.exe
+
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.11
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.11
+      id: go
```
deitch: Pretty old version, is it not?

.github/workflows/ci.yml
dave-tucker: I don't think it's do-able as actions doesn't support YAML anchors or anything like that.

.github/workflows/ci.yml
dave-tucker: Yes it is, but it's what's currently being used in CI. I didn't want to bump it here.

deitch: That is a very good point. I think I will open an issue separately.

justincormack:
Lets go ahead and then iterate on this.

.github/workflows/ci.yml
deitch: Looks like it. Oh well, it was a nice try.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
...
+        path: lkt
+        key: linuxkit-amd64-linux-${{hashFiles('src/**')}}
+
+    - name: Symlink Linuxkit
+      run: |
+        sudo ln -s `pwd`/lkt/linuxkit-amd64-linux /bin/linuxkit
+
+    - name: Build Packages
+      run: |
+        make -C pkg build
```
rn: does ubuntu-latest have docker pre-installed? If so which version? Is there a way to pin the version?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,322 @@
...
+
+    - name: Restore LinuxKit From Cache
+      uses: actions/cache@v1
+      with:
+        path: lkt
+        key: linuxkit-amd64-linux-${{hashFiles('src/**')}}
+
+    - name: Symlink Linuxkit
+      run: |
+        sudo ln -s `pwd`/lkt/linuxkit-amd64-linux /bin/linuxkit
```
rn: Is there a reason we link `linuxkit` to `/bin/linuxkit` but `rtf` to `/usr/local/bin/rtf`? Would be good to add a comment to the relevant commit if there is a good reason.

dave-tucker: No reason. I can fix this quickly before we merge.

.github/workflows/ci.yml
dave-tucker: According to [Github](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners) we have Docker version 3.0.8 🤔 
Sadly there is no way I know of to pin this.
I'd like very much to remove the Docker dependency anyway

.github/workflows/ci.yml
rn: SGTM

rn:
I noticed other PRs also failing on the GCP error. Not sure if anything changed behind the scenes or if this is a regression

dave-tucker:
Based on the error message it looks like it's something behind the scenes. The Python script that allocates machines in GCP gets a HTTP 503 when trying to list instances.

PR #3443, opened 2019-11-27
linuxkit: Remove QEMU Container Backend

dave-tucker:
**- What I did**

Removed the container backend for QEMU as discussed in #3442 

**- How I did it**

Deleted the code

**- How to verify it**

If you run on a system without QEMU installed, `lkt` will now throw an error that `qemu-system-x86_64` was not found on the `$PATH`

Removed Container Backend For QEMU

image


justincormack:
Fairly sure you can delete `tools/qemu` as well which is the image we ran for this.

dave-tucker:
@justincormack deleted!

PR #3446, opened 2019-12-09
Added prefix functionality

neacsu:
The initrd image can now be generated with a prefix path to the root
filesystem of the Docker image, using the '-prefix' option.

This is a useful feature when trying to create a custom ramfs build, as
it removes the constraint of placing the Docker filesystem at the root
of the resulting image.

The code necessary for this is mostly already present, it only needs to
link with a CLI option.

Signed-off-by: Bogdan Neacsu <bnneacsu@amazon.com>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/cmd/linuxkit/build.go
+++ b/src/cmd/linuxkit/build.go
@@ -48,6 +48,7 @@ func build(args []string) {
 	buildOutputFile := buildCmd.String("o", "", "File to use for a single output, or '-' for stdout")
 	buildSize := buildCmd.String("size", "1024M", "Size for output image, if supported and fixed size")
 	buildPull := buildCmd.Bool("pull", false, "Always pull images")
+	buildPrefix := buildCmd.String("prefix", "", "Add a prefix to the root filesystem of the image")
 	buildDisableTrust := buildCmd.Bool("disable-content-trust", false, "Skip image trust verification specified in trust section of config (default false)")
 	buildDecompressKernel := buildCmd.Bool("decompress-kernel", false, "Decompress the Linux kernel (default false)")
 	buildCmd.Var(&buildFormats, "format", "Formats to create [ "+strings.Join(outputTypes, " ")+" ]")
@@ -204,7 +205,7 @@ func build(args []string) {
 	if moby.Streamable(buildFormats[0]) {
 		tp = buildFormats[0]
 	}
-	err = moby.Build(m, w, *buildPull, tp, *buildDecompressKernel)
+	err = moby.Build(m, w, *buildPull, *buildPrefix, tp, *buildDecompressKernel)
 	if err != nil {
 		log.Fatalf("%v", err)
 	}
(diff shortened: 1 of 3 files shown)


deitch:
@neacsu thanks for this. 

How would the usage work? If I use this option, what does my resulting fs look like? Might be nice to capture in a document, but either that, or at least let's document here?

neacsu:
An example is running the following command:

`./bin/linuxkit build -name bootstrap -format kernel+initrd -prefix "rootfs/" linuxkit.yml`

while `linuxkit.yml` describes files and folders both inside the docker fs (e.g. `inside_file`) and outside of it (e.g. `outside_file`):
```
$ cat linuxkit.yml
init:
  - hello-world:latest
files:
  - path: rootfs/inside_file
     ...
  - path: outside_file
     ...
```

This will create an image with the following fs:
```
/outside_file
/rootfs/
├── inside_file
├── <rest of the hello-world:latest fs>
```

Hope this answers your question.

deitch:
So the prefix would apply to the images loaded in `init`, is that it?

I have a few questions:

1. Why is this better than capturing it in the config yml itself? I am concerned about reproducibility in this case, where the same `yml` can lead to 2 completely different resulting images. I think @rn would have some input on this one.
1. What is the need for it? You wrote that it "removes the constraint of placing the Docker filesystem at the root of the resulting image". Understood, but why is it helpful?

Assuming it is helpful - I probably just am not thinking about the right use case - I would lean towards having it as a configuration so it is reproducible rather than a CLI option.

rn:
@neacsu thanks for the patch. I don't quite understand the use case for this. Could you elaborate? Also, as @deitch commented we try to keep the root fs being described by the YAML as much as possible.

neacsu:
Thanks for the feedback. You are right, it might be better to place the option in the `yml` file to preserve reproducibility. 

As for the use-case of this, I'm afraid it is somewhat atypical: I used it so that I can have more control over how the ramfs is booted. This way, I can place a custom made `init` executable and other files needed by it in the initial root. Eventually the `init` process switches the rootfs to the one pointed by the Docker image.

deitch:
> Eventually the init process switches the rootfs to the one pointed by the Docker image.

So you use the `init` section or `files` to place a custom `init` in there and control how it is booted? I can see that it might be atypical, but not necessarily bad. I like the idea of supporting it, as long as we can capture it in the yml, and thus be reproducible.

neacsu:
Changed from a cli option to a YAML configuration option.

neacsu:
I've moved the prefix functionality from the CLI option into the configuration file.
@deitch what do you think? Is anything missing from the patches?

Thanks!

rn:
I'm still not sure how this would be used. Can you please add documentation to `docs/yaml.md` as well as a test (the test can produce a tar ball and then check the layout).

Also, this needs a rebase and the first two commits should be squashed (no need to first introduce a CLI option to then remove it with the next commit)

PR #3448, opened 2019-12-16
wireguard: update to 0.0.20191212

djs55:
**- What I did**

Bumped wireguard to the latest release.

**- How I did it**

Edited the URL and `sha256`

**- How to verify it**

Unsure -- I don't use wireguard myself. This was requested by a user of Docker Desktop and looked simple. I'm hoping that a CI run will be enough to check.

- Bump wireguard to 0.0.20191212


deitch:
I use WireGuard, but not on lkt. We should ask @zx2c4 who has contributed in the past and can comment.

I will admit that I am curious of the use case. WG from a desktop is used regularly. How are they using it from within the Docker Desktop VM?

zx2c4:
This PR is fine. I don't have much to offer in terms of "how to use Docker" for testing or whatever, but between September and December, the build system hasn't changed, so a simple bump like this should be fine.

djs55:
@deitch I'm also curious about the use-case -- I'll try to find out!

rn:
CI does not pick up the changes until I update the kernel (which I plan to do later this week).

PR #3455, opened 2019-12-30
wireguard: update to 0.0.20191226

zx2c4:
This new snapshot comes from the brand new linux-compat repo, which
follows the recent upstreaming into net-next. When Linux 5.6 lands in
LinuxKit, we'll be able to remove the module entirely.

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -38,9 +38,9 @@ ARG KERNEL_SERIES
 ARG EXTRA
 ARG DEBUG
 
-ENV WIREGUARD_VERSION=0.0.20191212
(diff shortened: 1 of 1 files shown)


zx2c4:
CC @rn

deitch:
Hi @zx2c4 . Does this resolve the points you raised in the email you sent last week?

zx2c4:
Yes.

deitch:
Great! What was the other half of that email? What needs to be done there?

zx2c4:
> Great! What was the other half of that email? What needs to be done there?

I'll handle it.

PR #3426, opened 2019-10-04
Bump containerd to v1.3.0

rn:
image

PR #3457, opened 2020-01-07
Update alpine to 3.11

simon-engledew:
**- What I did**

I have updated the `tools/alpine` image to Alpine 3.11 and got it building.

**- How I did it**

Looks like `libelf-dev` has been missing from the Alpine APK indexes since 3.9.

Tracking down the original `APKBUILD` for this package (https://git.alpinelinux.org/aports/tree/main/libelf/APKBUILD?id=26fcb98d3493e2f881998bfdacd76efff0479003), it appears that the website it was being built from has been offline for some time (http://www.mr511.de/software/).

I have added a updated `APKBUILD` recipe that uses a different mirror. The checksum is the same as the original source archive, so the new package should be an exact replacement.

**- How to verify it**

`cd tools/alpine && make build`

**- Why I did it**

I was trying to improve my understanding of how LinuxKit fits together!

I honestly have no idea if this is actually useful. I guess other images would also have to receive updates to test and verify this? I hope that if someone who knows what they are doing bumps into the 'libelf-dev issue' they can at least crib from my solution. :)

If there are next steps I can take with this I am happy to put more work in!

For what it's worth, I've also tried updating containerd to 1.3.2 with this image and it appears to work. Well, the system boots.

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.9 AS mirror
+FROM alpine:3.11 AS mirror
 
 # update base image
 RUN apk update && apk upgrade -a
@@ -6,6 +6,7 @@ RUN apk update && apk upgrade -a
 # Copy Dockerfile so we can include it in the hash
 COPY Dockerfile /Dockerfile
 COPY packages* /tmp/
+COPY libelf /libelf
 
 # mirror packages
 RUN cat /tmp/packages.$(uname -m) >> /tmp/packages && \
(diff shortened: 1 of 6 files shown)


rn:
@simon-engledew thanks for this PR. Unfortunately an update to tools/alpine more or less has to be to done by one of the maintainers as this is the base package for all packages and it needs to be signed etc.

The overall process for updating the Alpine base image is more or less the same as creating a release (see `docs/releasing.md`) and the order in which the updates are done is quite important.

Further, `linuxkit/alpine` is special in that it contains a mirror of all Alpine packages used by other linuxkit packages and the build process and tagging is different. To update, you first have to build, push and sign the x86 version and then git commit and git push the changes it makes to versions.x86_64. Then the same has to happen on a arm64 and s390x server. This gives you a `linuxkit/alpine` image with a particular hash which can then be used in other packages.

Updating contianerd requires similar steps, plus updates to some test packages. https://github.com/linuxkit/linuxkit/pull/3426 is kinda a template for this, but it fails because of contianerd devmapper tests failing. I'll try to update this soon and remove devmapper support.

As for `libelf`, I think we can probably just drop this package. It was used in the kernel build, but as of https://github.com/linuxkit/linuxkit/pull/3323 we are using elfutils-dev

simon-engledew:
@rn Thanks very much for the reply - it's fascinating how everything fits together!

I had a suspicion it wasn't going to be a very useful PR, but I figured if nothing else it would provide documentation for the libelf thing if anyone else searches for it. Good to know it's not needed anymore!

Sorry this ended up being a bit of a dead end - I appreciate the time you put into replying, cheers!

rn:
Yeah, knowing about libelf is useful for me and will safe time if I get around updateing to 3.11 (soon, hopefully).  Justin and I gave a presentation at DockerCon 2018 in Kopenhagen where we go into more details on how the package stuff works. I think the video is somewhere on YouTube and I might have the deck somewhere.

PR #3463, opened 2020-01-22
contrib: Add kubernetes manifest for open-vm-tools

### LINKED ISSUE #3462: Create open-vm-tools Kubernetes DaemonSet
**Description**
Currently we are running Kubernetes cluster on vSphere Hypervisor. To persist data from any container we use Container Storage Interface (CSI) driver for vSphere. It is mandatory for vSphere to have `open-vm-tools` installed. But if you do not have access to the host OS you are stuck. Here comes `linuxkit` provided `open-vm-tools` container which let all the functionality work (i.e. vSphere internal DNS resolution for Cloud Native Storage, graceful guest OS shutdown/reboot). To keep things tightly coupled, it would be nice to have a deployment file for Kubernetes next to the image source.

alex1989hu:
**- What I did**
This should allow end-users to gracefully reboot or shutdown Kubernetes
nodes (incuding control planes) running on vSphere Hypervisor

There are several use cases when cluster administrators are not able to
install extra packages onto the host OS

Fixes #3462

**- How I did it**
Create DaemonSet which runs on all Kubernetes node.

**- How to verify it**
`kubectl apply -f contrib/open-vm-tools/open-vm-tools-ds.yaml`

Add kubernetes manifest for open-vm-tools

### DIFF
--- a/contrib/open-vm-tools/README.md
+++ b/contrib/open-vm-tools/README.md
@@ -1,0 +1,10 @@
+# open-vm-tools
+This should allow end-users to gracefully reboot or shutdown Kubernetes nodes (incuding control planes) running on vSphere Hypervisor.
+
+Furthermore, it is also mandatory to have `open-vm-tools` installed on your Kubernetes nodes to use vSphere Cloud Provider (i.e. determinte virtual machine's FQDN).
+
+## Remarks:
+- `spec.template.spec.hostNetwork: true`: correctly report node IP address; required
+- `spec.template.spec.hostPID: true`: send the right signal to node, instead of killing the container itself; required
+- `spec.template.spec.priorityClassName: system-cluster-critical`: critical to a fully functional cluster
+- `spec.template.spec.securityContext.privileged: true`: gain more privileges than its parent process; required
--- a/contrib/open-vm-tools/open-vm-tools-ds.yaml
+++ b/contrib/open-vm-tools/open-vm-tools-ds.yaml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: open-vm-tools
+  name: open-vm-tools
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: open-vm-tools
+  template:
+    metadata:
+      labels:
+        app: open-vm-tools
+    spec:
+      hostNetwork: true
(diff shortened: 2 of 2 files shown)


tjfontaine:
I think it probably is more germane to https://github.com/linuxkit/kubernetes

However, I think you want to build your nodes with `LinuxKit` and then control the life cycle of `open-vm-tools` external to k8s (because what happens if k8s is down, and kubelet won't know to schedule open-vm-tools).

To that end following the example in https://github.com/linuxkit/linuxkit/blob/ef47cd9c98b76867673d112a42524ef68bcb36d1/docs/virtual-os-tools.md you can have a `vmware.yml` file that can be included on your `linuxkit` command line that adds the appropriate service when targeting VMware platforms.

rn:
@alex1989hu it looks like this file will not be included in the package. It's unclear to me if it should be in the package. Could you describe the use case a bit more in the context of LinuxKit?

alex1989hu:
The reason I created this PR to add the file next to the original source instead of keeping it in any additional repository. For instance, it will definitely have smaller audience and smaller chance to reach out people who also meet this situation. I thought a running open-vm-tools installed in a container means we have a special environment: hosting on vSphere. Thus, I wanted to extend the use-case and the usability of containerization in vSphere. Please consider it before finally closing this PR - it is also acceptable if you do not want this to have in the repository.

rn:
I would strongly prefer to only have files directly used by a package to be in the package directory. The main reason is that we hash the content of a package directory to compute the tag.So really only files directly used by the package should be int package directory. Some documentation/comment in the YAML or similar would also be useful.

Maybe add it under `./contrib/open-vm-tools` together with a README?

alex1989hu:
> I would strongly prefer to only have files directly used by a package to be in the package directory. The main reason is that we hash the content of a package directory to compute the tag.So really only files directly used by the package should be int package directory. Some documentation/comment in the YAML or similar would also be useful.
> 
> Maybe add it under `./contrib/open-vm-tools` together with a README?

I am happy to move it under `contrib` directory. Do you have any directory layout preference? For example: a separate  subdir `contrib/open-vm-tools/`

rn:
`contrib/open-vm-tools/` sounds good

alex1989hu:
> `contrib/open-vm-tools/` sounds good

Kindly ask you to check the recent push

PR #3482, opened 2020-03-23
workaround bad containerd bug

deitch:

**- What I did**

Replaced `/dev/null` for stdin in `cmd` with a temporary null file as `/run/containers-stdin/<servicename>/null`.

This is because of https://github.com/containerd/containerd/issues/4019, where containerd, on task delete, likes to delete everything in the parent directory of stdin. When using `/dev/null`, that means wiping out `/dev`. Oops.

**- How I did it**

Created a directory for all stdins in `/run`, then a directory per named service, then a null file with `mknod`

**- How to verify it**

CI. Run it.

Temporary workaround for bad containerd bug that can wipe out `/dev`

### DIFF
--- a/pkg/init/cmd/service/cmd.go
+++ b/pkg/init/cmd/service/cmd.go
@@ -6,7 +6,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	pathutil "path"
 	"path/filepath"
+
+	syscall "golang.org/x/sys/unix"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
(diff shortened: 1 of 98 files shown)


justincormack:
remove when https://github.com/containerd/containerd/pull/4123 is in a release

pkg/init/cmd/service/cmd.go
```diff
@@ -200,12 +203,31 @@ func start(ctx context.Context, service, sock, basePath, dumpSpec string) (strin
...
+	//
+	// To fix it temporarily, we are creating a tmpdir and creating a null
+	// device there, so that it can blow away the tempdir
+	stdinDir := pathutil.Join("/run", "containers-stdin", service)
+	if err := os.MkdirAll(stdinDir, 0700); err != nil {
+		return "", 0, "failed to create stdin directory", err
+	}
+	stdinFile := pathutil.Join(stdinDir, "null")
+	// make a dev null in stdinDir
+	if err := syscall.Mknod(stdinFile, uint32(os.FileMode(0660)|syscall.S_IFCHR), int(syscall.Mkdev(1, 3))); err != nil {
```
justincormack: you should probably use x/sys/unix not syscall as it is more portable, eg can be broken on some architectures. Although as it is a temporary workaround maybe its ok.

deitch: It is x/sys/unix; I just use syscall as the short form above.

justincormack: Oh, ugh, just call it unix.Mknod then!

deitch: Sigh. Old habits die hard.

deitch: It is a temporary workaround, and rebuilding and repushing on multiple platforms, plus letting the CI run, is a royal pain. Let's leave it for now.

PR #3483, opened 2020-03-28
Update kernels to 5.4.28/4.19.113/4.14.174 and switch to 5.4.x as default

rn:
The 5.4.x is the latest LTS kernel and has been out for a while. Switch to it as the default kernel.

This also fixes up some tests/examples which were still using 4.14.x kernel.

image


justincormack:
Hmm, why is circleci still there (and two DCOs?)

rn:
Hmm, the push of the new kernels seems to have failed. rebuilding/pushing

rn:
CircleCI is till there because we don't have admin rights to remove it. We have removed all the creds etc from github, but also need to disable something on the CircleCI end, I think

justincormack:
I can't see anything on circle end. There is one more place I can try to remove, it may break circle on rtf

justincormack:
Circle says I don't have org admin which is really odd as I do.

justincormack:
possibly fixed?

rn:
Updated the kernels on hub (pushed and signed this time). But, GH actions had a baby when i tried to rerun them (also quite possibly a user error). In any case, slightly changed the commit message of the last commit and force pushed to kick the actions again

Issue #2583, opened 2017-10-10
Switch kernel build to `linuxkit pkg`


ijc:
Once #2578 is put to bed I'd like to convert the kernel build over.

Looking at `kernel/Makefile` I can see 4 possible issues which may need enhancements to `linuxkit pkg`. 

1. Templating to handle multiple kernel versions. I think this will be fine, it'll be multiple invocations of `linuxkit pkg` against either a statically or automatically generated `build-foo.yml`.
2. The use of `--build-arg`. Should be simple enough to add a similar argument to `linuxkit pkg build` (and `push`) which is propagated to the build.
3. Invokes `docker push` and `push-manifest` more than once. This is because it builds and pushes `linuxkit/kernel:X.Y.Z$(EXTRA)-$(TAG)` by default (`X.Y.Z` is the kernel version, `$(EXTRA)` is an optional per image suffix like `-dbg`, `$(TAG)` is the usual tree-sh and `-dirty` suffix) but also tags and pushes `linuxkit/kernel:X.Y.Z$(EXTRA)` (no hash). This differs from the usual pattern of `linuxkit/pkg:$(TAG)`. For this we could add a new field to the `build.yml` called `tag: foo` where `foo` here would be `X.Y.Z$(EXTRA)` (per-kernel). If this field is set then the build/push will be of both `foo-$(TAG)` and `foo` (but not the default `$(TAG)`. Other names such as `alias` or `prefix` might be tempting but don't seem to match the semantics (in particular because the "default" name is not used so it is neither an alias nor is it strictly a prefix).
4. Secondary builds (`perf`, `zfs`) which use the just-now-built kernel image. I think the `build-arg` stuff above handles that?

/cc @rn any of that sound insane? Any other wrinkles with the kernel build I've missed?

rn:
- I think having `tag: foo` in the `build.yml` is fine. if present we should push to `<org>/<img>:<tag>` *and* `<org>/<img>:<tag>-<hash>`
- Specifying `--build-arg` in the `build.yml` would be good, I think. I prefer if I only have to change the kernel version in two places (currently `Makefile` and `kernel-config-X`). If this is an argument to `linuxkit` and is in `build.yml` it'll be three files.
[...]

PR #3484, opened 2020-03-29
eliminate manifest-push-script, jq dependency, odd logic

deitch:

**- What I did**

Replaced the exec of `manifest-push-script`, along with its logic to try and get credentials and dependencies on `jq`, with some golang that uses the native docker libraries.

**- How I did it**

Eliminate the script, use golang code

**- How to verify it**

Try to build and push a package, see if it works with manifests and signing.

Eliminate brittle script


deitch:
I got fed up with the jq headache on s390x (looking at your tweet @rn ), realized the script was a bit of a relic, so I refactored it out.

It now execs two things - manifest-tool and notary - but those might be rolled into the code itself. I would rather do it in two steps though. If this works, let us get it in, and I will open another PR to eliminate those. Part of the issue is that manifest-tool is quite out of date, doesn't use go modules.

One open question: the const that defines the script appears in the file [gen](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/pkglib/gen) which generates it again? I didn't understand what this is for, when it is run. I could not find a reference to it?

deitch:
Ouch. I eliminated the wrong file. Redoing and pushing.

deitch:
There we go. Now I fixed it. Much better.

deitch:
I put in the effort to convert from exec'ing `manifest-tool` to using it as a library. There are some version conflicts between what we include for github.com/docker/docker and what it includes, lots of transitive dependencies. It can be resolved, but should be a separate effort.

deitch:
Found some silly errors; still testing and updating. Will post here when ready.

deitch:
OK, finally all working nicely. Ready for review.

rn:
kernel build still requires `scripts/push-manifest.sh`, and hence `jq` and credentials etc, because it is not yet converted to `linuxkit pkg`. See https://github.com/linuxkit/linuxkit/issues/2583

But this is a goos step. I'll review in a bit

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -166,3 +177,142 @@ func (dr dockerRunner) save(tgt string, refs ...string) error {
...
+		return fmt.Errorf("failed to execute notary-tool: %v", err)
+	}
+
+	// report output
+
+	return nil
+
+}
+
+/*
```
rn: I would prefer to not have dead code here.

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -166,3 +177,142 @@ func (dr dockerRunner) save(tgt string, refs ...string) error {
...
+	log.Debugf("Executing: %v", cmd.Args)
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to execute notary-tool: %v", err)
+	}
+
+	// report output
+
+	return nil
+
```
rn: nit: Spare empty line

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -166,3 +177,142 @@ func (dr dockerRunner) save(tgt string, refs ...string) error {
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Env = append(os.Environ(), fmt.Sprintf("%s=%s", notaryDelegationPassphraseEnvVar, os.Getenv(dctEnvVar)), fmt.Sprintf("%s=%s", notaryAuthEnvVar, notaryAuth))
+	log.Debugf("Executing: %v", cmd.Args)
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to execute notary-tool: %v", err)
+	}
+
+	// report output
```
rn: Is this a a todo? If not remove. if yes, what's missing?

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -5,16 +5,32 @@ package pkglib
...
+const (
+	dctEnableEnv                     = "DOCKER_CONTENT_TRUST=1"
+	registry                         = "https://index.docker.io/v1/"
+	notaryServer                     = "https://notary.docker.io"
+	notaryDelegationPassphraseEnvVar = "NOTARY_DELEGATION_PASSPHRASE"
+	notaryAuthEnvVar                 = "NOTARY_AUTH"
+	dctEnvVar                        = "DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE"
+)
+
+var platforms = []string{
```
rn: This is currently only used in the invocation of `manifest-tool` and could be a constant string, comma separated.

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -166,3 +177,142 @@ func (dr dockerRunner) save(tgt string, refs ...string) error {
 	args := append([]string{"image", "save", "-o", tgt}, refs...)
 	return dr.command(args...)
 }
+
+func (dr dockerRunner) manifestPush(img string, trust bool) error {
```
rn: I may be missing this, but manifestPush() does not need to be a method (or whatever it is called) of dockerRunner, right?

Actually, can't you ditch the `trust` argument and use `dr.dct` in the function directly?

src/cmd/linuxkit/pkglib/docker.go
deitch: It is all a comment. I went through the effort of aligning it, then realized we had the vendor versioning issues. Is there a better place to keep it?

src/cmd/linuxkit/pkglib/docker.go
deitch: Fixed.

src/cmd/linuxkit/pkglib/docker.go
deitch: Shoot. Yes, it is missing. Adding.

deitch: Done.

deitch:
> The file gen should also be deleted

Removed.

src/cmd/linuxkit/pkglib/docker.go
deitch: > manifestPush() does not need to be a method (or whatever it is called) of dockerRunner, right?

You are correct. I will remove it from there.

> Actually, can't you ditch the trust argument and use dr.dct in the function directly?

I like your earlier approach of making it an independent func.

deitch: Done

src/cmd/linuxkit/pkglib/docker.go
deitch: It could, but since it describes all of our supported platforms, I suspect we will either consume it from elsewhere or provide it to elsewhere, so I kept it independent.

justincormack:
No, `docker manifest push` still doesn't support signing (needs the full containerd integration)

deitch:
> No, docker manifest push still doesn't support signing (needs the full containerd integration)

I thought `manifest-tool` doesn't do signing either. If you look at the code (or the original script), it does `manifest-tool` and then `notary` as 2 separate steps?

deitch:
In any case, I am more than happy to have this be one step better than before: eliminates a scripts and a dependency on `jq`. We can do another PR to replace other parts of it. I for one would be very happy to have `go` be our only dependency :-)

src/cmd/linuxkit/pkglib/docker.go
deitch: Removed

deitch:
All comments resolved. Back to you.

PR #3488, opened 2020-04-04
wireguard: bump to 1.0.20200401

zx2c4:


rn:
Thanks, will pick up with the next round of kernel updates

PR #3489, opened 2020-04-04
libunwind-dev workaround on x86 is no longer required

rvs:
Tiny fix for a small workaround that is no longer required


justincormack:
Sigh, how can I delete circleci hooks?

Issue #3498, opened 2020-04-16
perf package fails to build with alpine 3.11


rn:
While updating `linuxkit/alpine` to 3.11, the `kernel-perf` package fails to builds. The primary reason seems to be that alpine 3.11 is missing `libz.a` and we try to build a statically linked `perf` binary.

we will temporarily disable the `perf` build. An alternative would be to compile our own `libz`

thaJeztah:
From slack:

> we normally ship the static library in separate package. you can use `zlib-static`
> reason for splitting static is that alpine itself normally dont use it and sometimes the static lib can be very big

Issue #3496, opened 2020-04-13
go vet/go test/go install on arm64 require gcc


rn:
While updating to alpine 3.11, the script we use to compile go code fails in various places because `gcc` is not installed in the container we compile in on arm64. The same script works fine on amd64 and s390x with the same deps installed.

Repro (this uses `pkg/metadata` from LinuxKit.
```
docker run --rm -it -v $(pwd):/go/src/metadata alpine:3.11
```
then in the container:
```
% apk add go musl-dev linux-headers
% go version
go version go1.13.4 linux/arm64
% export GOPATH=/go
% cd /go/src/metadata
% go test
# runtime/cgo
exec: "gcc": executable file not found in $PATH

FAIL    metadata [build failed]
% export CGO_ENABLED=0
% go test
# metadata.test
loadinternal: cannot find runtime/cgo
/usr/lib/go/pkg/tool/linux_arm64/link: running gcc failed: exec: "gcc": executable file not found in $PATH

FAIL    metadata [build failed]
```

rn:
from alpine slack:
```
ncopa  7:22 AM
I believe this is due to our default -buildmode=pie   https://git.alpinelinux.org/aports/tree/community/go/default-buildmode-pie.patch?h=3.11-stable
7:25
@deitch can you try with alpine:edge  and do apk upgrade -U -a ? We pushed a change in go yesterday that removes our default -buildmode=pie
```
and 
```
ncopa  1 day ago
I don't think we will change that in 3.11, but it will be included in 3.12 which will be released soonish (May hopefully).
```

PR #3500, opened 2020-04-17
Update to Alpine 3.11

rn:
This PR updates `linuxkit/alpine` and all the dependencies to Alpine 3.11.

A couple of notable changes
- `kernel-perf` is disabled for now (see https://github.com/linuxkit/linuxkit/issues/3498). This fill be fixed in the next update
- We had to add `gcc` to a bunch of packages compiling with `go (see https://github.com/linuxkit/linuxkit/issues/3496). Likely will be fixed with Alpine 3.12

image


tools/alpine/Dockerfile
```diff
@@ -56,15 +56,12 @@ RUN go get -u github.com/LK4D4/vndr
 # `test/pkg/containerd/Dockerfile` when changing this.
 # For v1.3.x cherry-pick a upstream commit to allow disabling some tests.
 ENV CONTAINERD_REPO=https://github.com/containerd/containerd.git
-ENV CONTAINERD_COMMIT=v1.3.2
+ENV CONTAINERD_COMMIT=v1.3.3
```
deitch, line 59: Given that we know we need to bump it to v1.3.4 for the fifo fix, why not do it here?

rn: because it was released yesterday and I didn't feel like rebasing 20 odd patches and rebuilding everything again (otherwise commits won't line up). better have this in (assuming it works out) and then update containerd seperately.

deitch: Didn't realize it was quite the headache. No problem then.

rn:
Both failures (in `LinuxKit Build Tests` and `LinuxKit Build Tests`) are in building the GCP image. I think this is due to https://github.com/linuxkit/linuxkit/pull/3300 which I had forgotten to revert, well I had a revert PR which I didn't follow up on

rn:
Interestingly `linuxkit.packages.dm-crypt.luks` failed on this one. IIRC this passed before. Investigating

rn:
the `dm-crypt` test with `LUKS` needed a bigger disk. Pushed another commit

PR #3504, opened 2020-04-20
Add GUID Partition Table (GPT) support to the `format` package

federico-pellegatta:
**- What I did**
I've added GUID Partition Table (GPT) support to the `format` package in order to support partitions greater than 2 TiB.

**- How I did it**
I've added an *optional* `--partition` input parameter to the `format` command-line that can be used to select one of the supported partition table modes:
- `dos`: **(default)** master boot record (MBR), DOS partition table;
- `gpt`: GUID partition table (GPT).

Note: an error is triggered (w/t early termination) if an invalid partition table mode is supplied.
Note: it defaults to the "legacy" MBR/DOS behavior.

GUID Partition Table (GPT) support has been implemented by prepending `g\n` to the `fdisk` standard input string reader.

Screenshot 2020-04-20 at 15 33 35

**- How to verify it**

- use `format` as an on-boot package;
- attach an empty disk > 2 TiB (e.g. `/dev/xvdb`);
- set the partition flag to `gpt` in the command line parameter (e.g. `-partition gpt`);
```
    command: ["/usr/bin/format", "-partition", "gpt", "-verbose", "/dev/xvdb"]
```
- verify the partition table (e.g. `sfdisk -l /dev/xvdb`) to have `Disklabel type: gpt`;
- verify the partition size (e.g. to be greater than 2 TiB).

Added GUID Partition Table (GPT) support to the `format` package.

Signed-off-by: Federico Pellegatta <12744504+federico-pellegatta@users.noreply.github.com>

### DIFF
--- a/docs/external-disk.md
+++ b/docs/external-disk.md
@@ -52,9 +52,17 @@ onboot:
     command: ["/usr/bin/format", "-force", "-type", "xfs", "-label", "DATA", "-verbose", "/dev/vda"]
 ```
 
+```
+onboot:
+  - name: format
+    image: linuxkit/format:<hash>
+    command: ["/usr/bin/format", "-type", "ext4", "-partition", "gpt", "/dev/vda"]
+```
+
 - `-force` can be used to force the partition to be cleared and recreated (if applicable), and the recreated partition formatted. This option would be used to re-init the partition on every boot, rather than persisting the partition between boots.
 - `-label` can be used to give the disk a label
 - `-type` can be used to specify the type. This is `ext4` by default but `btrfs` and `xfs` are also supported
+- `-partition` can be used to specify the partition table type. This is `dos` by default but `gpt` is also supported
 - `-verbose` enables verbose logging, which can be used to troubleshoot device auto-detection and (re-)partitioning
 - The final (optional) argument specifies the device name
 
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -15,7 +15,7 @@ onboot:
   - name: sysfs
     image: linuxkit/sysfs:697f996868d6625a78bbbefb199d7b6b6ae3533a
   - name: format
-    image: linuxkit/format:80bd847a90842d3f19460a560d7d95d6fe13bd71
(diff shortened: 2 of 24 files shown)


deitch:
The document for this needs to be updated [here](https://github.com/linuxkit/linuxkit/blob/ef47cd9c98b76867673d112a42524ef68bcb36d1/docs/external-disk.md)

federico-pellegatta:
@deitch docs updated as per your comment

deitch:
I am going to nitpick on the documentation format; please update it. In the meantime, I will deploy the images so that CI can do the right thing (you need to be a maintainer to do that).

deitch:
Eh, I decided not to nitpick. It doesn't matter. Pushing out image and will need to update hashes. Look for a comment here.

deitch:
Please make the following changes:

1. Squash the commits together into a single commit and force-push back to github
1. Get the tag for format by running the following from the root directory:  `linuxkit pkg show-tag pkg/format`
1. Comment here and I will push out the packages

When that is done, I will comment here, and then you can:

1. Update all of the dependent packages in examples, tests, etc. by running the following from the root directory: `./scripts/update-component-sha.sh --image <output-from-show-tag>`
1. Amend the previous commit to include the changes (they should almost all be in `examples/` and `test/` and force-push
1. Comment here and I will push out the packages.

Thanks!

federico-pellegatta:
Commits squashed and force-pushed. The tag for the `format` package is `linuxkit/format:0b75e494eea0312f3015e6c6f7c5927620d56c96`

deitch:
Done, image pushed out with manifest and signed. Please run the next steps.

federico-pellegatta:
@deitch updated dependent packages and amended the previous commit.

On a side note, do you think it is worth amending this PR with a quick test for the `gpt` partitioning mode in the `linuxkit.packages.format_mount` test suite?

deitch:
Yeah it would be nice if you could. Force push and it will kick off CI again. It doesn’t change the commit for the pkg itself, so I don’t need to repush and sign the image. Please comment here when you’ve got it, and then when CI clears I’ll merge it in

PR #3508, opened 2020-04-23
Add GUID Partition Table (GPT) support to `extend` and `mount` packages

federico-pellegatta:
**- What I did**
I've added GUID Partition Table (GPT) support to extend and mount packages in order to support partitions greater than 2 TiB.

**- How I did it**
Both the `extend` and the `mount` packages are preliminary checking their targets to be Linux native partitions. The `partition.Type != "83"` check was implicitly assuming a DOS/MBR partition table.

In a GPT partitioned disk the partition type entry, in the partition table, for a Linux native partition is `0FC63DAF-8483-4772-8E79-3D69D8477DE4` instead.

The `extend` package was also always using `83` as partition type while re-creating the extended partition. I've updated it to use the one previously read from the partition table instead.

I've also updated all tests in the `extend` package test suite to avoid false positives: instead of checking all the mounted file systems they are now checking only the one mounted on `/var/lib/docker`.

**- How to verify it**
Run the test suite against updated `extend` and the `mount` packages.

Added GUID Partition Table (GPT) support to extend and mount packages.

Signed-off-by: Federico Pellegatta 12744504+federico-pellegatta@users.noreply.github.com


deitch:
I will take a look. Please squash the commits.

federico-pellegatta:
The CI is expected to fail, the newly added test will succeed only after `extend` and `mount` packages update.

@deitch this is based on #3504 and it is needed to fully support GPT partitioned disks.

federico-pellegatta:
Commits squashed and comments added. Tags for packages:
- `linuxkit/mount:19fa297189166206ac97261679c3e31fb140d48f`
- `linuxkit/extend:10bc501a6890be931c6f6602c8caca31337eafeb`

deitch:
Images out:

* `linuxkit/mount:19fa297189166206ac97261679c3e31fb140d48f`
* `linuxkit/extend:10bc501a6890be931c6f6602c8caca31337eafeb`

Kindly update with the script and repush here so CI runs correctly.

PR #3493, opened 2020-04-12
Native manifest

deitch:
**- What I did**

* Remove the dependency on an external [manifest-tool](https://github.com/estesp/manifest-tool) by using it as a library
* updated a lot of dependencies and the long tree (web?) it takes to get it there

**- How I did it**

painfully

**- How to verify it**

run it

**- Description for the changelog**

no longer need external manifest-tool

**Notes**

* I had to update a _lot_ of dependencies in `src/cmd/linuxkit/vendor/` based on `vendor.conf`. I need some real reviews on this.
* I did it in 3 commits to make it easier to review. Once we are satisfied, I will squash them into 1
* this only works with go 1.14 or higher, so our go-compile package will fail on this. I will get a separate PR in to update that

PR #3511, opened 2020-04-26
support merge yaml flags

deitch:

**- What I did**

Providing values for `binds`, etc. in the yaml file overrides the default. This isn't always the desired behaviour, and can lead to confusion. This adds merge capabilities to two fields via new syntax:

* `binds.merge` - merge into the default from the label
* `capabilities.merge` - merge into the default from the label

It also documents it

**- How I did it**

Modified the source

**- How to verify it**

Build and test. We probably should add a test for this. I will look at it.

Support adding capabilities and binds to the defaults, not just overriding them.


deitch:
Added some testing... which fails for reasons I don't understand. I just use the getty image (since it has default binds set up) and override the command, but I get a weird error:

```
Welcome to LinuxKit

                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
           \______ O           __/
             \    \         __/
              \____\_______/

Error creating 000-getty: exit status 1
```

deitch:
Just rebased on master after #3493 merged in. When this one is done, will be looking at push-without-sign and sign-only, to make all our lives easier.

deitch:
BTW, would appreciate if someone can look at the failing build. It is easy to run locally:

1. build `bin/linuxkit` from this
1. `cd test/cases/000_build/020_binds`
1. make sure `test.sh` is using that `linuxkit` binary you built above
1. `RT_PROJECT_ROOT=../../../ sh -x ./test.sh`
1. see the error

Any ideas?

federico-pellegatta:
My suggestion here is to pick another package, instead of `getty`, to perform your test.

The `Error creating 000-getty: exit status 1` issue always occurs, even using `getty` as-is in the `boot` section.

If you really want to stick with `getty`:
- add `linuxkit/containerd` to the `init` section of your `test.yml`, it is needed by `getty`;
- one of the default binds is `/var/lib/containerd:/var/lib/containerd` that is somehow failing when `getty` is placed in the `onboot` section (replacing it with `/usr/bin/containerd:/var/lib/containerd` fixes it).

deitch:
> If you really want to stick with getty:

Not necessarily; I just wanted something with default mounts so I could check the binds merge

> add linuxkit/containerd to the init section of your test.yml, it is needed by getty

I tried; it still gives the error. 

getty works fine in `services`, but not in `onboot` 🤷‍♂️

deitch:
In the end, it isn't the container at all. It is something to do with `check.sh`, which is not getting copied over correctly.

federico-pellegatta:
> getty works fine in services, but not in onboot 🤷‍♂️

It fails because it tries to bind mount `/var/lib/containerd:/var/lib/containerd`.

Here is my working version using `linuxkit/containerd` and a runtime `mkdir`:
```yml
kernel:
  image: linuxkit/kernel:5.4.30
  cmdline: "console=ttyS0 console=ttyAMA0"
init:
  - linuxkit/init:f670045ecb6ec31ea37dd10114366e9a1e915013
  - linuxkit/runc:1eef77f5963e44e491abfe392206769037d47ae2
  - linuxkit/containerd:8ee7a0d636fff9df7e13076f5492d06274e5f644
onboot:
  - name: getty
    image: linuxkit/getty:48f66df198981e692084bf70ab72b9fe2be0f880
    binds.merge:
      - /check.sh:/check.sh
      - /var/tmp:/var/tmp
    runtime:
      mkdir: ["/var/lib/containerd"]
    command: ["/check.sh"]
  - name: poweroff
    image: linuxkit/poweroff:06dd4e46c62fbe79123a028835c921f80e4855d3
    command: ["/bin/sh", "/poweroff.sh", "10"]
files:
  - path: check.sh
    source: ./check.sh
    mode: "0700"
trust:
  org:
    - linuxkit
    - library
```

deitch:
Try this one. Why would this fail?

```yml
kernel:
  image: linuxkit/kernel:5.4.30
  cmdline: "console=ttyS0 console=ttyAMA0"
init:
  - linuxkit/init:7195dc244cd92af01fd0895fd204249a6114c5e2
  - linuxkit/runc:f79954950022fea76b8b6f10de58cb48e4fb3878
onboot:
  - name: mount
    image: linuxkit/mount:19fa297189166206ac97261679c3e31fb140d48f
    binds.merge:
      - /check.sh:/check.sh
      - /var/tmp:/var/tmp
    command: ["/check.sh"]
  - name: poweroff
    image: linuxkit/poweroff:06dd4e46c62fbe79123a028835c921f80e4855d3
    command: ["/bin/sh", "/poweroff.sh", "10"]
files:
  - path: check.sh
    source: ./check.sh
trust:
  org:
    - linuxkit
    - library
```

FWIW, here are the defaults from `linuxkit/mount`:

```yml
- /dev:/dev
- /var:/var:rshared,rbind
- /:/hostroot
```

deitch:
If I add `getty` so I can get in and check the logs, I get:

```
time="2020-04-27T15:30:03Z" level=error msg="container_linux.go:349: starting container process caused \"exec: \\\"/check.sh\\\": permission denied\""
container_linux.go:349: starting container process caused "exec: \"/check.sh\": permission denied"
```

Which is strange, since it has the right perms.

federico-pellegatta:
It fails because the `mode: "0700"` is missing for `check.sh` in the `file` section.

deitch:
> It fails because mode: "0700" is missing for check.sh in the file section.

Oh drat! 🤦🏻‍♂️You are right. Works after that.

federico-pellegatta:
Alternatively you can remove `mode` (default is `0600`) and have `command: ["/bin/sh", "/check.sh"]`.

deitch:
Phew. Finally works. Thanks for the assist, @federico-pellegatta 

Ready for a review.

rn:
wouldn't `binds.add` and `capcbilities.add` a better name? This doesn't allow you to remove entries from the default, just adding them, right?

dave-tucker:
Funny, I was just thinking the same thing about not being able to remove binds :)

deitch:
> This doesn't allow you to remove entries from the default, just adding them, right?

Correct. If you want to remove, just use `binds` and it will replace them.

I don't know that it is worth the effort to come up with a syntax that will allow you to remove 1 of 4 and add 2 new ones. Want to add a few? Sure. Replace it all? Sure. Weird mixes? Just rewrite it as replace.

> wouldn't binds.add and capabilities.add a better name? 

I don't care all that much. If the consensus is change it, we can change it. Rolf, Dave, is that the consensus?

dave-tucker:
@deitch can you provide some more detail on the use case for merging binds?

playing devils advocate, but we could fix the documentation to say that anything in `binds` or `capabilities` always overrides those included in the image labels and that might address the usability issue.

I recall, albeit a long time ago, needing to add binaries to `linuxkit:sshd` and the advice I received was to build my own `sshd` image that builds on top the linuxkit one - which while it initially seemed like a PITA to maintain as I have to be responsible for image base bumps, also seems to more closely follow the "linuxkit way".

If you are needing to add binds or capabilities then your image is probably doing something different from what the base was intended for anyway?

My only worry with this would be say:
a) User uses `binds.add` in `linuxkit.yml`
b) User naively updates the image which has conflicting `binds` because of enhancements upstream
c) Build fails, or unexpected runtime behaviour.

dave-tucker:
> I don't care all that much. If the consensus is change it, we can change it. Rolf, Dave, is that the consensus?

I'm happy with `*.add`. It might allow for `*.drop` in future to allow for removing caps/binds.

deitch:
OK, replacing with `.add`

deitch:
Updated to `*.add`. Will respond to your longer comment shortly @dave-tucker

deitch:
> @deitch can you provide some more detail on the use case for merging binds?

> playing devils advocate, but we could fix the documentation to say that anything in binds or capabilities always overrides those included in the image labels and that might address the usability issue.

We already do. It still takes people by surprise over and over again. Truth is, I understand it. If an image ships with ten binds and four capabilities that make sense, and I want to add one more, why should it not "just work"? Now I have to read into it, figure out all of the binds or capabilities that exist, and copy them over. If a new version comes out, and it has more binds/capabilities, I lose those and something breaks.

> If you are needing to add binds or capabilities then your image is probably doing something different from what the base was intended for anyway?

Not necessarily. The biggest examples are init and sshd. I legitimately might want one more mount in there. Why would I want that to break everything?

> My only worry with this would be say:
> a) User uses binds.add in linuxkit.yml
> b) User naively updates the image which has conflicting binds because of enhancements  upstream
> c) Build fails, or unexpected runtime behaviour.

I can see that, but I am less worried about that theoretical behaviour, than about one that regularly trips people up nowadays.

PR #3513, opened 2020-04-26
Update containerd to v1.3.4

rn:
While at it:
- bump runc
- remove workaround in init
- add zlib-static to alpine base (wip https://github.com/linuxkit/linuxkit/issues/3498)

image


rn:
why dod you want to squash the commits? They are logical steps and one doesn't break the next?

deitch:
Thanks for doing this. I tested the wonky issue with containerd and it definitely is fixed.

PR #3518, opened 2020-05-07
Smaller qemu

justincormack:
Add Risc-V support and only ship binaries we use to keep the image small (accidentally shipping all the static images).
    
Risc-V ported from https://github.com/docker/binfmt/pull/21 by @carlosedp


carlosedp:
LGTM

PR #3532, opened 2020-05-24
options to split image steps and manifest steps

deitch:

**- What I did**

Split `pkg push` into 2 distinct stages, but can be executed as one:

1. Push the image from your local arch (e.g. amd64, arm64, etc.)
1. Create and push the manifest and sign it

The `docs/packages.md` has been updated to reflect how to do it. Short form:

1. Go to each arch and do 

```
lkt pkg push --manifest=false path/to/pkg
```

1. Go to wherever you have your signing keys and do

```
DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE=<whatever> lkt pkg push --image=false path/to/pkg
```

You can disable signing as well by doing `--sign=false`

The default remains the current behaviour 

**- How I did it**

Split up the code in `src/cmd/linuxkit/pkg_push.go`

**- How to verify it**

Create a package, test it.

Push packages and push manifest and sign manifest optionally distinct stages

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -7,7 +7,7 @@ packages, as it's very easy. Packages are the unit of customisation
 in a LinuxKit-based project, if you know how to build a container,
 you should be able to build a LinuxKit package.
 
-All LinuxKit packages are:
+All official LinuxKit packages are:
 - Signed with Docker Content Trust.
 - Enabled with multi-arch manifests to work on multiple architectures.
 - Derived from well-known (and signed) sources for repeatable builds.
@@ -15,6 +15,7 @@ All LinuxKit packages are:
 
 
 ## CI and Package Builds
+
 When building and merging packages, it is important to note that our CI process builds packages. The targets `make ci` and `make ci-pr` execute `make -C pkg build`. These in turn execute `linuxkit pkg build` for each package under `pkg/`. This in turn will try to pull the image whose tag matches the tree hash or, failing that, to build it.
 
 We do not want the builds to happen with each CI run for two reasons:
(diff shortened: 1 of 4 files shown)


deitch:
FYI I did think about making it two or three separate commands:

- pkg push - just for pushing the package
- pkg manifest - for creating the manifest 
- pkg sign - for signing the manifest

A different variant would be merging the last two, so that you would have 


- pkg push - just for pushing the package
- pkg manifest - for creating and signing the manifest, depending on `--sign=false|true`, default to true. If people prefer that, I’m happy to switch it.

rn:
I prefer one `pkg push` which by default pushes the package and manifest and signs it. This should be the default. Options to disable some of the steps are fine.

docs/packages.md
```diff
@@ -73,38 +74,89 @@ should also be set up with signing keys for packages and your signing
...
+* `linux/amd64`
+* `linux/arm64`
+* `linux/s390x`
 
-```
-DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="<passphrase>" linuxkit pkg push «path-to-package»
-```
+Official images *must* be built on all architectures. They can be built and
+pushed in parallel, but the manifest *must* be pushed on one platform only, when
+all of the images are done.
```
rn, line 85: Why? This makes the package push more complicated. By default a push should create the manifest list if not present or update the existing one, and then sign the new manifest list

docs/packages.md
```diff
@@ -73,38 +74,89 @@ should also be set up with signing keys for packages and your signing
...
 
-To build a package on an architecture:
+* `linux/amd64`
+* `linux/arm64`
+* `linux/s390x`
 
-```
-DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="<passphrase>" linuxkit pkg push «path-to-package»
-```
+Official images *must* be built on all architectures. They can be built and
```
rn, line 83: Not all official packages are available on all platforms

docs/packages.md
```diff
@@ -73,38 +74,89 @@ should also be set up with signing keys for packages and your signing
...
+
+Repeat the above on each platform where you need an image.
+
+This will do the following:
+
+1. Determine the name and tag for the image as follows:
+   * The tag is from the hash of the git tree for that package. You can see it by doing `linuxkit pkg show-tag «path-to-package»`.
+   * The name for the image is from `«path-to-package»/build.yml`
+   * The organization for the package is given on the command-line, default to `linuxkit`.
+1. Build the package in the given path using your local docker instance for the local platform. E.g. if you are running on `linux/arm64`, it will build for `linux/arm64`.
```
rn, line 118: shouldn't this be `2.` (and `3.` etc below). or does the markdown renderer do the right thing?

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -279,7 +300,7 @@ func signManifest(img, digest string, length int, auth dockertypes.AuthConfig) e
 	}
 
 	// report output
-	fmt.Printf("New signed multi-arch image: %s:%s\n", repo, tag)
+	fmt.Printf("Signed manifest: %s:%s\n", repo, tag)
```
rn, line 303: This should be `manifest list`

deitch:
That means leaving it the way it is now then @rn 

Take a look and review then please.

docs/packages.md
deitch: Is that how we want to do it? Part of the issue we had was pushing in parallel could create manifest indexes (technically, indices) that would trounce each other. By saying it this way, we encourage people to build and push in parallel, and then do the manifest at the end.

docs/packages.md
deitch: Oh. I had thought the language was there before. I can remove this.

docs/packages.md
deitch: Markdown does the right thing. Saves me lots of renumbering headache.

src/cmd/linuxkit/pkglib/docker.go
deitch: Sure

deitch:
Manifest list or index, per the OCI spec. I’m ok either way.

PR #3533, opened 2020-06-04
[WIP] Raspberry Pi4 support

dgiebert:
**Status**
- [x] Reworked the mkimage-pi3 to a more generic mkimage-rpi
- [x] Included agherzan u-boot fork as upstream did not work
- [x] Boots into getty
- [ ] Rework the kernel configuration
  - [x] Ethernet Compatibility
  - [ ] SD Card Support (currently working on)
  - [ ] Wifi Support (not tested yet)
- [ ] Documentation including a sample for RPi3 and RPi4
- [ ] Replace the current ubuntu boot script !

**- How I did it**
- Verified the build using a physical RPi4 to build and run


rn:
Thanks for taking a stab at this. Is this still WIP?

tools/mkimage-rpi/build.yml
```diff
@@ -1,4 +1,4 @@
-image: mkimage-rpi3
+image: mkimage-rpi4
```
rn: ideally the package/image name is the same as the directory name. by renaming it to `mkimage-rpi4` it implies it won't work on rpi3 anymore

dgiebert: I will rename this to mkimage-rpi

dgiebert:
I currently lack a RPi3 to test this on, will get to it next weekend and finish it

Should I put the example into the Readme or into the examples folder ?

rn:
Example would be good. Also, it might be better to have the kernel config change as a separate PR. kernel builds are different from the rest and I can compile/push something earlier (we have a arm64 server on packet.net for Linuxkit use). For the kernel config keep in mind that ist must be the exact config; we diff the generate config from `make oldconfig` and error if it is not the same as the one passed in to make sure we don't accidentally miss new kernel options.

dgiebert:
I will split the kernel into a new PR tonight and check the `make oldconfig` currently I only adapted 5.6 due to the better RPi4 support though.

I will create an example with comments in `example/rpi.yaml` then for both RPis and explain in the docs

rn:
Thanks. 5.4 support only is fine. Happy to then pull the changes into 5.6/5.7

dgiebert:
Just a short clearification with the support for the PI mainly arriving in 5.5 [Source](https://github.com/lategoodbye/rpi-zero/issues/43) it will be for 5.6 and up.

I can btw also implement and test for 5.7 if you have a base aarch64 config for me ?

rn:
Pick which ever kernel you think is best. 5.6 works for me. I haven't done the 5.7 work yet, but if the enablement is in 5.6 it should be picked up once I do the 5.7 work

docs/platform-rpi3.md
```diff
@@ -2,8 +2,8 @@
 
 LinuxKit supports building and booting images on a Raspberry Pi 3b
 using the mainline arm64 bit kernels. The LinuxKit arm64 kernel has
-support for some of the devices on the Raspberry Pi 3b, but notably it
-does *not* support:
+support for some of the devices on the Raspberry Pi 3b and 4, but
```
riconnon, line 5: Is it worth also adding 3b+ in this list?

yurinnick:
I'd like to start working on LinuxKit + RPi4 releated project, is there any change to get it merged anytime soon?

Nicola-Bini:
Any chance this will be reviewed? 
I suppose there was a problem with the check not using ARM, since Rasperry Pi 4 has an ARM processor

jmgilman:
Does anyone have a working solution while this sits in review?

dgiebert:
This week I will try to make this work again:
- Modules/Patches for 5.15 
- build an ISO image
- Finish the integration into the CLI

McCodeman:
Is this PR still active? Is there a timeline/roadmap for RPi4 support?

farmdawgnation:
Hey all, just adding my voice here - this would be helpful for some stuff I want to string together with Tinkerbell. Any hopes of this getting revived?

deitch:
I would like to see it as well. I am a maintainer, so if someone wants to bring this up to date, I can review and get it in. I sometimes have access to an rpi4, but I would rely mainly on whoever takes the lead on this to verify it.

Since this came out, we have 6.6.x series kernel support (and riscv64 kernel almost in, but that doesn't affect rpi4 😁 ), and many of the packages have had their dependencies updated. I do not know if those make it easier to support rpi4 or rpi5.

deitch:
> want to string together with Tinkerbell

Oh Tinkerbell! I did work with Packet for a time, including when Tinkerbell was being built, so I kept a close eye on it back then. That always has a warm spot in my heart.

deitch:
Looking at the current state of the PR, it doesn't look very complex. I see some changes and have some questions, for whoever wants to pick it up:

* I think the `docs/platform-rpi3.md` probably should be called `docs/platform-rpi.md`
* the kernel config changes for 5.10.x probably should be dropped
* do we need kernel config changes for 6.6.x to support rpi? Or will it work as is out of the box?
* do the `mkimage-rpi` tools work with any rpi? Or just the 4? Based on the changes to the doc, it looks like either?

frmrm:
Let me do some digging around and see how far I get. I might be able to configure my rpi to netboot my test images for easy testing. I think qemu might have some kind of rpi4 support now too...

frmrm:
Ha. Accidentally left toss comment from my other account but I and @farmdawgnation  are one and the same. :)

PR #3540, opened 2020-07-24
Make `pkg build` build the image

KenjiTakahashi:
**- What I did**

Fixed `pkg build` command not actually building anything. Think the problem originates from changes in https://github.com/linuxkit/linuxkit/commit/ea18be414eb85888f9990a7821608cdd505a2fe2.

**- How I did it**

Added `WithBuildImage` to opts in `pkg_build`.

**- How to verify it**

Running for example `linuxkit pkg build -org=example -disable-content-trust pkg/sshd` should result in a built image.

Fixed `pkg build` command not building the image.

### DIFF
--- a/src/cmd/linuxkit/pkg_build.go
+++ b/src/cmd/linuxkit/pkg_build.go
@@ -29,7 +29,7 @@ func pkgBuild(args []string) {
 
 	fmt.Printf("Building %q\n", p.Tag())
 
-	var opts []pkglib.BuildOpt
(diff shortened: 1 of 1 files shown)


deitch:
Well, wow. That was a sloppy one we missed. I think because we so rarely do `pkg build` - the vast majority of the time it is `pkg push` with its implicit build - that we missed it.

I ran tests to prove that the issue exists, and that this fixes it without unforeseen side effects. Looks good.

Thanks for stepping up on this @KenjiTakahashi

PR #3545, opened 2020-08-23
Move on to adapting Modern Go

stevefan1999-personal:
Signed-off-by: Steve Fan <29133953+stevefan1999-personal@users.noreply.github.com>

**- What I did**
See title

**- How I did it**
Putting the vendor.conf into the go.mod file and then fix some package incompatibility issue. Mainly docker where they use capital S in the author of logrus, why...

**- How to verify it**

- Updated Azure API code where they refactored it to OOP style.
  - Tried to update some method calls that uses context instead of old channel style. Since I had no knowledge of context before, I bet `context.Background()` should do the work
  - Enum rename
- Updated OpenStack code for field rename
- Fixed Packet function rename
- Removed vndr from Dockerbuild (probably)
- Major: Removed all vendor package code. This should impact the GOPATH, but this should be automatically fetched during `go build`
- Hacking some packages to accommodate API breaking changes
  - azure-vhd-utils hacked to use refactored Azure SDK
  - notary, manifest-tool hacked to point to the recent version of packages in the Docker ecosystem 

**- A ~picture~ video of a cute animal (not mandatory but encouraged)**
https://www.youtube.com/watch?v=bFvqf8Z13PU


deitch:
Wow. I am thrilled to see this move to modern go. I do have some questions about this.

* it looks like you kept the `vendor/` dir, just using `go -mod=vendor`, is that correct? 
* what is in the root directory `go.mod`? Everything that requires go imports is in `pkg/<something>/`, `tools/<something>/`, `src/cmd/linuxkit/`, etc.
* Most importantly, is it possible to have separate PRs for the major changes, or at least to group them? Reviewing something with massive file changes (modules) along with other functional changes is difficult. Further, if anything breaks, it will be hard to roll back part of it. Sometimes it isn't possible, due to specific versions not being available, but short of that, I would prefer to have 1 PR for the go modules switch, or better yet one for each package, as there are only 7 `vendor/` directories in here, but barring that, one for the `pkg/` subduers and one for `src/cmd/linuxkit`, and then 1 for each of the 5 or 6 other changes. We probably can group some of them together, if needed, but the big separation - if doable - should be done.

Thanks for this.

stevefan1999-personal:
@deitch I'm quite nooby in golang right now and I'm still learning haha. 

One thing I noticed is that the packages in pkg like metadata has to be vendor as of now due to CI pipeline not importing the GOPATH in the root. That's my hypothesis of why the build fails and therefore issued a module download in the builds containers. Then I tried separate specific dependencies to each package like using vendor.conf before and do `go mod vendor` as you mention, still to no avail...it keeps contacting go package server but there no network in the container...

I'm still yet into Go modules, worse now their v1 and v2 differences are somewhat confusing me...

deitch:
The fails might be because of that, or might be because a maintainer needs to build and push out new container images for CI to pass. 

Let's keep it simple. Can you keep this PR open, switch just one part (e.g. `src/cmd/linuxkit`) to go modules with `go --mod=vendor` so everything is in the `vendor/` dir, and we can work through the issues with just that one? Then we can do stuff in `pkg/`, then the rest?

rn:
@deitch can this be closed due to your recent effort?

deitch:
Yes I believe so.

rn:
@stevefan1999-personal I'm going to close this as we now have moved to go.mod.

PR #3547, opened 2020-08-31
add support for build-args

deitch:

**- What I did**

Added support for `--build-arg` to `lkt pkg build`, passing them on to `docker build --build-arg`

**- How I did it**

Added the command option to the `pkg build` flagSet, and then creating `WithBuildArg()` options

**- How to verify it**

Run it with and without the `--build-arg`

Support for `--build-arg`

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -204,6 +204,12 @@ linuxkit pkg build -org=wombat -disable-content-trust -hash=foo push
 and this will create `wombat/<image>:foo-<arch>` and
 `wombat/<image>:foo` for use in your YAML files.
 
+#### Build Args
+
+Like [docker build-time variables](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg), you can set build-time variables when running `linuxkit pkg build --build-arg=key=val`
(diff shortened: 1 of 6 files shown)


rn:
One of the key concepts behind the linuxkit package is that the build entirely depends on files checked into a git repository (with the git commit and repository being recorded in the package meta-data) and signed files from docker hub for the build. This PR seems to add the ability to change the build based on command line arguments passed into the build.

deitch:
Correct. We (you and I and some others) had discussed this verbally a while back. I wanted to raise the discussion again. 

This really has its roots in the docker build debate (years ago) about whether or not to include `--build-arg` for roughly the same reason. If I recall correctly, they ended up including it because practically won. It was just too necessary. Without it, you end up doing some really convoluted things that don't end up helping, like having a template file for the dockerfile checked in, adding the dockerfile to gitignore, and then generating the dockerfile separately prior to running `lkt pkg build`.

At a certain point we end up running against what people practically need to do. 

I'm coming down (obviously) on the docker-historical side of needing build args - in the end, `lkt pkg build` is just a fancy wrapper around `docker build` - but I would like to have the conversation here.

PR #3552, opened 2020-09-11
Added hash option to yaml

peter-pulik:
Signed-off-by: Peter Pulik <petopulik@gmail.com>

**- What I did**
I added the option to define package tag/hash in yaml file. I used the keyword "hash" for it to be consistent with a CLI argument name.

**- How I did it**
The priority is now:
CLI argument
YAML
hash value computed from GIT commit

**- How to verify it**
Try adding "hash: <HASH>" to any package yaml file

Added hash option for package yaml.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -37,6 +37,7 @@ A package source consists of a directory containing at least two files:
 
 - `image` _(string)_: *(mandatory)* The name of the image to build
 - `org` _(string)_: The hub/registry organisation to which this package belongs
+- `hash` _(string)_: The hash used for the resulting package (by default computed from the git commit)
 - `arches` _(list of string)_: The architectures which this package should be built for (valid entries are `GOARCH` names)
 - `extra-sources` _(list of strings)_: Additional sources for the package outside the package directory. The format is `src:dst`, where `src` can be relative to the package directory and `dst` is the destination in the build context. This is useful for sharing files, such as vendored go code, between packages.
 - `gitrepo` _(string)_: The git repository where the package source is kept.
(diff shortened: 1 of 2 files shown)


rn:
what would be the use case for this?

peter-pulik:
I personally find it better to tag packages based on the version of the application that's inside it. However, if I use --hash CLI option, I have to either remember the version of the application or keep a separate list of versions.

rn:
IIRC, the `-hash` option is only there for development purposes (I've never used it, if I build local i typically us `-dev`) and the `-release` option for `linuxkit pkg push` is there to tag packages with more human friendly names

peter-pulik:
I see. I still personally prefer my approach, however if this is the intended workflow I suppose we can close the PR now. Thank you for your input and hopefully my next PR will be better.

rn:
It's a tool. Different people have different workflows, just trying to explain the main/intended workflow. The hash is basically "these are all the bits this package was build from" and with the core packages we try to build them in a way that they are not pulling random bits from the internet. For LinuxKit releases we then tag the pacakges so the names are more readable.

I can see the use for having more flexible package tags. For example, instead of having `linuxkit/node_exporter:v0.8` for the Prometheus Node Exporter package, it might be useful to have the version of `node_exporter` somewhere we well. TBH, I have not thought much about how to best do this but think that using the `hash` for this is not quite right.

peter-pulik:
At first I thought about calling it a “tag”, not a “hash”. However, it achieves the same result as defining -hash CLI argument, so I thought it would be better this way. I wanted to put the version in the hash/tag to be consistent with how containers are usually versioned.

PR #3554, opened 2020-10-05
Update docker 20.10.6

thaJeztah:
Wasn't sure if anything else was needed to bring these up to date 😅 

/cc @rn @justincormack

### DIFF
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -31,7 +31,7 @@ services:
     image: linuxkit/openntpd:v0.8
 
   - name: docker
-    image: docker:19.03.8-dind
+    image: docker:20.10.6-dind
     capabilities:
       - all
     net: host
--- a/examples/docker-for-mac.yml
+++ b/examples/docker-for-mac.yml
@@ -83,7 +83,7 @@ services:
   # Bind mounts /var/run to allow vsudd to connect to docker.sock, /var/vpnkit
   # for vpnkit coordination and /run/config/docker for the configuration file.
   - name: docker-dfm
-    image: docker:19.03.8-dind
+    image: docker:20.10.6-dind
     capabilities:
      - all
     net: host
--- a/examples/docker.yml
+++ b/examples/docker.yml
@@ -28,7 +28,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:v0.8
   - name: docker
-    image: docker:19.03.8-dind
+    image: docker:20.10.6-dind
     capabilities:
      - all
     net: host
--- a/projects/compose/compose-dynamic.yml
+++ b/projects/compose/compose-dynamic.yml
@@ -29,7 +29,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:v0.8
   - name: docker
-    image: docker:19.03.8-dind
+    image: docker:20.10.6-dind
     capabilities:
      - all
     net: host
--- a/projects/compose/compose-static.yml
+++ b/projects/compose/compose-static.yml
@@ -29,7 +29,7 @@ services:
   - name: ntpd
     image: linuxkit/openntpd:v0.8
   - name: docker
-    image: docker:19.03.8-dind
+    image: docker:20.10.6-dind
     capabilities:
      - all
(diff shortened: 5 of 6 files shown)


thaJeztah:
Updated to docker 20.10.6 and the latest containerd v1.4.x release

@rn @djs55 @deitch ptal (hope I did this right)

deitch:
The `examples/` and `test/` all look fine. The change to `tools/alpine` is a lot more convoluted than that. I am in the process of updating everything (and I mean, _everything_) with an update to `tools/alpine`. It has been a few days of work.

Is containerd 1.4.4 required for docker 20.10.6? Or can it run on containerd 1.4.1? If it is not critically required, I would recommend taking out the containerd change, push the others through, and then we can do a separate PR just for that. If it is definitely required, then let's hold off on this until the other PR #3623 is in, and we can do a separate (but much smaller) alpine one.

thaJeztah:
> The change to tools/alpine is a lot more convoluted than that.

Ah thanks! Yes, that's the part I wasn't sure about

> Is containerd 1.4.4 required for docker 20.10.6? Or can it run on containerd 1.4.1? 

I think _technically_ 1.3.x would work as well, but we specify v1.4.1 as minimum version, as that's what we shipped/tested Docker 20.10.0 with; https://github.com/docker/docker-ce-packaging/blob/b30ac69f11d8046e41478c9e901a66894690a320/deb/common/control#L30

There were some bug fixes in patch releases after v1.4.1 though, so would be good to have.

>  let's hold off on this until the other PR #3623 is in, and we can do a separate (but much smaller) alpine one.

Sounds ok to me; I just realised that I also forgot to update runc, which looks to be handled separately; https://github.com/linuxkit/linuxkit/blob/45429986290f3811b6da093e8fdbbaffce96616e/pkg/runc/Dockerfile#L14

About runc; let me add some info (for when you'll be working on updating containerd); 

If you are updating containerd, and want to know the "matching" runc version (that containerd has been tested against); upcoming releases will have the version specified in a separate file; see this pull request for more details: https://github.com/containerd/containerd/pull/5036

Docker currently ships runc version v1.0.0-rc93 because it fixes some issues that were commonly hit, _however_ depending on how it's used, you _may_ want to use v1.0.0-rc92 or (hopefully soon..) v1.0.0-rc94, because while rc92 fixed some issues, it also introduced a regression in other regions; see https://github.com/containerd/containerd/pull/5289 and https://github.com/opencontainers/runc/issues/2865) - yeah we're a bit between a rock and a hard place until rc94 is release 😞

thaJeztah:
Reverted the containerd update 👍

deitch (approved):

deitch:
Sounds like we should get this in (just did "LGTM"), and then, when #3623 is done and merged, do a separate update for containerd 1.4.4, and one for runc.

Isn't containerd 5.x supposed to be on the cusp?

thaJeztah:
Yes, containerd 1.5.0 is getting close, but containerd maintains a 6 month support overlap between releases, so version 1.4 will continue to be maintained until (at least) 6 months after v1.5.0; https://github.com/containerd/containerd/blob/master/RELEASES.md#support-horizon

Docker will likely switch to 1.5 with the docker 21.xx release (date TBD); we have done some tests in CI to verify containerd doesn't break there, but we're good at "finding the odd bugs" once people start using it en-masse, so perhaps using "latest 1.4.x" as a stepping-stone between would be good

rgl:
Ping?

Maybe also include latest containerd 1.5.2?

PS: I'm looking forward for using cgroup v2!

djs55:
I think we can merge this now.

Since #3623 has been merged, we can make a separate PR for containerd 1.4.4 and one for runc as @deitch suggests above.

djs55 (approved):

thaJeztah:
> Edit: looks like containerd 1.4.4. was included in 95ce638

Probably should be updated to containerd 1.4.6 and runc v1.0.0-rc95

djs55:
@thaJeztah I've made separate PRs for containerd.1.4.6 and runc-v1.0.0-rc95 as you suggested!

PR #3555, opened 2020-10-05
docker: set oom score adjust to -500

thaJeztah:
relates to https://github.com/moby/moby/pull/41528

Currently, dockerd sets a "-500" oom-score-adjust for itself by default.

We're removing this default, and instead:

- for dockerd running as systemd unit, set the oom-score through systemd
- when manually running dockerd, require users to explicitly set a score

This patch is in preparation of those changes.

### DIFF
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -44,7 +44,8 @@ services:
       - /var/run:/var/run
       - /lib/modules:/lib/modules
       - /etc/docker/daemon.json:/etc/docker/daemon.json
-    command: ["/usr/local/bin/docker-init", "/usr/local/bin/dockerd"]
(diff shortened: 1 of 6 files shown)


thaJeztah:
@justincormack ptal

justincormack:
Not sure this is the right way to do it, linuxkit can set the oom score directly in the config which seems better

thaJeztah:
Oh, yes, I'd prefer that as well; this was mostly to keep the old behavior, and I'm not really familiar with LinuxKit; can you point me to an example? Happy to update then

rn:
`examples/docker.yml` and others contain an example on how to add a custom `daemon.json` file.  Is this what you are looking for?

PR #3556, opened 2020-10-07
bump alpine version, add openssh-client, wireguard apk, containerd 1.4.1

deitch:
**- What I did**

Upgraded the base linuxkit/alpine to:

* be based on alpine:3.12
* include `sudo` (needed for `abuild`)
* add `openssh-client` package 
* replace manual build of wireguard with apk package
* bump containerd to 1.4.1

**- How I did it**

Followed the guide [here](https://github.com/linuxkit/linuxkit/blob/972c41c4fa56e69b4f15a9b5637fce866da46b86/docs/releasing.md#update-linuxkitalpine), and then @justincormack kindly signed the manifest because my signing doesn't work (thanks Justin!)

**- How to verify it**

CI

Bump key package versions


justincormack:
Does that also switch containerd to shim v2? (or was that done before?)

deitch:
> Does that also switch containerd to shim v2? (or was that done before?)

I don't actually know. Is that a switch from 1.3.x to 1.4.x?

justincormack:
Actually I dont think it matters here, this will be set in any config in the containerd package.

justincormack (approved):

tools/alpine/packages
```diff
@@ -87,12 +87,13 @@ open-iscsi
...
 openrc
 openresolv
+openssh-client
 openssh-server
 openssl
 openssl-dev
 patch
 pigz
-python
+python2
```
justincormack, line 96: Hmm, no we should not install python 2 it is out of support

deitch:
I will wait for CI to clear; once it is green, I will merge it in. @rn if you have any comments you want on it before that, feel free.

tools/alpine/packages
justincormack: What requires it if anything?

deitch: This is a good question. Do we depend on it anywhere? I was trying to make as little of a change as possible, in this case making sure that python2 was where it was expected

deitch:
```
grep -r python . 2>/dev/null | grep -v -w vendor | grep -v 'tools\/alpine'
./tools/grub/Dockerfile:  python3 \
./tools/grub/Dockerfile:# because python is not available
./tools/grub/Dockerfile:RUN ln -s python3 /usr/bin/python
./projects/miragesdk/examples/https-unikernel/Dockerfile:RUN sudo apt-get install -y screen python-pip python-setuptools python-dev --no-install-recommends
./projects/miragesdk/examples/https-unikernel/README.md:$ python
./projects/miragesdk/examples/https-unikernel/README.md:(or just do `python src/test_store.py`)
./projects/memorizer/docs/scripts/README.txt:Ex: python memorizer.py "ls" "mkdir dir"
./projects/memorizer/docs/scripts/README.txt:Ex: python test_memorizer.py -e
./scripts/kconfig-split.py:#!/usr/bin/env python
./kernel/Dockerfile.bcc:  python \
./kernel/Dockerfile.bcc:RUN mkdir -p /out/usr/lib/python2.7 && \
./kernel/Dockerfile.bcc:    cp -a /usr/lib/python2.7/site-packages /out/usr/lib/python2.7/
./kernel/Dockerfile.bcc:  python \
./kernel/Dockerfile.perf:    python3 \
```

So:

* `tools/grub` uses python3, so that's fine
* `projects/miragesdk/examples/https-unikernel/Dockerfile` uses its own `FROM`, so it doesn't count
* `kernel/Dockerfile.perf` uses python3, so that's fine
* `scripts/kconfig-split.py` is used on the developer platform, so that's fine
*  `kernel/Dockerfile.bcc` uses python

Basically only `kernel/Dockerfile.bcc`. @rn might know?

PR #3557, opened 2020-10-07
include openssh-client in sshd pkg

deitch:

**- What I did**

Added the `openssh-client` package to `linuxkit/sshd`. This enables people to ssh _from_ being on the host (as opposed to ssh to, which you already can do), as well as scp from it and to it.

**- How I did it**

1. Added the `openssh-client` apk package from the linuxkit/alpine cache
1. built and pushed `pkg/sshd` on all platforms, and signed the manifest
1. ran `update-components-sha.sh`, which updated `projects/` and `examples/`

**- How to verify it**

CI

ssh client tools in linuxkit/sshd

### DIFF
--- a/examples/aws.yml
+++ b/examples/aws.yml
@@ -18,7 +18,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:v0.8
   - name: sshd
-    image: linuxkit/sshd:v0.8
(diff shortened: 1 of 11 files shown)


rn (approved):
pull the new image with DCT on arm64 and x86 and seems to be signed

PR #3559, opened 2020-10-08
add support for containerd args

deitch:
**- What I did**

Add support for passing command-line args to containerd

**- How I did it**

Modifying `pkg/init` to accept args in an optional file `/etc/containerd/cli-opts`

Also add a line to the `docs/faq.md`

**- How to verify it**

CI

containerd accepts arguments

### DIFF
--- a/pkg/init/cmd/service/system_init.go
+++ b/pkg/init/cmd/service/system_init.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -15,6 +16,10 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+)
+
+const (
+	containerdOptsFile = "/etc/containerd/cli-opts"
 )
 
 func cleanupTask(ctx context.Context, ctr containerd.Container) error {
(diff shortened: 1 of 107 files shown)


pkg/init/cmd/service/system_init.go
```diff
@@ -78,8 +82,14 @@ func systemInitCmd(ctx context.Context, args []string) {
 	// remove (unlikely) old containerd socket
 	_ = os.Remove(*sock)
 
+	// look for containerd options
+	ctrdArgs := string{}
+	if b, err := ioutil.ReadFile(containerdOptsFile); err != nil {
+		ctrdArgs = strings.Fields(string(b))
```
justincormack, line 88: in theory you might want to support quoting so `--arg 'test value'` is not split, it might be worth at least adding a note about this, not sure Go has a shell-like parser you can use?

justincormack:
This seems very clean and simple.

pkg/init/cmd/service/system_init.go
deitch: I like the idea, but until I see an actual use case, I don't have the time to hunt something down. I will add a note about it in the faq.

deitch: Updated the faq

deitch:
Done. Pushed out and signed package, updated all of the tests. Ready for a review.

justincormack (approved):

PR #3563, opened 2020-10-19
Fix equals error

deitch:
Fix stupid equals error. Also know is `==` and `!=` are not the same. 🤦🏻‍♂️


justincormack (approved):

justincormack (approved):

deitch:
Interesting. There is a single failure, and it isn't related to this change as far as I can tell.

deitch:
Now it all is clean. Go figure.

PR #3573, opened 2020-11-24
pull and cache images locally without docker

deitch:

**- What I did**

Replaced `docker image pull` and `docker container run/export/etc.` for pulling down and caching images with a local cache.

This essentially removes docker (and any daemon) as a dependency from the `lkt build` process. 

We still need docker to output images, so `dockerRun()` continues to exist. For now.

We also need docker for `lkt pkg build`, but that always will exist.

Also creates support for cross-arch building. So I can run `lkt build --arch=arm64` to build an arm64 image while running on amd64, etc.

And docs (of course).

**- How I did it**

Used the [go-containerregistry](https://godoc.org/github.com/google/go-containerregistry/) library to pull, store and update images, as well as create tar streams. It took a few PRs on the underlying library, but all in by now.

For the `ImageTar()`, to apply the diffs to images, I did a two-pass, thanks to a good suggestion from @justincormack : 

1. Run through each layer, collecting filenames in a map in memory, and removing them per whiteout
1. Run through each layer, only writing a file if it is in the final map

**- How to verify it**

CI should do nicely.

Eliminate docker and daemon as an image build-time dependency for docker images.

**Additional information**

This is not quite ready to get merged in. There are a few things still missing, but I would like to get feedback on all of this first.

- [x] cache management - this will be storing lots of stuff in your cache. We should have a way to see what is in the cache, fetch images to the cache (so we can work offline), remove images from the cache, garbage-collect the cache (remove any unreferenced blobs). I know how to do this, will follow up with an immediate PR once this is good.
- [x] duplicate entries - the cache `index.json` might get duplicate entries if it finds a partial image - e.g. root manifest and 2 out of 3 blobs, so when it downloads the third, it might accidentally create two entries. I am working on that.
- [x] local image build - with the removal of docker, the following can happen: I build an image locally using `lkt pkg build` (i.e. it is in my docker cache); I do `lkt build` and it fails, because the built image is in my docker cache, but lkt is looking in its own cache. Assuming we are agreed, I would have `lkt build` attempt to look in docker first; if it fails (no docker) or cannot find the image, then go to the lkt cache. If user passed `--pull`, then it always pulls to lkt cache from the remote registry


deitch:
Just rebased on latest `master` after #3577 and #3575 are in. Hopefully CI goes green.

deitch:
It builds! Excellent!

I need some design feedback from maintainers on the general design, and the three outstanding issues, before I move ahead on those.

rvs:
From the peanut gallery: this makes perfect sense to me @deitch -- and this is also very much desired on the EVE side.

deitch:
@justincormack had asked what the cache would look like.

It follows the [v1 layout spec](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) precisely. I think that following the OCI spec is important here, as it will enable us to use other libraries and/or tools, as time goes on, and make it easier for others to understand it.

For example, if I pull `docker.io/library/alpine:3.11` and `docker.io/linuxkit/init:305e6125c5a36e175e6e97bb101579df72ef9fe0`, then I would get the following structure:

```
.
├── blobs
│   └── sha256
│       ├── 04fb172885a8d8f6aacfef173b9adf5641bc14745ee240e6abd44ba830838281
│       ├── 0ff8a9dffabb5ed8dcba4ee898f62683305b75b4086f433ee722db99138f4f53
│       ├── 19c4e520fa84832d6deab48cd911067e6d8b0a9fa73fc054c7b9031f1d89e4cf
│       ├── 223fd74c41cf93c8b7a2fc1d93d02a1bc682be7c307382cce89b8b23d1ea5823
│       ├── 2826c1e79865da7e0da0a993a2a38db61c3911e05b5df617439a86d4deac90fb
│       ├── 29e5d40040c18c692ed73df24511071725b74956ca1a61fe6056a651d86a13bd
│       ├── 39eda93d15866957feaee28f8fc5adb545276a64147445c64992ef69804dbf01
│       ├── 3a20da8b220d13725eb708c1d442d0a9535dca1d0b480019691ad1f1b7ec6312
│       ├── 3cfb62949d9d8613854db4d5fe502a9219c2b55a153043500078a64e880ae234
│       ├── 41ba0806c6113064dd4cff12212eea3088f40ae23f182763ccc07f430b3a52f8
│       ├── 45fb88dad5a457a5ba5f3bc5f3931e5a436fa53c98516886add30deaeaaad402
│       ├── 4b858171dd2c97fe3f2909cbc6935856b5a7a0a9b5bb82a41907a815ec497e53
│       ├── 51964c254c5a63abe88654776036ad9bfc2b6a2c20c959c96834a5247227240d
│       ├── 7184c046fdf17da4c16ca482e5ede36e1f2d41ac8cea9c036e488fd149d6e8e7
│       ├── 85cb214b669c0ecf0b49067e3c156f6ec6f22b848eb3c29936d8666f109eb5cc
│       ├── 9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54
│       ├── 9a8fdc5b698322331ee7eba7dd6f66f3a4e956554db22dd1e834d519415b4f8e
│       ├── a88ebc66a3d0e2163937d2514ae788053e4209891e0fa6eb98d601b7e787f0ea
│       ├── ad295e950e71627e9d0d14cdc533f4031d42edae31ab57a841c5b9588eacc280
│       ├── b28e271d721b3f6377cb5bae6cd4506d2736e77ef6f70ed9b0c4716da8bdf17c
│       ├── b9b76886efefcd83f4b686230ac80276b1af2637aeadefb1680e97591fcb52e7
│       ├── b9e3228833e92f0688e0f87234e75965e62e47cfbb9ca8cc5fa19c2e7cd13f80
│       ├── bd9e7c0ac7cbd9ee0e28973f5d4db8df7b9a89568fce9bd48f7e4c693170cdb9
│       ├── c20d2a9ab6869161e3ea6d8cb52d00be9adac2cc733d3fbc3955b9268bfd7fc5
│       ├── c4fe9b047d1506377235b1dbcf01fb4a98cdf780554530decc99ef9893408ca6
│       ├── cbdbe7a5bc2a134ca8ec91be58565ec07d037386d1f1d8385412d224deafca08
│       ├── cfa7ecd5609292dd8f08176e45b32151410683ce34ef12a8600c6c6bdffa85e4
│       ├── e095eb9ac24e21bf2621f4d243274197ef12b91c67cde023092301b2db1e073c
│       ├── ec30e5377f42dcfcf36047553dab57ed38b0b28babdeecbc34a165b7f3778814
│       ├── f346f2698be24b0658565bdd732c703ae9771cac21859d105fc6577a1bbcaee5
│       └── f70734b6a266dcb5f44c383274821207885b549b75c8e119404917a61335981a
├── index.json
└── oci-layout
```

Where the contents of `index.json are:

```
{
   "schemaVersion": 2,
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
         "size": 1638,
         "digest": "sha256:9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54",
         "annotations": {
            "org.opencontainers.image.ref.name": "docker.io/library/alpine:3.11"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
         "size": 403,
         "digest": "sha256:45fb88dad5a457a5ba5f3bc5f3931e5a436fa53c98516886add30deaeaaad402",
         "annotations": {
            "org.opencontainers.image.ref.name": "docker.io/linuxkit/init:305e6125c5a36e175e6e97bb101579df72ef9fe0"
         }
      }
   ]
}
```

The annotation used for the name is the [official annotation for a reference to an image](https://github.com/opencontainers/image-spec/blob/master/annotations.md). 

In the above example, the first entry in the list is for `docker.io/library/alpine:3.11` and has the digest `sha256:9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54`. If we look in `blobs/sha256/9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54`, we will see:

```json
{
  "manifests": [
    {
      "digest": "sha256:39eda93d15866957feaee28f8fc5adb545276a64147445c64992ef69804dbf01",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 528
    },
    {
      "digest": "sha256:0ff8a9dffabb5ed8dcba4ee898f62683305b75b4086f433ee722db99138f4f53",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v6"
      },
      "size": 528
    },
    {
      "digest": "sha256:19c4e520fa84832d6deab48cd911067e6d8b0a9fa73fc054c7b9031f1d89e4cf",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v7"
      },
      "size": 528
    },
    {
      "digest": "sha256:ad295e950e71627e9d0d14cdc533f4031d42edae31ab57a841c5b9588eacc280",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "arm64",
        "os": "linux",
        "variant": "v8"
      },
      "size": 528
    },
    {
      "digest": "sha256:b28e271d721b3f6377cb5bae6cd4506d2736e77ef6f70ed9b0c4716da8bdf17c",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "386",
        "os": "linux"
      },
      "size": 528
    },
    {
      "digest": "sha256:e095eb9ac24e21bf2621f4d243274197ef12b91c67cde023092301b2db1e073c",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      },
      "size": 528
    },
    {
      "digest": "sha256:41ba0806c6113064dd4cff12212eea3088f40ae23f182763ccc07f430b3a52f8",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "s390x",
        "os": "linux"
      },
      "size": 528
    }
  ],
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "schemaVersion": 2
}
```

You can replicate this with [ocidist](https://github.com/deitch/ocidist) with the commands:

```
ocidist pull --format v1-layout --path /tmp/cache docker.io/library/alpine:3.11
ocidist pull --format v1-layout --path /tmp/cache docker.io/linuxkit/init: 305e6125c5a36e175e6e97bb101579df72ef9fe0
```

docs/image-cache.md
```diff
@@ -0,0 +1,61 @@
...
+  ]
+}
+```
+
+## How LinuxKit Uses the Cache and Registry
+
+For each image that linuxkit needs to read, it does the following. Note that if the `--pull` option
+is provided, it always will pull, independent of what is in the cache.
+
+1. Check in the cache for the image name in the cache `index.json`. If it does not find it, pull it down and store it in cache, using content trust, if enabled.
```
justincormack, line 55: Should there be a note about whether it checks with a HEAD request if the image has changed?

deitch: It does not. Behaviour is:

1. If `--pull` is not specified, check in local cache for an entry with the matching name.
    * If it finds it, use it. Pull nothing unless some blob is missing, in which case it tries to fill them in from remote registry.
    * If it does not find it, pul everything from remote
1. If `--pull` is specified, pull everything from remote

But the definition of "pull everything" is smart. If it encounters a blob, and that blob is available hash-verified locally, then it will not pull that blob from remote, just use it locally.

justincormack:
What is the locking situation for updating the index file? At a glance google/g-containerregistry uses Go mutexes, which is fine if all the code is running in the same binary, but not if you are running multiple independent copies of linuxkit at once? Should we also take a file lock before updating the index?

deitch:
> What is the locking situation for updating the index file? At a glance google/go-containerregistry uses Go mutexes, which is fine if all the code is running in the same binary, but not if you are running multiple independent copies of linuxkit at once? Should we also take a file lock before updating the index?

Funny you ask that. I was just discussing it offline with some people. 

The answer is probably, yes, but beyond the scope for now. The closest analogy is containerd, which runs a daemon with exclusive control over its boltdb database file. 

My approach to this would be to keep it as is for now, accept the risk, and then have the underlying library add the capability, perhaps optional.

justincormack:
That means you can't share a cache in CI for example, and building multiple linuxkit images using say parallel make might randomly break. I don't think wrapping a file lock around the one function would be too hard, would be nicer if upstream does it but I can see why they might not (maybe open an issue to ask?)

deitch:
> That means you can't share a cache in CI for example, and building multiple linuxkit images using say parallel make might randomly break. 

You can have each have its own cache. It is a bit inefficient, but not terrible. 

But if you really want a shared cache, and running in parallel, then, yes, you need to have some form of locking.

Also, keep in mind that it only writes to the `index.json` if it does not have the image; If it has it, then it does not need to do anything.

> would be nicer if upstream does it but I can see why they might not (maybe open an issue to ask?)

Already in discussion, but it will take a bit of time.

deitch:
This now has an option to do `lkt build --docker` so it will pull images from the local docker cache before looking in the lkt cache.

deitch:
cache management - at a basic level - also now covered. Can clean up the cache, or list the names of images and their root manifest hashes.

deitch:
@justincormack I managed to get the entire interface to the cache into one package, `github.com/linuxkit/linuxkit/src/cmd/linuxkit/cache`, so all of the usage of `github.com/google/go-containerregistry` is encapsulated there. This should make it easier to continue to use, or to swap out if we ever want to.

deitch:
This is ready. Finally. Got the last piece in - replacing rather than duplicating entries in the `index.json` of the v1 layout spec - a few hours ago.

It seems to stumble on the `amd64-windows.exe` tests, though. Anyone know what/why?

deitch:
Oh! The windows stuff passed. 🎉

deitch:
I have no idea what is wrong with CI here.

justincormack:
Getting errors building linuxkit.yml as busybox:latest is arch linux/arm64/v8 rather than linux/arm64... these seem to be inconsistent

justincormack:
golint is complaining

```
src/cmd/linuxkit/cache/image.go:8:1: exported function ListImages should have comment or be unexported
src/cmd/linuxkit/cache/pull.go:11:1: exported function ValidateImage should have comment or be unexported
src/cmd/linuxkit/cache/resolvabledescriptor.go:44:1: exported function FindRoot should have comment or be unexported
src/cmd/linuxkit/cache/source.go:14:6: exported type ImageSource should have comment or be unexported
src/cmd/linuxkit/cache/source.go:20:1: exported function NewSource should have comment or be unexported
src/cmd/linuxkit/cache/source.go:29:1: exported method ImageSource.Config should have comment or be unexported
src/cmd/linuxkit/cache/source.go:51:1: exported method ImageSource.TarReader should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:16:1: exported function Client should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:26:1: exported function HasImage should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:37:1: exported function InspectImage should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:50:1: exported function Create should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:70:1: exported function Export should have comment or be unexported
src/cmd/linuxkit/docker/cmd.go:84:1: exported function Rm should have comment or be unexported
src/cmd/linuxkit/docker/source.go:24:6: exported type ImageSource should have comment or be unexported
src/cmd/linuxkit/docker/source.go:28:1: exported function NewSource should have comment or be unexported
src/cmd/linuxkit/docker/source.go:34:1: exported method ImageSource.Config should have comment or be unexported
src/cmd/linuxkit/docker/source.go:54:1: exported method ImageSource.TarReader should have comment or be unexported
src/cmd/linuxkit/moby/image.go:27:6: exported type ImageSource should have comment or be unexported
src/cmd/linuxkit/util/util_unix.go:9:1: exported function HomeDir should have comment or be unexported
src/cmd/linuxkit/util/util_windows.go:7:1: exported function HomeDir should have comment or be unexported
```

justincormack:
ok arm64 issues fixed and linting errors fixed

src/cmd/linuxkit/docker/cmd.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	log "github.com/sirupsen/logrus"
+)
+
+// Client get a docker client.
+func Client() (*client.Client, error) {
+	// for maximum compatibility as we use nothing new
+	// 1.30 corresponds to Docker 17.06, supported until 2020.
+	err := os.Setenv("DOCKER_API_VERSION", "1.30")
```
justincormack, line 20: Maybe we should increment this a bit? 2020 is kind of ending...

src/cmd/linuxkit/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+func cacheUsage() {
+	invoked := filepath.Base(os.Args[0])
+	fmt.Printf("USAGE: %s cache command [options]\n\n", invoked)
+	fmt.Printf("Supported commands are\n")
+	// Please keep these in alphabetical order
+	fmt.Printf("  clean\n")
+	fmt.Printf("  import\n")
```
justincormack, line 18: This doesn't exist?

justincormack:
cache import doesnt seem to exist and cache ls is pretty unreadable... for ls could we iterate over the map and print it maybe so at least there are line breaks?

src/cmd/linuxkit/cache.go
deitch: Ah, I never implemented it. Let's leave it out, I will do a separate much smaller PR for it.

deitch:
> cache import doesnt seem to exist

I got rid of it. I will implement on a separate lightweight PR. It doesn't have to be in this one.

> cache ls is pretty unreadable

oh wow that is horrid. OK, fixing.

src/cmd/linuxkit/docker/cmd.go
deitch: What version should be here? I am happy to change it to whatever you think is right.

deitch:
Fixed `cache ls`

src/cmd/linuxkit/docker/cmd.go
deitch: Let's leave this for now. If we want to change it, it should get its own PR so that anything that breaks is isolated.

justincormack (approved):

deitch:
I will merge as soon as CI is green.

PR #3574, opened 2020-11-24
fix deprecation of set-env

deitch:

**- What I did**

fix the GitHub Actions [set-env deprecation](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) that was causing CI to fail

**- How I did it**

Replaced `set-env` in our `ci.yaml` with a new step, per the GitHub recommendation

**- How to verify it**

CI runs

update CI per GitHub set-env deprecation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,13 @@ jobs:
       with:
         path: ./src/github.com/linuxkit/linuxkit
 
+    - name: Set path
+      run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+      env:
+         GOPATH: ${{runner.workspace}}
+
     - name: Get pre-requisites
       run: |
-            echo "::set-env name=PATH::$PATH:$(go env GOPATH)/bin"
             go get -u golang.org/x/lint/golint
             go get -u github.com/gordonklaus/ineffassign
       env:


deitch:
I think this does it, anyways. 😄

.github/workflows/ci.yml
```diff
@@ -27,16 +27,23 @@ jobs:
       with:
         path: ./src/github.com/linuxkit/linuxkit
 
+    - name: Set path
+      run: |
+            echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
```
thaJeztah, line 32: Looks like previously it was _appending_ `GOPATH` to `PATH`, correct? (`echo "::set-env name=PATH::$PATH:$(go env GOPATH)/bin"`)

thaJeztah: In case it's useful; here's how these are set in containerd/cri (but perhaps that CI differs); https://github.com/containerd/cri/blob/release/1.4/.github/workflows/ci.yml#L22-L23

deitch: Yeah, it was appending `${GOPATH}/bin`, but it looks like we overrode `GOPATH` in the actions, which I missed in converting it. I think I got it now. I hope.

deitch: Thanks for hopping in @thaJeztah 😄

.github/workflows/ci.yml
```diff
@@ -27,9 +27,13 @@ jobs:
       with:
         path: ./src/github.com/linuxkit/linuxkit
 
+    - name: Set path
+      run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
```
thaJeztah, line 31: I guess (looking at `GOPATH: ${{runner.workspace}}` below), this would now be the equivalent of

```suggestion
      run: echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
```

correct? Not sure if there's advantage of one over the other though.

deitch: Technically, no advantage. Given that we already set `GOPATH: {{runner.workspace}}` in the yaml, I would rather consume what it is set to via `go env GOPATH` than use it explicitly again. Sort of like setting a var and using the var (sort of).

thaJeztah: gotcha

thaJeztah (approved):
LGTM

justincormack (approved):

PR #3579, opened 2020-12-04
localize lk-wg-config.sh

allfro:
**- What I did**
This patch adds:
* a local copy of the `lk-wg-config.sh` which was normally pulled from a GitHub gist. 
* support for `PostUp` and `PreUp` configuration directives in WireGuard.

**- How I did it**
* Copied the version of the `lk-wg-config.sh` script from the GitHub gist specified in the Dockerfile. 
* Downloaded a copy of the original wg-quick script and extracted the relevant sections to support Pre- and Post-Up hooks.
* Modified the Dockerfile to include the local updated copy of `lk-wg-config.sh`

**- How to verify it**
* I tested the modifications with a lab setup at home using iptables masquerading to enabled once wireguard successfully connected. Everything worked as expected.

* Added support for PreUp and PostUp in wireguard `lk-wg-config` script.

### DIFF
--- a/pkg/ip/lk-wg-config.sh
+++ b/pkg/ip/lk-wg-config.sh
@@ -1,0 +1,209 @@
+#!/bin/bash
+#
+# Copyright (C) 2016-2017 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
+#
+# This file is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2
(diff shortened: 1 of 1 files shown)


rn:
Thanks for the patch, but , the reason this file was not checked into the repository is that it is only available under GPL and we did not want a GPL licensed file in the LinuxKit repository

rn:
Closing this due to the GPL issue mentioned above

Issue #2447, opened 2017-08-22
Improve Azure support


radu-matei:
There are things that need to get improved in the Azure support:

- [ ] migrate the creation of resources to the [ARM deployment model in the Azure Go SDK](https://github.com/Azure/azure-sdk-for-go/blob/master/arm/resources/resources/deployments.go)

- [ ] remove non-official vendored dependencies

- [ ] variables for hard coded values (networking, (unused) initial admin username and password for VM)

Please feel free to also include any other relevant thing I might have missed, feature requests and others.

radu-matei:
There is a [new release for Azure Go SDK](https://github.com/Azure/azure-sdk-for-go/releases/tag/v12.1.0-beta) - will first target this version for the creation of resources (storage account, network interfaces, VM).

PR #3585, opened 2020-12-28
Add Azure providers

schnerring:
This affects:

* issue #2447

**- What I did**

1. Add Azure WireServer client
2. Add Azure IMDS provider
3. Add Azure OVF provider

**- How I did it**

1. I implement a simple WireServer client `provider_azure_wire.go` that reports ready to Azure. Both IMDS and OVF provider depend on it, otherwise provisioning won't succeed. See also in the Microsoft docs: [Creating generalized images without a provisioning agent](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/no-agent#generic-steps-without-using-python)
2. Implement Azure IMDS metadata provider `provider_azure_imds.go` that utilizes the [Instance Metadata Service API](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service). User data is currently disabled, see:
    * https://github.com/MicrosoftDocs/azure-docs/issues/64154
    * https://github.com/MicrosoftDocs/azure-docs/issues/30370 (OP)
3. Implement Azure OVF metadata `provider_azure_ovf.go` which extracts user data from the `ovf-env.xml` file on the Azure attached provisioning DVD. The IMDS provider also uses this provider as a fallback to get user data. The implementation is inspired by [cloud-init's Azure docs](https://cloudinit.readthedocs.io/en/latest/topics/datasources/azure.html) and https://github.com/Azure/WALinuxAgent

**- How to verify it**

I'm successfully creating managed [k3OS](https://github.com/rancher/k3os) images for Azure. k3OS uses linuxkit's metadata package.

I forked k3OS, [created a Packer template](https://github.com/schnerring/k3os/tree/packer-azure/package/packer/azure) and and [swapped linuxkit/linuxkit for schnerring/linuxkit](https://github.com/schnerring/k3os/commit/b24105d98e094a5f467c3df81c239abb976ee761).

You can find the Packer template using the forked versions on the  [packer-azure-test](https://github.com/schnerring/k3os/tree/packer-azure-test/package/packer/azure) branch.

Add IMDS and OVF Azure providers to metadata package

**- Questions / Thoughts**

I don't know how the metadata package integrates with the rest of linuxkit but it feels very much like reinventing the wheel. Why not bundle cloud-init or WALinuxAgent agents with linuxkit? Reimplementations will always be inferior compared to these. Is it because those agents are too bloated?

The metadata package's distinction between `cdromProvider`s and `netProvider`s doesn't make sense for Azure providers. The IMDS provider (`netProvider`) uses the OVF provider (technically `cdromProvider`) as a fallback for getting user data. [cloud-init also uses this fallback mechanism for e.g. SSH keys](https://cloudinit.readthedocs.io/en/latest/topics/datasources/azure.html#imds):

> Cloud-init will first try to utilize SSH keys returned from IMDS, and if they are not provided from IMDS then it will fallback to using the OVF file provided from the CD-ROM

Thus far, the OVF provider only extracts user data but not hostname or SSH keys, so it's not usable as standalone provider, only as a fallback for the IMDS provider.

I've implemented the fallback mechanism in the IMDS provider itself but maybe this "hybrid" approach should be handled by `main.go`?

Any feedback is appreciated.


justincormack:
Hey thanks for this. Yes cloud-init is very large and assumes it can run shell scripts to configure things on the host, which doesn't generally work, and conflicts with immutability. WALinuxAgent is gigantic and tries to guess how to poke around to configure things based on distro, so does not work. The aim of the metadata package is to gather minimal config info that is useful, and to be workable across multiple providers for a base setup. You can use other options if you like if they work, LinuxKit always gives you a choice.

pkg/metadata/provider_azure_ovf.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+func (p *ProviderAzureOVF) String() string {
+	return "Azure-OVF"
+}
+
+// Probe returns true if DVD is successfully mounted, false otherwise
+func (p *ProviderAzureOVF) Probe() bool {
+	if err := retry(6, 5*time.Second, p.mount); err != nil {
```
justincormack, line 53: Is this needed? Does the virtual CD take same time to appear??

schnerring: Yes, I used  the WALinuxAgent's osutil `default.py` implementation as reference, see:

https://github.com/Azure/WALinuxAgent/blob/v2.2.52/azurelinuxagent/common/osutil/default.py#L579-L618

On my test system `ls /tmp | grep cdrom` returns:

``` text
cdrom003897275
cdrom310295156
cdrom860156597
```

This would mean that two attempts failed. I thought about how nice it'd be to have regression tests for metadata packages. But how? Some thoughts (sorry, if I'm going offtopic):

* The tests would need to be labeled and not run by default. Not every host can run tests for all platforms
* For Azure you'd have to provision a VM / many VMs with the metadata package and verify its/their state after. Is that approach even feasible? Any inputs?

The impact of the `retry()` method is substantial. I also use it inside `ProviderAzureIMDS`. So an image that uses neither `AzureProviderIMDS` nor `AzureProviderOVF` would probe 30 seconds for each provider, amounting to 60 seconds of probing. So worst case, when `metadata` is run without args, this would amount to 60 seconds of just probing because of Azure providers.

schnerring:
> Hey thanks for this. 

Thanks for the quick response. Fun opportunity to learn some Go :)

> The aim of the metadata package is to gather minimal config info that is useful, and to be workable across multiple providers for a base setup. You can use other options if you like if they work, LinuxKit always gives you a choice.

So this would be `hostname`, SSH key(s) and user data?

stevefan1999-personal:
Hi I would like to merge your PR with mine, is that ok to you ;p

schnerring:
> Hi I would like to merge your PR with mine, is that ok to you ;p

What do you mean by "merge"? If this PR gets merged, you can rebase your PR off of new `master`, no reason to include the same set of changes in two PRs.

stevefan1999-personal:
@schnerring I remember I also had an implementation for Azure, though [I have reported](https://github.com/MicrosoftDocs/azure-docs/issues/64154) the metadata API isn't working as intended, and even after months Microsoft aren't doing anything about it. I just wait and left it in the cold until you also popped up and shown an alternative implementation.

But fine, I will just drop mine so you can go first.

schnerring:
Is there anything I can do to get this merged? #3593 was merged 14 days ago, I don't know what's keeping this PR from getting merged besides rebasing. Any input is appreciated.

pkg/metadata/provider_azure_imds.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Probe checks if Azure IMDS API is available
+func (p *ProviderAzureIMDS) Probe() bool {
+	// "Poll" VM Unique ID
+	// See: https://azure.microsoft.com/en-us/blog/accessing-and-using-azure-vm-unique-id/
+	pollVMID := func() error {
+		_, err := p.imdsGet("compute/vmId")
+		return err
+	}
+	if err := retry(6, 5*time.Second, pollVMID); err != nil {
```
rn, line 44: As far as I can see no other `Probe` function does a retry. Is this really necessary? If all providers would probe like this it potentially will take a long time to cycle through the list of all providers until we hit a match.

schnerring: You're right, for the IMDS provider it's probably not needed, I'll have time to test this tomorrow. For the OVF provider, as mentioned earlier, it is needed and I pretty much mimmick what WALinuxAgent does. I'll remove the retry logic for the IMDS provider and move the retry function to `provider_azure_ovf.go`.

pkg/metadata/main.go
```diff
@@ -101,6 +116,14 @@ func main() {
...
 		case p == "digitalocean":
 			netProviders = append(netProviders, NewDigitalOcean())
+		case p == "azure-imds":
+			netProviders = append(netProviders, NewAzureIMDS())
+		case p == "azure-ovf":
+			// TODO not every provider should create a separate http client
+			client := &http.Client{
+				Timeout: time.Second * 2,
+			}
+			netProviders = append(netProviders, NewAzureOVF(client))
```
schnerring, line 126: At the moment, each provider creates a separate http client, but [the docs state:](https://golang.org/pkg/net/http/)

> Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

Should I open a separate PR for this?

schnerring: I think we should entirely remove `azure-ovf` here, because `azure-imds` already uses it as a fallback.

pkg/metadata/provider_azure_imds.go
rn: Ah, I missed your earlier comment. Can we remove the OVF provider from the default list? I suspect a common usage for the `metadata` package would be just add it without arguments. Having it taking 30s longer depending on which provider is used and where in the list the provider happens to be does not look like a good user experience. `Probe()` is really meant to be a quick check. Is there a different way to figure out which provider to use?

schnerring:
So, as previously discussed, I have implemented the following improvements:

* move `retry()` function to OVF provider
* remove `retry()` from IMDS provider
* remove OVF provider from default provider list in `metadata/main.tf`. The default provider list now only contains a single `azure` provider

Unfortunately we cannot  get rid of the `retry()` logic. As long as https://github.com/MicrosoftDocs/azure-docs/issues/64154 is not resolved, there's no chance of getting user-data from IMDS directly. This means we have to rely on the `ovf-env.xml` file to retrieve user-data which is provided by a CD-ROM attached by the Azure platform.

The couple of times I've tested, mounting the device always succeeded with the 3rd try, so it takes around 10 seconds for probing to succeed. I guess this could take longer when the Azure platform has latency issues. I just stuck to what WALinuxAgent does.

So the best we can do here @rn is to move the Azure provider to the end of the provider list, so it gets probed last in case no arguments were provided.

pkg/metadata/provider_azure_imds.go
schnerring: @rn see my latest comment, please.

* I removed the OVF provider from the default list
* I moved the Azure provider to the end of the provider list which is the best we can do IMO.

This way using `metadata` without args, the 30 second delay never occurs _if running on a supported platform_.

However, when using the `metadata` package without args, in an environment where none of the providers are supported, a 30 second delay will occur. I don't know if this is an actual use case, though. If it is, we'd have to _always_ exclude the `azure` provider by default and in case one wanted to use it, `explicitly include it` which is not a nice user experience, too, IMO.

schnerring:
@justincormack @rn Please don't code review, yet. There has been an update regarding the Microsoft IMDS API. See my comment in the GitHub issue: https://github.com/MicrosoftDocs/azure-docs/issues/64154#issuecomment-826399932

The `customData` feature has been deprecated in favor of the `userData` which was introduced with version `2021-01-01` of the IMDS API. This means that the OVF provider is obsolete which solves the `retry()` problem.

I'm excited to test this and I'll be refactoring the code within the next two weeks.

schnerring:
Finally got around doing this and I'm ready for a review @rn @justincormack 

I'm excited to tell you guys that all the previous issues regarding `retry()` logic are now resolved because the OVF provider is not required anymore, now that `userData` is available through the IMDS.

The provider is now as lightweight as any other provider, like GCP or AWS. One issue I'm having is that `customData` is not obsolete as I previously suspected:

* `customData` is only available during initial provisioning and would require the OVF provider if we wanted to support it because it will never be available through the IMDS API.
* `userData` is available and mutable during the lifecycle of a VM and available through the IMDS API.

We could keep the OVF provider but just not include it in the default provider list to also support `customData`, what do you guys think?

Itxaka:
@rn @justincormack Is there something we can do to move this forward? Is something missing from the PR? We are also using linuxkit and we are missing on the azure functionality.

PR #3589, opened 2021-01-06
provide wildcards to ineffassign

deitch:

**- What I did**

Replaced the arg to `ineffassign` from a list of individual `.go` files to `./src/cmd/linuxkit/...`

The latest ineffassign doesn't accept individual files that are in different directories, so have to switch to `/...`
See [here](https://github.com/gordonklaus/ineffassign/commit/8eed68eb605f2ddf507412accb8dd7a595b5c114)

**- How I did it**

Replaced a single line in Makefile

**- How to verify it**

1. Change a few files to introduce the error
2. Run `ineffassign`
3. See that it catches them

Support latest ineffassign args

cc @giggsoff


deitch:
Without this, all PRs will break, since CI installs ineffassign from scratch.

justincormack:
why are you doing src/cmd/... not just ...?

justincormack (approved):

deitch:
> why are you doing src/cmd/... not just ...?

I don't want it picking up all of the packages. The prior version found everything `*.go` under `src/cmd/linuxkit/` and passed it on with explicit filenames.

PR #3594, opened 2021-01-18
switch to go modules

deitch:

**- What I did**

Replaced `vndr` with go modules.

**- How I did it**

1. Run `go mod init`
2. Clean up lots of painful dependencies
3. Eventually conclude we need to update to latest azure-sdk-for-go library
4. Fix up azure.go to fit it

**- How to verify it**

CI. Plus I built the binaries manually too

CLI uses go modules


rn:
This looks painful...

could you also update the docs? In particular https://github.com/linuxkit/linuxkit/blob/master/docs/vendoring.md

Might be good to have a short description (or link to decent documentation) on how to use this...

justincormack (approved):

deitch:
> This looks painful...

It was. Quite. But it is done.

> could you also update the docs? In particular https://github.com/linuxkit/linuxkit/blob/master/docs/vendoring.md
> Might be good to have a short description (or link to decent documentation) on how to use this...

Yes, a good idea. Since @justincormack approved, the changes are only docs ("only" in the sense of "easy to change and test and guaranteed to pass CI"), I am going to merge this in as-is, and immediately open a doc PR. Just easier.

the-maldridge:
This makes me very happy.  I'll probably go in and update the terraform provider in a bit so that I can rip out some of my hacks needed to compile the old way.

deitch:
It still gets weird if you do `go get` for reasons I do not understand yet. You can do `git clone` and `cd src/cmd/linuxkit` and `go install .`, but a simple `go get` blows up. But at least we are in go modules, so we are in the right playground (or plague, as autocorrect tried to correct me...)

PR #3595, opened 2021-01-19
update docs

deitch:

**- What I did**

Updated the vendoring doc to describe the newer go modules workflow 

**- How I did it**

Changed a markdown file

**- How to verify it**

Read the file. If it makes sense, it passes.

Vendoring doc updated for go modules

### DIFF
--- a/docs/vendoring.md
+++ b/docs/vendoring.md
@@ -2,16 +2,24 @@ Vendoring
 =========
 
 The Go code in this repo depends on a number of Go libraries.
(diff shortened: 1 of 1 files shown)


deitch:
As requested [here](https://github.com/linuxkit/linuxkit/pull/3594#issuecomment-762471308)

rn (approved):
Thanks. Look simpler (hm) than the vndr dance but I could be so wrong

deitch:
In theory, it is simpler. If there are odd clashes of specific dependency versions, it can be a bit hard to resolve.

PR #3606, opened 2021-03-08
Bump Intel microcode version to 20210216

Ricordel:
Signed-off-by: Yoann Ricordel <yoann.ricordel@qarnot-computing.com>

**- What I did**

Bump the version of the pulled Intel microcode from `20191115` to `20210216`

**- How I did it**

Changed the checked out tag in `kernel/Dockerfile`

**- How to verify it**

I guess by trusting that the change is trivial (look at the diff) and just check that a kernel it still builds.

Bump Intel microcode from `20191115` to `20210216`


rn (approved):
Thanks for the update

PR #3609, opened 2021-03-19
Remove Notary and Content Trust

dave-tucker:
**- What I did**
This commit removes Notary and Content Trust.
Notary v1 is due to be replaced with Notary v2 soon.
There is no clean migration path from one to the other.
For now, this removes all signing from LinuxKit.
We will look to add this back once a new Notary alternative
becomes available.

**- How I did it**
- Removed any command line flags relating to signing and content trust
- Removed `trust.go`
- Removed all references to signing and content trust from makefiles, scripts and documentation
- Ensured that everything still compiles

**- How to verify it**
If CI is still green, then all should be well.

- Removed support for Docker Content Trust and Notary Signing

image


deitch:
Hats off @dave-tucker  I will look early next week.

rn:
Thanks. I'll take a look later as well. May I suggest that to merge this PR it requires approval from all active maintainers which currently seem to be @deitch @justincormack and myself to make sure we all (publicly) agree with this (@dave-tucker I assume your approval is implicit in making this PR in the first place)

dave-tucker:
Yes @rn, my +1 is implicit

kernel/Makefile
```diff
@@ -144,7 +138,7 @@ forcepush: forcepush_$(2)$(3)$(4)
 show-tags: show-tag_$(2)$(3)$(4)
 
 # 'docker build' with the FROM image supplied as --build-arg
-# *and* with DOCKER_CONTENT_TRUST=1 currently does not work
+# *and* with  currently does not work
```
deitch, line 141: This line is odd.

src/cmd/linuxkit/moby/images.go
```diff
@@ -39,16 +37,5 @@ func imagePull(ref *reference.Spec, alwaysPull bool, trust bool, cacheDir string
 // imageLayoutWrite takes an image name and pulls it down, writing it locally
 func imageLayoutWrite(cacheDir string, ref *reference.Spec, architecture string, trust bool) (ImageSource, error) {
```
deitch, line 38: We removed how this uses `trust`, but kept it as an arg. I think we should just get rid of it here, too.

src/cmd/linuxkit/build.go
```diff
@@ -223,7 +217,7 @@ func build(args []string) {
 		}
 
 		log.Infof("Create outputs:")
-		err = moby.Formats(filepath.Join(*buildDir, name), image, buildFormats, size, !*buildDisableTrust, cacheDir)
+		err = moby.Formats(filepath.Join(*buildDir, name), image, buildFormats, size, false, cacheDir)
```
deitch, line 220: Why do we pass `false` to trust, as opposed to just getting rid of it as an arg entirely?

src/cmd/linuxkit/cache/push.go
```diff
@@ -12,8 +12,6 @@ import (
 // PushWithManifest push an image along with, optionally, a multi-arch index.
 func PushWithManifest(dir string, name, suffix string, pushImage, pushManifest, trust, sign bool) error {
```
deitch, line 13: We probably should remove `trust` as an arg here.

src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	dctEnableEnv        = "DOCKER_CONTENT_TRUST=1"
+	dctEnableEnv        = ""
```
deitch, line 23: Why not just get rid of it?

deitch (changes requested):
Two requests:

1. One of the comment changes is confusing
2. This PR removes the `trust` env var or arg from lots of places, but leaves it in a few places, always setting it to `false` or `""` and not passing it anywhere. We should just complete the process and get rid of those.

dave-tucker:
@deitch good catched. most of it was some clumsy `sed` on my behalf :blush: I've gone through, fixed up the comments in question and removed the `trust bool` from `pkg/moby`.

deitch (dismissed):
LGTM, and thank you for taking care of this.

We should wait for a LGTM from all of the core maintainers.

docs/packages.md
```diff
@@ -56,7 +54,7 @@ A package source consists of a directory containing at least two files:
 Before you can build packages you need:
 - Docker version 19.03 or newer, which includes [buildx](https://docs.docker.com/buildx/working-with-buildx/)
 - If you are on a Mac you also need `docker-credential-osxkeychain.bin`, which comes with Docker for Mac.
-- `make`, `notary`, `base64`, `jq`, and `expect`
+- `make`, `base64`, `jq`, and `expect`
 - A *recent* version of `manifest-tool` which you can build with `make
```
justincormack, line 58: Do we still need this?

scripts/push-manifest.sh
```diff
@@ -62,24 +56,4 @@ OUT=$(manifest-tool $MT_ARGS push from-args \
                     --template "$TARGET"-ARCH \
                     --target "$TARGET")
 
-echo "$OUT"
-if [ -z "$TRUST" ]; then
-    echo "Not signing $TARGET"
-    exit 0
-fi
-
-# Extract sha256 and length from the manifest-tool output
```
justincormack, line 71: I think you can remove all the manifest-tool parts above as it is only used here

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -11,7 +11,7 @@ import (
 	"runtime"
 
 	"github.com/containerd/containerd/reference"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
justincormack, line 14: thats a horrible name to give it!

tools/alpine/push-manifest.sh
```diff
@@ -4,14 +4,11 @@ set -e
 # This script creates a multiarch manifest for the 'linuxkit/alpine'
 # image, pushes and signs it. The manifest is pushed with the tag of
```
justincormack, line 5: remove "signs it"

docs/packages.md
deitch: No, we do not. manifest-tool now is embedded as a library inside the linuxkit binary. We no longer exec out to it.

For that matter, what to we use expect, base64 and jq for?

tools/alpine/push-manifest.sh
```diff
@@ -4,14 +4,11 @@ set -e
...
-# credentials helper to extract the Hub credentials. We need to
-# manually sign the manifest using 'notary'.
+# credentials helper to extract the Hub credentials.
 #
 # This script is specific to 'linuxkit/alpine'. For normal packages we
 # use a different scheme.
 #
-# For signing, DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE must be set.
-#
 # This should all be replaced with 'docker manifest' once it lands.
```
justincormack, line 12: docker manifest did land a long time ago, can we use instead of manifest tool?

deitch: We should consider it as a replacement for the embedded library stuff (although I usually prefer embedded), but that definitely should be subject of a separate PR

docs/packages.md
rn: kernel build still uses this

docs/packages.md
```diff
@@ -179,56 +173,12 @@ To perform _all_ of the steps at once - build and push out the image for whateve
...
-
-```console
-IMAGE=docker.io/linuxkit/containerd
-TAG=a4aa19c608556f7d786852557c36136255220c1f
-SIZE=1052
-HASH=66b3d74aeb855f393ddb85e7371a00d5f7994cc26b425825df2ce910583d74dc
-notary -s https://notary.docker.io -d ~/.docker/trust addhash -p $IMAGE $TAG $SIZE --sha256 $HASH  -r targets/releases
-```
 
 ##### Getting Size and Hash
```
rn, line 183: IIRC the size was only needed for signing. so can probably remove this section as well

kernel/Makefile
```diff
@@ -143,12 +137,6 @@ push: push_$(2)$(3)$(4)
 forcepush: forcepush_$(2)$(3)$(4)
 show-tags: show-tag_$(2)$(3)$(4)
 
-# 'docker build' with the FROM image supplied as --build-arg
```
rn, line 146: I wouldn't delete this comment yet, since it sounds like by removing content trust we can simplify the bits below

scripts/push-manifest.sh
rn: we can probably remove this script. it;'s only used for kernel build and the alpine base. if we can replace the manifest tool with docker than there is no need for this script

kernel/Makefile
dave-tucker: Addressed. Added a FIXME to the top to show that the comment isn't current anymore.

docs/packages.md
dave-tucker: Done

tools/alpine/push-manifest.sh
dave-tucker: Done

scripts/push-manifest.sh
dave-tucker: I'll keep it here. We can remove it in a follow-up PR that removes manifest-tool and related bits

deitch (approved):

rn (approved):
Thanks for working on this.

justincormack (approved):

PR #3610, opened 2021-03-19
make: Fix for local builds outside of GOPATH

dave-tucker:
Go can be weird about tools having to run in a directory with
go.mod. This commit moves the linuxkit makefile to the same
directory as the code. To fix it.

Note: I've put off fixing the in-container builds until we can bump the alpine/go versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
 
     - name: Build
       run: |
-        make GOARCH=${{matrix.target.arch}} GOOS=${{matrix.target.os}} LOCAL_TARGET=bin/linuxkit-${{matrix.target.suffix}} local-build
+        make GOARCH=${{matrix.target.arch}} GOOS=${{matrix.target.os}} LOCAL_TARGET=$(pwd)/bin/linuxkit-${{matrix.target.suffix}} local-build
         file bin/linuxkit-${{matrix.target.suffix}}
       env:
         GOPATH: ${{runner.workspace}}
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,11 @@
 VERSION="v0.8+"
-GIT_COMMIT=$(shell git rev-list -1 HEAD)
 
 GO_COMPILE=linuxkit/go-compile:7b1f5a37d2a93cd4a9aa2a87db264d8145944006
 
 ifeq ($(OS),Windows_NT)
-LINUXKIT?=bin/linuxkit.exe
 RTF?=bin/rtf.exe
 GOOS?=windows
 else
-LINUXKIT?=bin/linuxkit
 RTF?=bin/rtf
 GOOS?=$(shell uname -s | tr '[:upper:]' '[:lower:]')
 endif
@@ -22,10 +19,14 @@ endif
 
 PREFIX?=/usr/local
 
+LOCAL_TARGET?=$(CURDIR)/bin/linuxkit
+
+export VERSION GO_COMPILE GOOS GOARCH LOCAL_TARGET
+
 .DELETE_ON_ERROR:
 
 .PHONY: default all
-default: $(LINUXKIT) $(RTF)
+default: linuxkit $(RTF)
 all: default
 
 RTF_COMMIT=2351267f358ce6621c0c0d9a069f361268dba5fc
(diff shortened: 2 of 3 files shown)


deitch (commented):
I didn't see the current issue, but I have no objection to moving parts of the Makefile into the src directory for the lkt binary, with simple wrappers at the root.

You do have some loss of flexibility, though, as a bunch of useful targets move from the root of the directory to `src/cmd/linuxkit/`.

dave-tucker:
@deitch I think the reproduction case for this issue is to check out `linuxkit` outside your gopath and to attempt `make`.
The other alternative I considered was to prefix the `go build`, `go vet` `ineffassign` commands with `cd ./src/cmd/linuxkit && ...` which led to some inconsistent behaviour with the package specifier having to be one of `./...` or the full `github.com/linuxkit/linuxkit/src/cmd/linuxkit`... and in some cases, not providing it at all made it work :confused: 
Putting everything in `src/cmd/linuxkit` was the tidier option although I agree it's a little less flexible.

Alas CI is claiming that the hash is different from the build stage and test stages so I need to debug that before this can proceed

deitch:
I think it is good enough, though. As long as we make it clear, nothing terrible about `cd src/cmd/linuxkit` to do things

rn:
https://github.com/linuxkit/linuxkit/pull/3616 should fix the incontainer builds, at least for me on ar,64, s390x, amd64 (darwin and linux)

rn:
But local build is still broken (i don't have the code in gopath), so maybe this fixes it...need to check it out

dave-tucker:
@rn CI is now green if you'd like to try this out :tada:

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
 
     - name: Build
       run: |
-        make GOARCH=${{matrix.target.arch}} GOOS=${{matrix.target.os}} LOCAL_TARGET=bin/linuxkit-${{matrix.target.suffix}} local-build
+        make GOARCH=${{matrix.target.arch}} GOOS=${{matrix.target.os}} LOCAL_TARGET=`pwd`/bin/linuxkit-${{matrix.target.suffix}} local-build
```
rn, line 60: nit: This should be `$(pwd)` instead of ``pwd``

Makefile
```diff
@@ -1,14 +1,9 @@
-VERSION="v0.8+"
```
rn, line 1: any chance we can keep these in the toplevel `Makefile`? I would strongly prefer having them at the toplevel...if not `docs/releasing.md` needs updating.

Makefile
```diff
@@ -50,56 +49,25 @@ bin/manifest-tool: tmp_mt_bin.tar | bin
...
 
-LINUXKIT_DEPS=$(wildcard src/cmd/linuxkit/*.go) $(wildcard src/cmd/linuxkit/*/*.go) Makefile
-$(LINUXKIT): tmp_linuxkit_bin.tar
-	tar xf $<
-	rm $<
-	touch $@
-
-tmp_linuxkit_bin.tar: $(LINUXKIT_DEPS)
-	tar cf - -C src/cmd/linuxkit . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)" -o $(LINUXKIT) > $@
+linuxkit:
```
rn, line 52: this needs to be PHONY since it's not a file being created. Alternatively any reason to not use `bin/linuxkit` since this is the real target

Makefile
```diff
@@ -50,56 +49,25 @@ bin/manifest-tool: tmp_mt_bin.tar | bin
...
-	$(MAKE) clean
-	$(MAKE) -j 3 GOOS=windows tmp_rtf_bin.tar tmp_mt_bin.tar tmp_linuxkit_bin.tar
-	$(MAKE) clean
-	$(MAKE) -j 3 GOOS=linux tmp_rtf_bin.tar tmp_mt_bin.tar tmp_linuxkit_bin.tar
-	$(MAKE) clean
-
-LOCAL_LDFLAGS += -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)
-LOCAL_TARGET ?= $(LINUXKIT)
-
-.PHONY: local-check local-build local-test local-static-pie local-static local-dynamic local
```
rn, line 75: you still have `local` and other targets below. These are also PHONY but that is removed here

dave-tucker:
@rn thanks. all comments addressed.

deitch (approved):
I don't see any issues.

dave-tucker:
Rebased to pick up changes in #3645

dave-tucker:
@deitch your +1 vanished after rebase. if you could re-add it i'll go ahead and merge

deitch (approved):

PR #3614, opened 2021-04-01
Ignore detected block sizes in cdrom provider

mraerino:
**- What I did**

Ignored detected blocksizes in go-diskfs.
We were running into an issue where the detection is wrong (https://github.com/diskfs/go-diskfs/issues/104) which causes a panic in `metadata` (https://github.com/diskfs/go-diskfs/issues/103)

**- How I did it**

**- How to verify it**


deitch:
This makes sense to me. There are open PRs on go-diskfs to fix this specific issue (the panics), but I see no reason not to rely on default sizes here.

deitch (approved):

PR #3618, opened 2021-04-04
Move metadata package to go modules

mraerino:
**- What I did**

Migrate the `pkg/metadata` program to go modules

**- How I did it**

Ran:

```
cd pkg/metadata
go mod init github.com/linuxkit/linuxkit/pkg/metadata
go mod tidy
go mod vendor
rm vendor.conf
```

**- How to verify it**

It should still build i guess

**- Description for the changelog**

Migrate `pkg/metadata` to use Go modules

**- A picture of a cute animal (not mandatory but encouraged)**

[](https://unsplash.com/photos/TG43wDuNPj0)

PR #3612, opened 2021-03-23
cross build packages

deitch:
**- What I did**

Support cross-building packages (cross-building OS images already is complete)

**- How I did it**

Modify the linuxkit binary to support `--platform` options

**- How to verify it**

CI

**- Description for the changelog**

Support image cross-build

PR #3623, opened 2021-04-13
Full update Alpine: base to 3.13, go to 1.16, mod=vendor support

deitch:

**- What I did**

Added support for go modules with vendoring to our standard `go-compile.sh`

**- How I did it**

Added a few lines that check for `go.mod` and `vendor/` dir

**- How to verify it**

CI

Support for go modules with vendor in linuxkit package built via `linuxkit pkg build`

**- Notes**

I have to refresh on how to push out alpine.


deitch:
Helps with #3618

mraerino (approved):
this makes sense

rn:
LGTM but need to push a new alpine base and update to whichever packages needs it and the packages need updating...

deitch:
I don't recall how to build and push out this alpine package.

rn:
It's a bit intricate since `make push` creates the versions files, but `docs/releasing.md` has the instructions

deitch:
> It's a bit intricate since make push creates the versions files, but docs/releasing.md has the instructions

Ouch. I just read it through, I can see why I forgot how to do that. So many back and forth and back and forth, and it is part of a larger release.

I am starting to go through it, not sure I will succeed all the way. Maybe the package cross-build #3612 once it is in can be extended here and simplify it?

deitch:
Ugh, `make push` failed:

```
Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
docker run --rm  sh -c 'echo Dockerfile /lib/apk/db/installed $(find /mirror -name '*.apk' -type f) $(find /go/bin -type f) | xargs cat | sha1sum' | sed 's/ .*//' | sed 's/$/-amd64/'> hash
Unable to find image 'sh:latest' locally
docker: Error response from daemon: pull access denied for sh, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```

deitch:
Updated to include latest go, also documented how to update packages.

deitch:
And alpine base updated to `alpine:3.13`

tools/alpine/Dockerfile
```diff
@@ -7,11 +7,18 @@ RUN apk update && apk upgrade -a
 COPY Dockerfile /Dockerfile
 COPY packages* /tmp/
 
-# mirror packages
+# mirror packages - both generic and reposiroty specific ones
```
dave-tucker, line 10: `s/reposiroty/repository/g`

dave-tucker (changes requested):

deitch:
I still am running through my own minor tweaks on this, so even if someone approves and it passes CI, please let me do the merge.

rn:
please don't update alpine this way. Every alpine update has slight issues (packages renamed etc). Here you just update the alpine base for a small number of packages. If you bump alpine it's best to update all the uses of `linuxkit/alpine` and rebuild the users of it in one go. Otherwise you'll run into issues later on and need to update alpine again etc.

deitch:
@rn you confused me.

This is trying to do several things:

* get us off of alpine 3.11/3.12 to modern 3.13
* update `go-compile.sh`
* update go to 1.16

Once this works, and I can do `make push` successfully, we can run an update on all of the packages and rerun CI to rebuild everything.

What are you suggesting we do?

rn:
If you are willing to bite the bullet of upgrading everything to newer alpine, that's great.

deitch:
"willing" is an interesting choice of words. I will try, though.

deitch:
That was brutal, but I finally got this thing to build the `linuxkit/alpine` image. Got caught up in versions conflicts and seccomp and musl.

Going to try to update the packages on different archs and then update per `releasing.md`

deitch:
I think I have the final version of this working. I will push out the images and update downstream dependencies. Before I go through that headache, I would like some maintainers to review the changes to `tools/alpine`.

The rest will be in another commit, to be squashed immediately prior to merge.

tools/alpine/README.md
```diff
@@ -0,0 +1,43 @@
...
+the [Dockerfile](./Dockerfile) in this directory.
+
+The packages installed come from several sources:
+
+* [packages](./packages) - this file contains the list of packages to mirror locally in `linuxkit/alpine`, and will be available to all downstream users of `linuxkit/alpine`. These are installed using the default `apk` package version for the specific version of alpine. For example, if the line starts with `FROM alpine:3.13` and `packages` contains `file`, then it will run simply `apk add file`. The packages listed in [packages](./packages) are installed on all architectures.
+* `packages.<arch>` - these files contain the list of packages to mirror locally in `linuxkit/alpine`, like `packages`, but only for the specified architecture. For example, [packages.x86_64](./packages.x86_64) contains packages to be installed only on `linuxkit/alpine` for `x84_64`.
+* `packages.repo.<name>` - these files contain the list of packages to mirror locally in `linuxkit/alpine`, like `packages`, but to pull those packages from the provided `<name>` of Alpine's `apk` repo. For example, `packages.repo.edge` installs packages from Alpine's `edge` package repository.
```
rn, line 37: IIRC Mixing Alpine packages from different repos has been problematic in the past.

tools/alpine/Dockerfile
```diff
@@ -7,11 +7,30 @@ RUN apk update && apk upgrade -a
...
+        repo=${repopkgs##*repo.} && archspecific=/tmp/packages.$(uname -m).repo.${repo} && mergedfile=/tmp/packages.merged.${repo} && repofile=/tmp/repositories.${repo} && \
+	cachedir=/tmp/cache/${repo} && \
+	mkdir -p ${cachedir} && \
+        cp ${repopkgs} ${mergedfile} && \
+        if [ -f ${archspecific} ]; then cat ${archspecific} >> ${mergedfile}; fi && \
+        sed "s#alpine/[^\/]*/#alpine/${repo}/#g" /etc/apk/repositories > ${repofile} && \
+        apk update --repositories-file=${repofile} --cache-dir ${cachedir} && \
+        apk fetch --repositories-file=${repofile} --cache-dir ${cachedir} --recursive -o /mirror/$(apk --print-arch) $(cat ${mergedfile}); done
+
+# we CANNOT allow musl-dev or musl > 1.2.2-r0, which ships with alpine:3.13, because 1.2.2-r2, which ships with alpine:edge / alpine:3.14
```
rn, line 26: Ughh, this is ugly (no, I don't have a better solution)

rn:
This looks painful. BTW, can you change the commit message for the first commit. It currently says: `use mod=vendor when available, latest go from edge` but should really say that this is a full blown Alpine update. IN the commit body it would be good to mentioned that go is picked up from edge and go-compile is updated. The latter I would ideally have as a separate commit, but I know how painful that would be to change so not suggesting it :)

tools/alpine/README.md
deitch: You are right, it has. But this is the only way to get go1.16. This is all it is being used for for now, and I don't have a better answer.

tools/alpine/Dockerfile
deitch: "ugly" is an understatement, but I also don't have something better.

deitch:
> BTW, can you change the commit message for the first commit

Done

> The latter I would ideally have as a separate commit, but I know how painful that would be to change so not suggesting it :)

Brutally so. I don't mind other minor updates later.

deitch:
Everything updated. Hoping and prating that CI goes clean now.

deitch:
The only remaining failing test is containerd, which is because version 1.4.1 doesn't correctly handle `index=off` in its options. It is fixed in 1.4.2, see [here](https://github.com/containerd/containerd/blob/v1.4.2/snapshots/overlay/overlay_test.go#L177).

So, I am going to disable that specific test so this is clean, and we can reenable it in a separate PR when we bump containerd up.

deitch:
Closing in favour of #3637

PR #3627, opened 2021-04-19
Go Module-ify all the things

dave-tucker:
**- What I did**

I added a `go.mod` file to the root of each `pkg` directory. (except metadata, as that was handled in #3618 which this PR includes)
This is a deliberate departure from how we handled vendoring previously fot two reasons.

1. Any go-based packages will *always* have a `go.mod` at the root such that go doesn't behave weirdly when compiling
1. We can very easily add tags to version modules if we choose:

Before: `git tag pkg/init/cmd/init/v1.0.0`
After: `git tag pkg/init/v1.0.0`

**- How I did it**
For each package:

1. Moved vendor folder to new location
1. `go mod init` - imports existing deps
1. `go mod tidy`
1. `go mod vendor`
1. Remove old vendor file

**- How to verify it**

Packages will still compile

image

Requires #3623 
Includes #3618


deitch:
I am quite in favour of this. Not bothering to click "Approve", since you will have to rebase anyways. Will "Approve" after that.

Important to note that #3623 uses go1.16, which has `GO111MODULE=on` as default. Every build failed, since there was no `go.mod`, so many of them have added an explicit `GO111MODULE=off` to the packages (done in each relevant package's `Dockerfile`).

Once that is in, as part of this PR, you will need to remove that.

Definitely should wait until #3623 is in before even trying, though. I have had to rebuild everything multiple times, since you often are pretty far down before realizing that some deep-down dependency requires an update (see `apk add libseccomp-static`).

deitch:
The sheer number of files! The vast majority are changes in `pkg/<something>/vendor/`. IIRC, `vndr` also did a decent job trimming the contents of `vendor/`, so unclear why this is different. But if it works, I don't complain.

dave-tucker:
@deitch some are simple renames because the vendor directory has moved from `pkg/foo/cmd/foo/vendor` to `/pkg/foo/vendor`... and go mod seems to have slightly different rules than `vndr` for what files are included or not.... anyways, this is *almost* there now I added `binfmt` to CI. Just debugging some failed packages.

dave-tucker:
Ok I'm 99.9% confident that this will pass CI now :tada: 

I've included the commit from #3618 in this PR (thanks for the contribution @mraerino :tada:) as it too relies on the CI and UX changes in this patch set. 

For reviewers:

1. CI now sets up `binfmt_misc` support for `arm64`. It wasn't done explicitly before and I saw some intermittent errors.
1. CI now *does not build s390x packages*. QEMU emulation for this arch is too unreliable. I've tried with `QEMU_CPU=max` on my machine and still end up with `Illegal Instruction` errors
1. The UX of `lkt pkg build` has changed slightly in order to accommodate this

Prior to this change, `lkt pkg build --platform linux/amd64, linux/arm64 pkg/format` would build the package for both platforms regardless of whether or not the package was supported on that architecture.

The change in this PR will now only build only supported architectures from the provided `--platforms` and emits a WARN log message if the arch/package is not supported. This allows for us to pass `OPTIONS= --platforms linux/amd64,linux/arm64` and build the combination of supported packages for those two architectures.

@deitch i'll definitely need your review on the `lkt pkg build` change because I expect that I may also need to update docs?

deitch:
@dave-tucker let's talk through the UI. We probably should update the docs to reflect this once settled.

The existing behaviour is as follows:

* If you specify `--platforms`, it builds for those platforms. It doesn't care what is in `build.yaml`, it just builds. ("flags have power")
* if you do not specify `--platforms`, it builds for every platform in `build.yaml`
* if you do not specify `--platforms` and there are no specific platforms in `build.yaml`, it builds for every default platform in linuxkit (`linux/amd64`,`linux/arm64`,`linux/s390x`)

Essentially, `build.yaml` is _advisory_, telling you what to build if nothing else is specified, rather than _mandatory_ or restrictive, saying, "build for these platforms and no other".

You are proposing changing it so that `build.yaml` is restrictive: you only can build for the platforms specified in `build.yaml`. If you try to build for anything else, you will get an error (or, as suggested above, a warning).

I think both of the UX structures - current and proposed - have limitations and don't hit all use cases:

* I want to build for all supported platforms - works in both cases
* I want to build for one or more specific supported platforms - works in both use cases
* I want to have a generic list and have it build only for the supported ones out of that list - works only in your proposed use case
* I want to build for one or more specific platforms even if unsupported - works only in the current structure

(yes, the last one is real; I just saw someone build some packages for risc-v)

Can we come up with a UX that can handle all of the use cases? If we could do that, I think it would be great. Maybe your proposed change with an additional `--platform-overide=true` or similar flag? Not sure.

Separately, I really would prefer the UX change to linuxkit binary in a separate smaller PR, if we can do it.

dave-tucker:
@deitch ack, i'll move this over to a new PR.

> Essentially, build.yaml is advisory, telling you what to build if nothing else is specified, rather than mandatory or restrictive, saying, "build for these platforms and no other".

I like that... but in practice it hasn't always been used that way. the `arches` in the `build.yml` of `bpftrace` is empty because it doesn't compile and should always be skipped - one of the many issues that led me towards this version of the UX.

> Can we come up with a UX that can handle all of the use cases? If we could do that, I think it would be great. Maybe your proposed change with an additional --platform-overide=true or similar flag? Not sure.

On reflection, I think I can cover my use case by adding a `--skip-unsupported` flag. This fits the spirit of "flags have power", doesn't change the existing behaviour, and should be a trivial change.

deitch:
Makes sense. Any sane UX that covers all of the use cases should be OK.

dave-tucker:
Ok, waiting on https://github.com/linuxkit/linuxkit/pull/3648 before I rebase this one and get it ready for merge.

.github/workflows/ci.yml
```diff
@@ -91,6 +91,10 @@ jobs:
       with:
         path: ./src/github.com/linuxkit/linuxkit
 
+    - name: Set up binfmt
+      # Only register arm64 as we are on amd64 already. s390x is not reliable
+      run: docker run --privileged --rm tonistiigi/binfmt --install arm64
```
justincormack, line 96: why arent we using our own binfmt package?

dave-tucker: because the buildx instructions said to use tonis's one :joy: i can change it

pkg/host-timesync-daemon/Dockerfile
```diff
@@ -1,16 +1,9 @@
...
+ENV GOPATH=/go PATH=$PATH:/go/bin
 
-ENV VIRTSOCK_COMMIT=f1e32d3189e0dbb81c0e752a4e214617487eb41f
-RUN mkdir -p $GOPATH/src/github.com/linuxkit && \
-  cd $GOPATH/src/github.com/linuxkit && \
-  git clone https://github.com/linuxkit/virtsock.git && \
-  cd virtsock && \
-  git checkout $VIRTSOCK_COMMIT
-
-COPY .  /go/src/host-timesync-daemon
```
deitch, line 13: The move to go modules made this able to just import it, and all of this build stuff goes away?

dave-tucker: It's vendored at that commit. It could have been done without modules. Not sure why we did it this way originally.

deitch: In any case, we now have it in `go.sum`, so that should be good enough.

deitch (approved):
Other than resolving the binfmt question, looks good to me. I am happy to see this done. Hats off.

rn:
where are we with this? I see one approval and a green CI, however there are no updates to any uses of the changed packages so we do not know if they actually work. Can we rebuild all changed packages, and update the YAMLs before merging?

dave-tucker:
I can update all the package shas here but unfortunately I can't push them so they can be used in the tests....

dave-tucker:
Update: I can now push images so I'll add a commit that updates the SHAs tomorrow

the-maldridge:
What wound up happening here, this is a changeset that I'm interested in seeing merged, and it sounds like it is done.

deitch:
Yes, I think this is about ready. @dave-tucker do you want to rebase and resolve the `pkg/mount/Dockerfile` conflict? I was about to do it, realized I probably will miss some context.

Once that is in and pushed, we can approve, let CI go green, and merge.

dave-tucker:
I think I got stuck figuring building the packages and updating the SHAs.
Let me go and see where my branch is, rebase and see where we're at.

dave-tucker:
@deitch rebased, updated package shas and CI is green

deitch (approved):

deitch:
Approved. I leave you the honour of merging @dave-tucker

PR #3632, opened 2021-04-22
Update binaries path in the `auditd` package

federico-pellegatta:
**- What I did**
Fixed the binaries path used in the `auditd` package; updated from `/sbin/` to `/usr/sbin/`.

The `auditd` package currently fails with the following error:
```
2021-04-22T10:36:15Z auditd /usr/bin/runaudit.sh: exec: line 5: /sbin/auditd: not found
```

**- How to verify it**
Run the newly added `auditd` package test case.

Updated binaries path from `/sbin/` to `/usr/sbin/` in the `auditd` package.

Signed-off-by: Federico Pellegatta <12744504+federico-pellegatta@users.noreply.github.com>


deitch:
I am in favour of this, but we are updating the base of most of the packages (including auditd) as part of #3623. Once that is in (hopefully another day or two), we should rebase this on that, and then get this one in.

deitch:
Those updates came in. Do you want to rebase this @federico-pellegatta and then see if it still is needed?

PR #3636, opened 2021-04-26
performance regression on hyperkit

fredericdalleau:
While testing a newer 5.10 kernel on docker desktop we found hyperkit was performing slower
than virtualization framework by about 30%.

The benchmark mode of p7zip provided get some measurements :
BENCH="apt -y update && apt -y install p7zip-full && 7z b"
docker run -it --rm debian:buster-slim sh -c "$BENCH"
it also occured with `linuxkit run kyperkit -cpus 2`

Curiously, /proc/cpuinfo was showing a bogomips difference between multiple cpus:
(ns: getty) linuxkit-025000000002:~# uname -srv ; grep bogomips /proc/cpuinfo
Linux 5.10.31-linuxkit #1 SMP Sun Apr 18 17:44:33 UTC 2021
bogomips	: 6999.82
bogomips	: 1250.81
(also occurs with 5.11.15)

The 4.19 kernel not showing this difference, a bisect pointed to:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2420a0b1798d
A revert fixed the issue.

**- What I did**

Identify and fix an issue occuring on hyperkit (only)

**- How I did it**

Kernel bisect and revert
Tested hyperkit and apple virtualization

**- How to verify it**

Run the latest kernel on hyperkit with multiple processors, and check the bogomips in /proc/cpuinfo, or benchmark the performances.
Compare with other VM on same machine.

Fix performance regression on hyperkit

image

### DIFF
--- a/kernel/patches-5.10.x/0001-Revert-x86-tsc-Set-LAPIC-timer-period-to-crystal-clo.patch
+++ b/kernel/patches-5.10.x/0001-Revert-x86-tsc-Set-LAPIC-timer-period-to-crystal-clo.patch
@@ -1,0 +1,35 @@
+From a08c878b14c36a07fc2cfd0e1e3fb668428c0b78 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Fr=C3=A9d=C3=A9ric=20Dalleau?= <fred@dalleau.re>
+Date: Fri, 19 Mar 2021 09:15:03 +0100
+Subject: [PATCH] Revert "x86/tsc: Set LAPIC timer period to crystal clock
+ frequency"
+
+This reverts commit 2420a0b1798d7a78d1f9b395f09f3c80d92cc588.
+---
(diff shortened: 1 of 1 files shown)


djs55 (approved):
Disclaimer: I'm not a kernel expert :)

I think that there must be a bug in `hyperkit` somewhere which this commit triggers, but we haven't managed to find it. This particular revert has been shipped in Docker Desktop and seems to fix / work-around the issue, without causing problems under Hyper-V or qemu. I think it's worth taking the patch while `hyperkit` remains a common platform for LinuxKit kernels. In the longer term we can hopefully either fix `hyperkit` properly or transition people to another hypervisor.

An additional benefit of taking the patch would be that Docker Desktop can switch back to vanilla linuxkit kernels again, which is always nice.

rn:
I posted a fix for HyperKit https://github.com/moby/hyperkit/pull/311 so closing this.

PR #3637, opened 2021-04-27
Full update Alpine: base to 3.13, go to 1.16, mod=vendor support

deitch:

**- What I did**

Added support for go modules with vendoring to our standard `go-compile.sh`

**- How I did it**

Added a few lines that check for `go.mod` and `vendor/` dir

**- How to verify it**

CI

Support for go modules with vendor in linuxkit package built via `linuxkit pkg build`

**- Notes**

I have to refresh on how to push out alpine.

This should be a replacement for #3623 ; I am doing a separate PR because I was concerned about getting it right with all of the changes and rebasing. If this one passes clean, we will merge this in and close that one.

This adds one more thing: containerd 1.4.4 instead of 1.4.1.


deitch:
@thaJeztah I managed to roll containerd v1.4.4 into this too :-)

tools/alpine/go-compile.sh
```diff
@@ -13,6 +13,9 @@ dir="$1"
 
 cd "$dir"
 
+# Use '-mod=vendor' for builds which have switched to go modules
+[ -f go.mod -a -d vendor ] && export GOFLAGS="-mod=vendor"
```
dave-tucker, line 17: nit: this is true for builds that have switches to go modules AND use `go mod vendor`  to populate a vendor directory.

It's a little bothersome to me me that we have a script here and also one in `pkg/go-compile`.
It might be nice to deprecate this one and just have the latter.

tools/grub/Dockerfile
```diff
@@ -1,4 +1,6 @@
+# this is really hard to build. Do not change this version unless you must
```
dave-tucker, line 1: I appreciate the warning, but some context on why it's hard would be nice.

dave-tucker (changes requested):
a couple of nits. mainly because the change is quite large, and i'd like to ensure that the things learned from doing it are persisted in documentation or the commit history.

i'd also prefer that we get rid of alpines `go-compile.sh` and prefer the one in the go-compile image.

dave-tucker:
> This is the other half of it. By declaring it `extern` in `main.go`, it has to be declared properly somewhere, or gcc will complain. When compiling on linux for amd64 or arm64, that is in `rng_linux_amd64.go` and `rng_linux_arm64.go`, respectively. On any other platform - in our case, just s390x, but could be anything - we don't declare it anywhere, but we do include `rng_unsupported.go`.
> 
> Now it is even on all platforms:
> 
> * everyone includes `main.go`, where it is declared as `extern int`
> * everyone includes `rng_<something>` for their platform, where it is declared as `int`
> 
> And gcc is happy.

@deitch ack. would have been handy to have something in the commit message when i was reviewing, but appreciate that this change is huuuuuge! thanks for doing this by the way. rather you than me

tools/alpine/go-compile.sh
deitch: Agreed. At this point, though, it was _so_ much effort over several weeks just to get everything working, I wasn't going to try refactoring that.

Similar nit: only two packages downstream use containerd; so why have it in alpine, where the entire base changes if we bump alpine version? I plan on breaking it out... _after_ this is in

tools/grub/Dockerfile
deitch: I basically gave up on getting it to work with the newer versions. We are using a pretty old grub, no longer maintained, and will not get fixed. The warning was basically, "leave this alone, do not even try."

I looked about a year ago at replacing this with upstream grub. I think almost everything we customize already was upstreamed. Something was missing, but too much to remember now.

In any case, big red warning lights: "do not change this or suffer what you have brought upon yourself!"

deitch:
> appreciate that this change is huuuuuge! thanks for doing this by the way. rather you than me

😆 

👍 

Looking for some nice approvals (of the PR; not of me).

tools/alpine/go-compile.sh
deitch: changing this would require a **full** day of work, as the `go-compile.sh` is in alpine, which means changing everything downstream, a long slow process.

tools/grub/Dockerfile
deitch: Similar issue. If we change a single thing in `tools/grub`, then we rebuild it and push it out, then change the downstream dependencies `tools/mkimage-iso-efi`, `tools/mkimage-qcow2-efi`, `tools/mkimage-raw-efi`, then rebuild those and push them out, then etc. etc.

We should avoid doing that at all costs here.

dave-tucker (approved):
I've opened issues so we don't lose track of the nits that aren't really practical to address here.
Thanks again for this @deitch

tools/alpine/Dockerfile
```diff
@@ -7,11 +7,30 @@ RUN apk update && apk upgrade -a
...
+	cachedir=/tmp/cache/${repo} && \
+	mkdir -p ${cachedir} && \
+        cp ${repopkgs} ${mergedfile} && \
+        if [ -f ${archspecific} ]; then cat ${archspecific} >> ${mergedfile}; fi && \
+        sed "s#alpine/[^\/]*/#alpine/${repo}/#g" /etc/apk/repositories > ${repofile} && \
+        apk update --repositories-file=${repofile} --cache-dir ${cachedir} && \
+        apk fetch --repositories-file=${repofile} --cache-dir ${cachedir} --recursive -o /mirror/$(apk --print-arch) $(cat ${mergedfile}); done
+
+# we CANNOT allow musl-dev or musl > 1.2.2-r0, which ships with alpine:3.13, because 1.2.2-r2, which ships with alpine:edge / alpine:3.14
+#   uses the new faccessat2 system call which gives errors, see https://wiki.alpinelinux.org/wiki/Draft_Release_Notes_for_Alpine_3.14.0#faccessat2
```
djs55, line 27: Are the errors coming from a Docker Desktop shipping an old `libseccomp`?

djs55 (approved):
LGTM!

tools/alpine/Dockerfile
deitch: I am not sure. @justincormack understands it in a way that I do not. I just accepted it as the issue. :-)

rn:
Heavy birth this one. Thanks for digging through this. I guess you have not updated the kernel build yet. I'll do that with the next kernel update.

rn (approved):

deitch:
> I guess you have not updated the kernel build yet. I'll do that with the next kernel update.

Correct. It was too much.

I do plan on attacking that "tag the kernel builder with the kernel" in the next week or so. If you want to wait to get that in, might be a good idea.

PR #3638, opened 2021-04-28
handle image already in cache

deitch:

**- What I did**

* Add check if an image exists locally before building with `lkt pkg build`. Beforehand, it had just looked on the registry; now it checks both, and validates the local image if found (all components exist, hashes valid, diffids valid). As before, can be overridden by `--force`.
* Add pull/push logging to `lkt pkg` activities via the underlying library, using the existing streams and config

**- How I did it**

* when calling `cache.ImagePull()`, have an option to `alwaysPull` and, if false, check locally and validate if found before going to remote and, eventually, build/
* Tied the logging to the container registry libraries to our own logging setup in `main.go

**- How to verify it**

CI

Check local cache before building images with override option; verbose push logging

FYI, I would like #3637 (or #3623, which #3637 should replace) to go in first, in case it has any conflicts and causes a rebase. #3637 is an enormous lift to rebase. So this is good for review and CI, but let me handle the merge once approved and CI is green.

### DIFF
--- a/src/cmd/linuxkit/cache/write.go
+++ b/src/cmd/linuxkit/cache/write.go
@@ -28,8 +28,9 @@ const (
 )
 
 // ImagePull takes an image name and pulls it down, writing it locally. It should be
-// efficient and only write missing blobs, based on their content hash.
-func (p *Provider) ImagePull(ref *reference.Spec, trustedRef, architecture string) (lktspec.ImageSource, error) {
+// efficient and only write missing blobs, based on their content hash. If the ref already
+// exists in the cache, it will niot pull anything, unless alwaysPull is set to true.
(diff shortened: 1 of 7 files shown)


src/cmd/linuxkit/cache/write.go
```diff
@@ -28,15 +28,27 @@ const (
 )
 
 // ImagePull takes an image name and pulls it down, writing it locally. It should be
-// efficient and only write missing blobs, based on their content hash.
-func (p *Provider) ImagePull(ref *reference.Spec, trustedRef, architecture string) (lktspec.ImageSource, error) {
+// efficient and only write missing blobs, based on their content hash. If the ref already
+// exists in the cache, it will niot pull anything, unless alwaysPull is set to true.
```
rn, line 32: nit niot -> not

rn (approved):

PR #3642, opened 2021-04-29
document and simplify some releasing

deitch:

**- What I did**

* Created `docs/alpine.md` to describe how to update base alpine and its dependencies. Most of it is taken from `docs/releasing.md`.
* Updated the text in `docs/alpine.md` to be more generic and handle both releases and general updates of alpine.
* Simplified the text in `docs/alpine.md` to reflect that `lkt pkg push` handles cross-building now, and no need to hop between build servers, except for the base package itself.
* Added makefile logic to `pkg/Makefile` so that `lkt pkg push` will automatically add the release CLI flag if `LK_RELEASE` is set, thus further simplifying the documentation and unifying the process between general updates and releases.
* Added makefile target to the root `Makefile` to update the packages, with "check LK_RELEASE var" logic similar to the above, thus further simplifying the documentation and unifying the process between general updates and releases.
* Updated `docs/alpine.md` to reflect the makefile simplifications.
* Updated `docs/releasing.md` to remove lots of redundant text, simplify it, and reference the unified language in `docs/alpine.md`

This whole process could use further simplification, but one step at a time.

**- How I did it**

Wrote markdown files, wrote makefile text

**- How to verify it**

I ran them manually to check, but CI should catch it as well.

Clearer alpine and release documentation, simplification of base updates


Makefile
```diff
@@ -130,3 +130,15 @@ ci-pr: test-cross
 .PHONY: clean
 clean:
 	rm -rf bin *.log *-kernel *-cmdline *-state *.img *.iso *.gz *.qcow2 *.vhd *.vmx *.vmdk *.tar *.raw
+
+update-package-tags:
```
rn, line 134: This mostly is a shell script...I don't think this should be in the toplevel Makefile..maybe it should be a shellscript

pkg/Makefile
```diff
@@ -3,11 +3,17 @@ DIRS = $(dir $(shell find . -maxdepth 2 -mindepth 2 -type f -name build.yml))
 
 OPTIONS ?=
 
+PUSHOPTIONS =
```
rn, line 6: This is not necessary since we already have OPTIONS, which was added for this purpose and is more generic. See commit
```
1fadf13ec4a2 ("pkg: Allow specifying addition options to 'linuxkit' invocations")
```

docs/alpine.md
```diff
@@ -0,0 +1,212 @@
+# Updating Alpine Base
```
rn, line 1: I'm in two minds about factoring this out, but if we do the file name needs to be more descriptove. like `alpine-base-update.md`

deitch: I don't mind changing the name.

> I'm in two minds about factoring this out, 

I was, as well, until I looked back on what I went through updating alpine. This was 100% _not_ a release, yet had to go through all of the process, and while IW as at it, figure out where to start, which parts of the doc made sense just for a release vs for updating alpine, etc. It really needed a standalone "here is how to update alpine and its dependencies" doc.

pkg/Makefile
deitch: I tried every which way, but make kept stymieing me. How would you do it such that it would work with all 4 combinations of: `LK_RELEASE` set and not set; `OPTIONS` set and not set?

If you do:

```make
OPTIONS += -release $(LK_RELEASE)
```

then it only works if `OPTIONS` was not set, i.e. `make LK_RELEASE=foo` but not `make LK_RELEASE=foo OPTIONS=abc`.

If you do

```make
OPTIONS = $(OPTIONS) -release $(LK_RELEASE)
```

you end up with a recursive issue, etc.

If you can find a way to do it such that it works in all circumstances, I am happy to use it. But for now, this does just work.

Makefile
deitch: It was in a document, which meant copy-paste and everything with it. This certainly is a step up. I would take this in, then we can iterate if we want.

pkg/Makefile
rn: if you need to specify more than one option (where do you need this?) can'y you do: `make OPTIONS="--foo -release $(LK_RELEASE)"`?

Makefile
rn: As said, I'd much rather have this as script. The Makefile currently is about building stuff. We already have scripts to manage tags and such in `scripts` so might as well put this there.

pkg/Makefile
deitch: I actually do not know that we do. I am trying to keep the current behaviour while adding capabilities.

I also don't want to have to tell people `make push`, unless you are doing a release, in which case do `make OPTIONS="--release $(LK_RELASE)" push`. This alpine and its dependency updating and pushing process is so brutal and complicated that anything that makes it easier is of value.

If I can tell people (which I did in the doc): follow the steps in here whatever you are doing. If you are creating a release first, `export LK_RELEASE=foo` and then follow the _exact same steps_. I am 200% for it.

Makefile
deitch: You want a script in `scripts/` that then calls this `update-component-sha.sh`? 

> Makefile currently is about building stuff

make is about combining common tasks, so I can do `make foo` and `make bar`. That might be "building stuff", and it might be updating stuff. We have lots of non-"build stuff" targets: `bin` creates a directory; `test-cross` runs tests; `clean` removes artifacts. 

Personally, I think the entire alpine and dependency build process should be simplified into a single script with options. But I am not about to tackle that now; maybe after I have had some distance from the last two weeks of painful building and rebuilding (maybe). In the meantime, I just want to capture what I went through and simplify it with low-hanging fruit so the next person doesn't go through it.

pkg/Makefile
rn: we do push without release. any package update should be a push (and update of tags) and release should have once in a while once  we have a good set up updated packages.

The alpine part is really a separate issue which has little to do with the Makefile changes.

deitch: > we do push without release. 

Not according to the existing docs. They just said "run `make OPTIONS="-release ${LK_RELEASE" push`. So I could not even copy-paste, as confusing as that would be, because I would create a release by accident, so I had to carefully copy, edit, paste.

In documenting what is needed, it was clear to me that there are two use cases:

* push when not releasing `make push`
* push when releasing `make OPTIONS="-release rel_${LK_RELEASE}" push`

Trying to simplify this into something that requires zero cognitive energy is exactly what I am after. Always run `make push`, but sometimes you set `LK_RELEASE` at the very beginning of the process.

rn: the doc is about releasing but in between releases we update packages and then update the YAMLs with the new tags, so a release is really something we do two times a year or so while some packages do get updated much more frequent.

deitch: Which is why I broke it out into a separate doc for alpine, and tried to simplify just two of the commands. I am all up for a script for all of it, but can we do this one step at a time? I most certainly am not getting to scripting it all up for a long time. I would rather capture some simplifications, however small, right now.

Which reminds me; I need to rename the file like you suggested.

docs/alpine.md
deitch: File renamed.

deitch:
With #3658 merged in, this gets simplified further.

deitch:
We should make some progress on this @rn

deitch:
With the distance of time, I really don't want to lose the benefits of having documented and simplified (at least a little but) what we do with alpine. Can we get this in?

deitch:
I just did another cleanup of the docs here. Can we get this merged in please?

dave-tucker (approved):
Mostly LGTM but with a couple of nits.

docs/alpine-base-update.md
```diff
@@ -0,0 +1,231 @@
...
+
+```sh
+LK_ROOT=$(pwd)
+LK_REMOTE=origin         # or whatever your personal remote is
+LK_BRANCH=alpine-update  # or whatever the name of the branch on which you are working is
+```
+
+Note that if you are cutting a release, the `LK_BRANCH` may have a release-type name, e.g. `rel_v0.4`.
+
+Make sure that you have a recent version of the `linuxkit`
```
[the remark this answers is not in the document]
deitch: Done

docs/alpine-base-update.md
```diff
@@ -0,0 +1,231 @@
...
+utility in the path. Either a previous release or compiled from
```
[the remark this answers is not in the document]
deitch: Much better language, thanks. Done.

docs/alpine-base-update.md
```diff
@@ -0,0 +1,231 @@
...
+push the manifest.
+
+#### Make alpine changes
+
+Make any changes in `tools/alpine` that you desire, then commit them:
+
+```sh
+cd tools/alpine
+# make changes
+git commit -s -a -m "Update linuxkit/alpine"
```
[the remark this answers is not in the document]
deitch: I added some text right beforehand, please take a look.

docs/alpine-base-update.md
```diff
@@ -0,0 +1,231 @@
...
+git checkout $LK_BRANCH
+cd $LK_ROOT/tools/alpine
+make push
+```
+
+Repeat on each platform.
+
+#### Commit Changed Versions Files
+
+When all of the platforms are done, copy the changed `versions.<arch>` from each platform to one please, commit and push.
```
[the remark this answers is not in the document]
deitch: Being a little too polite; must be my Canadian side.

pkg/Makefile
```diff
@@ -3,19 +3,25 @@ DIRS = $(dir $(shell find . -maxdepth 2 -mindepth 2 -type f -name build.yml))
...
+
+ifneq ($(LK_RELEASE),)
+PUSHOPTIONS += -release $(LK_RELEASE)
```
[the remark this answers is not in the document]
dave-tucker: FWIW I find them all equally unappealing :joy:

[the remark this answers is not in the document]
deitch: With that, I cannot argue

PR #3643, opened 2021-04-29
pass linkmode=external only for Linux

deitch:

**- What I did**

When upgrading to go1.16 and gcc10.x, we ran into issue of conflict with `-buildmode=pie` creating dynamic linux binaries rather than static, since it uses go's new built-in linker, which creates dynamically-linked linux binaries. The solution is to put `-linkmode=external`, which causes it to use the external linker.

What we did worked, but it set `-linkmode=external` for _all_ builds, not just for linux. gcc handles most of those fine... except for Darwin arm64 (macOS M1). 

In truth, `-linkmode=external` really should only be used for the scenario when all of the above is true:

* building for linux
* `-buildmode=pie`

The only affected area is `tools/go-compile` and its dependencies, since everything that uses `go-compile.sh` from within `linuxkit/alpine` is building linux packages anyways. The only area that uses `tools/go-compile` is `Makefile`, and is referenced in `docs/`

This PR removes `-linkmode=external` from `Makefile`, places it inside `tools/go-compile/compile.sh`, but only when building for linux with `-buildmode=pie`.

**- How I did it**

1. Change `tools/go-compile/compile.sh` and check it in
2. Push out the `linuxkit/go-compile` image
3. Run `scripts/update-component-sha.sh`, which updates Makefile and one doc
4. Remove `-linkmode=external` from `Makefile`

**- How to verify it**

Run it. I did. CI is nice too.

Handle building from makefile for non-Linux when run from Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 VERSION="v0.8+"
 GIT_COMMIT=$(shell git rev-list -1 HEAD)
 
-GO_COMPILE=linuxkit/go-compile:ceff1817e4c670d253c7bd38da05daecaa513f5f
+GO_COMPILE=linuxkit/go-compile:9f76f29606aec51f2f568984c4c6fe55da2dde10
 
 ifeq ($(OS),Windows_NT)
 LINUXKIT?=bin/linuxkit.exe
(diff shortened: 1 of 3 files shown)


deitch:
Also fixes `ineffassign` warnings.

dave-tucker (approved):
LGTM

PR #3647, opened 2021-05-01
ci: Cache Package Builds

dave-tucker:
**- What I did**
Added the default linuxkit cache directory to the GitHub Actions cache.
This will ensure that we don't pull images that already exist in the cache, or build them if we've already done so. 
It should speed up CI.

**- How I did it**
Editied the ci.yml file

**- How to verify it**
Output from the CI run should show the cache directory being cached

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,14 @@ jobs:
         sudo ln -s $(pwd)/bin/linuxkit-amd64-linux /usr/local/bin/linuxkit
         /usr/local/bin/linuxkit version
 
+    - name: Cache Packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.linuxkit/cache/
+        key: ${{ runner.os }}-linuxkit-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-linuxkit-
+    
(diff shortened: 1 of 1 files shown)


dave-tucker:
First run took 2 mins to build packages and around 1hr to test packages. Hoping that the 2nd run will yield a small improvement.
The larger improvement would be gained from situations where a PR changes one or more packages, and then pushes another commit that only touched one of them...

BUT interestingly, it should also work for cases where you build a package + update the package SHA in the tests or examples as the new image should be in the cache. That might get us a step closer to making it easier for folks without push access to the `linuxkit` org being able to make changes to packages.

dave-tucker:
And we have a 10 second speed up for building packages :tada: 
I've just noticed that the `Package Tests` are using alpine:3.11 so perhaps there will be no speed up there in this run, but we can improve on that later.

deitch:
Oh my. Anything and everything that speeds up CI, especially the package tests.

dave-tucker:
Ok it doesn't make a huge difference right now because the images used in `Package Tests` don't always match the latest (but they should). I'd still like this merged for the reasons mentioned in other comments.

deitch (approved):

PR #3648, opened 2021-05-02
pkg_build: Allow skipping unsupported platforms

dave-tucker:
**- What I did**

Fixed a bug to ensure to that `--platforms` **always** overrides build.yml whether or not the platform is declared as supported or not.

Fixed a missing comment on `FullTag` which was upsetting the linter.

Fixed a messy import from being named `v1` to being named `registry`.

Added a `--skip-platforms` flag to `lkt pkg build`. This allows for use cases where you want to build only supported combinations from a list of packages. For example, in CI we wish to skip s390x emulation as it's buggy, so we'd like to do `lkt pkg build --skip-platforms=linux/s390x ...`

**- How I did it**
Edited `pkg_build.go` and other files in `pkglib` and added documentation

**- How to verify it**

Before, attempting to force `pkg/bpftrace` to build on amd64 would fail:
```
$ linuxkit pkg build --platforms linux/amd64 --disable-cache --force pkg/bpftrace
Building "linuxkit/bpftrace:a5c57a3791a87e96aea456f78fd88973ea9904d0"
arch amd64 not supported by this package, skipping build
```

Now it at least attempts to build it (and fails because it's broken).

To solve my use case, the following command will skip s390x builds.

```
lkt pkg build --skip-platforms linux/s390x ...
```

Added `--skip-platforms` flag to `lkt pkg build` to allow for skipping one or more platforms

image

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -105,6 +105,15 @@ As of this writing, those are:
 * `linux/arm64`
 * `linux/s390x`
 
+You can choose to skip one of the platforms from `build.yml` or those selected
+by default using the `--skip-platforms` flag.
+
+For example:
+
+```
+linuxkit pkg build --skip-platforms linux/s390x ...
+```
+
 You can override the target build platform by passing it the `--platforms` option:
 
 ```
--- a/src/cmd/linuxkit/pkg_build.go
+++ b/src/cmd/linuxkit/pkg_build.go
@@ -28,6 +28,7 @@ func pkgBuild(args []string) {
 	force := flags.Bool("force", false, "Force rebuild")
 	docker := flags.Bool("docker", false, "Store the built image in the docker image cache instead of the default linuxkit cache")
 	platforms := flags.String("platforms", "", "Which platforms to build for, defaults to all of those for which the package can be built")
+	skipPlatforms := flags.String("skip-platforms", "", "Platforms that should be skipped, even if present in build.yml")
 	builders := flags.String("builders", "", "Which builders to use for which platforms, e.g. linux/arm64=docker-context-arm64, overrides defaults and environment variables, see https://github.com/linuxkit/linuxkit/blob/master/docs/packages.md#Providing-native-builder-nodes")
 	buildCacheDir := flags.String("cache", defaultLinuxkitCache(), "Directory for storing built image, incompatible with --docker")
 
@@ -47,13 +48,34 @@ func pkgBuild(args []string) {
 	if *docker {
 		opts = append(opts, pkglib.WithBuildTargetDockerCache())
 	}
+
+	// skipPlatformsMap contains platforms that should be skipped
+	skipPlatformsMap := make(map[string]bool)
+	if *skipPlatforms != "" {
+		for _, platform := range strings.Split(*skipPlatforms, ",") {
+			parts := strings.SplitN(platform, "/", 2)
+			if len(parts) != 2 || parts[0] == "" || parts[0] != "linux" || parts[1] == "" {
+				fmt.Fprintf(os.Stderr, "invalid target platform specification '%s'\n", platform)
+				os.Exit(1)
+			}
+			skipPlatformsMap[strings.Trim(parts[1], " ")] = true
+		}
+	}
 	// if platforms requested is blank, use all from the config
 	var plats []imagespec.Platform
 	if *platforms == "" {
 		for _, a := range p.Arches() {
+			if _, ok := skipPlatformsMap[a]; ok {
+				continue
+			}
 			plats = append(plats, imagespec.Platform{OS: "linux", Architecture: a})
 		}
 	} else {
+		// don't allow the use of --skip-platforms with --platforms
+		if *skipPlatforms != "" {
+			fmt.Fprintln(os.Stderr, "--skip-platforms and --platforms may not be used together")
+			os.Exit(1)
+		}
 		for _, p := range strings.Split(*platforms, ",") {
 			parts := strings.SplitN(p, "/", 2)
 			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/reference"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
+	registry "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/cache"
 	lktspec "github.com/linuxkit/linuxkit/src/cmd/linuxkit/spec"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/version"
@@ -235,7 +235,7 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 		fmt.Fprintf(writer, "building %s\n", ref)
 		var (
 			args  []string
-			descs []v1.Descriptor
+			descs []registry.Descriptor
 		)
 
 		if p.git != nil && p.gitRepo != "" {
@@ -354,9 +354,9 @@ func (p Pkg) Build(bos ...BuildOpt) error {
 }
 
 // buildArch builds the package for a single arch
-func (p Pkg) buildArch(d dockerRunner, c lktspec.CacheProvider, arch string, args []string, writer io.Writer, bo buildOpts) (*v1.Descriptor, error) {
+func (p Pkg) buildArch(d dockerRunner, c lktspec.CacheProvider, arch string, args []string, writer io.Writer, bo buildOpts) (*registry.Descriptor, error) {
 	var (
-		desc    *v1.Descriptor
+		desc    *registry.Descriptor
 		tagArch string
 		tag     = p.Tag()
 	)
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/reference"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/cache"
 	lktspec "github.com/linuxkit/linuxkit/src/cmd/linuxkit/spec"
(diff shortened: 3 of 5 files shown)


rn:
This option seem the wrong way round. By default `pkg build` should only build for architectures specified in `build.yml`. We maybe want to add a flag to override this for testing but then you may as well edit build.yml to check...

deitch:
@rn by default:

* if you specify no arches, then it build for whatever is in `arches` in `build.yaml`
* if you specify no arches, and there are none in `build.yaml`, then it builds for all the built-in defaults (`amd64`, `arm64`, `s390x`)
* if you specify specific arches that are in `build.yaml`, then it builds those

The question becomes, what to do when you specify an arch that is not listed in `build.yaml`. We need to support two use cases:

* do not build if it is not in the file (warn)
* override: "I know what I am doing, if I said `--platforms=linux/dec`, then try to build for it!"

The current behaviour matches the second, but doesn't cover the first. @dave-tucker is adding support for the first, so it becomes:

* if I said `--platform linux/blah`, I know what I am doing, build for `linux/blah`, whether or not it is in `build.yaml`. This makes sense, as CLI flags usually override config files (second case above)
* if I said `--platform linux/blah --skip-unsupported`, I am saying, "I want linux/blah if you support it, but skip/warn if you do not" (first case above)

The behaviour can be summed up as:

1. By default, do what is in `build.yaml`
2. If I override it with a CLI, build for the explicit arch
3. If I override the override, build for the arch only if supported

This gives the ability to say (in a script, e.g.), "build the following 10 packages for arm64 and amd64, but only if supported". That allows you to skip platforms that you might not want built now (e.g. s390x in CI, which does poorly in emulation), and get amd64 and arm64 wherever it is supported.

I think this makes a lot of sense. Does that explain it?

In theory, we could have an alternate route: `--skip-platforms linux/powerpc`, which would mean, "build all platforms in arches except for the ones I specified (`linux/powerpc`), but I defer to whichever @dave-tucker thinks works best here.

dave-tucker:
> By default pkg build should only build for architectures specified in build.yml

Reproducing @deitch's comment on #3627 to aid discussion (emphasisi mine)

> let's talk through the UI. We probably should update the docs to reflect this once settled.
>
> The existing behaviour is as follows:
>
>If you specify --platforms, it builds for those platforms. It doesn't care what is in build.yaml, it just builds. **("flags have power")**
>if you do not specify --platforms, it builds for every platform in build.yaml
>if you do not specify --platforms and there are no specific platforms in build.yaml, it builds for every default platform in linuxkit >(linux/amd64,linux/arm64,linux/s390x)
**Essentially, build.yaml is advisory, telling you what to build if nothing else is specified, rather than mandatory or restrictive, >saying, "build for these platforms and no other".**
>
>You are proposing changing it so that build.yaml is restrictive: you only can build for the platforms specified in build.yaml. If you >try to build for anything else, you will get an error (or, as suggested above, a warning).
>
>I think both of the UX structures - current and proposed - have limitations and don't hit all use cases:
>
>I want to build for all supported platforms - works in both cases
>I want to build for one or more specific supported platforms - works in both use cases
>I want to have a generic list and have it build only for the supported ones out of that list - works only in your proposed use case
>I want to build for one or more specific platforms even if unsupported - works only in the current structure
>(yes, the last one is real; I just saw someone build some packages for risc-v)
>
>Can we come up with a UX that can handle all of the use cases? If we could do that, I think it would be great. Maybe your >proposed change with an additional --platform-overide=true or similar flag? Not sure.

When I thought about it,  we should either decide:

a) Flags have power and consistently override `build.yml`
b) Flags have power (sometimes) and inconsistently override `build.yml`

I ended up thinking that b) doesn't seem like a great UX:
`lkt pkg build --platforms=linux/mips64el pkg/format` would do nothing but print an error message, where it would currently try and build it.

deitch (changes requested):
Looks fine to me.

Can you add a test case for it? The test cases are ~L301-309 in `build_test.go`. The last 3 are arches, so should be pretty straightforward to add for yours.

dave-tucker:
> In theory, we could have an alternate route: --skip-platforms linux/powerpc, which would mean, "build all platforms in arches except for the ones I specified (linux/powerpc), but I defer to whichever @dave-tucker thinks works best here.

The only problem with this route is that `--skip-platforms linux/s390x` only applies to the default "magic" of building for all arches with first class support in lkt. It would be odd to use with  like this `--platforms linux/s390x,linux/arm64 --skip-platforms linux/s390x`. It also wouldn't cover cases where you might want to build for only supported combinations from a selection of non-default arches/

rn:
From @deitch above:
> The behaviour can be summed up as:
> By default, do what is in build.yaml
> If I override it with a CLI, build for the explicit arch
> If I override the override, build for the arch only if supported

I think that if you need an override for an override the UX is wrong.

From @dave-tucker:
> a) Flags have power and consistently override build.yml
> b) Flags have power (sometimes) and inconsistently override build.yml

I'm fine with this a) but b) seems wrong.

from @dave-tucker above:
> For example, in CI we wish to skip s390x emulation as it's buggy, so we'd like to do lkt pkg build --skip-unsupported --platforms=linux/amd64,linux/arm64 ...

So, how about, we had a `--skip-platforms` option where you can specify platforms from `build.yml` which should not be build and we error out when a user specifies both `--skip-platforms` and `--platforms`?

I only notice now that the `--platforms=` argument takes architectures in a different format to `build.yml`. That doesn't look like good UX to me.

deitch:
Agreed. Which is why I think what you have here is the best combination.

dave-tucker:
> Can you add a test case for it? The test cases are ~L301-309 in build_test.go. The last 3 are arches, so should be pretty straightforward to add for yours.

@deitch I thought I could, but I can't (easily). The tests in `build_test.go` pertain to `pkglib` and this logic sits outside of it.
i.e we remove unsupported architectures before they hit `pkg.Build`. I could add a proper `rtf` style test for it, but we don't have any as yet for package builds.

deitch:
Dang. I forgot about that @dave-tucker 

Well, if you can find a way to throw tests in, then great. If not, 🤷‍♂️

rn:
Too many simultaneous replies. @deitch:
> Agreed. Which is why I think what you have here is the best combination.

which statement do you agree with here?

Can we agree that the default and common use case should be that `build.yml` specifies the supported architectures for a given package and that command line arguments should only be used to override the default behaviour?

In my opinion, if you specify `--platforms` the arches specified in `build.yml` should be completely ignored. So you can build for a subset of arches or new arches as you see fit.

If you use `--skip-platforms` you take the set from `build.yml` and filter out the arches which should be skipped.

I don't see a use case where you want to do both, ie specify architectures and then skip them.

One could make an argument for a `--additional-platforms` argument which takes the platforms from `build.yml` and then adds additional ones. Then it may make sense to allow combination with `--skip-platforms` but to me this looks like a pretty esoteric use case.

deitch:
> which statement do you agree with here?

Dave's.

> Too many simultaneous replies

YesNoYesNoYNeos. 😆 

> Can we agree that the default and common use case should be that build.yml specifies the supported architectures for a given package and that command line arguments should only be used to override the default behaviour?

We all are in agreement on that.

> In my opinion, if you specify --platforms the arches specified in build.yml should be completely ignored. So you can build for a subset of arches or new arches as you see fit.

Yes, but. The but is the one Dave brought up: I have 10 packages (or 50) and I want to specify to build them all for the following 2 arches (e.g. amd64 and arm64), but only if they are supported for those platforms. It is a "considerate override".

If Dave can hit his use case by doing `--skip-platforms` instead of `--platforms=... --skip-unsupported`, I would be fine with that too. All I care about is hitting the primary use cases:

* no platforms in flags (which we all agree on and is unchanged)
* I want to specify precisely what to build for, ignoring whatever is in `build.yaml` - current behaviour
* I want to build for a subset of what is in `build.yaml` (all except `foo`)

> One could make an argument for a --additional-platforms argument which takes the platforms from build.yml and then adds additional ones. Then it may make sense to allow combination with --skip-platforms but to me this looks like a pretty esoteric use case

Agreed, let's not go too far down the rabbit hole now.

rn:
> Yes, but. The but is the one Dave brought up: I have 10 packages (or 50) and I want to specify to build them all for the following > 2 arches (e.g. amd64 and arm64), but only if they are supported for those platforms. It is a "considerate override".

I think `--platforms` override is more something  you'd do on individual packages as a developer and `--skip-platforms` would be something you do more on many packages, like skipping s390 cross build in a CI because it's not stable.

We currently only support three arches, so `--skip-platforms` is easy. But if we, say, would support sparc, you could still use `--skip-platforms=s390,sparc` to just build for platforms `arm64`/`amd64`. Doesn't seem too complicated and to me cleaner than an override for an override

deitch:
If @dave-tucker is good with that and it fits his use case, I am on board. 😄

dave-tucker:
@rn @deitch comments now addressed and PR description updated:
TL;DR

1. Fixed a bug so that `--platforms` does indeed always override build.yml as discussed
1. Implemented `--skip-arches` flag that works as @rn suggested

NOTE: I've stuck with `arches` because that's what we call it in `build.yml`, and it expects arches to be `s390x,arm64` etc...
Not `linux/s390x,linux/arm64`. This is by design.
Per Rolf's comment, we may wish to change the behaviour of `--platforms` to match, and perhaps even `s/--platforms/--arches` because (as the docs state) it will never not be a variant of `linux/*`.

deitch:
There was some robust discussion about `--platforms` vs `--arches` when that feature was implemented. I seem to remember a comment somewhere about it as well, but I cannot find it now.

The upshot of it was that just about everything everywhere - docker, go, etc. - use platforms `<OS>/<ARCH>`, so we should as well. We use a shorthand of `arches` inside `build.yaml` (because disk space is super expensive? 😄 More likely because we are building linuxkit, whose OS always is linux), but I would be wary to change either of them at this point. 

That having been said, `--platforms` and `--skip-arches` is jarring. I would go for `--skip-platforms`, but my opinion.

rn:
+1 on `--skip-platforms` and then, at some point can introduce `platforms` with `OS/arch` to `build.yml` and phases out `arches`. I was never a big fan of the `arches` name

dave-tucker:
Alright, 3rd time is a charm :joy: Updated now. Will update PR description too.

deitch (dismissed):

dave-tucker:
Pushing again to fix a typo, but with Avi's LGTM on the previous revision and green CI this should be good to merge.
The platforms tests were spinning for almost 4 hours n that last run :scream: hopefully that was a oneoff, but if not I'll have to think about adding a timeout.

rn (approved):
Thanks. I think the new options are much cleaner

PR #3650, opened 2021-05-02
alpine carry its own hash

deitch:

**- What I did**

Small change: have linuxkit/alpine calculate and carry its own hash, and use it from makefile, rather than calculating it in the Makefile. As a nice side effect, it allows us to be working inside any linuxkit/alpine image and know precisely what tag ran it.

Also, `docker build` sets `TARGETARCH` for us, which gets rid of a few lines in the `Makefile`

Finally, if we ever change how we calculate it, that change will be inherent inside the image build itself 

**- How I did it**

Move some logic from `tools/alpine/Makefile` to `tools/alpine/Dockerfile`, and then change `Makefile` to consume it. 

**- How to verify it**

Run it.

linuxkit/alpine carries its own calculated hash tag with it.

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -93,6 +93,8 @@ RUN set -e && \
 
 FROM alpine:3.13
 
+ARG TARGETARCH
+
 COPY --from=mirror /etc/apk/repositories /etc/apk/repositories
 COPY --from=mirror /etc/apk/repositories.upstream /etc/apk/repositories.upstream
 COPY --from=mirror /etc/apk/keys /etc/apk/keys/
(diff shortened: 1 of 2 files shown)


tools/alpine/Dockerfile
```diff
@@ -93,6 +93,8 @@ RUN set -e && \
 
 FROM alpine:3.13
 
+ARG TARGETARCH
```
dave-tucker, line 96: What populates `TARGETARCH`?
I also wonder whether we should rename this to be `IMAGE_SUFFIX` as it better described it's purpose, especially as the arch != what you'd expect from `uname -m`

dave-tucker (changes requested):

tools/alpine/Dockerfile
dave-tucker: Ugh I should read commit messages better. I see where this is populated now :joy:

dave-tucker (approved):

deitch:
Thanks @dave-tucker

rn:
I personally have a strong preference on having the details in the git commit message rather than the PR summary. I can do `git show <hash>` and get the explanation. but here i only get the summary line and the details are on some web page. Might be a bit old fashioned...

deitch:
I don't understand Rolf. To what do you refer, the git commit message vs PR?

rn:
```
git show 1e99ba3dcee3
commit 1e99ba3dcee3cfffe53e1a0f13cdf4a61bd4874b
Author: Avi Deitcher <avi@deitcher.net>
Date:   Sun May 2 19:44:39 2021 +0300

    alpine carry its own hash

diff --git a/tools/alpine/Dockerfile b/tools/alpine/Dockerfile
```
has none of the details you mentions above, like
> Also, docker build sets TARGETARCH for us, which gets rid of a few lines in the Makefile

PR #3651, opened 2021-05-03
tag kernel with builder version; simplify Makefile

deitch:

**- What I did**

A few things in `kernel/`:

* some simplifications in `Makefile` to eliminate redundant code (mainly so it was easier for me to grasp)
* removed all references to `linuxkit/alpine:<hash>` in all of the kernel Dockerfiles, replaced with an `ARG`, which is passed by `docker build` (which already had some `--build-arg`) from the `Makefile`
* save the name of the builder image in the final kernel image as `/kernel-builder`
* save the name of the builder image in the final kernel image as a label `org.mobyproject.linuxkit.kernel.builder ${BUILDER}`

The purpose is to have the image carry its matched builder in the appropriate location, so you can get it, whether you are already inside the container (from the file) or outside (from the docker label).

**- How I did it**

Removed all references to linuxkit/alpine from the kernel Dockerfiles, replaced them with `ARG BUILDER`, and then passed it from Makefile.

Added two lines at the end, one a `RUN echo` and one a `LABEL`.

**- How to verify it**

We should build the kernels, some suggestion from @rn would be appreciated

Include kernel builder image version with kernel

### DIFF
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -1,4 +1,6 @@
-FROM linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba AS kernel-build
+ARG BUILD_IMAGE
+FROM ${BUILD_IMAGE} AS kernel-build
+ARG BUILD_IMAGE
 RUN apk add \
     argp-standalone \
     automake \
@@ -209,7 +211,8 @@ RUN DVER=$(basename $(find /tmp/kernel-modules/lib/modules/ -mindepth 1 -maxdept
     ( cd /tmp && tar cf /out/kernel-dev.tar usr/src )
 
 RUN printf "KERNEL_SOURCE=${KERNEL_SOURCE}\n" > /out/kernel-source-info
-
+RUN printf "${BUILD_IMAGE}" > /out/kernel-builder
+LABEL org.mobyproject.linuxkit.kernel.buildimage ${BUILD_IMAGE}
 
 FROM scratch
 ENTRYPOINT []
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -18,6 +18,7 @@ IMAGE:=kernel
 IMAGE_BCC:=kernel-bcc
 IMAGE_PERF:=kernel-perf
 IMAGE_ZFS:=zfs-kmod
+IMAGE_BUILDER=linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba
 
 # You can specify an extra options for the Makefile. This will:
 # - append a config$(EXTRA) to the kernel config for your kernel/arch
@@ -73,6 +74,11 @@ KERNEL_VERSIONS=
 build:
 push:
 
+.PHONY: notdirty
+notdirty:
+	@if [ x"$(DIRTY)" !=  x ]; then echo "Your repository is not clean. Will not push image"; exit 1; fi
+
+
 # A template for defining kernel build
 # Arguments:
 # $1: Full kernel version, e.g., 4.9.22
@@ -99,8 +105,10 @@ build_$(2)$(3)$(4): Dockerfile Makefile $(wildcard patches-$(2)/*) $(wildcard co
 			--build-arg KERNEL_SERIES=$(2) \
 			--build-arg EXTRA=$(3) \
 			--build-arg DEBUG=$(4) \
+			--build-arg BUILD_IMAGE=$(IMAGE_BUILDER) \
 			$(LABELS) \
 			--no-cache -t $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) .
+
 
 forcebuild_$(2)$(3)$(4): Dockerfile Makefile $(wildcard patches-$(2)/*) $(wildcard config-$(2)*) config-dbg
 	docker build \
@@ -108,11 +116,11 @@ forcebuild_$(2)$(3)$(4): Dockerfile Makefile $(wildcard patches-$(2)/*) $(wildca
 		--build-arg KERNEL_SERIES=$(2) \
 		--build-arg EXTRA=$(3) \
 		--build-arg DEBUG=$(4) \
+		--build-arg BUILD_IMAGE=$(IMAGE_BUILDER) \
 		$(LABELS) \
 		--no-cache -t $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) .
 
-push_$(2)$(3)$(4): build_$(2)$(3)$(4)
-	@if [ x"$(DIRTY)" !=  x ]; then echo "Your repository is not clean. Will not push image"; exit 1; fi
+push_$(2)$(3)$(4): notdirty build_$(2)$(3)$(4)
 	docker pull $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) || \
 		(docker push $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) && \
 		 docker tag $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) $(ORG)/$(IMAGE):$(1)$(3)$(4)$(SUFFIX) && \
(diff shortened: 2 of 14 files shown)


kernel/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM linuxkit/alpine:e2391e0b164c57db9f6c4ae110ee84f766edc430 AS kernel-build
+ARG BUILDER
```
rn, line 1: `BUILDER` is an odd name. Maybe `BUILD_BASE` or `BUILD_IMAGE` is more descriptive.

deitch: I like `BUILD_IMAGE`. Done.

deitch: Updated

deitch:
I added the tagging. When building the kernel, it tags the builder image with a `linuxkit/kernel:...-builder` tag.

For example, if we build `linuxkit/kernel:5.4.113` with `linuxkit/alpine:abcd123`, then it also tags `linuxkit/alpine:abcd123` with `linuxkit-kernel:5.4.113-builder`.

deitch:
I rebased on your latest changes @rn and ensure the (now single) place in the Makefile that has the hash is per your latest commit.

Can we get this reviewed and in?

kernel/Makefile
```diff
@@ -93,48 +99,42 @@ KERNEL_VERSIONS+=$(1)
...
 build_$(2)$(3)$(4): Dockerfile Makefile $(wildcard patches-$(2)/*) $(wildcard config-$(2)*) config-dbg
-	docker pull $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) || \
-		docker build \
-			--build-arg KERNEL_VERSION=$(1) \
-			--build-arg KERNEL_SERIES=$(2) \
-			--build-arg EXTRA=$(3) \
-			--build-arg DEBUG=$(4) \
-			$(LABELS) \
-			--no-cache -t $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) .
+	docker pull $(ORG)/$(IMAGE):$(1)$(3)$(4)-$(TAG)$(SUFFIX) || $(MAKE) forcebuild_$(2)$(2)$(4)
```
rn, line 102: I'm a fan of this pattern of calling make from a make file. It looks very much like an antipattern.

deitch: I assume you meant that you are _not_ a fan of this pattern?

It isn't an anit-pattern. It actually is quite common, which is why there is a `$(MAKE)` special variable. As one example, it appears ~45 times in the root `Makefile` for the linux kernel [here](https://github.com/torvalds/linux/blob/master/Makefile)

deitch: Any other comments on it?

deitch: By the way, even if `$(MAKE)` _were_ an anti-pattern, I would encourage it here. Small anti-patterns are better than wholesale code repetition. But in any case, it isn't an anti-pattern; it is the correct way to call other make targets when they are conditional dependencies.

rn: I'm not a fan. most calls to `$(MAKE)` in the Linux kernel top level Make file are with either `-f` to call with another Makefile or `-C` to call `make` in another directory. This is definitely different to call `make` on the same `Makefile` again with another target. That is the part I don't like. The more native `make` way would be to create (temporary) files or just live with a little repetition (which is what I chose). These things don't change much so it's ok

deitch: I disagree, but I honestly do not care enough about this. Most of the time, I don't touch the kernel or its Makefile. I can revert that part and we can get the rest in.

deitch: Updated; just the other simplifications remain. Of course, also rebased to latest `master`. Take a look and let us get it through @rn

rn (commented):
Thanks, i like this a lot better. What's missing though, is: "How would I use this?" and "Does it actually work?". As I understand it, the primary motivation for this is to make external kernel module builds easier/simpler. So what would the kernel module compile tests we already have look like? Can they be updated as part of this PR?

deitch:
> Thanks, i like this a lot better

😄 

> As I understand it, the primary motivation for this is to make external kernel module builds easier/simpler.

Yes, although some of it is just simple cleanup, like having the builder image in a single place in the `Makefile`, or reducing a bit of complexity.

> So what would the kernel module compile tests we already have look like? Can they be updated as part of this PR?

You know, that is a good point. We definitely could. I am going to update them in the next few minutes as part of this PR.

I do have a concern though. This PR only updates _future_ kernel builds, but _existing_ ones don't have any of the builder references. I think we can fix that by pushing out tags manually for now? Thoughts?

rn:
I'm fine with just having this for the current and future builds

deitch:
@rn I added two changes in the tests:

* explicitly using the builder tag to make the modules, rather than the `linuxkit/alpine:<hash>`, which exercises it
* a test for tags, to make sure that they are all there and in place and right

I expect the latter to fail, and I only added for 5.4.x, but you can take a look and let's see if we agree on the direction.

test/cases/020_kernel/211_tags_5.4.x/test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+# SUMMARY: Test build and insertion of kernel modules
```
rn, line 2: Looks like a copy/paste issue

test/cases/020_kernel/211_tags_5.4.x/test.sh
```diff
@@ -0,0 +1,57 @@
...
+clean_up() {
+	docker rm ${ctrid}
+	/bin/rm -f ${BUILDERFILE}
+}
+trap clean_up EXIT
+
+# check the kernel images for tags, labels, files
+KERNEL=linuxkit/kernel:5.4.116
+BUILDER=${KERNEL}-builder
+BUILDERFILE=/tmp/kernel-builder-${KERNEL}
```
rn, line 23: I would do this in the local directory and add a `.gitignore`, but `/tmp` works too

rn (commented):
Yeah, I think that works.

test/cases/020_kernel/211_tags_5.4.x/test.sh
deitch: It is. Fixed.

deitch:
I updated per the comments, and added "tags tests" for the other 3 series. Funny, the test files are identical, I think I will move it all into `test/cases/_lib/`.

There are a few remaining questions and issues.

* the kmod tests pass for 5.4.x but not for the others, because I didn't actually tag the other images quite yet.
* the tagging tests fail, because the tags exist, but the labels do not and the file inside the image certainly does not.

I will add the other tags, to get those tests to pass; not a big deal (but they do have to be multi-arch index).

For the tagging tests, I don't want to rebuild the kernel images to add the files, and adding `LABEL` does mean modifying the config, which means making some changes.

I see two options:

First option:
As a one-time, create a local file name `kernel-builder` with contents of `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`, then create a `Dockerfile` as follows:

```Dockerfile
FROM linuxkit/kernel:5.4.39 
COPY kernel-builder /etc/kernel-builder
LABEL org.mobyproject.linuxkit.kernel.buildimage linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba
```

Then do `docker pull` and `docker build`, then tag and push to override, for each of the 3 arches, then create the multi-arch index.

Future ones won't have an issue, since this PR makes the above part of the pipeline.

Second option: disable the tags tests until the pipeline has run at least once with new kernels. I don't like that, as we are likely to forget it.

Thoughts @rn ?

deitch:
OK, created the tags, and extracted all of the common tag tests into a single library. Can you take a look at it @rn ?

test/cases/020_kernel/tags.sh
```diff
@@ -0,0 +1,65 @@
...
+
+clean_up() {
+	docker rm ${ctrid}
+	/bin/rm -f ${BUILDERFILE}
+}
+trap clean_up EXIT
+
+# check the kernel images for tags, labels, files
+BUILDER=${KERNEL}-builder
+BUILDERFILE=/tmp/kernel-builder-${KERNEL}
```
rn, line 31: CI is failing because KERNEL is of the form `linuxkit/kernel:5.11.18` so this assumes a subdirectory of `/tmp/kernel-builder-linuxkit`. you might want to extract just the version number with `KERNEL_VERSION=$(echo $KERNEL | cut -d: -f 2)`, check that it's non-zero in length...or whatever your favourite string processing utility is ...

rn (commented):
Looks pretty good to me. Thanks for sticking this out. CI is brafing on the thing I commented on. Should be an easy fix

test/cases/020_kernel/tags.sh
deitch: Nice one. Thanks

deitch: I am just going to make it a random file and be done with it.

deitch:
I updated it to just use a random pid file. It seems to make more sense that way.

I think CI still will complain because the existing images do not have the labels or the file, although I already created the tag. I don't mind retrofitting the label and file as a one time for each of the 3 latest kernels. See my suggestion [here](https://github.com/linuxkit/linuxkit/pull/3651#issuecomment-846576403), let me know what you think.

rn:
can't you just force push kernel images for all the kernels with the new labels/tags

deitch:
> can't you just force push kernel images for all the kernels with the new labels/tags

I could, but I need to add the new labels. I don't think there is a `docker label` command, hence the dockerfile above. I will proceed with those.

Tags are done already.

deitch:
OK, did it. Going to kick off CI again and hope that it works. I added the label and file and builder tags to the current latest kernels:

* `5.4.116`
* `5.10.34`
* `5.11.18`

deitch:
Yes! Kernel tests pass! @rn  take another pass at this, and maybe we can get it in?

Issue #3652, opened 2021-05-03
bcc fails to compile with alpine 3.13


rn:
With the soon to be updated kernel build to the `linuxkit/alpine` based on 3.13, `lixnukti/kernel-bcc` fails to compile.

After a few warnings about `size_t` and `const counter_type` size mismatches we get:

```
/build/bcc/src/cc/frontends/b/lexer.h:61:38: warning: comparison of integer expressions of different signedness: 'size_t' {aka 'long unsigned int'} and 'const counter_type' {aka 'const int'} [-Wsign-compare]
   61 |       } else if (i > begin.line && i == end.line) {
      |                                    ~~^~~~~~~~~~~
/build/bcc/src/cc/frontends/b/codegen_llvm.cc: In member function 'virtual ebpf::StatusTuple ebpf::cc::CodegenLLVM::visit_string_expr_node(ebpf::cc::StringExprNode*)':
/build/bcc/src/cc/frontends/b/codegen_llvm.cc:429:55: warning: 'llvm::CallInst* llvm::IRBuilderBase::CreateMemCpy(llvm::Value*, unsigned int, llvm::Value*, unsigned int, uint64_t, bool, llvm::MDNode*, llvm::MDNode*, llvm::MDNode*, llvm::MDNode*)' is deprecated [-Wdeprecated-declarations
]
  429 |   B.CreateMemCpy(ptr, 1, global, 1, n->val_.size() + 1);
      |                                                       ^
In file included from /usr/lib/llvm10/include/llvm/Support/AlignOf.h:16,
                 from /usr/lib/llvm10/include/llvm/ADT/SmallVector.h:17,
                 from /usr/lib/llvm10/include/llvm/ADT/Twine.h:12,
                 from /usr/lib/llvm10/include/llvm/IR/BasicBlock.h:17,
                 from /build/bcc/src/cc/frontends/b/codegen_llvm.cc:21:
/usr/lib/llvm10/include/llvm/IR/IRBuilder.h:523:17: note: declared here
  523 |       CallInst *CreateMemCpy(Value *Dst, unsigned DstAlign, Value *Src,
      |                 ^~~~~~~~~~~~
/usr/lib/llvm10/include/llvm/Support/Compiler.h:305:3: note: in definition of macro 'LLVM_ATTRIBUTE_DEPRECATED'
  305 |   decl __attribute__((deprecated))
      |   ^~~~
[...]

Issue #3644, opened 2021-04-29
question: linuxkit/kernel-perf for 5.10.25?


srenatus:
Hi!

Looking through the tags on https://hub.docker.com/r/linuxkit/kernel-perf/tags/?page=1&ordering=last_updated, I can't find one for the 5.10.25 kernel, or any 5.10.x.

Am I missing something? 🤔 

Also, what I'm really after is a not-too-difficult way to run `perf` in a docker container running on docker-for-mac, which currently reports using the kernel `5.10.25-linuxkit`. Am I wrong assuming that I'd need a matching `kernel-perf` image of that version?

Thanks for your time
Stephan

srenatus:
Sorry, thinking about this, I should address docker, not linuxkit. And look at that, someone already has: https://github.com/docker/for-mac/issues/5500

Excuse me for the noise 😅

PR #3654, opened 2021-05-03
Update kernels

### LINKED ISSUE #3620: Deprecated some kernel versions
We are currently supporting:
- 4.14.x for x86_64 EOL Jan 2024
- 4.19.x for x86_64 EOL Dec 2024 (Currently used for Docker Desktop)
- 5.4.x for x86_64, arm64, s390x EOL Dec, 2025
- 5.6.x x86_64, arm64, s390x EOLed

There are already patches out for 5.10.x (EOL Dec 2022) and the latest stable is 5.11 with 5.12 in progress.

Apart from removing the already EOLed 5.6.x, I suggest to also remove support for 4.14.x and 4.19.x. since 5.4.x is also on long long term support.

@justincormack @djs55 any objections from the Docker Desktop side?

rn:
- Make 5.10.x the default kernel
- Remove support for 4.19.x (we have 5.4.x and 5.10.x as LTS kernels already) (resolves #3620)
- Update kernels to 5.11.18/5.10.34/5.4.116
- Update linuxkit/alpine for kernel builds
- Fix `perf` builds and support for 5.10.x (see #3644)
- Disable `bcc` builds. They are broken again (see #3652)
 
blue-ringed-octopus


deitch (approved):
I see no issues. I will gave to rebase #3651 after this, but that is fine.

PR #3658, opened 2021-05-05
Build or push multiple packages at once

deitch:

**- What I did**

* Enabled `lkt pkg build ./dir/one ./dir/two ... ./dir/n` and `lkt pkg push ./dir/one ./dir/two ... ./dir/n`  - multiple builds in sequence
* Refactored the massive amount of duplication between `pkgBuild()` and `pkgPush()` into a single func. `Push()` always was just a superset of `build()`

None of this is parallelized using goroutines, though we could do that. This just gets the basic serial builds and semantics in place.

**- How I did it**

Refactored and added

**- How to verify it**

Run it. Also CI. We like CI.

Build and push multiple packages with a single command.


rn:
Not generally opposed to this change, but like to understand the use case better before adding more functionality. It would be yet another feature for which we should have tests (like we also don't have tests for the `--platforms`).

For automation something like `make` will do the same thing (and do parallel builds)

deitch:
Agreed. It would be nice to have tests. Actually, there's quite a bit in the lkt binary that doesn't have tests. 

Use case: building or pushing multiple packages. I agree that make or scripting can provide automation in some circumstances. I have found that when I don't have the right script or makefile (eg working elsewhere than in linuxkit/linuxkit) I need to hand craft a loop, which is silly and wasteful. 

Even when I do work in linuxkit/linuxkit, and need to build or push a lot of packages, I have to do some odd, and fragile, copy m-pasting to get the right combination over and over. 

Continuing that last point, this may be more here, but it would allow us to simplify lots of make targets.

src/cmd/linuxkit/pkg_build.go
```diff
@@ -102,9 +130,38 @@ func pkgBuild(args []string) {
 		os.Exit(1)
 	}
 	opts = append(opts, pkglib.WithBuildBuilders(buildersMap))
-	if err := p.Build(opts...); err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
-		os.Exit(1)
+
+	for _, p := range pkgs {
+		// unless overridden, platforms are specific to a package, so this needs to be inside the for loop
+		if len(plats) == 0 {
```
dave-tucker, line 136: The second time this loops around, won't `plats` now be > 0 and therefore the comment above is no longer true? Effectively if not specified, we'll always build packages to the arches supported in the first package.

src/cmd/linuxkit/pkg_build.go
```diff
@@ -102,9 +130,38 @@ func pkgBuild(args []string) {
...
+			for _, a := range p.Arches() {
+				if _, ok := skipPlatformsMap[a]; ok {
+					continue
+				}
+				plats = append(plats, imagespec.Platform{OS: "linux", Architecture: a})
+			}
+		}
+		opts = append(opts, pkglib.WithBuildPlatforms(plats...))
```
dave-tucker, line 144: Same here with opts as the variable exists outside this for loop.

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -80,175 +80,189 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+	argDisableCache := fs.Bool("disable-cache", piBase.DisableCache, "Disable build cache")
+	argEnableCache := fs.Bool("enable-cache", !piBase.DisableCache, "Enable build cache")
+	argNoNetwork := fs.Bool("nonetwork", !piBase.Network, "Disallow network use during build")
+	argNetwork := fs.Bool("network", piBase.Network, "Allow network use during build")
 
-	argOrg := fs.String("org", pi.Org, "Override the hub org")
+	argOrg := fs.String("org", piBase.Org, "Override the hub org")
 
-	// Other arguments
+	// Other argumentss
```
dave-tucker, line 92: It was better before :joy:

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -80,175 +80,189 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) (Pkg, error) {
...
+			pi.Org = os.Getenv("USER")
+			if hash == "" {
+				hash = "dev"
+			}
 		}
 
-		srcHashes += h
-		sources = append(sources, pkgSource{src: srcPath, dst: dstPath})
-	}
+		// Go's flag package provides no way to see if a flag was set
```
dave-tucker, line 169: I do wonder whether it would be worthwhile using another flags library - maybe `spf13/viper`

dave-tucker (changes requested):

src/cmd/linuxkit/pkg_build.go
```diff
@@ -61,16 +96,10 @@ func pkgBuild(args []string) {
...
 	var plats []imagespec.Platform
-	if *platforms == "" {
-		for _, a := range p.Arches() {
-			if _, ok := skipPlatformsMap[a]; ok {
-				continue
-			}
-			plats = append(plats, imagespec.Platform{OS: "linux", Architecture: a})
-		}
-	} else {
+	if *platforms != "" {
```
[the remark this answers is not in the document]
deitch: Sure. I tried it, it looks slightly better. Happy to do so.

src/cmd/linuxkit/pkglib/pkglib.go
deitch: Not if you speak Parseltongue!

src/cmd/linuxkit/pkglib/pkglib.go
deitch: I think it probably would be, but I am unwilling to spend the time doing that surgery now.

src/cmd/linuxkit/pkg_build.go
deitch: This is a really good point. I got that with the `fiBase`, but not this one. Fixing.

src/cmd/linuxkit/pkg_build.go
deitch: Ditto.

src/cmd/linuxkit/pkg_build.go
deitch: Fixed

deitch:
@dave-tucker I think I addressed your questions

dave-tucker (approved):
Code LGTM. To echo @rn's earlier comments, I'll state for the record that I too would like to see some better tests for this. Unfortunately that's not easy given the current state of the code, and would probably mean more `rtf` style tests. I've opened https://github.com/linuxkit/linuxkit/issues/3660 to track looking at alternatives

deitch:
With that approval, I am going to merge it in. We should keep on top of #3660 to handle flags and test

PR #3659, opened 2021-05-05
fix: make command

### LINKED ISSUE #3447: targets/{justin|riyaz} metadata is nearing expiry, you should re-sign the role metadata
**Description**

When building linuxkit images I get:

```
WARN[0052] targets/justin metadata is nearing expiry, you should re-sign the role metadata
WARN[0052] targets/riyaz metadata is nearing expiry, you should re-sign the role metadata
```
For at least:

* linuxkit/sysfs:v0.7
* linuxkit/openntpd:v0.7
* linuxkit/node_exporter:v0.7

Since v0.7 is the latest major in most cases, it seems they might all soon trigger this warning.

linuxkit build -format kernel+initrd build.yml

WARN[0052] targets/{justin|riyaz} metadata is nearing expiry, you should re-sign the role metadata

No warnings.

edvinerikson:
fixes #3447 

**- What I did**

Updated the install process when building from source

**- How I did it**

Moving `src/cmd/linuxkit/bin` to `bin/` when building linuxkit

**- How to verify it**

run `make` and `make install` and verify that `linuxkit` command exists

Updated `make` to move `src/cmd/linuxkit/bin` to `bin/`


rn:
Thanks for pointing out the new location of the `linuxkit` binary. I'd much rather address this my moving the compiled binary to `./bin` so that all binaries build are there.

edvinerikson:
Is this more in line with what you want?

dave-tucker (changes requested):
@edvinerikson thanks for the contribution.

the fix should be as follows:
```diff
diff --git a/Makefile b/Makefile
index 3c47141dd..56c022538 100644
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,11 @@ VERSION="v0.8+"
 GO_COMPILE=linuxkit/go-compile:7b1f5a37d2a93cd4a9aa2a87db264d8145944006
 
 ifeq ($(OS),Windows_NT)
+LINUXKIT?=$(CURDIR)/bin/linuxkit.exe
 RTF?=bin/rtf.exe
 GOOS?=windows
 else
+LINUXKIT?=$(CURDIR)/bin/linuxkit
 RTF?=bin/rtf
 GOOS?=$(shell uname -s | tr '[:upper:]' '[:lower:]')
 endif
@@ -18,10 +20,9 @@ CROSS+=-e GOARCH=$(GOARCH)
 endif
 
 PREFIX?=/usr/local
-
 LOCAL_TARGET?=$(CURDIR)/bin/linuxkit
 
-export VERSION GO_COMPILE GOOS GOARCH LOCAL_TARGET
+export VERSION GO_COMPILE GOOS GOARCH LINUXKIT LOCAL_TARGET
 
 .DELETE_ON_ERROR:
 
@@ -52,7 +53,7 @@ tmp_mt_bin.tar: Makefile
        docker run --rm --log-driver=none -e http_proxy=$(http_proxy) -e https_proxy=$(https_proxy) $(CROSS) $(GO_COMPILE) --clone-path github.com/estesp/manifest-tool --clone $(MT_REPO) --commit $(MT_COMMIT) --package github.com/estesp/manifest-tool --ldflags "-X main.gitCommit=$(MT_COMMIT)" -o bin/manifest-tool > $@
 
 .PHONY: linuxkit
-linuxkit:
+linuxkit: bin
        make -C ./src/cmd/linuxkit
 
 .PHONY: test-cross
 ```
 
 That will ensure that when the `linuxkit` target is run from the top-level makefile, that `LINUXKIT` is set to `$(CURDIR)/linuxkit` (or + `.exe` for windows).
 This is pretty much the same mechanism we use for the `local` targets to put their binaries in `./bin` - via the `LOCAL_TARGET` variable.
 
 I would also revert the README changes as from the diff it just looks like you've added/removed some newlines/spaces which doesn't alter the way it renders. If it's required, feel free to put it in a separate commit along with some context in the commit message as to what it's fixing.

File: src/cmd/linuxkit/Makefile [hunk 1]
```diff
@@ -26,12 +26,12 @@ all: default
...
 LINUXKIT_DEPS=$(wildcard *.go) $(wildcard */*.go) Makefile
 $(LINUXKIT): tmp_linuxkit_bin.tar | bin
-	tar xf $<
+	tar xfP $<
 	rm $<
 	touch $@
 
 tmp_linuxkit_bin.tar: $(LINUXKIT_DEPS)
-	tar cf - -C . . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)" -o $(LINUXKIT) > $@
+	tar cf - -C . . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)" -o bin/linuxkit > $@
```
edvinerikson, line 34: Not sure if this is the correct way. I struggled with the extraction where it treated an absolute path (CURDIR) as a relative one.

dave-tucker (changes requested):
I tested this and it didn't work. The binary in the top level dir was empty, created by the `touch` command, and the real binary was still in `src/cmd/linuxkit/bin`

This should fix it (and fixes the way we build `rtf` to be the same too:

```diff
diff --git a/Makefile b/Makefile
index dfd2a40a4..9cd91ec7d 100644
--- a/Makefile
+++ b/Makefile
@@ -35,12 +35,12 @@ RTF_COMMIT=2351267f358ce6621c0c0d9a069f361268dba5fc
 RTF_CMD=github.com/linuxkit/rtf/cmd
 RTF_VERSION=0.0
 $(RTF): tmp_rtf_bin.tar | bin
-       tar xf $<
+       tar -C $(dir $(RTF)) -xf $<
        rm $<
        touch $@
 
 tmp_rtf_bin.tar: Makefile
-       docker run --rm --log-driver=none -e http_proxy=$(http_proxy) -e https_proxy=$(https_proxy) $(CROSS) $(GO_COMPILE) --clone-path github.com/linuxkit/rtf --clone https://github.com/linuxkit/rtf.git --commit $(RTF_COMMIT) --package github.com/linuxkit/rtf --ldflags "-X $(RTF_CMD).GitCommit=$(RTF_COMMIT) -X $(RTF_CMD).Version=$(RTF_VERSION)" -o $(RTF) > $@
+       docker run --rm --log-driver=none -e http_proxy=$(http_proxy) -e https_proxy=$(https_proxy) $(CROSS) $(GO_COMPILE) --clone-path github.com/linuxkit/rtf --clone https://github.com/linuxkit/rtf.git --commit $(RTF_COMMIT) --package github.com/linuxkit/rtf --ldflags "-X $(RTF_CMD).GitCommit=$(RTF_COMMIT) -X $(RTF_CMD).Version=$(RTF_VERSION)" -o rtf > $@
 
 # Manifest tool for multi-arch images
 MT_COMMIT=bfbd11963b8e0eb5f6e400afaebeaf39820b4e90
diff --git a/src/cmd/linuxkit/Makefile b/src/cmd/linuxkit/Makefile
index cbeeb96fa..fd4c0e4f9 100644
--- a/src/cmd/linuxkit/Makefile
+++ b/src/cmd/linuxkit/Makefile
@@ -26,12 +26,12 @@ all: default
 
 LINUXKIT_DEPS=$(wildcard *.go) $(wildcard */*.go) Makefile
 $(LINUXKIT): tmp_linuxkit_bin.tar | bin
-       tar xf $<
+       tar -C $(dir $(LINUXKIT)) -xf $<
        rm $<
        touch $@
 
 tmp_linuxkit_bin.tar: $(LINUXKIT_DEPS)
-       tar cf - -C . . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)" -o bin/linuxkit > $@
+       tar cf - -C . . | docker run --rm --net=none --log-driver=none -i $(CROSS) $(GO_COMPILE) --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=$(GIT_COMMIT) -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version=$(VERSION)" -o linuxkit > $@
 
 .PHONY: test-cross
 test-cross:
 ```
 
 This will ensure that the the go-compile image outputs a binary a single binary in the tar archive.  We then change to the `$(dir $(THE_BINARY))` when we extract it.

**Review by dave-tucker — DISMISSED** [review]
LGTM, but one of the commits is mine :laughing: 
@deitch as we're trading reviews if you wouldn't mind taking a look I'd appreciate it

File: src/cmd/linuxkit/Makefile [hunk 2]
```diff
@@ -6,6 +6,7 @@ export GO_FLAGS=-mod=vendor
 ifeq ($(OS),Windows_NT)
 LINUXKIT?=bin/linuxkit.exe
 GOOS?=windows
+SUFFIX?=.exe
```
deitch, line 9: I do not understand this. As `LINUXKIT?=bin/linuxkit.exe`, what do we need the `SUFFIX` for? The change to the last line to make it `-o $(LINUXKIT)$(SUFFIX)` would make it `linuxkit.exe.exe`, wouldn't it?

deitch (changes requested):
It mostly looks fine to me - I like passing the intended target to `src/cmd/Makefile` with `LINUXKIT` env var - but the `SUFFIX` thing looks duplicate. I assume this is just for windows, but don't we already set it?

src/cmd/linuxkit/Makefile
dave-tucker: `go build -o linuxkit$(SUFFIX)` will produce `linuxkit` on everything except windows where it will be `linuxkit.exe`
Therefore when we `tar -C bin -xf tmp.tar` on linux we create `bin/linuxkit` and on windows we create `bin/linuxkit.exe`

I just realised I can achieve the same result with `$(notdir $(LINUXKIT))` so I'll do that instead to save the extra variable

dave-tucker:
@deitch comment addressed. I use `$(notdir $(LINUXKIT))` to save using another variable

deitch (approved):

PR #3663, opened 2021-05-07
Include open-vm-tools packages

### LINKED ISSUE #3662: open-vm-tools does not report ip information back to vcenter
**Description**
Vcenter shows that open-vm-tools is running, but ip address and hostname is not being reported back. It seems that in alpine 3.12 the open-vm-tools package got split into sub packages. 

The open-vm-tools-guestinfo package should be supported in the alpine base image I think.

Run iso image including open-vm-tools container in vcenter.

No ip information available

Ip information visible in vcenter

I got it working locally by building my own version of open-vm-tools image that installs this package. However it would be nice to get it fixed upstream as well.

edvinerikson:
closes #3662 

In alpine version 3.12, the open-vm-tools package got split into new
smaller sub-packages. The implication of this is that features such as
reporting of hostname and ip address to vCenter stopped working.

**- What I did**
Installed the open-vm-tools packages in the open-vm-tools image. 

Added the following packages to the open-vm-tools image:
* open-vm-tools-guestinfo
* open-vm-tools-deploypkg
* open-vm-tools-static
* open-vm-tools-vmbackup

### DIFF
--- a/pkg/open-vm-tools/Dockerfile
+++ b/pkg/open-vm-tools/Dockerfile
@@ -1,9 +1,13 @@
-FROM linuxkit/alpine:0c069d0fd7defddb6e03925fcd4915407db0c9e1 AS mirror
+FROM linuxkit/alpine:b3028815dbf119404ab68a25acfcf84c585ab432 AS mirror
 RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out \
     alpine-baselayout \
     busybox \
-    open-vm-tools 
+    open-vm-tools \
+    open-vm-tools-guestinfo \
+    open-vm-tools-deploypkg \
+    open-vm-tools-static \
+    open-vm-tools-vmbackup 
 
 # Remove apk residuals
 RUN rm -rf /out/etc/apk /out/lib/apk /out/var/cache


rn:
@edvinerikson adding new packages to `tools/alpine` is not quite as straight forward as we need to push the new base image for all three supported architectures. I've done that part in https://github.com/linuxkit/linuxkit/pull/3664

In order to update `pkg/open-vm-tools` you also need to update:
```
FROM linuxkit/alpine:0c069d0fd7defddb6e03925fcd4915407db0c9e1 AS mirror
```
to newer version of the base image, which, after the above PR is merge, is:
```
linuxkit/alpine:b3028815dbf119404ab68a25acfcf84c585ab432
```

You can then test the new package by using `linuxkit pkg build ./pkg/open-vm-tools`

edvinerikson:
Thanks for explaining it. I couldn't figure out how to build a new base image, it always failed due to some missing image manifest when I tested locally. I did however managed to build a new open-vm-tools image locally by switching the base alpine to use alpine:3.13 from docker hub. This worked great to verify that the new packages work as they should.

I did however notice one thing, which I don't know if it is related to LinuxKit. Somehow the VM loses its ipv4 address after a while, but ipv6 is kept.

rn:
loosing ipv4 address might be related to how you run dhcpd. Most sample images just it run in one-shot mode as an on boot thing and if the lease expires it might change. If you add a dhcpd service does it still loose the connectivity?

PR #3664, opened 2021-05-09
Include open-vm-tools packages

rn:
In alpine version 3.12, the open-vm-tools package got split into new
smaller sub-packages. The implication of this is that features such as
reporting of hostname and ip address to vCenter stopped working.

relates to #3662
part of #3663

chimp

### DIFF
--- a/tools/alpine/versions.aarch64
+++ b/tools/alpine/versions.aarch64
@@ -1,4 +1,4 @@
-# linuxkit/alpine:3f3f1159e07077cdcf1544f536cfee5052fffa89-arm64
+# linuxkit/alpine:4967a81f63c6fab8cfa686ba051ebe8e0f0fd8ee-
 # automatically generated list of installed packages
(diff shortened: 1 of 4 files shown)


tools/alpine/versions.aarch64
```diff
@@ -1,4 +1,4 @@
-# linuxkit/alpine:3f3f1159e07077cdcf1544f536cfee5052fffa89-arm64
+# linuxkit/alpine:4967a81f63c6fab8cfa686ba051ebe8e0f0fd8ee-
```
djs55, line 1: Forgive my ignorance :) is this a generated list of package versions in the image? Is it expected that the `arm64` suffix is removed?

rn: Yeah, this has now been fixed with https://github.com/linuxkit/linuxkit/pull/3665. I'll merge this anyway, since it's mostly cosmetic.

PR #3665, opened 2021-05-10
correctly use target-arch

deitch:

**- What I did**

In PR #3650 we moved the logic for calculating the hash for an arch-specific alpine image from inside `tools/alpine/Makefile` to inside `tools/alpine/Dockerfile` and then consumed it. This was a good idea, but the implementation had two issues:

* The actual usage of `TARGETARCH` did so in a way that left it as a literal `${TARGETARCH}` rather than interpolated correctly by the Dockerfile
* The file was named `/etc/alpine-hash`, which implied that it was the generic, multi-arch index, when really it is just the arch-specific one

This PR corrects both of the above. It ensures that `TARGETARCH` is interpolated by the Dockerfile, and it saves the file as `/etc/alpine-hash-arch`.

This does not resolve the question of having an alpine image carry the multi-arch tag, likely in `/etc/alpine-hash`, as that is a more complicated issue, and something we never had before (one step up). Instead, it just fixes a regression from #3650 so that this all works correctly.

If I can figure out a sane way to get `/etc/alpine-hash` before this PR goes in, I will update this PR, else it will be the subject of a separate one (potentially).

**- How I did it**

Modified `tools/alpine/Makefile` and `tools/alpine./Dockerfile`

**- How to verify it**

CI is your friend. Also, you can do what I did, which is to build `linuxkit/alpine` manually on different platforms and check the content of `/etc/alpine-hash-arch`.

Fix alpine hash file generation

### DIFF
--- a/tools/alpine/Dockerfile
+++ b/tools/alpine/Dockerfile
@@ -106,6 +106,6 @@ COPY --from=mirror /iucode_tool /usr/bin/
 
 RUN apk update && apk upgrade -a
 
-RUN echo Dockerfile /lib/apk/db/installed $(find /mirror -name '*.apk' -type f) $(find /go/bin -type f) | xargs cat | sha1sum | sed 's/ .*//' | sed 's/$/-${TARGETARCH}/' > /etc/alpine-hash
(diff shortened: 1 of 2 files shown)


dave-tucker (approved):
LGTM

deitch:
Thanks. Glad to get this regression out of the way.

PR #3666, opened 2021-05-10
calculate manifest using same method as lkt pkg push via git ls-tree

deitch:

**- What I did**

Changed the calculation of `linuxkit/alpine` tag from "let's just use whatever is in x86_64" to "let's reuse `lkt pkg show-tag`, which calculates it based on the git tree.

This has 2 advantages:

1. It ensures we capture any changes - all of the hashes of per-arch ones already are checked in in `versions.<arch>`; this ensures that if we change, e.g. arm64 but not amd64 that we capture the change.
2. Keeps it consistent across the packages

This doesn't quite bring us to being able to do `lkt pkg build tools/alpine` - we may never get there - but it at least reuses some of the logic while doing a better job capturing changes.

**- How I did it**

Added  `tools/alpine/build.yml` (necessary for `show-tag` to work), then changed `push-manifest.sh` to just call `lkt pkg show-tag`

**- How to verify it**

Build it, try it.

Unify hash calculation for alpine index with the rest of linuxkit

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -62,44 +62,54 @@ master.
 
 This step is not necessarily required if the alpine base image has
 recently been updated, but it is good to pick up any recent bug
-fixes. Updating the alpine base image is different to other packages
-and it must be performed on `x86_64` first:
+fixes. Updating the alpine base image is different to other packages.
+You must perform the arch-specific image builds, pushes and updates on each
+architecture first - these can be done in parallel, if you choose. When done,
+you then copy the updated `versions.<arch>` to one place, commit them, and
+push the manifest.
 
-```sh
-cd $LK_ROOT/tools/alpine
-make push
-```
+Details:
 
-This will update `linuxkit/alpine` and change the `versions.x86_64`
-file. Check it in and push to GitHub:
+#### Build and Push Per-Architecture
 
-```sh
-git commit -a -s -m "tools/alpine: Update to latest"
-git push $LK_REMOTE rel_$LK_RELEASE
-```
-
-Now, on each build machine for the other supported architectures, in turn:
+On each supported platform, build and update `linuxkit/alpine`, which will update the `versions.<arch>`
+file.:
 
 ```sh
 git fetch
 git checkout rel_$LK_RELEASE
 cd $LK_ROOT/tools/alpine
 make push
-git commit -a --amend
-git push --force $LK_REMOTE rel_$LK_RELEASE
 ```
 
-With all supported architectures updated, head back to the `x86_64`
-machine and update the release branch:
+Repeat on each platform.
+
+#### Commit Changed Versions Files
+
+When all of the platforms are done, copy the changed `versions.<arch>` from each platform to one please, commit and push.
+In the below, replace `linuxkit-arch` with each build machine's name:
 
 ```sh
-git fetch && git reset --hard $LK_REMOTE/rel_$LK_RELEASE
+# one of these will not be necessary, as you will likely be executing it on one of these machines
+scp linuxkit-s390x:$LK_ROOT/tools/alpine/versions.s390x $LK_ROOT/tools/alpine/versions.s390x
+scp linuxkit-aarch64:$LK_ROOT/tools/alpine/versions.aarch64 $LK_ROOT/tools/alpine/versions.aarch64
+scp linuxkit-x86_64:$LK_ROOT/tools/alpine/versions.x86_64 $LK_ROOT/tools/alpine/versions.x86_64
+git commit -a -s -m "tools/alpine: Update to latest"
+git push $LK_REMOTE rel_$LK_RELEASE
+```
+
+#### Update and Push Multi-Arch Index
+
+Push out the multi-arch index:
+
+```sh
+make push-manifest
 ```
 
 Stash the tag of the alpine base image in an environment variable:
 
 ```sh
-LK_ALPINE=$(head -1 alpine/versions.x86_64 | sed 's,[#| ]*,,' | sed 's,\-.*$,,' | cut -d':' -f2)
+LK_ALPINE=$(make show-tag)
 ```
 
 
--- a/tools/alpine/Makefile
+++ b/tools/alpine/Makefile
@@ -15,10 +15,19 @@ ifeq ($(ARCH), s390x)
 DEPS += packages.s390x
 endif
 
+ORG?=linuxkit
+IMAGE?=alpine
+TAG?=$(shell git ls-tree --full-tree HEAD -- $(PWD) | awk '{print $$3}')
+DIRTY?=$(shell git diff-index HEAD -- $(PWD))
+ifneq ($(DIRTY),)
+TAG := $(TAG)-dirty
+endif
+
+
 default: push
 
 show-tag:
-	@sed -n -e '1s/# \(.*\/.*:[0-9a-f]\{40\}\)/\1/p;q' versions.$(ARCH)
+	@echo "$(ORG)/$(IMAGE):$(TAG)"
 
 iid: Dockerfile Makefile $(DEPS)
 	docker build --no-cache --iidfile iid .
@@ -34,8 +43,10 @@ push: hash iid versions.$(ARCH)
 	docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
 		(docker tag $(shell cat iid) $(ORG)/$(IMAGE):$(shell cat hash) && \
 		 docker push $(ORG)/$(IMAGE):$(shell cat hash))
-	./push-manifest.sh $(ORG) $(IMAGE)
 	rm -f iid
+
+push-manifest:
+	./push-manifest.sh $(ORG) $(IMAGE)
 
 build: hash iid versions.$(ARCH)
 	docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
--- a/tools/alpine/push-manifest.sh
+++ b/tools/alpine/push-manifest.sh
@@ -17,8 +17,13 @@ IMAGE=$2
 IMG_X86_64=$(head -1 versions.x86_64 | sed 's,[#| ]*,,')
 IMG_ARM64=$(head -1 versions.aarch64 | sed 's,[#| ]*,,')
 IMG_s390x=$(head -1 versions.s390x | sed 's,[#| ]*,,')
-# Extract the TAG from the x86_64 name and build the manifest target name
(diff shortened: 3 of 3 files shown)


rn:
Thanks for taking a stab at this. The current build is not perfect but I'm not sure if this PR improves upon it. The `build.yml` only seems to be there to make `show-tag` work but none of the other `linuxkit pkg` options are supported and and will have very confusing error modes (if any errors). `tools/alpine` is simply not a linuxkit package and making it appear to be that for a small subset of package options is more confusing than just accepting that it is special and handle it special. The other thing is that you still need to go around each arch to build `linuxkit/alpine` and this just adds more tags which are not fully useable to docker hub since every platform build changes the git tree hash.

deitch:
I really like the idea of making `tools/alpine` into a true linuxkit package, allowing us to do cross-builds and builds from other builders, etc. not to mention being able to just do `lkt pkg push tools/alpine` and be done with it.

But there is a lot more to it than this. There are a few divergences:

* we would need to calculate the tags of the per-arch images differently (although arguably they don't really need tags, when they are referenced in an index)
* we would need to have a step to extract the unique per-arch hash to place in `versions.<arch>` and then check them in, then calculate a new hash (using `git ls-tree`), before creating and pushing the multi-arch index

I think to get the above to work, we would need to create some stage hooks in `lkt pkg push`, which we may get to someday, but not now.

I am going to change this PR to just use the `ls-tree` rather than x86_64-specific hash.

deitch:
@rn take a look at it now. It uses the same logic as `lkt pkg push` but does so directly in the script

rn:
`make show-tag` would need updating as well, but the main issue, as I now realise, is that the build will always dirty the directory by creating the `versions` file. So you need to `make build` && `git commit` && `make push`, with the `make push` rebuilding again, and hopefully not updating any packages. You have to do this then for all three architectures with the `git push` in between if you build this on different machines. This seems even more complex than what we currently have.

deitch:
> make show-tag would need updating as well

Very good point. Fixed.

deitch:
> is that the build will always dirty the directory by creating the versions file. So you need to make build && git commit && make push, with the make push rebuilding again, and hopefully not updating any packages. You have to do this then for all three architectures with the git push in between if you build this on different machines. This seems even more complex than what we currently have

I agree wholeheartedly that it is complicated. I don't think it is any more complicated than the current setup. Here are the current docs:

<pre>
 ```sh
cd $LK_ROOT/tools/alpine
make push
```

This will update `linuxkit/alpine` and change the `versions.x86_64`
file. Check it in and push to GitHub:

```sh
git commit -a -s -m "tools/alpine: Update to latest"
git push $LK_REMOTE rel_$LK_RELEASE
```

Now, on each build machine for the other supported architectures, in turn:

```sh
git fetch
git checkout rel_$LK_RELEASE
cd $LK_ROOT/tools/alpine
make push
git commit -a --amend
git push --force $LK_REMOTE rel_$LK_RELEASE
```

With all supported architectures updated, head back to the `x86_64`
machine and update the release branch:

```sh
git fetch && git reset --hard $LK_REMOTE/rel_$LK_RELEASE
```
</pre>

Isn't that exactly what you are describing already?

I still would like to get this to be able to be built via `lkt pkg build`, because that nicely does runners on remote machines or emulation, and so solves all of the bouncing around between machines, saving it all to linuxkit cache. But as you noted elsewhere, that is a job for another day.

rn:
Maybe I'm missing something or am misreading the patch, but I think this would push `-dirty` tags *and* requires an extra build/push.

For example, on x86_64:
```
make push
```
This will build the image and create `versions.x86_64`. Then will call `push-manifest.sh`. At this point the git tree is dirty and this pushes a `-dirty` tag.

Then I `git commit`, and `git push` and go to next architecture. There I do `make push`...again a dirty tag would be pushed. `git commit/push` and next architecture. Now I have three `-dirty` tags pushed to hub and I need to do another `make push` (which also does another full build), and hope that `versions.XXX` does not change.  So this seem to be another `make push` being required over what we currently have.

deitch:
> Now I have three -dirty tags pushed to hub and I need to do another make push (which also does another full build), and hope that versions.XXX does not change. So this seem to be another make push being required over what we currently have

This is a good explanation. I see what you are getting it; I think you may be right.

The right answer to this appears to be to separate `make push` from `make push-manifest` (this stuff all is so much easier with lkt packages, which is why I did all of that hard work to get it local cache and indexes and such). Then you can just append to the instructions with:

<pre>
and push the manifest:

```sh
make push-manifest
```
</pre>

I am going to push that here, see how that works out.

deitch:
Take a look at it now. This way, in `tools/alpine/`, calling `make push` only does the arch-specific image, and `make push-manifest` gets saved until the end, being called once. It always was sort of strange that we pushed the manifest multiple times.

docs/releasing.md
```diff
@@ -96,6 +96,12 @@ machine and update the release branch:
 git fetch && git reset --hard $LK_REMOTE/rel_$LK_RELEASE
 ```
 
+Finally, with all of the arch-specific images out, push out the multi-arch index:
```
rn, line 99: I think with this, the comment above regarding that this must be done on `x86_64` first is no longer true. The only requirement is to build this in turns...you can even build in parallel and copy the `versions.<arch>` from the build machines

docs/releasing.md
```diff
@@ -96,6 +96,12 @@ machine and update the release branch:
...
+
+```sh
+make push-manifest
+```
+
 Stash the tag of the alpine base image in an environment variable:
 
 ```sh
```
rn, line 107: This also needs changing. Best to replace with `make show-tag`

tools/alpine/Makefile
```diff
@@ -34,8 +43,9 @@ push: hash iid versions.$(ARCH)
 	docker pull $(ORG)/$(IMAGE):$(shell cat hash) || \
 		(docker tag $(shell cat iid) $(ORG)/$(IMAGE):$(shell cat hash) && \
 		 docker push $(ORG)/$(IMAGE):$(shell cat hash))
+
+push-manifest:
 	./push-manifest.sh $(ORG) $(IMAGE)
-	rm -f iid
```
rn, line 38: I think we still need to delete this on normal push

tools/alpine/push-manifest.sh
```diff
@@ -17,8 +17,10 @@ IMAGE=$2
 IMG_X86_64=$(head -1 versions.x86_64 | sed 's,[#| ]*,,')
 IMG_ARM64=$(head -1 versions.aarch64 | sed 's,[#| ]*,,')
 IMG_s390x=$(head -1 versions.s390x | sed 's,[#| ]*,,')
-# Extract the TAG from the x86_64 name and build the manifest target name
-TAG=$(echo "$IMG_X86_64" | sed 's,\-.*$,,' | cut -d':' -f2)
+# Extract the TAG from the tree hash - just like how "linuxkit pkg show-tag" does it - name and build the manifest target name
+TAG=$(git ls-tree --full-tree HEAD -- $(pwd) | awk '{print $3}')
+DIRTY=$(git diff-index HEAD -- $(pwd))
+[ -n "$DIRTY"] && TAG=$(TAG)-dirty
```
rn, line 23: We want to prevent `-dirty` tags being pushed. IIRC with `linuxkit pkg push` this errors out unless the user explicitly forces a push. I don't think we need the override here, just refuse to push `-dirty`

rn (commented):
This is going in the right place. Thanks. A few more nits, esp wrt to documentation and not pushing `-dirty`

docs/releasing.md
deitch: Good point. I updated.

docs/releasing.md
deitch: Good catch. Updated.

tools/alpine/Makefile
deitch: Dang, I thought I had left it. Fixed.

tools/alpine/push-manifest.sh
deitch: I like that. Two mins.

deitch: Updated. Take a look.

rn: Maybe I'm missing this in the web-ui but I don't see this erroring

deitch: Of course, I didn't actually push it before. 🤦‍♂️ 

_Now_ it does it.

docs/releasing.md
```diff
@@ -62,23 +62,27 @@ master.
 
 This step is not necessarily required if the alpine base image has
 recently been updated, but it is good to pick up any recent bug
-fixes. Updating the alpine base image is different to other packages
-and it must be performed on `x86_64` first:
+fixes. Updating the alpine base image is different to other packages.
+You must perform the arch-specific image builds, pushes and updates on each
+architecture first - these can be done in parallel, if you choose. When done,
+you then copy the updated `versions.<arch>` to one place, commit them, and
```
rn, line 68: This description does not match the the steps below where we still go from one arch for the initial build and then other arches with git forced pushes etc

rn (commented):
Minor nit on the documentation but other wise looks good. I would also really prefer to have the description in the PR summary in the commit message, explaining in plain english what this commit does. This makes it a lot easier when you look at this commit in a a few months and wonder what it does. Just read the commit message rather than trying to hunt the PR and read the PR message

docs/releasing.md
deitch: Good point. I updated the steps to reflect the above: build and push on each, collect the updated files, commit.

deitch:
Synced the documents and changed the commit message. Have another look @rn

rn (approved):
Thanks for updating the docs. I think i would prefer if I could just cut and past commands and now I have to adjust it to whatever my shortcuts and paths are for the given build machine but the gist of how to use it is much clearer and this is definitely an improvement

deitch:
>  I think i would prefer if I could just cut and past commands and now I have to adjust it to whatever my shortcuts and paths are for the given build machine

Agreed. If you can think of an easier way to document it, that would be great, another simple PR.

I really would like to do something like we do with `lkt pkg build`/`lkt pkg push`, where we run it on one and it knows how to build on the others. The only really differences between `lkt pkg build` and this are:

* the `versions.<arch>` files, which are generated, and then need to be checked in, but that could easily be handled via a two-step process: `lkt pkg build` and then `git commit -a` and then push
* the hash calculation is different

PR #3668, opened 2021-05-17
simplify nobuild/force/build logic; simplify docker command; fix docker load issue

deitch:

**- What I did**

Simplified the `lkt pkg build` and `push` logic to reduce flags, and ensure it always builds when it is supposed to.

We had cases where it was supposed to build and wasn't.

**- How I did it**

Eliminated the duplicate `-nobuild` and `-image` flags on `lkt pkg build` or `pkg push`. Now, the logic is straightforward. This also is in the code:

* `pkg build`                   - always builds unless is in cache
* `pkg build --force`           - always builds even if is in cache
* `pkg push`                    - always builds unless is in cache
* `pkg push --force`            - always builds even if is in cache
* `pkg push --nobuild`          - skips build; if not in cache, fails
* `pkg push --nobuild --force`  - nonsensical

**- How to verify it**

Run it. CI too.

Simplified flags


deitch:
Interesting. `linuxkit/trim-after-delete` build failed on arm64:

```
error: failed to solve: rpc error: code = Unknown desc = executor failed running [/dev/.buildkit_qemu_emulator /bin/sh -c apk add --no-cache --initdb -p /out     alpine-baselayout     busybox     util-linux]: exit code: 2
504
Error building "linuxkit/trim-after-delete:510a407410a9cd2ebbf9e6cc445a57055189e580": error building for arch arm64: exit status 1
505
make: *** [Makefile:13: build] Error 1
506
```

I think you did recent work on that package @djs55 ? Could it be related?

deitch:
This now has 3 commits, each of which stands on its own

1. Simplified the force/nobuild/image logic, which makes it easier to use and removes a bug wherein builds sometimes wouldn't happen
2. refactored the docker "build context" to just pass in stdin when needed by commands, and have the appropriate command set its own args
3. fixed an issue with docker load wherein it was trying to load from a flattened filesystem image in tar, rather than a properly laid out tar image file

djs55:
@deitch I merged an update to `trim-after-delete` but haven't managed to `linuxkit pkg push` because of an emulation failure on s390 :( It sounds like `/dev/.buildkit_qemu_emulator` is another emulator failure?

deitch:
Rebased to catch the fix for `trim-after-delete` from #3669

djs55 (approved):
I tried various options `pkg build` and `pkg push` and it seemed to be working as advertised (although I had a `push` problem which is an issue with my account and not the code here). So LGTM!

PR #3670, opened 2021-05-18
Allowing linuxkit to recongnize linux/riscv64 arch

rvs:
This is inspired by EVE OS's port to riscv64 https://github.com/lf-edge/eve/pull/2083 

**- What I did**

added linux/riscv64 as a recognizable architecture

**- How I did it**

just followed the breadcrumbs of what already is being done for `linux/s390x`

**- How to verify it**

one can try building EVE packages along the lines of https://github.com/lf-edge/eve/pull/2083 but until the Docker qemu-user bug for Mac OS is fixed -- this will require either workarounds on Docker Desktop side OR running it on Linux and Windows.

Adding linux/riscv64 as a recognizable architecture

### DIFF
--- a/src/cmd/linuxkit/pkglib/build.go
+++ b/src/cmd/linuxkit/pkglib/build.go
@@ -352,12 +352,7 @@ func (p Pkg) buildArch(d dockerRunner, c lktspec.CacheProvider, arch string, arg
 		tagArch string
 		tag     = p.Tag()
 	)
-	switch arch {
-	case "amd64", "arm64", "s390x":
-		tagArch = tag + "-" + arch
-	default:
-		return nil, fmt.Errorf("Unknown arch %q", arch)
-	}
+	tagArch = tag + "-" + arch
 	fmt.Fprintf(writer, "Building for arch %s as %s\n", arch, tagArch)
 
 	if !bo.force {
--- a/src/cmd/linuxkit/pkglib/docker.go
+++ b/src/cmd/linuxkit/pkglib/docker.go
@@ -26,10 +26,6 @@ const (
 	buildkitBuilderName = "linuxkit"
 )
 
-var platforms = []string{
-	"linux/amd64", "linux/arm64", "linux/s390x",
-}
-
 type dockerRunner interface {
 	buildkitCheck() error
 	tag(ref, tag string) error


src/cmd/linuxkit/pkglib/docker.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 var platforms = []string{
-	"linux/amd64", "linux/arm64", "linux/s390x",
+	"linux/amd64", "linux/arm64", "linux/s390x", "linux/riscv64",
```
deitch, line 31: I actually don't think this is used anywhere. I cannot find it referenced, and when I commented it out and ran `go vet`, I had no complaints. Maybe it should just be removed?

src/cmd/linuxkit/pkglib/pkglib.go
```diff
@@ -69,7 +69,7 @@ func NewFromCLI(fs *flag.FlagSet, args ...string) ([]Pkg, error) {
 	// Defaults
 	piBase := pkgInfo{
 		Org:          "linuxkit",
-		Arches:       []string{"amd64", "arm64", "s390x"},
+		Arches:       []string{"amd64", "arm64", "s390x", "riscv64"},
```
deitch, line 72: This makes it a _default_ for building, but not enable it. Can you check if your riscv builds work even without this change? As I see the logic, this should not get in the way.

This really should go somewhere more explicit, but that is beyond the scope of this PR.

src/cmd/linuxkit/pkglib/build.go
```diff
@@ -361,7 +361,7 @@ func (p Pkg) buildArch(d dockerRunner, c lktspec.CacheProvider, arch string, arg
 		tag     = p.Tag()
 	)
 	switch arch {
-	case "amd64", "arm64", "s390x":
+	case "amd64", "arm64", "s390x", "riscv64":
```
deitch, line 364: This should be the one piece necessary to enable building on `riscv64`. Personally, I would like to see it gone. If someone explicitly selected to build for arch `foobar128`, then `buildArch` should not be filtering it out. I don't think we should have the restricted set of archs hard-coded into `buildArch()`.

I am going to propose that you just remove this entire `switch arch` clause, replacing it with just `tagArch = tag+"-"+arch` that is in the successful case.

src/cmd/linuxkit/pkglib/docker.go
deitch: Yes, can you remove it, and see that everything works?

deitch (changes requested):
You have highlighted some weaknesses in the code. Can you make those changes - remove the whole `var platforms` section; do not change the `Arches: []string{...}`, remove the entire `switch arch` to be replaced with just setting the `tagArch` - and see that it all works?

src/cmd/linuxkit/pkglib/docker.go
rvs: It does! Thanks for the hint @deitch

src/cmd/linuxkit/pkglib/pkglib.go
rvs: You were right -- so far I think it is fine to leave it as-is. It is confusing tho

src/cmd/linuxkit/pkglib/build.go
rvs: Done.

rvs:
Followed up on all your advice @deitch -- can you please give it another look?

deitch (approved):

PR #3679, opened 2021-05-26
Update kernels to 5.11.22/5.10.39/5.4.121

rn:
while at it, also fix the remove special case for WireGuard builds, fix `KERNEL_SOURCE`, fix `ARG` handling for `BUILD_SOURCE` and remove remnants of `DOCKER_CONTENT_TRUST` from the kernel builds

ibex


deitch (commented):
Don't the various kernel test cases need to be updated for the versions as well?

deitch:
@rn I see why you didn't change the kernel versions in the tests; because they failed.

I see 2 potential issues causing the test failures. Both likely are due to my own work earlier 🤦‍♂️ 

First, [here](https://github.com/linuxkit/linuxkit/pull/3679/files#diff-a32d0aec1031dd50b4f1728240d347e72c01312b4ae4ed8503985ffcd7c1d492R198-R200) we add the `LABEL` to the earlier stage image, but then do `FROM scratch` afterwards. The `LABEL` doesn't get carried over. It probably should change this:

```Dockerfile
RUN printf "${BUILD_IMAGE}" > /out/kernel-builder
LABEL org.mobyproject.linuxkit.kernel.buildimage ${BUILD_IMAGE}

FROM scratch
```

to this:

```Dockerfile
RUN printf "${BUILD_IMAGE}" > /out/kernel-builder

FROM scratch
LABEL org.mobyproject.linuxkit.kernel.buildimage ${BUILD_IMAGE}
```

Which probably also means we need to add the `ARG` later as well.

Second, I think the tags tests are looking for the file in the wrong place.

We create the builder [here](https://github.com/linuxkit/linuxkit/pull/3679/files#diff-a32d0aec1031dd50b4f1728240d347e72c01312b4ae4ed8503985ffcd7c1d492R197) in `/out/kernel-builder` on the `kernel-build`, and then copy it all over [here](https://github.com/linuxkit/linuxkit/pull/3679/files#diff-a32d0aec1031dd50b4f1728240d347e72c01312b4ae4ed8503985ffcd7c1d492R204) as `COPY --from=kernel-build /out/* /`. That means the file ends up in `/kernel-builder`, but our tests [here](https://github.com/linuxkit/linuxkit/blob/70ac373206642339ff5cc9e9b091723156082499/test/cases/020_kernel/tags.sh#L39) look for it in `/etc/kernel-builder`. This should be changed to look for it in `/kernel-builder` rather than `/etc/kernel-builder`.

rn:
Thanks. Added a fix for both and currently compiling/pushing new version and then will git push the update to kick CI again

deitch:
@rn yeah, that is because it still is testing against the existing kernels, the ones I retrofitted manually. Those I did (accidentally) to `/etc/kernel-builder` rather than `/kernel-builder`. I will go back and manually retrofit those correctly, but you should update this PR to test your new versions.

deitch:
I manually updated all of the existing kernel versions (5.4.116, 5.10.34, 5.11.18) on each of three arches (amd64, arm64, s390x), as well as the multi-arch index for each version. I will kick off the tests again, but in any case, this should just update the versions in the tests.

rn:
Doh, I had a commit to update the all the YAMLs/test to latest kernels but then removed it when I had to do more fixes (and forgot about it). Anyway, not added a commit to update the kernels.. Let's see where we get to this time

deitch:
I looked at the one remaining failing test. It isn't happy because it has a different manifest for `linuxkit/kernel:5.4.121-builder` and `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`. (FWIW, it does looks solely at the architecture manifests, and sorts and compacts them so that irrelevant diffs are ignored).

Taking a closer look, here are the contents of each. 

alpine:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 2411,
      "digest": "sha256:71bfb0d6b054c9fdc99235ead2eaaf086ce77bda6446b8b4e23a11e0396b4fd0",
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 2414,
      "digest": "sha256:1bb15c0ae605b6b4b4b1a7f62a004d1459ec9a707ad2d289ca61c5e948d145fc",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 2411,
      "digest": "sha256:39ec5efb1b9dc265e1ea44f88f094ed87437b93b6622016b6915e71fb6e88fb7",
      "platform": {
        "architecture": "s390x",
        "os": "linux"
      }
    }
  ]
}```

and linuxkit/kernel:5.4.121-builder:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "size": 2414,
      "digest": "sha256:1bb15c0ae605b6b4b4b1a7f62a004d1459ec9a707ad2d289ca61c5e948d145fc",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```

It has the right manifest for amd64, but is missing the ones for s390x and arm64. Did the multiarch get pushed out?

rn:
I hadn't pushed the other kernels yet. Let me kick that off

olljanat:
Any possibility to enable CONFIG_SQUASHFS_ZSTD as part of this? I would like to be able to use Zstandard compression on Squashfs instead of Gzip as it looks to be very good compromise with size and decompression speed like which why example Alpine switched all their packages to using it https://archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/ (I can open PR about mkimage-squashfs update).

rn:
@olljanat sorry, not as part of this PR. I really just would like to get this over the line after having spent quite a few days on it already. Please open an issue and hopefully then I won't forget for the next kernel update

rn:
@deitch it looks like these two lines in the tags test seem to wrongly assume that all kernels are build for all architectures:
```
sumtag=$(docker manifest inspect ${BUILDER} | jq  -c '.manifests | sort_by(.digest)' | sha256sum | awk '{print $1}')
sumlabel=$(docker manifest inspect ${BUILDERLABEL} | jq  -c '.manifests | sort_by(.digest)' | sha256sum | awk '{print $1}')
```
If I read this right, `sumtag` is the sha256 of the manifest build (which does exist for all arches) and `sumlabel` is the sha256 for the labels for all kernels, which does *not* exist for all all arches.

Above the 5.4.x chekc fails because we only build 5.4.x for x86 but not for arm64 and s390x.

Also, i ran the above on macOS where it passed because `sha256sum` does not exit, which should clearly be an error for the test.

deitch:
> the tags test seem to wrongly assume that all kernels are build for all architectures:
> If I read this right, sumtag is the sha256 of the manifest build (which does exist for all arches) and sumlabel is the sha256 for the labels for all kernels, which does not exist for all all arches.

right idea, but not exactly how it does it.

Unfortunately, Docker hub doesn't support one tag pointing to another in the same repo, let alone another repo. So I cannot have `linuxkit/kernel:5.11.22-builder` just be a pointer to `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`.

So `linuxkit/kernel:5.11.22-builder` has its own index, which has the same manifests as the index at `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`. We check that the manifests in the index are the same for both.

> Above the 5.4.x check fails because we only build 5.4.x for x86 but not for arm64 and s390x.

Close enough. It fails because `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba` is an index that has manifests for amd64, arm64, s390x, while `linuxkit/kernel:5.11.22-builder` is an index that has manifests for amd64 but not arm64 or s390x. The manifests for amd64 _are_ identical, so that is good, but the entire list is not identical because some are missing.

The reason that `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba` has 3 arches, and `-builder` has just 1, is as you said. We create that `-builder` tag [here](https://github.com/linuxkit/linuxkit/blob/master/kernel/Makefile#L138-L142): 

```Makefile
tagbuilder_$(2)$(3)$(4): notdirty
        docker tag $(IMAGE_BUILDER) $(ORG)/$(IMAGE):$(1)$(3)$(4)-builder$(SUFFIX) && \
        docker push $(ORG)/$(IMAGE):$(1)$(3)$(4)-builder$(SUFFIX) && \
        $(PUSH_MANIFEST) $(ORG)/$(IMAGE):$(1)$(3)$(4)-builder $(DOCKER_CONTENT_TRUST)
```

That adds an arch-specific builder tag `linuxkit/kernel:version-builder-amd64`, etc. for the arch on which it is running, then creates the index `linuxkit/kernel:version-builder` after doing so. If we only run the make on a subset of arches, the index only will contain some of the arches.

I am more than open to a better way to do the test, or to generate the index.

What it really is testing is: "My builder is `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`, the tag for me is `linuxkit/kernel:5.11.22-builder`. Do these point to the same thing?"

We don't want to just test for a single arch (just amd64, e.g.), because then the index for `-builder` might actually be missing something and broken. 

Ideally, the index would be identical to `linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba`. If it were easily doable, I would replace the above makefile code with something like (pseudo-code, not doable using docker AFAIK):

```sh
for arch in amd64 arm64 s390x; do
  docker tag linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba-${arch} linuxkit/kernel-version-builder-${arch}
  push-manifest linuxkit/kernel-version-builder
done
```

Don't even pull them, just tag remotely, without even building, and do all of the arches from one place.

Open to ideas.

rn:
Thanks for clarifications.

> We don't want to just test for a single arch (just amd64, e.g.), because then the index for -builder might actually be missing something and broken.

We don't do this for any other tests. We only test on the architecture we are executing on. So, for example for any package, which do not check if the manifest reflects what the `build.yml` says. Now it is debatable if this is the right thing to do, but in an ideal world we'd have a CI runner for each architecture which checks if that architecture works. Alternatively we'd extend the CI to be more cross platform aware. But for now, I think we should stick with what the other tests are doing and I'll see if I can adjust the tag test

rn:
Rewrite the tag test to iterate through each architecture for which we have a give kernel image and make sure label and builder tag exist and are the same. Worked in a local test, lets see what the CI says

deitch (approved):
I like how you made the tags test more accurate.

PR #3685, opened 2021-06-04
Makefile: default to local arch, not amd64

djs55:
**- What I did**

Made `make` and `make local-build` default to the local machine architecture. It's still possible to override by setting `GOARCH` in the environment as normal.

**- How I did it**

I borrowed runs from the `kernel/` Makefile and took @dave-tucker 's advice

> maybe we could add the default GOARCH?=amd64 to the existing Windows branch and use uname -m and map the result to GOARCH on mac/linux

**- How to verify it**

```
dave@m1 linuxkit % make local-build
make -C ./src/cmd/linuxkit local-build
CGO_ENABLED=0 go build -o /Users/dave/github.com/djs55/linuxkit/bin/linuxkit  --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=c54ace566a81fa4c6094e971baf19ca2a0c338c8 -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version="v0.8+""

dave@m1 linuxkit % file bin/linuxkit
bin/linuxkit: Mach-O 64-bit executable arm64

dave@m1 linuxkit % GOARCH=amd64 make local-build
make -C ./src/cmd/linuxkit local-build
CGO_ENABLED=0 go build -o /Users/dave/github.com/djs55/linuxkit/bin/linuxkit  --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=c54ace566a81fa4c6094e971baf19ca2a0c338c8 -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version="v0.8+""

dave@m1 linuxkit % file bin/linuxkit
bin/linuxkit: Mach-O 64-bit executable x86_64
```

- `make` and `make local-build` default to building for the local architecture, previously `amd64` was assumed. The architecture can be overridden via the `GOARCH` environment variable.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,23 @@ ifeq ($(OS),Windows_NT)
 LINUXKIT?=$(CURDIR)/bin/linuxkit.exe
 RTF?=bin/rtf.exe
 GOOS?=windows
+GOARCH?=amd64
 else
 LINUXKIT?=$(CURDIR)/bin/linuxkit
 RTF?=bin/rtf
 GOOS?=$(shell uname -s | tr '[:upper:]' '[:lower:]')
+MACHINE := $(shell uname -m)
+ifeq ($(MACHINE),x86_64)
+GOARCH?=amd64
 endif
+ifeq ($(MACHINE),$(filter $(MACHINE),aarch64 arm64))
+GOARCH?=arm64
+endif
+ifeq ($(MACHINE),s390x)
+GOARCH?=s390x
+endif
+endif
+
 GOARCH?=amd64
 ifneq ($(GOOS),linux)
 CROSS+=-e GOOS=$(GOOS)


Makefile
```diff
@@ -6,11 +6,23 @@ ifeq ($(OS),Windows_NT)
...
+endif
+ifeq ($(MACHINE),$(filter $(MACHINE),aarch64 arm64))
+GOARCH?=arm64
 endif
+ifeq ($(MACHINE),s390x)
+GOARCH?=s390x
+endif
+endif
+
 GOARCH?=amd64
```
thaJeztah, line 26: is this one still needed?

djs55: It does look like we have one to many of those :) Maybe I should delete the first `GOARCH?=amd64` in the Windows section at the top and then Windows could fall through to the one at the bottom? This would also default to `amd64` on machines which aren't Intel/Arm/s390x as before. WDYT?

rn:
I'm not yet convinced there is a good solution for this. We want to support cross platform build and local as well as docker builds. Docker builds in Docker Desktop means local GOOS/GOARCH/uname is not the same as in the build container.

djs55:
I was mainly surprised how I ended up with an `amd64` binary on my `arm64` machine. I thought, "who uses amd64 nowadays?" :) It's not very important, I'll close this for now.

PR #3688, opened 2021-06-07
Push arch-specific tags, always build index from registry

deitch:

**- What I did**

* ensured that tags for arch-specific pkg images exist. We used to have `linuxkit/foo:abc-amd64`, `linuxkit/foo:abc-arm64`, etc., in addition to the multi-arch index `linuxkit/foo:abc`. With the move to linuxkit cache, we stopped pushing those tags out. They turn out to be rather useful, so this pushes the tags out again.
* always rebuilds the multi-arch index (after pushing) from every known architecture in the registry. We used to do that when we had to build and push on each arch. Now that we don't _have_ to, it would just push whatever was local. However, if you _wanted_ to - or even just pushed out on one arch - it would replace the index that was there with one that pointed just to that arch. Oops. With this change, whenever it creates an index, it will look for every arch known under the sun and see if a tag exists for it, then create the index. This provides the least chance of error.

**- How I did it**

Changes to the `linuxkit` command, specifically `cache.Push()` and `registry.PushManifest()`

**- How to verify it**

Push out some images. I did. It was fun.

Ensure tagged images per architecture, and index always includes every arch in registry

### DIFF
--- a/src/cmd/linuxkit/cache/push.go
+++ b/src/cmd/linuxkit/cache/push.go
@@ -7,64 +7,8 @@ import (
 	namepkg "github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/registry"
+	log "github.com/sirupsen/logrus"
 )
-
-// PushWithManifest push an image along with, optionally, a multi-arch index.
-func (p *Provider) PushWithManifest(name, suffix string, pushImage, pushManifest bool) error {
-	var (
-		err     error
-		options []remote.Option
-	)
-	imageName := name + suffix
-	ref, err := namepkg.ParseReference(imageName)
-	if err != nil {
-		return err
-	}
-
-	if pushImage {
-		fmt.Printf("Pushing %s\n", imageName)
-		// do we even have the given one?
-		root, err := p.FindRoot(imageName)
-		if err != nil {
-			return err
-		}
-		options = append(options, remote.WithAuthFromKeychain(authn.DefaultKeychain))
-		img, err1 := root.Image()
-		ii, err2 := root.ImageIndex()
-		switch {
-		case err1 == nil:
-			if err := remote.Write(ref, img, options...); err != nil {
-				return err
-			}
-			fmt.Printf("Pushed image %s\n", imageName)
-		case err2 == nil:
-			if err := remote.WriteIndex(ref, ii, options...); err != nil {
-				return err
-			}
-			fmt.Printf("Pushed index %s\n", imageName)
-		default:
-			return fmt.Errorf("name %s unknown in cache", imageName)
-		}
-	} else {
-		fmt.Print("Image push disabled, skipping...\n")
-	}
-
-	auth, err := registry.GetDockerAuth()
-	if err != nil {
-		return fmt.Errorf("failed to get auth: %v", err)
-	}
-
-	if pushManifest {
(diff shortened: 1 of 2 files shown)


src/cmd/linuxkit/cache/push.go
```diff
@@ -7,65 +7,9 @@ import (
 	namepkg "github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/registry"
+	log "github.com/sirupsen/logrus"
 )
```
deitch, line 12: For the curious, we haven't used `PushWithManifest()` in some time; all of the functionality therein already is in just `Push()`.

src/cmd/linuxkit/registry/manifest.go
```diff
@@ -11,15 +11,23 @@ import (
...
+// these platforms are used only as the source for registry.PushManifestList(). Since it is
+// configured to ignore missing, we could include dozens if we wanted; it doesn't hurt, adding maybe a few seconds to
+// the whole run.
+// Ideally, we could just look for all tags that start with linuxkit/foo:<hash>-*, but the registry API
+// only supports "list all the tags" and "get a specific tag", no "get by pattern". The "get a specific tag"
+// is exactly what registry.PushManifestList() uses, so no benefit to use doing that in advance,
+// while "list all tags" is slow, and has to cycle through all of the (growing numbers of) tags
+// before we know what exists. Might as well leave it as is.
+var platformsToSearchForIndex = []string{
+	"linux/amd64", "linux/arm64", "linux/s390x", "linux/riscv64", "linux/386", "linux/ppc64le", "linux/arm/v6", "linux/arm/v7",
```
justincormack, line 23: Lets not put the 32 bit platforms, we have never built them and I don't want to support 32 bit. We will have enough fun with armv9 and stuff no doubt...

This is kind of horrible, we should pull the existing manifest and add/replace the architectures really.

deitch: > This is kind of horrible, we should pull the existing manifest and add/replace the architectures really.

It isn't that simple. In principle, we are not doing, "I have added the following local arch, so let me add that arch," it is a more generic, "tell manifest-tool to add every arch I know about, skipping the ones I don't know." 

If hub had a "get tags that match x-*", then I would get those and add them. But that doesn't exist, only get a specific one or list all (see the comment right above this line).

If someone pushes out just the arch-specific on platform A, then on B, then C, then D, and then runs the manifest, it should create for all of them. I don't know of any way to do it other than to list every arch we could think of.

deitch: Removed the 32-bit ones.

justincormack: Yeah, there is something wrong with the whole manifest list design...

justincormack (approved):

src/cmd/linuxkit/registry/manifest.go
deitch: I think the issue actually is with the registry. If I could do queries, like "get tags by starts with x" or "by architecture" or "by annotation", then many more things become possible and flexible.

justincormack: I don't think this should be the role of tags. Ideally you would want to query by tags (which images and architectures are built from this git commit?) or equivalent.

deitch: That would be fine too. But I don't think any of that exists.

PR #3696, opened 2021-06-21
check for arch when pulling to cache, push by descriptor

### LINKED ISSUE #3694: pkg push can fail when an index exists
**Description**

If you do `lkt pkg push <image>`, (without `--force`), and the index _already_ exists in the registry, then push will fail with the error

```
could not find arch-specific image in cache
```

The key issue is the following. When we do `lkt pkg build` or `lkt pkg push` (which is just `build` followed by pushing the arch-specific image and then creating/updating the multi-arch index), linuxkit does the following. None of this applies if we do `--force`, which _always_ will build a local arch-specific image:

1. Look in the cache: does the image exist?
1. If not, try to pull it down.
1. If we found it in the cache, or successfully pulled it down, then we are done with build phase; if not, or we had `--force`, build it, and we are done with build phase
1. If we asked for `pkg push`:
   1. push an arch-specific image for each arch in the index
   2. push a multi-arch index for everything in the registry

There are two problems.

First, how it does the second-to-last step, "push an arch-specific image for each arch in the index". linuxkit goes through each arch in the index, then tries to find an image in local cache named `<image>-arch`. That _will_ exist if we built it locally. However, if we pulled down `<image>`, then the only tag that exists locally is `<image>`, not `<image>-arch`, and it will fail. All of the blobs and manifests exist, but the reference in `index.json` does not exist.

Second, we found and pulled `<image>`, and all of its subsidiary blobs. Those might not have included all of the platforms for which we are building, yet it assumes success.

For example, we ran

```
lkt pkg build --platforms linux/amd64,linux/riscv64 foo/bar:123
```

The index `foo/bar:123` already exists, and references just an image for `linux/arm64`, which was built earlier.

`lkt pkg build` (and `pkg push`) will see successful download of `foo/bar:123`, and think it is done, even though it actually only has images for `arm64`, when the user explicitly requested `amd64` and `riscv64`

Note: You _always_ can solve this by passing `--force`, but that is quite wasteful if it already exists.

Proposed solutions:

First, after we pull an image to the cache, check that it references the requested platforms for which the user wants to build. If not found, then continue to build.

Second, when pushing for each arch in the index manifest list, instead of searching for the arch-specific manifest in the cache by using `<image>-<arch>`, where `<image>` is the index tag, actually tag the root based on the digest in the same manifest entry that gave the arch.

I can put together a PR for this.

deitch:

**- What I did**

Fixes #3694 

Two changes:

1. When pulling an image or index to the cache, don't assume it fulfills the requested architecture. It might have pulled an index with other arches. Explicitly check for the requested architecture.
2. When pushing from cache to registry, don't look by `<imageName>-<arch>`, as it might not be there, as the root might just be an index with descriptors per architecture. Instead, look in the index manifests for the right descriptor and push that. This is more accurate anyways.

**- How I did it**

Changes to `src/cmd/linuxkit/cache/`

**- How to verify it**

* CI
* I am running tests

More reliable cache pull/push

### DIFF
--- a/src/cmd/linuxkit/cache/pull.go
+++ b/src/cmd/linuxkit/cache/pull.go
@@ -29,29 +29,12 @@ func (p *Provider) ValidateImage(ref *reference.Spec, architecture string) (lkts
 			if desc, err = partial.Descriptor(img); err != nil {
 				return ImageSource{}, errors.New("image could not create valid descriptor")
 			}
-			if desc.Platform == nil || desc.Platform.Architecture != architecture || desc.Platform.OS != "linux" {
-				return ImageSource{}, fmt.Errorf("image was not for requested architecture: linux/%s", architecture)
-			}
 		} else {
 			ii, err := root.ImageIndex()
 			if err == nil {
 				imageIndex = ii
 				if desc, err = partial.Descriptor(ii); err != nil {
 					return ImageSource{}, errors.New("index could not create valid descriptor")
-				}
-			}
-			// check that the index has a manifest for our arch
-			im, err := imageIndex.IndexManifest()
-			if err != nil {
-				return ImageSource{}, fmt.Errorf("could not get index manifest: %v", err)
-			}
-			for _, m := range im.Manifests {
-				if m.Platform != nil && m.Platform.Architecture == architecture && m.Platform.OS == "linux" {
-					return p.NewSource(
-						ref,
-						architecture,
-						desc,
-					), nil
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


deitch:
I am satisfied with this. Ready for review.

chrisatkins2019 (approved):

chrisatkins2019 (commented):
Nicely though of

chrisatkins2019 (approved):

justincormack (approved):

PR #3700, opened 2021-06-29
enable 5.11.x-rt

TiejunChina:
This just helps get -rt back with 5.11.x-rt.


rn (approved):
Thank you. Compile tested will push proper image on the next kernel update

PR #3701, opened 2021-06-29
Disable CONFIG_BPFILTER and CONFIG_BPFILTER_UMH

fredericdalleau:
CONFIG_BPFILTER is aimed to provide a replacement for netfilter.

When CONFIG_BPFILTER is enabled, the kernel tries to contact a user mode helper
for each iptable rule update. However the implementation of this helper has not
been upstreamed yet. The communication thus fails and the kernel then falls back
to netfilter.
As a result, the rule update takes more than ten times the duration of the
netfilter implementation alone.

This has been reported by Docker Desktop users for whom it can take minutes to
start a container sharing a few hundred ports. https://github.com/docker/for-mac/issues/5668

More details on the situation is described in https://lwn.net/Articles/822744/.

**- What I did**

After Docker Desktop upgraded to kernel 5.x, users reported slowdown in iptables rule update
https://github.com/docker/for-mac/issues/5668 
The trace below trace was also showing : 
```
bpfilter: Loaded bpfilter_umh pid
bpfilter: write fail -32
```

**- How I did it**

From the trace shown above, I tried first to blacklist the bpfilter module, but blacklist didn't seem to work.
After removing bpfilter.ko, the iptables rules started being faster, but it was still not on par with an earlier kernel.
The kernel code showed the CONFIG_BPFILTER would still try to load the (now missing) bpfilter.ko module and fail before falling back.
This load tentative has a cost, and disabling CONFIG_BPFILTER brings the number back to the expected values.

**- How to verify it**

Measure execution time of iptables rule update, with and without CONFIG_BPFILTER.

Disable CONFIG_BPFILTER

image


djs55:
Very interesting lwn article about `bpfilter` and "user mode helpers" -- I didn't know about these!

I tested this change locally on my M1 with the Docker Desktop builds you shared on the docker/for-mac ticket.

Before:
```
dave@m1 % time docker run -p 30000-30100:30000-30100 -it busybox sh -c "exit"
docker run -p 30000-30100:30000-30100 -it busybox sh -c "exit"  0.08s user 0.08s system 0% cpu 3:25.80 total
```
After:
```
dave@m1 % time docker run -p 30000-30100:30000-30100 -it busybox sh -c "exit"
docker run -p 30000-30100:30000-30100 -it busybox sh -c "exit"  0.07s user 0.05s system 2% cpu 4.501 total
```
so from 205s (!) down to 4.5s. Quite an improvement.

Reading the lwn article it sounds like it might be a good idea to disable this as you suggest, until they decide whether to complete the features or remove them upstream.

WDYT @rn ?

rn:
SGTM, let me pick this up on the next kernel builds. Hopefully I find the time this weekend

rn:
another option would be add a dummy usermode helper, but disabling for now is fine with me

rn (approved):

PR #3684, opened 2021-06-03
Update containerd to 1.4.6 and runc to v1.0.0-rc95

rn:
Based on https://github.com/linuxkit/linuxkit/pull/3681 but with all images build and pushed (as well as fixes to alpine build).

Images are only build/pushed for amd64 and arm64, not s390x since there where several issues (including `linuxkit` not building). Since we are going to deprecate s390x support (see https://github.com/linuxkit/linuxkit/issues/3676) I did not bother investigating further.

@djs55 please take a look

heavy-cat

PR #3702, opened 2021-07-02
bump alpine to 3.14, remove go version hack

deitch:

**- What I did**

Updated alpine from 3.13 to 3.14.

Since 3.14 includes go 1.16.5, we no longer need that hack to grab it from edge, so got rid of it

**- How I did it**

Updated `tools/alpine/Dockerfile` and some `packages` files

**- How to verify it**

CI

alpine 3.14 as base

I didn't yet push out the images. I will do so and then update the PR.


deitch:
Also removed `gummiboot` from `packages.x86_64` for two reasons:

1. It is no longer available in Alpine packages from 3.14 onwards
2. I did a search, and we don't use it anywhere in the entire repository

deitch:
This _does_ clean up the golang repo dependency, which is great. But it does use musl and musl-dev 1.2.2-r3 (i.e. greater than 1.2.2-r0), which brings back the faccess2 issue, see [here](https://wiki.alpinelinux.org/wiki/Draft_Release_Notes_for_Alpine_3.14.0#faccessat2)

@justincormack and @thaJeztah , I thought the two of you knew this issue? Is it fixed in the version of runc we are using?

thaJeztah:
I _think_ the faccess2 issue should be resolved in runc v1.0.0-rc95 and up (at least, when running on a kernel version that supports it, if I'm not mistaken). I see https://github.com/linuxkit/linuxkit/pull/3684 was still pending (what version is currently used?)

deitch:
[Looks like](https://github.com/linuxkit/linuxkit/blob/master/pkg/runc/Dockerfile#L15) v1.0.0-rc10. So we need #3684 as you pointed out. What was the issue there?

deitch:
Pushed out images and manifests. But I did not update downstream dependencies.

rn:
@deitch where are you with this? The base seems updated but I don't seen any uses of it. Are you planning to do that at some point?

deitch:
We are blocked by #3684. With 3.14, we have the updated musl packages, which create the containerd/runc issues. When that one (#3684) is in, we can move this one in.

rn:
Thanks. Looks like @djs55 made some progress on it

djs55:
@deitch Hopefully this is unblocked by the runc/containerd upgrade. Let me know if you're too busy -- I can try to rebase it and regenerate the images if that's helpful.

deitch:
Yeah, I will try to tackle next week.

deitch:
OK, updated. I rebuilt based on alpine 3.14. @djs55 does this take advantage of your changes in #3684? What do I need to do to ensure it gets them?

djs55:
This LGTM. Do we just need to update the alpine yamls in the packages (and then rebuild the packages and then update the package hashes)? I think everything should just work, as far as I can see.

rn:
Yes, this still requires the full dance of updating tools then packages and then tests to the new version.

deitch:
"the full dance" - well put.

deitch:
Oh that is such a painful "full dance".

Finally updating the packages in `pkg/`. I assume that will run all night, have to be restarted a few times. Hopefully sometime tomorrow it will be ready.

djs55:
A previous CI run failed pulling `runc` but when I checked with `linuxkit pkg push` it was already built and pushed. So I hit the re-run tests button. Hopefully this time ...

djs55 (approved):

Issue #3676, opened 2021-05-26
Remove Support For s390x


dave-tucker:
Linuxkit has supported s390x for a while, but we are not aware of any users of this architecture.
While IBM have graciously helped provide resources for both the initial port and providing machines for build/test we are but a small group of volunteers and the additional overhead of supporting an architecture, kernel/package builds etc... is hard to justify without any users.

As such I propose that we remove support for s390x for now. This is not forever, and we'll be happy to open discussions about bringing it back in future if there are users that want it. For now though we have to prioritse where our maintainers can spend their (limited) time.

/cc @justincormack @deitch @rn @djs55 as i'd like you to ack this

thaJeztah:
makes sense to me

deitch:
Agreed.

djs55:
OK!

justincormack:
SGTM

justincormack:
Note that things should still work on s390x, this mainly means no prebuilt binaries for now.

rn:
I agree with Justin. Happy to remove it from the builds and as a default arch from `linuxkit` but would be good to keep support where needed to build binaries and packages. This does imply, though removing it from `linuxkit/alpine` and the kernel builds, which are special.

FWIW, I currently have some changes to kernel builds etc which I'm still debugging and I'd appreciate if I don't get a conflicting PR while doing so :)

PR #3704, opened 2021-07-04
Update kernels to 5.10.47/5.4.129, remove 5.11.x and add 5.12.x

rn:
Also:
- Remove s390x kernel WIP #3676
- fix kconfig build
- add CONFIG_SQUASHFS_ZSTD

seaslug


rn:
The CI failures in `LinuxKit Build Tests` seem transient.

The CI failure for kernel tests does not seem to have enough info on the failure, but when I ran it locally it complained:
```
starting kernel config sanity test with /proc/config.gz
FAILURE: mqueue filesystem missing
kernel config test failed!
Kernel config test suite FAILED
```
which is a bit odd since the `mqueue` FS is actually present when I checked with a getty container

rn:
It's actually not `mqueue` but `oprofilefs`. Test test had a bug

rn:
Pushed new version of the kernel config test

djs55 (approved):
LGTM!

PR #3716, opened 2021-10-13
Add OCI devices to yaml (needed by getty with runc v1.0.0-rc95)

djs55:
**- What I did**

Added explicit Linux devices to the image spec and added `/dev/console` (and friends) to the `getty` package, as suggested in the discussion [here](https://github.com/linuxkit/linuxkit/pull/3684#issuecomment-895081096). These will be needed when we upgrade runc as `/dev/console` was removed from the defaults in https://github.com/opencontainers/runc/commit/60e21ec26e15945259d4b1e790e8fd119ee86467

There's another relevant change in runc, described in this [security advisory](https://github.com/opencontainers/runc/security/advisories/GHSA-g54h-m393-cpwq). Previously the [cgroup (v1) device whitelist controller](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/devices.html) granted `read`, `write` and `mknod` to every container by default. The new default is to allow nothing by default, so we must grant access to everything in `devices:`.

**- How I did it**

I extended the schema in the `linuxkit` tool to add `devices`. Entries look like this:
```
  devices:
    - path: "/dev/console"
      type: c
      major: 5
      minor: 1
      mode: 0666
```
and are translated into `[]specs.LinuxDevice` (aside: I would have used the `specs.LinuxDevice` type itself but I needed to translate the `mode` from a string into an `os.FileMode`). I based this on the code in [moby/moby which creates a `specs.LinuxDevice` and a `specs.LinuxDeviceCgroup`](https://github.com/moby/moby/blob/master/oci/devices_linux.go).

For packages which need access to arbitrary block devices (e.g. `pkg/format`), since it's not really feasible to enumerate all possible block devices, this PR adds support for a wildcard rule like this:

```
  devices:
    - path: all
      type: b
```

A paranoid user who needs to grant access to only a specific device could override this in their own yml.

**- How to verify it**

You can check it does no harm by:

```
linuxkit build examples/getty.yml
linuxkit run getty
# observe a getty
```

and

```
linuxkit build examples/swap.yml
linuxkit run -disk size=4096M  swap
# observe a getty
free
# observe swap is in-use
```

and then you can observe it fixes the problem in https://github.com/linuxkit/linuxkit/pull/3684 by rebasing these patches on top, and verifying that the `getty` and `swap` still works.

- Add a `devices:` section to the yaml, to allow devices like `/dev/console`, `/dev/ttyS0` etc to be added to containers like `getty`. This is required by `runc` v1.0.0-rc95.

image

### DIFF
--- a/examples/docker-for-mac.yml
+++ b/examples/docker-for-mac.yml
@@ -26,7 +26,7 @@ onboot:
     command: ["/usr/bin/mountie", "/var/lib"]
   # make a swap file on the mounted disk
   - name: swap
-    image: linuxkit/swap:0028aeae2741d28120e4d2c6efcc5af003eae395
+    image: linuxkit/swap:7f7074c05bad414af39f2374301b72bc67314715
     command: ["/swap.sh", "--path", "/var/lib/swap", "--size", "1024M"]
   # mount-vpnkit mounts the 9p share used by vpnkit to coordinate port forwarding
   - name: mount-vpnkit
--- a/examples/swap.yml
+++ b/examples/swap.yml
@@ -18,7 +18,7 @@ onboot:
     image: linuxkit/mount:a9a5f731261891bd880e108e6fa2be5bac2f63ba
     command: ["/usr/bin/mountie", "/var/external"]
   - name: swap
-    image: linuxkit/swap:0028aeae2741d28120e4d2c6efcc5af003eae395
+    image: linuxkit/swap:7f7074c05bad414af39f2374301b72bc67314715
     # to use unencrypted swap, use:
     # command: ["/swap.sh", "--path", "/var/external/swap", "--size", "1G"]
     command: ["/swap.sh", "--path", "/var/external/swap", "--size", "1G", "--encrypt"]


djs55:
@deitch I think it's only needed for `getty`, which accesses `/dev/console` (etc) directly. Since the `pkg/getty/build.yml` has the `devices:` section, no-one should have to change their own yaml.

I might do a few manual tests to check that assertion :)

deitch:
Makes sense to me, then, @djs55 . Just need to clean up the history, run the tests, and see about the failing tests.

djs55:
@rn sorry about that, I realised I hadn't run the update sha script for `getty` properly and had to update a few more of them. The hash should be
```
% ./bin/linuxkit pkg build pkg/getty 
Building "linuxkit/getty:ebe3397aa9b08e8f310121fdb0aac9406e1c8e73"
```
which we're now using
```
dave@m1 linuxkit % git grep linuxkit/getty: | grep -v ebe3397
pkg/getty/README.md:    image: linuxkit/getty:<hash>
```

djs55:
@rn thanks for the merge! I think there might still be some tweaks needed for the new containerd and runc, see https://github.com/linuxkit/linuxkit/pull/3718 So far I think `poweroff` was failing but seemed to work when I removed `readonly: true` (not sure why). I'm seeing some hanging in `linuxkit.platforms.qemu.run_efi` which I'm investigating.

PR #3718, opened 2021-10-15
Update runc, containerd, add devices: and fix readonly

djs55:
This subsumes  #3684 : update runc, containerd

Update to `runc` 1.0.2, the latest release.

Update to `containerd` 1.4.11, the latest release.

This allows (and requires!) us to remove the `faccessat` workaround.

The patches are mostly updates to packages which require `devices:`, and associated hash updates.

The CI is finally green!

The most significant change is removing the `ro` mount flag from `/dev` in the `config.json` passed to `runc`. `runc` versions `1.0.0-rc93` and later fail to start the container because they try to `mkdir` inside the root filesystem while setting up the environment, which fails because it has been mounted readonly. The fix is to remove the `ro` flag, but I think this is probably ok because the container still has a `readonly: true` root filesystem.

image

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -4,7 +4,7 @@ kernel:
 init:
   - linuxkit/init:78fb57c7da07c4e43c3a37b27755581da087a3b6
   - linuxkit/runc:5f9941eed05f58293f928c9f2f0b6a3f9f6f55c1
-  - linuxkit/containerd:1c4d860d198095565baa3d8d60d434f034f0fef0
+  - linuxkit/containerd:b991571dd383351dbb2873f9ae3b70475adb6816
   - linuxkit/ca-certificates:4df823737c9bf6a9564b736f1a19fd25d60e909a
 onboot:
   - name: sysctl
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -4,7 +4,7 @@ kernel:
 init:
   - linuxkit/init:78fb57c7da07c4e43c3a37b27755581da087a3b6
   - linuxkit/runc:5f9941eed05f58293f928c9f2f0b6a3f9f6f55c1
-  - linuxkit/containerd:1c4d860d198095565baa3d8d60d434f034f0fef0
+  - linuxkit/containerd:b991571dd383351dbb2873f9ae3b70475adb6816
   - linuxkit/ca-certificates:4df823737c9bf6a9564b736f1a19fd25d60e909a
 onboot:
   - name: sysctl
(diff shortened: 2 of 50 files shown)


djs55:
I notice the `/dev/pts` directory already exists inside the `rootfs`, so I don't know why `runc` tries to create it.

I removed the `/dev/pts` entry from the `config.json` but now it says:
```
linuxkit-0e465ebdb34b:~# runc create --bundle /containers/onboot/000-binfmt foobar
[   27.379103] IPVS: ftp: loaded support on port[0] = 21
ERRO[0000] container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:88: creating device nodes caused: read-only file system 

```

djs55:
Looking for the version with the behaviour change:
- 1.0.0-rc94: FAIL
- 1.0.0-rc93: FAIL
- 1.0.0-rc92: PASS
- 1.0.0-rc91: PASS
- 1.0.0-rc90: PASS

Looking for the PR Merge commit between rc92 and rc93 (total 129 PRs):
- opencontainers/runc@e8498d3ad580b31d8755253e60570466be5a66ba (60/129): FAIL
- opencontainers/runc@9d4c02cf290f58080c17fcdfd91e2fe56ccaf57a (64/129):  FAIL
- opencontainers/runc@6a28ca3eb4aa313309db9dbd978e5e703d92e507 (65/129): PASS
- opencontainers/runc@07e35a7a409c14d665d06733c21aaae37810d63a (67/129): PASS
- opencontainers/runc@44f221e2fce70dd6a75e7d67e9bdde36706aef79 (75/129): PASS
- opencontainers/runc@35c3555946d09a811d8a14021376395bf0b70666 (90/129): PASS

djs55:
Oops, I've accidentally reverted the containerd alpine change. Let me fix that.

djs55:
@fredericdalleau thanks!

At the moment the processes inside these specific containers are all running as root so I don't think the `perms` currently have any effect e.g.
```
/ # ps uax | grep kmsg
 1293 root      0:00 /usr/bin/containerd-shim-runc-v2 -namespace services.linuxkit -id kmsg -address /run/containerd/containerd.sock
 1327 root      0:00 /usr/bin/kmsg
```
The access control is based on whether the device is in the [cgroup whitelist](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/devices.html.) or not. I think restricting the perms might cause the reader to assume the perms are more important than they currently are.

I originally based this code on similar code in [moby](https://github.com/moby/moby/blob/master/oci/devices_linux.go), which allows the perms (and owner/group) of the devices to be customized. Perhaps it's more useful in `docker` where you are encouraged to run containers without root? It's an interesting question whether it's worth removing root from some of these packages. We already restrict the capabilities. I guess containers like `format` are ultimately trusted not to completely erase all system disks so perhaps being non-root wouldn't really help. It might be more useful for read-only things like `kmsg`. I'll have to think about it a bit more!

djs55:
@deitch I'm happy with it now. I think we can look again at the `runc` `ro` mount option problem when we get feedback on https://github.com/opencontainers/runc/issues/3248, but I don't think it's a blocker.

PR #3723, opened 2021-10-28
update-component-sha --pkg option

deitch:

**- What I did**

Added an option to `update-component-sha.sh` that allows you to do `--pkg`, so if you want to update everything to whatever the latest is in `pkg/init`, you just need to do:

```
./scripts/update-component-sha.sh --pkg ./pkg/init
```

**- How I did it**

Modified the shell script

**- How to verify it**

Run it. I did.

Easier to use update-component-sha

### DIFF
--- a/scripts/update-component-sha.sh
+++ b/scripts/update-component-sha.sh
@@ -11,71 +11,93 @@ usage() {
     cat >&2 <<EOF
 $0 --<mode> <how-to-find> <new-hash>
 
-Available modes: --hash and --image
+Available modes: --hash, --image, --pkg
 
 Replace by hash:
-	$0 --hash <OLD> <NEW>
-	Example: $0 --hash 8675309abcdefg abcdef567899
-    	   Will replace all instances of 8675309abcdefg with abcdef567899
+        $0 --hash <OLD> <NEW>
+        Example: $0 --hash 8675309abcdefg abcdef567899
+           Will replace all instances of 8675309abcdefg with abcdef567899
 
-Replace by image: $0 --image <IMAGE> <NEW>
-	$0 --image <IMAGE> <NEW>
-	Example: $0 --image linuxkit/foo abcdef567899
-	   Will tag all instances of linuxkit/foo with abcdef567899
+Replace by image:
+        $0 --image <IMAGE> <NEW>
+        Example: $0 --image linuxkit/foo abcdef567899
+           Will tag all instances of linuxkit/foo with abcdef567899
 
         $0 --image <IMAGE>:<NEW> is accepted as a convenient shortcut for cutting
         and pasting e.g.the output of linuxkit pkg show-tag
 
-By default, for convenience, if no mode is given (--image or --hash), the first method (--hash) is assumed. 
-Thus the following are equivalent:
-	$0 <OLD> <NEW>
-	$0 --hash <OLD> <NEW>
+Replace by pkg directory:
+        $0 --pkg <PATH/TO/PKG> <NEW>
+        Example: $0 --pkg ./pkg/xen-tools
+           Will use linuxkit pkg show-tag on the directory ./pkg/xen-tools, and then
+     tag all instances with the result
 
 EOF
 }
 
+updateImage() {
+        local image
+        local hash
+
+        case $# in
+        1)
+                image=${1%:*}
+                hash=${1#*:}
+                ;;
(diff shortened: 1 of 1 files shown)


scripts/update-component-sha.sh
```diff
@@ -11,71 +11,93 @@ usage() {
...
+        1)
+                image=${1%:*}
+                hash=${1#*:}
+                ;;
+        2)
+                image=$1
+                hash=$2
+                ;;
+        esac
+        git grep -E -l "[[:space:]]$image:" | grep -v /vendor/ | grep Dockerfile | xargs sed -i.bak -E -e "s,([[:space:]])($image):([^[:space:]]+), $image:$hash,g"
```
djs55, line 52: I made a change to `pkg/format` and committed it so I have
```
% linuxkit pkg show-tag pkg/format
linuxkit/format:438c3e4d4a6ee94e5d0d190ff4af1ba8224f766f
```
Unfortunately running `./scripts/update-component-sha.sh --pkg ./pkg/format` didn't do anything. I'm a bit suspicious of this line. When I run:
```
git grep -E -l "[[:space:]]linuxkit/format:"                                      
docs/external-disk.md
examples/cadvisor.yml
examples/dm-crypt-loop.yml
examples/dm-crypt.yml
examples/docker-for-mac.yml
examples/docker.yml
examples/swap.yml
projects/compose/compose-dynamic.yml
projects/compose/compose-static.yml
test/cases/030_security/000_docker-bench/test.yml
test/cases/040_packages/003_containerd/test.yml
test/cases/040_packages/005_extend/000_ext4/test-create.yml
test/cases/040_packages/005_extend/001_btrfs/test-create.yml
test/cases/040_packages/005_extend/002_xfs/test-create.yml
test/cases/040_packages/005_extend/003_gpt/test-create.yml
test/cases/040_packages/006_format_mount/000_auto/test.yml
test/cases/040_packages/006_format_mount/001_by_label/test.yml
test/cases/040_packages/006_format_mount/002_by_name/test.yml.in
test/cases/040_packages/006_format_mount/003_btrfs/test.yml
test/cases/040_packages/006_format_mount/004_xfs/test.yml
test/cases/040_packages/006_format_mount/005_by_device_force/test.yml
test/cases/040_packages/006_format_mount/006_gpt/test.yml
test/cases/040_packages/006_format_mount/010_multiple/test.yml
```
the `grep -v /vendor/` is ok, but the `grep Dockerfile` generates an empty list. If I remove the `grep Dockerfile` then it's almost perfect, except it also hits a few cases in `docs/external-disk.md` where we have examples containing `linuxkit/format:<hash>` -- perhaps .md files should be filtered out.

rn: fine with filtering out `.md` files or alternativelu match on hex after the `:`, though the latter will skip packages referred to by label/version.

deitch: 🤦‍♂️   I first did this in something else that I needed to work on Dockerfiles. So I copied `update-component-sha.sh`, worked it all out, then copied it back... and forgot to remove the `grep Dockerfile`.

Fixing.

deitch: OK, fixed that.

deitch: As for markdown. I haven't actually changed the grep logic in this PR, so it functions the way it did before.

I am going to suggest a different tack. I propose we leave this as is for now - this PR includes solely the addition of the `--pkg` option (working correctly, now that @djs55 found my mistake) - and then after we do another PR that restricts all of the changes in `update-component-sha.sh` to _just_ yaml files. Those are the only valid input files to `lkt build` anyways.

The only exception to that is the builders in `lkt` binary itself, specifically the list in [here](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/moby/output.go#L18-L31).

I propose that we do this as 3 PRs then, in the following order:

1. This one, just adding `--pkg` without changing the grep list (e.g. to exclude `.md` files)
2. One to extract the static list of images from [output.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/moby/output.go#L18-L31) and instead include them in a yaml file that gets included build-time via `go:embed`.
3. One to restrict `update-component-sha.sh` to only modify yaml files

Feedback, please.

djs55 (approved):
LGTM

PR #3724, opened 2021-10-28
service: handle creating cgroupv2 cgroups

djs55:
**- What I did**

Following on from #3719 which added cgroupv2 support, this PR fixes runtime config e.g.:
```
  runtime:
    cgroups:
    - systemreserved
    - podruntime
    - kubepods
```

Previously we would try to create cgroupv1-style entries, instead of cgroupv2.

**- How I did it**

To detect whether the system is running with cgroupv1 or cgroupv2, I used the suggestion from [stackexchange](https://unix.stackexchange.com/a/668244) about testing for the `cgroup.controllers` file which is also apparently Docker's recommended check.

When in cgroupv2 mode, all we need to do is a single `mkdir` to create the cgroup.

**- How to verify it**

I've extended the cgroupv2 test case to declare a `runtime: cgroups:` entry and check it is created.

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:5.4.30
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
-  - linuxkit/init:7e3d51e6ab5896ecb36a4829450f7430f2878927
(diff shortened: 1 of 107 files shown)


rn:
LGTM but looks like a rebase is needed

nustiueudinastea:
Hey @djs55 any chance we can get this merged?

nustiueudinastea:
Thank you so much @djs55!

PR #3715, opened 2021-09-29
Enable CONFIG_VIRTIO_FS

fredericdalleau:
Hello, for testing virtiofs I enabled CONFIG_VIRTIO_FS as suggested in https://virtio-fs.gitlab.io/howto-qemu.html

**- What I did**
Enable CONFIG_VIRTIO_FS=y in the configuration

**- How I did it**
Using sed, though I only tested kernel 5.10.47

**- How to verify it**
refer to https://virtio-fs.gitlab.io/.
It requires starting a virtiofs-enabled hypervisor and adding appropriate configuration to create the share. The virtiofs can then be mounted:
`$ mount -t virtiofs <tag> <mountpoint>`

**- Description for the changelog**
Enable CONFIG_VIRTIO_FS

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #3726, opened 2021-11-01
Run CI with updated hashes from #3715 (CONFIG_VIRTIOFS)

djs55:
This is a temporary PR to run the CI.


kernel/config-5.11.x-x86_64
```diff
@@ -421,8 +421,7 @@ CONFIG_X86_INTEL_TSX_MODE_OFF=y
 # CONFIG_X86_INTEL_TSX_MODE_AUTO is not set
 # CONFIG_X86_SGX is not set
 CONFIG_EFI=y
-CONFIG_EFI_STUB=y
-CONFIG_EFI_MIXED=y
+# CONFIG_EFI_STUB is not set
```
rn, line 424: This looks a bit odd to be disabled here and elsewhere. I can't remember but I think we may have used this on Windows.

djs55: @fredericdalleau was this an intended change or an accident?

fredericdalleau: Mmm weird, not something expected. I think I ran `make oldconfig` let me double check

fredericdalleau: So it turns out this has been introduced by cross generating the x86 configs files from arm64.
The reverse also true, if I cross generate arm64 configs from x86 I get a different output for the arms configs.
I'm reworking it, i'll update soon.

djs55: I've rebased this branch on @fredericdalleau 's latest, rebuilt the kernels (with `make push` on amd64 and arm64) and updated the shas for testing.

djs55: OK, the CI is green for the rebuilt kernels 😅 . WDYT @rn?

kernel/config-5.11.x-x86_64
```diff
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 5.11.22 Kernel Configuration
+# Linux/x86 5.11.4 Kernel Configuration
```
rn, line 3: This does not look right.

djs55: I was wondering about that too. The config currently has .22 but the Makefile mentions .4 https://github.com/linuxkit/linuxkit/blob/master/kernel/Makefile

fredericdalleau: My understanding was that the previous configuration was incorrect

rn: Ah, looked more closely. We used to have a 5.11.22 kernel but I removed it with:
```
1e97e29be338 ("kernel: Remove 5.11.x as it is not longer maintained")
```
we still keep the config file around, though, since the real-time kernel (`-rt`) is still based on 5.11.x. For normal kernel we do check that the generated config matches the checked-in config to make sure we do not accidentally enable/disable kernel options, but for debug (`-dbg`) and real-time (`-rt`) kernels we don't because we append stuff to the config and expect to have difference.

Long story short, this looks l right

rn (approved):

PR #3727, opened 2021-11-04
extract hard-coded default image builders into file

deitch:

**- What I did**

Extract the hard-coded image builders out of `moby/output.go` and into an `images.yaml` file, which is go-embedded.

Also required bumping go to 1.16, but we all use it anyways.

**- How I did it**

Create the new file, use go embedding

**- How to verify it**

Let CI run. I ran manually, both to build an image and to debug step through to make sure it gets created correctly. I might have missed a case (don't think so, but that is what CI is there for).

Extract hard-coded builder images from go file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.13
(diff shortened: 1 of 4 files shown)


djs55 (approved):
LGTM

deitch:
CI is running. Will merge when done.

PR #3728, opened 2021-11-05
restrict update to just yaml files

deitch:

**- What I did**

Restrict `update-component-sha.sh` to only change files that end in `yml`/`yaml`/`yml.in`/`yaml.in`, as discussed [here](https://github.com/linuxkit/linuxkit/pull/3723#discussion_r739785144) with @djs55 and @rn 

By the way, this is not _strictly_ necessary. It would continue to work without this. But it seems to make sense.

**- How I did it**

Modified the single script.

**- How to verify it**

Run it. It works. I did. 

update-component-sha.sh only affects yaml files

**Note:** I found one other place where we have the hashes embedded in a `.go` file, PR for that should go in first #3729

### DIFF
--- a/scripts/update-component-sha.sh
+++ b/scripts/update-component-sha.sh
@@ -49,7 +49,7 @@ updateImage() {
                 hash=$2
                 ;;
         esac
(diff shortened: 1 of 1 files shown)


deitch:
I extended it to include Dockerfile and Makefile, as we need those for alpine updates.

dave-tucker (approved):
LGTM

rn (approved):

PR #3729, opened 2021-11-05
extract more hard-coded yaml

deitch:

**- What I did**

Continuing #3727 , this extracts one more chunk of hard-coded yaml out of a `.go` file and places it in a yaml that is pulled in using go embed.

**- How I did it**

Extract and go embed.

**- How to verify it**

Ran manually, ran via debugger to see that it was populated, let CI run.

Last hard-coding out of go files.


dave-tucker (approved):
LGTM

Issue #3738, opened 2021-11-28
acpid: operation not permitted


sprat:
**Description**

I am pretty sure that ACPID does not work properly, my machine does not stop when I hit the power button and I've seen a problem in the ACPID logs:
```
(ns: sshd) linuxkit-b8aeed7efdbc:/# cat /var/log/acpid.log
acpid: /dev/input/event0: Operation not permitted
```
I suspect that it is due to the runc version update that was performed recently: acpid probably need to be allowed to access some devices in order to work. However, I have no idea how to fix that: which devices need to be enabled?

**Steps to reproduce the issue:**

Here is a part of my configuration:
```
kernel:
  image: linuxkit/kernel:5.10.76
  cmdline: console=tty0
  ucode: intel-ucode.cpio
init:
  - linuxkit/init:eb597ef74d808b5320ad1060b1620a6ac31e7ced
  - linuxkit/runc:21dbbda709ae138de0af6b0c7e4ae49525db5e88
  - linuxkit/containerd:2f0907913dd54ab5186006034eb224a0da12443e
services:
  - name: acpid
    image: linuxkit/acpid:e9a94e593d6be2fc1b3eb6d566f23ac9ca9807fd
```

**Describe the results you received:**

There's an error in /var/log/acpid.log

**Describe the results you expected:**

My machine should stop when I hit the power button.

**Additional information you deem important (e.g. issue happens only occasionally):**

None

sprat:
It seems to work if I add
```
    devices:
      - path: all
        type: c
```
But maybe this can be restricted a little bit

sprat:
Someone can confirm if we need to allow access to more or less devices? I am not very familiar with acpid. Then I will make a pull request to fix that.

sprat:
Anyone?

PR #3740, opened 2021-12-09
Fix acpid devices

sprat:
**- What I did**

I fixed the bug #3738 by adding missing "devices" field on the acpid `build.yml`

**- How I did it**

By typing on my keyboard

**- How to verify it**

I checked the fix manually on real hardware, I can cleanly shutdown the machine by pressing the power button shortly.

I also checked the acpid logs, and verified that there was no errors

Add missing "devices" field on the acpid service to allow input devices

tapir

### DIFF
--- a/pkg/acpid/build.yml
+++ b/pkg/acpid/build.yml
@@ -1,5 +1,9 @@
 image: acpid
 config:
   binds:
-  - /dev:/dev
+    - /dev:/dev
+  devices:
+    - path: all
+      type: c
+      major: 13
   pid: host
--- a/examples/docker-for-mac.yml
+++ b/examples/docker-for-mac.yml
@@ -49,7 +49,7 @@ onboot:
 services:
   # Enable acpi to shutdown on power events
   - name: acpid
-    image: linuxkit/acpid:e9a94e593d6be2fc1b3eb6d566f23ac9ca9807fd
+    image: linuxkit/acpid:c05a368754f6436b326945dc16135ba547568d8d
   # Enable getty for easier debugging
   - name: getty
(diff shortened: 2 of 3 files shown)


rn:
Thanks. I build/pushed the updated images and kicked CI

sprat:
Thanks! Seems to be green ;)

pkg/acpid/build.yml
```diff
@@ -1,5 +1,9 @@
 image: acpid
 config:
   binds:
-  - /dev:/dev
+    - /dev:/dev
+  devices:
+    - path: all
+      type: c
+      major: 13
```
rn, line 8: could you please explain in the commit message why this fixes it or if there potentially can be a smaller set up devices mapped?

sprat: Sure, I can amend the commit message. But as you can see in the bug, I'm not really sure which devices needs to be allowed for ACPId. I know that "input" devices have a major number set to 13, but I don't know if ACPId needs to access other types of devices. Anyway, it seems to be ok on my machine, that's why I pushed this change. Even if it's not enough, it's still better than the current state of ACPID configuration.

sprat: I looked into busybox' acpid code, and it seems that it needs access to /dev/input/event* only. So I guess that allowing all "input" devices is ok. So I amended the relevant commit, should be ok now.

sprat:
I think it should be ok now, can someone review my changes and merge them?

rn (approved):
Thanks for the commit update (and the original fix)

sprat:
Thanks!

PR #3745, opened 2021-12-21
Upgrade runc to 1.0.3

fredericdalleau:
**- What I did**

- Update runc package run 1.0.3
- Remove the RO workaround that prevented container start
- update all examples and yamls to use images of new runc (with help from @djs55)

**- How I did it**

- update the version number
- revert the workaround patch (git revert)
- run scripts/update-component-sha.sh

**- How to verify it**
First ensure the RO issues is still a reality on runc 1.0.2 by checking out only the revert (only the first patch of this PR) and then using the command below :
 `cd ~/linuxkit/src/cmd/linuxkit && go build . && ./linuxkit build ../../../linuxkit.yml && ./linuxkit run qemu -kernel linuxkit`

The system will boot and show error.
```
INFO[2021-12-21T15:26:42.136250559Z] shim disconnected                             id=rngd
ERRO[2021-12-21T15:26:42.136718017Z] copy shim log                                 error="read /proc/self/fd/23: file already closed"
ERRO[0000] failed to create task                         error="OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting \"devpts\" to rootfs at \"/dev/pts\" caused: mkdir /containers/services/rngd/rootfs/dev/pts: read-only file system: unknown"
time="2021-12-21T15:26:42.141640559Z" level=info msg="starting signal loop" namespace=services.linuxkit path=/run/containerd/io.containerd.runtime.v2.task/services.linuxkit/shutdown pid=525
```
We can see in `ctr -n services.linuxkit t ls` that rngd service failed to start.

Then checking out this PR to upgrade runc and upgrade the runc package SHA in all recipes, and run the command line again.
No error message appears.
The rngd container started (and stopped) properly.

Upgrade runc to 1.0.3

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -3,7 +3,7 @@ kernel:
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
   - linuxkit/init:eb597ef74d808b5320ad1060b1620a6ac31e7ced
-  - linuxkit/runc:21dbbda709ae138de0af6b0c7e4ae49525db5e88
+  - linuxkit/runc:9f7aad4eb5e4360cc9ed8778a5c501cce6e21601
   - linuxkit/containerd:2f0907913dd54ab5186006034eb224a0da12443e
   - linuxkit/ca-certificates:c1c73ef590dffb6a0138cf758fe4a4305c9864f4
 onboot:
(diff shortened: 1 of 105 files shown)


rn (approved):
Thanks for following up on the fix

PR #3747, opened 2021-12-22
Fix sshd

sprat:
**- What I did**

Fixed some problems with sshd package:
- allow access to all devices
- fix a tini warning

**- How I did it**

No devices appeared when running blkikd, had a hard time to figure out the problem
By checking the logs, I also noticed a warning with tini because it does not run with pid 0, so I fixed the problem as described in the tini warning by adding the "-s" flag

**- How to verify it**

CI

SSHd: allow access to all devices + fixed a warning with tini

https://www.santelog.com/sites/santelog.com/www.santelog.com/files/images/accroche/adobestock_276208008_lama.jpeg

### DIFF
--- a/examples/openstack.yml
+++ b/examples/openstack.yml
@@ -19,7 +19,7 @@ services:
   - name: rngd
     image: linuxkit/rngd:a44c6230f1620f1d6310b95a9e9f585de73a8bd7
   - name: sshd
-    image: linuxkit/sshd:782720b1c320f4b31abd89711ac235d1d441217c
+    image: linuxkit/sshd:4696ba61c3ec091328e1c14857d77e675802342f
     binds:
      - /run/config/ssh/authorized_keys:/root/.ssh/authorized_keys
   - name: nginx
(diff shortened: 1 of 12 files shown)


rn:
Thanks for tracking this down. I'm not sure we want to give access to all device by default. In my opinion we want the packages locked down as much as possible by default and then use the image YAML to relax it on a per case basis. This is, for example, what we do with getty.

@djs55 @justincormack @deitch any strong opinion either way?

deitch:
I'm mixed on this one. In general, I would prefer locked down. On the other hand, the purpose of the sshd package is administrative access. If it cannot access devices, it isn't much good. If everyone will have to enable it, does that mean it might as well have it as a default?

rn:
@deitch fair point. 

@sprat looks like the second commit for updating the hashes has no signed-off-by line. Could you force push an update?

sprat:
I agree with @deitch: after noticing that the sshd service has full priviledges & host pid already, I think that we should also grant access to devices too, for administrative tasks. It was the case before the introduction of per-services "devices" configuration.

@rn : I've pushed-force to add the signef-off line, I should have enabled that by default.

rn (approved):

sprat:
Thanks!

PR #3751, opened 2022-01-10
Update Moby schema for architecture

### LINKED ISSUE #3750: Architecture parameter appears to break external consumers of `moby.Moby`
I import the linuxkit package in resinstack/terraform-provider-linuxkit, and it appears that the architecture parameter which was added in the cache changes (0b7502f130) which has no yaml tags is breaking the ability to import the config.  This key doesn't have a yaml tag to suppress it, so when I generate the yaml I am exporting it with an `architecture` value that's empty.  This gets picked up by the import as an invalid additional property.

I think that the easy solution here is to just tag the attribute, but so far in my testing this hasn't worked.   Any additional suggestions or debugging ideas would be appreciated.

the-maldridge:

Resolves #3750 

**- What I did**

Updated the goschema schema.

**- How I did it**

Sheer dumb luck and guesswork.

**- How to verify it**

Honestly not sure, this really should have broken some test when the cache PR was merged, and the fact that it didn't really is a defect in and of itself.

Provide schema value for architecture parameter.

The schema still includes lots of references to trust config.  I would argue that these should be removed, but that would break anyone who has trust stanzas in their config.  Whether or not that's something to care about is another matter, and I'd argue that since the default for linuxkit is to build from master it isn't something to worry about.

### DIFF
--- a/src/cmd/linuxkit/moby/config.go
+++ b/src/cmd/linuxkit/moby/config.go
@@ -24,7 +24,7 @@ type Moby struct {
 	Onshutdown   []*Image     `yaml:"onshutdown" json:"onshutdown"`
 	Services     []*Image     `yaml:"services" json:"services"`
 	Files        []File       `yaml:"files" json:"files"`
-	Architecture string
+	Architecture string       `yaml:"architecture" json:"architecture"`
 
 	initRefs []*reference.Spec
 }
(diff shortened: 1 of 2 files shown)


deitch:
As discussed in #3750 , you are right that it shouldn't break it, but the fix is not to put it in here, but rather to pass the arg separately. `type Moby` should not have the property `Architecture`. It was put there to make it easier to pass around, but without thinking of the yaml consequences.

the-maldridge:
Obsoleted by 5b7466732a90aa7394112906c0ce769fda2605c9.

PR #3752, opened 2022-01-10
pass target architecture explicitly

### LINKED ISSUE #3750: Architecture parameter appears to break external consumers of `moby.Moby`

deitch:

**- What I did**

Fixed #3750 

`type Moby` contained a field `Architecture` to pass around what arch we are building for. However, that does not correctly reflect the yaml file, which should be usable across architectures. It only was added as a convenience, to avoid passing another parameter to several functions.

This removes that non-yaml property and passes it around explicitly where needed.

**- How I did it**

Removed the field, added it as arg to several functions.

**- How to verify it**

CI is your friend.

Align Moby type with yaml by moving extraneous fields out and passing explicitly.

### DIFF
--- a/src/cmd/linuxkit/build.go
+++ b/src/cmd/linuxkit/build.go
@@ -180,7 +180,6 @@ func build(args []string) {
 		if err != nil {
 			log.Fatalf("Invalid config: %v", err)
 		}
-		c.Architecture = *buildArch
 		m, err = moby.AppendConfig(m, c)
 		if err != nil {
 			log.Fatalf("Cannot append config files: %v", err)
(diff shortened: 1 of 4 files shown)


dave-tucker (approved):
LGTM

PR #3753, opened 2022-01-10
replace Build many args with opts struct

deitch:

**- What I did**

Replaced the signature of `Moby.Build(lots, of args, here, and, keep, changing)` with an `BuiltOpts struct`

**- How I did it**

Replace, test

**- How to verify it**

CI

Simpler `moby.Build` signature

**- Notes for maintainers**

I am not sure how much this buys. We never really intended `moby.Build()` to be used by the public, but it is; implicit API and all that.

This does clean it up at least somewhat, by giving it a signature of:

```go
func Build(m Moby, w io.Writer, opts BuildOpts) error {
```

That strikes me as better and easier to understand.

However, once inside `moby.Build`, we call `ImageTar()` once and `outputImage()` 3 times, each of which just takes lots of those parameters.

The question for maintainers:

1. Is this worth it? I think yes, but open question.
2. Should we also change how we pass things to the non-exported `outputImage()` and the exported `ImageTar()`?

### DIFF
--- a/src/cmd/linuxkit/moby/build.go
+++ b/src/cmd/linuxkit/moby/build.go
@@ -83,14 +83,14 @@ func OutputTypes() []string {
 	return ts
 }
 
-func outputImage(image *Image, section string, prefix string, m Moby, idMap map[string]uint32, dupMap map[string]string, pull bool, iw *tar.Writer, cacheDir string, dockerCache bool, arch string) error {
+func outputImage(image *Image, section string, prefix string, m Moby, idMap map[string]uint32, dupMap map[string]string, iw *tar.Writer, opts BuildOpts) error {
 	log.Infof("  Create OCI config for %s", image.Image)
 	imageName := util.ReferenceExpand(image.Image)
 	ref, err := reference.Parse(imageName)
 	if err != nil {
 		return fmt.Errorf("could not resolve references for image %s: %v", image.Image, err)
 	}
-	src, err := imagePull(&ref, pull, cacheDir, dockerCache, arch)
+	src, err := imagePull(&ref, opts.Pull, opts.CacheDir, opts.DockerCache, opts.Arch)
 	if err != nil {
 		return fmt.Errorf("Could not pull image %s: %v", image.Image, err)
 	}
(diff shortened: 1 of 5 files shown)


src/cmd/linuxkit/moby/build.go
```diff
@@ -116,7 +116,8 @@ func outputImage(image *Image, section string, prefix string, m Moby, idMap map[
 }
 
 // Build performs the actual build process
-func Build(m Moby, w io.Writer, pull bool, tp string, decompressKernel bool, cacheDir string, dockerCache bool, arch string) error {
+func Build(m Moby, w io.Writer, opts BuildOpts) error {
+	//pull bool, tp string, decompressKernel bool, cacheDir string, dockerCache bool, arch string
```
dave-tucker, line 120: I think this can go

src/cmd/linuxkit/moby/opts.go
```diff
@@ -0,0 +1,10 @@
+package moby
+
+type BuildOpts struct {
```
dave-tucker, line 3: If the struct is to be part of an API, I would document it too.

dave-tucker (changes requested):

dave-tucker:
@deitch 

> The question for maintainers:
>
> Is this worth it? I think yes, but open question.

IMO yes. It's cleaner to use a struct here and it'll keep the API ~stable across releases

> Should we also change how we pass things to the non-exported outputImage() and the exported ImageTar()?

I would say also yes. They may as well take `BuildOpts` as an argument.

src/cmd/linuxkit/moby/build.go
deitch: Oh, yeah, that was to make it easier for me to convert things. Oops.

src/cmd/linuxkit/moby/opts.go
deitch: Fair enough.

deitch:
Done.

There is a part of me that wants to create opts-type structs for some of the others, but I think this is good enough for now. Incremental improvements. Back at you @dave-tucker

dave-tucker (approved):
LGTM thanks @deitch

Issue #3756, opened 2022-01-15
Docker example does not work


the-maldridge:
**Description**

The docker example appears broken.  The config.json doesn't appear to be included in the build, among other issues.  

**Steps to reproduce the issue:**
1. Build linuxkit from master.  I built from `1df038e1b`.
2. Copy the docker.yml example to a new empty directory.
3. Build:

```
./linuxkit build docker.yml 
Extract kernel image: docker.io/linuxkit/kernel:5.10.76
Add init containers:
Process init image: docker.io/linuxkit/init:7e3d51e6ab5896ecb36a4829450f7430f2878927
Image docker.io/linuxkit/init:7e3d51e6ab5896ecb36a4829450f7430f2878927 not found in local cache, pulling
Process init image: docker.io/linuxkit/runc:9f7aad4eb5e4360cc9ed8778a5c501cce6e21601
Process init image: docker.io/linuxkit/containerd:2f0907913dd54ab5186006034eb224a0da12443e
Process init image: docker.io/linuxkit/ca-certificates:c1c73ef590dffb6a0138cf758fe4a4305c9864f4
Add onboot containers:
  Create OCI config for linuxkit/sysctl:bdc99eeedc224439ff237990ee06e5b992c8c1ae
Image docker.io/linuxkit/sysctl:bdc99eeedc224439ff237990ee06e5b992c8c1ae not found in local cache, pulling
  Create OCI config for linuxkit/sysfs:c3bdb00c5e23bf566d294bafd5f7890ca319056f
Image docker.io/linuxkit/sysfs:c3bdb00c5e23bf566d294bafd5f7890ca319056f not found in local cache, pulling
  Create OCI config for linuxkit/format:7efa07559dd23cb4dbebfd3ab48c50fd33625918
Image docker.io/linuxkit/format:7efa07559dd23cb4dbebfd3ab48c50fd33625918 not found in local cache, pulling
  Create OCI config for linuxkit/mount:422b219bb1c7051096126ac83e6dcc8b2f3f1176
Image docker.io/linuxkit/mount:422b219bb1c7051096126ac83e6dcc8b2f3f1176 not found in local cache, pulling
Add service containers:
  Create OCI config for linuxkit/getty:3c6e89681a988c3d4e2610fcd7aaaa0247ded3ec
  Create OCI config for linuxkit/rngd:4f85d8de3f6f45973a8c88dc8fba9ec596e5495a
[...]

PR #3761, opened 2022-01-29
pkg/runc: Revert to 1.0.0-rc90

the-maldridge:
This commit reverts runc to version 1.0.0-rc90 which is the last known
version to not have any degraded functionality with linuxkit.  The
first known broken version is 1.0.0-rc91, and a bisect between these
two bounds shows the breakage first appearing at
60e21ec26e15945259d4b1e790e8fd119ee86467.  As I don't have the time
nor knowledge to debug this further, my best suggestion I can offer is
to revert to the most recent version that worked.

**- What I did**
Reverted runc

**- How I did it**
by typing?

**- How to verify it**
build the package, observe that the version number is different.

  * Revert runc due to cgroup regressions.

### DIFF
--- a/pkg/runc/Dockerfile
+++ b/pkg/runc/Dockerfile
@@ -12,7 +12,7 @@ RUN \
   make \
   && true
 ENV GOPATH=/go PATH=$PATH:/go/bin GO111MODULE=off
-ENV RUNC_COMMIT=v1.0.3
+ENV RUNC_COMMIT=v1.0.0-rc90
 RUN mkdir -p $GOPATH/src/github.com/opencontainers && \
   cd $GOPATH/src/github.com/opencontainers && \
   git clone https://github.com/opencontainers/runc.git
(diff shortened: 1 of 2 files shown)


djs55:
Thanks for the extra information. I think the next step I’d like to take is
to add a CI test which demonstrates the bug and then we can more easily
test potential fixes, and prevent problems recurring in future.

the-maldridge:
Sure, is there an existing framework within linuxkit for tests that boot VMs or is there some other paradigm you'd like to see for a test here?  The way I did the bisect was by building a VM with a suspect version in it and looking to see if acpid is running or not since that failed at the same time as dockerd did.

djs55:
The tests are shell scripts run by a python tool called “rtf”. For example
here’s a cgroupv2 test:
https://github.com/linuxkit/linuxkit/blob/master/test/cases/040_packages/003_cgroupv2/test.sh
They’re run by the GitHub actions so hopefully the logs are visible. The
biggest gotcha at the moment is that changes to packages (eg runc) are not
automatically picked up because all the image hashes are pinned. In the
past I’ve pushed an image to somewhere I have write access to, then
reference that temporarily in the PR as a test. This is a limitation of the
CI setup that we’re hoping to improve.

the-maldridge:
While writing the test I came across that something in linuxkit is generally pretty broken right now.  Please review and comment on #3756.

the-maldridge:
Obsoleted by the information provided in #3735

PR #3771, opened 2022-03-19
Containerd separate

deitch:
**- What I did**

* separated the download and build of containerd from `tools/alpine` into its own package `linuxkit/containerd-dev`
* update dependencies on containerd in alpine - `pkg/containerd` and `pkg/init` - to source from `linuxkit/containerd-dev`
* update containerd to 1.6.1

This enables us to update containerd versions in the future without depending on updating alpine (i.e. we get more modular). This becomes important, e.g., when we want to bump containerd to address a CVE or bug rapidly without having to update the entire alpine.

**- How to verify it**

CI

containerd build standalone, containerd to 1.6.1


deitch:
I need to push out `linuxkit/init` again after the rebase, then push out the downstream `test/`, `examples/`, etc. I will comment here when this is ready for proper review and testing.

deitch:
OK, pushed out an updated version of `linuxkit/init`, updated all of the dependencies here (`examples/`, `test/` etc.), CI has been kicked off. This is ready for review.

deitch:
Strange. I reran all of the tests (test run [here](https://github.com/linuxkit/linuxkit/actions/runs/2009401313)), and it was green, but it will not update this page.

This is green and good to go, pending review.

djs55 (approved):
Sounds like a good idea! If the tests are green then I'm happy.

rn:
Seem like a good thing to do. 

If `pkg/containerd-dev` not supposed to be used as a final package, but more like a package used to build other packages, I think `./tools/containerd-dev` would be a better place.

Also, https://github.com/linuxkit/linuxkit/tree/master/test/pkg/containerd needs updating.

deitch:
Yeah, I can go for it being in `tools/` rather than `pkg/`. And since it doesn't change the `pkg/` tag, I won't need to push it out.

Good catch on `test/pkg/`

deitch:
OK, moved to `tools/containerd-dev/`; as expected, the pkg hash didn't change.

Also updated `test/pkg/containerd/`.

This forces a CI run, which hopefully will be green, and we can merge in.

I do need a new review after the changes, though @djs55 @rn

deitch:
CI nice and green now.

djs55 (approved):

PR #3775, opened 2022-04-21
add docs about how to find the builder

deitch:

**- What I did**

Added docs about the three ways to find the kernel builder image

**- How I did it**

Updated `docs/kernels.md`

**- How to verify it**

Just read it; it is docs.

Document how to find the builder for kernels.

### DIFF
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -45,22 +45,36 @@ Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a m
 ## Compiling external kernel modules
 
 This section describes how to build external (out-of-tree) kernel
-modules. It is assumed you have the source available to those modules,
-and require the correct kernel version headers and compile tools.
+modules. You need the following to build external modules. All of
+these are to be built for a specific version of the kernel. For
+the examples, we will assume 5.10.104; replace with your desired
+version.
 
-The LinuxKit kernel packages include `kernel-dev.tar` which contains
+* source available to your modules - you need to get those on your own
+* kernel development headers - available in the `linuxkit/kernel` image as `kernel-dev.tar`, e.g. `linuxkit/kernel:5.10.104`
+* OS with sources and compiler - this **must** be the exact same version as that used to compile the kernel
+
+As described above, the `linuxkit/kernel` images include `kernel-dev.tar` which contains
 the headers and other files required to compile kernel modules against
 the specific version of the kernel. Currently, the headers are not
 included in the initial RAM disk, but it is possible to compile custom
 modules offline and then include the modules in the initial RAM disk.
 
-There is a [example](../test/cases/020_kernel/011_kmod_4.9.x), but
+The source is available as the same name as the `linuxkit/kernel` image, with the addition of `-builder` on the tag.
+For example:
+
+* `linuxkit/kernel:5.10.92` has builder `linuxkit/kernel:5.10.92-builder`
+* `linuxkit/kernel:5.15.15` has builder `linuxkit/kernel:5.15.15-builder`
+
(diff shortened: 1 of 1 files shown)


docs/kernels.md
```diff
@@ -45,22 +45,42 @@ Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a m
...
 This section describes how to build external (out-of-tree) kernel
-modules. It is assumed you have the source available to those modules,
-and require the correct kernel version headers and compile tools.
+modules. You need the following to build external modules. All of
+these are to be built for a specific version of the kernel. For
+the examples, we will assume 5.10.104; replace with your desired
+version.
+
+* source available to your modules - you need to get those on your own
+* kernel development headers - available in the `linuxkit/kernel` as `kernel-dev.tar`, e.g. `linuxkit/kernel:5.10.104`
```
rn, line 54: nit: "In the linuxkit/kernel" image

docs/kernels.md
```diff
@@ -45,22 +45,42 @@ Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a m
...
+* OS with sources and compiler - this **must** be the exact same version as that used to compile the kernel
+
+You have 3 places to determine the correct version of the OS image with sources and compiler:
+
+* `-builder` tag added to the kernel version, e.g. `linuxkit/kernel:5.10.104-builder`
+* labels on the kernel image, e.g. `docker inspect linuxkit/kernel:5.10.104 | jq -r '.[].Config.Labels["org.mobyproject.linuxkit.kernel.buildimage"]'`
+* `/kernel-builder` file in the kernel image
+
+Use any of the above the the `<builder>`.o
```
rn, line 63: nit, double "the". Also not a complete sentence

docs/kernels.md
```diff
@@ -45,22 +45,42 @@ Most kernel modules are autoloaded with `mdev` but if you need to `modprobe` a m
...
-There is a [example](../test/cases/020_kernel/011_kmod_4.9.x), but
+There is an [example](../test/cases/020_kernel/011_kmod_4.9.x), but
 basically one can use a multi-stage build to compile the kernel
 modules:
 
 ```
-FROM linuxkit/kernel:4.9.33 AS ksrc
-FROM linuxkit/alpine:<hash> AS build
+FROM linuxkit/kernel:5.10.104 AS ksrc
+FROM linuxkit/alpine:2be490394653b7967c250e86fd42cef88de428ba AS build
```
rn, line 81: should we default to the `-builder` case. It seems more intuitive and requires one less step, ie does not require looking up the alpine hash? The alpine hash  should only be necessary of we retag or something

rn (commented):
Thanks for adding/updating the docs. As mentioned in one of the comments, I think it might easier to write/understand if we just default to the `-builder` images and only use the alpine images as backup. They are more like a implementation detail

docs/kernels.md
deitch: Good catch. Updated.

docs/kernels.md
deitch: Yeah, strange that. Fixed.

docs/kernels.md
deitch: Yes, that not only is a good point, but is exactly what it is there for. The others are backups... which turned out to be a good thing. I had to track one down last week that was missing the `-builder` tag, and could not remember how to do it. This saved me (and prompted me to write this doc).

I will reword this.

deitch: Done

deitch:
Restructured it to reference the `-builder` as primary. Take another look @rn

rn (approved):

PR #3778, opened 2022-05-13
Fix a merge conflict in openstack.yml example

NeilGerring:
In openstack.yml, there was the left overs from a merge conflict that hadn't been resolved properly. This PR resolves it by using the current change as the change looks to be the same.


djs55 (approved):
LGTM, thanks

PR #3781, opened 2022-05-24
remove need for maintainers to build packages

deitch:

**- What I did**

Removed the need for maintainers to push out packages.

**- How I did it**

2 changes:

1. On PR, it already did `lkt pkg build`, which forces it to download the image to lkt cache or, if not there, build it (or fail). Here we just need to tell maintainers "do NOT push it", which this does. Eventually, we should remove manual pushing capability.
2. Added a new GH actions workflow file that, when the build ci.yaml is done, and we are on `master` or `main` branch, pushes out every package. Since only changed ones will not already be in registry, this will push out only the changed ones.

**- How to verify it**

Requires a new PR that changes a package, have no one push it out, see that it works.

Eliminate the need for maintainers to manually build and push out packages.

**- some comments**

I am not 100% sure that this works, or if the approach is correct. It may slow things down, since it now depends on emulated builds, whereas maintainers were using Packet build servers.

On the other hand, getting human hands out of the critical path *absolutely* is the right thing.

Looking forward to feedback.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,42 @@
+# publish changes that are merged to master
+name: Packages Push
+on:
+  workflow_run:
+    workflows: [LinuxKit CI]
+    types: [completed]
+    branches: [master, main]
+
+jobs:
+  packages:
+    name: Publish Changed Packages
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v1
+      with:
+        path: ./src/github.com/linuxkit/linuxkit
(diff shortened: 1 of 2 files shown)


djs55 (approved):
This looks great. I think we should merge and test it.

deitch:
I was thinking, once we merge it in, we create a `pkg/dummy/` and we can test it there

deitch:
Actions post-merge-to-master is running. Letting it complete, then I will push a pkg/dummy.

deitch:
See #3783 to fix this. It might not work, it might only solve the download issue, leaving caching an issue. One issue at a time.

PR #3783, opened 2022-05-25
download artifact from other workflow

deitch:

**- What I did**

Fix the cross-workflow download of artifact

**- How I did it**

Replace the standard GitHub `actions/download-artifact` (which does not support cross-workflow downloads) with an explicit `actions/github-script` which does.

**- How to verify it**

CI will not change. We will need to merge into `master` to know that it works.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -16,10 +16,26 @@ jobs:
       with:
         path: ./src/github.com/linuxkit/linuxkit
     - name: Download linuxkit
-      uses: actions/download-artifact@v2
+      uses: actions/github-script@v3.1.0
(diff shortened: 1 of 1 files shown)


deitch:
CI is green, this one is ready to go. Until this gets merged in, CI will be broken.

rn (approved):
I would be lying if I said I'd understand all of this, but since CI is green and this kinda makes sense LGTM

deitch:
Ok merged in. Only when merge-to-master Actions run to completion will we know if it worked.

deitch:
Welp, it didn't fix the problem. I'll keep digging. Expect new PRs soon.

PR #3788, opened 2022-06-07
test pkg/dummy

deitch:
Just a test of the build system. Will let CI run, tests should pass, and then will merge to PR, should push out new `linuxkit/dummy` image.

Once it is done, I will open a PR to delete `pkg/dummy` and manually delete the image.

### DIFF
--- a/pkg/Makefile
+++ b/pkg/Makefile
@@ -1,4 +1,4 @@
-DIRS = $(dir $(shell find . -maxdepth 2 -mindepth 2 -type f -name build.yml))
+DIRS = $(dir $(shell find . -maxdepth 2 -mindepth 2 -type f -name build.yml | sort))
 .PHONY: push force-push build forcebuild show-tag clean
 
 OPTIONS ?=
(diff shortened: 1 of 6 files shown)


rn:
Might be worth actually having unit tests for the various `build.yml` options and check for regressions in `linuxkit build`. They probably should not be in `pkg` but somewhere else under tests

rn (approved):

deitch:
I figured it out. labels. Had to dig into the rtf code.

Going to let CI re-run, but it needs a new review.

deitch:
The CI failed, because it couldn't find the `linuxkit/dummy` image. Which, for some reason, was not built.

Building packages starts [here](https://github.com/linuxkit/linuxkit/runs/6775769164?check_suite_focus=true#step:7:1). It calls 

```
make OPTIONS="--skip-platforms linux/s390x" -C pkg build
```

and only seems to build 19 of the packages, not including `dummy`, hence the issue. When I run `make -n OPTIONS="--skip-platforms linux/s390x" -C pkg build`, it wants to build 41 packages (which is all of them in `pkg/`).

If we can figure out why, we can get it to build correctly.

deitch:
I updated the `Makefile` to echo the build line, it now gives:

```
linuxkit pkg build --skip-platforms linux/s390x ./sysfs/ ./qemu-ga/ ./binfmt/ ./open-vm-tools/ ./metadata/ ./tss/ ./dhcpcd/ ./auditd/ ./format/ ./firmware/ ./trim-after-delete/ ./swap/ ./vpnkit-expose-port/ ./runc/ ./host-timesync-daemon/ ./logwrite/ ./openntpd/ ./node_exporter/ ./modprobe/ ./bpftrace/ ./memlogd/ ./ca-certificates/ ./extend/ ./dummy/ ./firmware-all/ ./rngd/ ./sysctl/ ./mount/ ./acpid/ ./dm-crypt/ ./kmsg/ ./containerd/ ./vpnkit-forwarder/ ./sshd/ ./ip/ ./vsudd/ ./losetup/ ./mkimage/ ./getty/ ./cadvisor/ ./init/
```

This has all 41, which means that it isn't the Makefile or make target, but the actual build that is failing silently. Time to investigate why.

deitch:
I figured it out. New PR on its way.

deitch:
The build is working correctly this time, post #3789. Once CI is green, we can merge in and be sure that merge-to-master works correctly.

PR #3789, opened 2022-06-07
when no archs for pkg to build, should continue rather than return

deitch:

**- What I did**

#3787 skips a build if it has no archs. Which is good. Except instead of continuing to the next package, it exits.
This fixes it.

**- How I did it**

Change `return` for `continue`

**- How to verify it**

CI, and I ran it manually

Continue for all packages after skipping packages with no architectures


deitch:
I can confirm from CI that this fixes it.

djs55 (approved):
OK!

Issue #3799, opened 2022-07-08
Broken build in latest (806c5c418df3753f4d5a62231a036009422f71f3)


waldner:
**Description**

Building the tool with `make` fails.

**Steps to reproduce the issue:**

Just run `make` at the root of the latest tree.

**Describe the results you received:**

```
$ make
mkdir -p bin
make -C ./src/cmd/linuxkit
make[1]: Entering directory '/tmp/linuxkit/src/cmd/linuxkit'
tar cf - -C . . | docker run --rm --net=none --log-driver=none -i  linuxkit/go-compile:7b1f5a37d2a93cd4a9aa2a87db264d8145944006 --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit=806c5c418df3753f4d5a62231a036009422f71f3 -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version="v0.8+"" -o linuxkit > tmp_linuxkit_bin.tar
gofmt...
govet...
golint...
./pkglib/docker.go:27:2: a blank import should be only in a main or test package, or have a comment justifying it
sh: a: unknown operand
make[1]: *** [Makefile:34: tmp_linuxkit_bin.tar] Error 2
make[1]: *** Deleting file 'tmp_linuxkit_bin.tar'
make[1]: Leaving directory '/tmp/linuxkit/src/cmd/linuxkit'
make: *** [Makefile:58: linuxkit] Error 2
```

**Describe the results you expected:**

Build should complete successfully.

deitch:
I find it so interesting that previous CI passed when that was changed. How odd. I will comment o the PR.

PR #3743, opened 2021-12-14
fix: #3742 golint requires comments on non-main package for blank import

aoxn:
Signed-off-by: 谢瑶瑶 <yaoyao.xyy@alibaba-inc.com>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3800, opened 2022-07-09
Fix: #3799 golint requires comments on non-main package for blank import

waldner:
Signed-off-by: Davide Brini <waldner@katamail.com>


deitch:
Of course, it would be my commit that broke this. The original PR is [here](https://github.com/linuxkit/linuxkit/pull/3791). What I find surprising is that CI was clean for that; checks are [here](https://github.com/linuxkit/linuxkit/pull/3791/checks).

Either way, we should fix it.

deitch (changes requested):
Can you remove the line:

```go
// fix: #3799 
```

We shouldn't have that directly in the code. Other that that, LGTM/.

waldner:
Well, I copied it from #3743, that's why I included that line. But I'll remove it, of course.

deitch:
Thanks. CI takes quite a while to run, but once it is clear, I will merge it in

PR #3803, opened 2022-07-11
Bump alpine to 3.16

rakeshrgvn:
Signed-off-by: Rakesh Raghavan <rakesh.raghavan@abbott.com>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -4,7 +4,7 @@ on:
   workflow_run:
     workflows: [LinuxKit CI]
     types: [completed]
-    branches: [master, main]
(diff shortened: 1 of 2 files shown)


deitch:
Thanks for the PR, @rakeshrgvn 

Unfortunately it isn't that simple. You need to update every downstream component as well, which is pretty much everything in pkg/ and tools/, many in test/ as well. 

Without that, package changes in versions of alpine can break things downstream and we won't know until much later.

deitch:
Documentation is https://github.com/linuxkit/linuxkit/blob/master/docs/alpine-base-update.md

PR #3807, opened 2022-07-22
Write log entries as json

dgageot:
This is the first step in supporting structured success/error data in `memlogd` log files.

Those log files are currently made of semicolon separated fields and, for each line, the list of fields is the same.

We would like to support multiple kinds of entries. Application logs, like we do already, but also structured success/error data. Those entries would have some fields in common and some based on the type of entry. 

Switching from semicolon separated fields to json gives us this flexibility without much added complexity. In fact, it even removes a constraint we currently have on `Source` field not containing a semicolon.

### DIFF
--- a/pkg/memlogd/cmd/logread/main.go
+++ b/pkg/memlogd/cmd/logread/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
-	"bufio"
+	"encoding/json"
 	"flag"
+	"fmt"
 	"net"
-	"os"
+	"strings"
+	"time"
 )
 
 const (
@@ -12,6 +14,16 @@ const (
 	logFollow
 	logDumpFollow
 )
+
+type LogEntry struct {
+	Time   time.Time `json:"time"`
+	Source string    `json:"source"`
+	Msg    string    `json:"msg"`
+}
+
+func (msg *LogEntry) String() string {
+	return fmt.Sprintf("%s;%s;%s", msg.Time.Format(time.RFC3339Nano), strings.ReplaceAll(msg.Source, `;`, `\;`), msg.Msg)
+}
 
 func main() {
 	var err error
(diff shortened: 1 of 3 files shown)


pkg/memlogd/cmd/logread/main.go
```diff
@@ -49,7 +61,13 @@ func main() {
 		panic(err)
 	}
 
-	r := bufio.NewReader(conn)
-	r.WriteTo(os.Stdout)
+	var entry LogEntry
+	decoder := json.NewDecoder(conn)
+	for {
+		if err := decoder.Decode(&entry); err != nil {
+			panic(err)
```
djs55, line 68: I almost panicked at the `panic` but then I saw the `panic` above: I guess this is how we handle errors in this utility (which is fine)

dgageot: Yes...

pkg/memlogd/cmd/memlogd/main.go
```diff
@@ -213,27 +217,20 @@ func readLogFromFd(maxLineLen int, fd int, source string, logCh chan logEntry) {
...
-				if err := syscall.Close(msg.fd); err != nil {
-					doLog(logCh, fmt.Sprintf("ERROR: failed to close fd: %s", err))
-				}
-				continue
-			}
-			go readLogFromFd(lineMaxLength, msg.fd, msg.name, logCh)
-		}
+func loggingRequestHandler(lineMaxLength int, logCh chan LogEntry, fdMsgChan chan fdMessage) {
+	for msg := range fdMsgChan {
+		go readLogFromFd(lineMaxLength, msg.fd, msg.name, logCh)
```
djs55, line 233: This looks more concise!

pkg/memlogd/cmd/memlogd/main_test.go
```diff
@@ -113,63 +112,6 @@ func TestFinite2(t *testing.T) {
 	}
 }
 
-func TestGoodName(t *testing.T) {
-	// Test that the source names can't contain ";"
```
djs55, line 117: Ah it's a nice feature that we don't need to test this any more.

djs55 (approved):
It looks like a nice improvement to me!

deitch:
Except that `logwrite` wasn't synced with this. Oops. It caused a bunch of failures for CI now.

We need to get logwrite synced.

PR #3831, opened 2022-10-03
Support remote docker over ssh

dgageot:
I've got a remote docker daemon that I connect to over ssh.
Something like `DOCKER_HOST=ssh://docker-remote docker ps` works well

When I use linuxkit, with `DOCKER_HOST` set to `ssh://docker-remote` it'll silently fail to connect.

This introduced the support for a remote docker over ssh.

### DIFF
--- a/src/cmd/linuxkit/docker/cmd.go
+++ b/src/cmd/linuxkit/docker/cmd.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"io"
+	"os"
 
 	"github.com/containerd/containerd/reference"
+	"github.com/docker/cli/cli/connhelper"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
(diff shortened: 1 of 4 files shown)


dgageot:
Hi @deitch, do you think this one can be merged too? It pretty important when using a remote docker to be able to connect over ssh.

dgageot:
@deitch When one sets DOCKER_HOST or creates a docker context, linuxkit will indeed pick it up correctly. However, it the docker daemon that this points to is reached over ssh, linuxkit will fail to actually talk to the daemon.
Connecting over ssh is something not too old and the standard docker library doesn't do that auto magically.

dgageot:
It's for `lkt pkg build` and `lkt build`. Anytime linuxkit tries to connect to Docker.
Let me try to rephrase what this PR is doing:

It adds the support to `ssh:` scheme to connect to a remote Docker.

Let's say I run those commands to create a dedicated context:
```
$ docker context create mybuilder --docker host=ssh://docker-remote
$ docker context use mybuilder
$ linuxkit build -docker linuxkit.yml
Extract kernel image: docker.io/linuxkit/kernel:5.10.104
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Because linuxkit doesn't support `ssh:*` as a daemon url, it'll just connect locally, totally ignoring the docker context. And it'll do that silently because in the code, we swallow all the errors for the (good) reason that we should be able to build without docker.

deitch:
OK got it.

PR #3832, opened 2022-10-06
Use faster gzip library

dgageot:
This drop in replacement for the `gzip` library reduces the time needed to build an image. 

Before this patch, my two images are built in 34s.
After this patch, my two images are built in 28s.


deitch:
I am not going to complain about a 17% improvement in build times. I wish we could use [zstd](https://github.com/facebook/zstd), but too many things do not support it (including things that need to read a kernel).

src/cmd/linuxkit/initrd/initrd.go
```diff
@@ -3,13 +3,13 @@ package initrd
...
 	"archive/tar"
 	"bytes"
-	"compress/gzip"
 	"errors"
 	"io"
 	"io/ioutil"
 	"path/filepath"
 	"strings"
 
+	gzip "github.com/klauspost/pgzip"
```
deitch, line 12: I think a comment here that it is a drop-in 100% compatible replacement and 17% faster than `compress/gzip` would be helpful.

src/cmd/linuxkit/moby/build.go
```diff
@@ -18,6 +17,7 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/reference"
+	gzip "github.com/klauspost/pgzip"
```
deitch, line 20: Ditto

src/cmd/linuxkit/push_packet.go
```diff
@@ -11,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 
+	"github.com/klauspost/compress/gzip"
```
deitch, line 13: Why do we use `klauspots/compress/gzip` here and `klauspost/pgzip` elsewhere?

deitch (changes requested):
Some comments please.

src/cmd/linuxkit/push_packet.go
dgageot: Thanks, it's a mistake

Issue #3833, opened 2022-10-06
Release a binary for darwin/arm64


dgageot:
I believe the release process should include releasing a binary for darwin/arm64.

I would like to help make that happen but I'm unsure which steps to follow.

deitch:
I just hunted through all of our workflows, and was surprised that there is nothing for cutting releases based on tags and sharing artifacts. I didn't do the last releases, but @rn did, so perhaps he knows?

I think we should standardize this. Do you want to create a workflows file `release.yaml` for releases? Sees a tag, generates a release, loads up release artifacts.

rn:
we didn't really had the auto-push for any of the previous releases. so it was all done manually.

dgageot:
I'll try to take a look at it tomorrow or next week

dgageot:
This is fixed now

PR #3835, opened 2022-10-07
Add an automated release workflow

dgageot:

**- What I did**

This adds a GitHub Actions workflow that should trigger on each tag creation, 
if the tag starts with the letter `v`.
It'll build linuxkit for all os/arch targets, create a checksum file and create a draft release on GitHub.

**- How to verify it**

I couldn't test the very last part because I'm not authorised to create a release.
We should try and create a v0.9.0 tag or something.

See https://github.com/linuxkit/linuxkit/issues/3833

otis

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: LinuxKit CI
+
+on:
+  create:
+    tags:
+      - v*
+
+jobs:
+  build:
+    name: Build all targets
+    runs-on: macos-latest
+    steps:
+
+    - name: Set up Go 1.16
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16.7
+      id: go
+
+    - name: Check out code
+      uses: actions/checkout@v1
+      with:
+        path: ./src/github.com/linuxkit/linuxkit
+
+    - name: Set path
+      run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+      env:
+         GOPATH: ${{runner.workspace}}
+
+    - name: Build
+      run: |
+        make build-all-targets
+      env:
+        GOPATH: ${{runner.workspace}}
+
+    - name: GitHub Release
+      uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        draft: true
+        files: bin/*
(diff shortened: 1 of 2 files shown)


djs55 (approved):
LGTM although I’m not a github actions expert

deitch (approved):
This looks like it should be using internal makefile constructs to get all of the targets, but good enough for now.

dgageot:
@deitch Yes. I couldn't find how, though

deitch:
Yeah, getting the syntax just right (and understandable) can be a pain. Sometimes, the non-native way is just fine. 😁

dgageot:
@deitch would you like to try the preparation of a draft release?

deitch:
Yes, let's merge this in, and then I will add a tag, which we can delete.

deitch:
Created v0.8.1 and pushed it; let's see if it plays out.

deitch:
@dgageot I missed two things in here that I should have caught. Can you submit a new PR to fix them?

1. The `name` field in L1 of the workflow is "Linuxkit CI", which already is used; it should be changed to "Release" or similar.
2. The `runs-on` for the job in L11 is `macos-latest`. This should run on their default `ubuntu-latest`.

deitch:
Sure looks like it worked, see https://github.com/linuxkit/linuxkit/actions/runs/3211331714/jobs/5249425890 .

In addition to the above 2, I would request a third change: why make these only "draft" releases?

dgageot:
@deitch
+ I'll change the `name`
+ I can't change the `runs-on`. It needs to be `macos` in order to build the darwin binaries. `ubuntu-latest` can't do it
+ Do you see the binaries on the draft release? I don't have access
+ I think creating a draft is safer so that there's always one simple human step. wdyt?

deitch:
You can leave it draft for now, but at some point we will just make it a real release. 

Yeah the artifacts are there. Are draft releases not public?

I didn't think about the macOS build issue. That's because of the virtualization framework dependency. Ugh. 

Can we break it down into different jobs and then merge the results? Eh good enough for now.

dgageot:
> Yeah the artifacts are there. Are draft releases not public?

Not public, no. At least not by default. I can't see it.
 
> I didn't think about the macOS build issue. That's because of the virtualization framework dependency. Ugh.

Yes. CGO is enabled on Darwin and requires a macOS compiler I guess.

> Can we break it down into different jobs and then merge the results? Eh good enough for now.

That would probably be way more complicated.

deitch:
And since I'm the one who knowingly made cgo required to enable visualization framework, I probably should remember it. 😁

Let's leave that part as is.

PR #3841, opened 2022-10-08
Upgrade to go 1.19

dgageot:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,10 @@ jobs:
     runs-on: ${{ matrix.target.runner }}
     steps:
 
-    - name: Set up Go 1.16
+    - name: Set up Go 1.19
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.7
+        go-version: 1.19.2
       id: go
 
     - name: Check out code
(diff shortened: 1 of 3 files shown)


thaJeztah:
Looks like it's failing to install `golint`

```bash
go get -u golang.org/x/lint/golint
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

However, it's deprecated, so I'm not sure if it should even be used (see https://github.com/golang/lint). Perhaps better if we would use golangci-lint with some good linters enabled.

deitch:
Yeah we should. I've switched to golangci-lint pretty much everywhere.

deitch:
I have an open PR #3849 to add modern golangci-lint. But I am going to wait until your cleanup changes are in @dgageot in the various PRs before I merge it, since it raises a lot of the same issues.

deitch:
Once #3849 is clean and in, you should be able to rebase this and get good results.

PR #3843, opened 2022-10-08
Remove dead or redundant code (src/cmd/linuxkit)

dgageot:


deitch:
Can you split these into two, one for src/cmd/linuxkit and one for the packages? If we have any issues with the changes, it'll make it far simpler to track down. 

Shouldn't `go vet` pick any of this up?

pkg/metadata/provider_vultr.go
```diff
@@ -31,7 +31,7 @@ func (p *ProviderVultr) String() string {
 func (p *ProviderVultr) Probe() bool {
 	// Getting the index should always work...
 	_, err := vultrGet(vultrMetaDataURL)
-	return (err == nil)
```
deitch, line 34: This should go with packages.

deitch:
Needs a rebase

deitch (approved):

PR #3842, opened 2022-10-08
Remove 99% of deprecated ioutil usage (src/cmd/linuxkit)

dgageot:
Signed-off-by: David Gageot <david.gageot@docker.com>

PR #3844, opened 2022-10-08
Prefer nil empty slices (src/cmd/linuxkit)

dgageot:

### DIFF
--- a/src/cmd/linuxkit/moby/build.go
+++ b/src/cmd/linuxkit/moby/build.go
@@ -72,7 +72,7 @@ var additions = map[string]addFun{
 
 // OutputTypes returns a list of the valid output types
 func OutputTypes() []string {
-	ts := []string{}
+	var ts []string
 	for k := range streamable {
 		ts = append(ts, k)
 	}
--- a/src/cmd/linuxkit/moby/config.go
+++ b/src/cmd/linuxkit/moby/config.go
@@ -486,7 +486,7 @@ func mergeStrings(v1, v2 *[]string) *[]string {
 		return v2
 	}
 	// merge the two uniquely
-	ret := []string{}
+	var ret []string
 	m := make(map[string]bool)
 	for _, s := range *v1 {
 		if m[s] {
@@ -768,7 +768,7 @@ func ConfigToOCI(yaml *Image, config imagespec.ImageConfig, idMap map[string]uin
 			return oci, runtime, fmt.Errorf("Cannot parse tmpfs, too many ':': %s", t)
 		}
 		dest := parts[0]
-		opts := []string{}
+		var opts []string
 		if len(parts) == 2 {
 			opts = strings.Split(parts[1], ",")
 		}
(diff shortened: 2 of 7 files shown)


pkg/init/cmd/rc.init/main.go
```diff
@@ -81,7 +81,7 @@ func mkdir(path string, perm os.FileMode) {
 
 // list of all enabled cgroups
 func cgroupList() []string {
-	list := []string{}
+	var list []string
```
thaJeztah, line 84: While we're updating this, I'm wondering if we should also optimise allocations, so instead of returning `list` (on error), just return `nil`, and when we successfully read the list, then initialise the slice with (`make([]string, 0, <size>)`.

For this function, something like:

```go
// list of all enabled cgroups
func cgroupList() []string {
	f, err := os.Open("/proc/cgroups")
	if err != nil {
		log.Printf("cannot open /proc/cgroups: %v", err)
		return nil
	}
	defer f.Close()
	reader := csv.NewReader(f)
	// tab delimited
	reader.Comma = '\t'
	// four fields
	reader.FieldsPerRecord = 4
	cgroups, err := reader.ReadAll()
	if err != nil {
		log.Printf("cannot parse /proc/cgroups: %v", err)
		return nil
	}
    list := make([]string, 0, len(cgroups))
	for _, cg := range cgroups {
		// see if enabled
		if cg[3] == "1" {
			list = append(list, cg[0])
		}
	}
	return list
}
```

pkg/init/cmd/rc.init/main.go
```diff
@@ -126,7 +126,7 @@ func read(path string) string {
 
 // read a directory
 func readdir(path string) []string {
-	names := []string{}
+	var names []string
 	files, err := ioutil.ReadDir(path)
```
thaJeztah, line 130: Same here.

We should also replace `ioutil`, as it's deprecated 😅 


```go
// read a directory
func readdir(path string) []string {
	files, err := os.ReadDir(path)
	if err != nil {
		log.Printf("cannot read directory %s: %v", path, err)
		return nil
	}
	names := make([]string, 0, len(files))
	for _, f := range files {
		names = append(names, f.Name())
	}
	return names
}
```

thaJeztah: - Oh! Just spotted you opened https://github.com/linuxkit/linuxkit/pull/3842 👍

pkg/init/cmd/rc.init/main.go
dgageot: Yes, we could do that but that's a whole new level to find them all!

deitch:
Can you please split out the changes to `src/cmd/linuxkit` and the ones to stuff in `pkg/`?

deitch:
So this certainly doesn't hurt. Why does it help? Won't it get optimized away anyways?

dgageot:
In terns of performances, I have no idea. I know it's more idiomatic and good linters will usually complain.

deitch:
Then I don't care.

deitch (approved):

deitch:
When it is ready, switch it to non-draft and I will merge it in.

PR #3845, opened 2022-10-08
Fix comments

dgageot:

### DIFF
--- a/pkg/init/cmd/service/logging.go
+++ b/pkg/init/cmd/service/logging.go
@@ -83,7 +83,7 @@ func (f *fileLog) Dump(n string) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

deitch (commented):
Why? The phrase "symlinks to x" has been in common use as long as I can remember (which I won't admit how far)

dgageot:
@deitch It's just that function comments in Go should start with the name of the function followed by a verb.
It's really a nit but here the function is `Symlink` and the comment starts with `Symlinks`.

deitch:
Oh from the go comments perspective? OK, sure, that makes sense.

deitch (approved):

PR #3848, opened 2022-10-09
Remove dead or redundant code (packages)

dgageot:

### DIFF
--- a/pkg/host-timesync-daemon/main.go
+++ b/pkg/host-timesync-daemon/main.go
@@ -108,7 +108,7 @@ func main() {
 		d := time.Date(int(t.tmYear+1900), time.Month(t.tmMon+1), int(t.tmMday), int(t.tmHour), int(t.tmMin), int(t.tmSec), 0, time.UTC)
 		log.Printf("Setting system clock to %s", d)
 		tv := syscall.Timeval{
-			Sec:  int64(d.Unix()),
+			Sec:  d.Unix(),
 			Usec: 0, // the RTC only has second granularity
 		}
 		if err = syscall.Settimeofday(&tv); err != nil {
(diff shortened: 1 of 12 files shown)


pkg/host-timesync-daemon/main.go
```diff
@@ -51,10 +51,10 @@ func rtcReadTime() rtcTime {
 	defer f.Close()
 	result := rtcTime{}
 	arg := uintptr(0)
-	arg |= (iocREAD << iocDIRSHIFT)
-	arg |= (rtcRDTIMETYPE << iocTYPESHIFT)
-	arg |= (rtcRDTIMENR << iocNRSHIFT)
-	arg |= (unsafe.Sizeof(result) << iocSIZESHIFT)
```
deitch, line 57: unless lint or vet complains about these, I prefer to leave them with the `(   )`. Operation precedence in go should do bitwise shift before bitwise OR and especially before assignment, but it can be confusing.

dgageot: I'll remove those changes for now

deitch: Just watch. As soon as I enable golangci-lint it will complain 😆 

We will cross that bridge when we come to it.

dgageot: 🥇

pkg/host-timesync-daemon/main.go
```diff
@@ -108,7 +108,7 @@ func main() {
 		d := time.Date(int(t.tmYear+1900), time.Month(t.tmMon+1), int(t.tmMday), int(t.tmHour), int(t.tmMin), int(t.tmSec), 0, time.UTC)
 		log.Printf("Setting system clock to %s", d)
 		tv := syscall.Timeval{
-			Sec:  int64(d.Unix()),
```
thaJeztah, line 111: Could this one have been for non-x86 or 32-bit arches? I know we have similar casts in other repositories (but if it's deliberate; we need to add a nolint comment with an explanation

deitch: I am not sure. I think that as long as it is not complaining, let's leave it as is. If someone has an issue somewhere on a 32-bit arch, let's worry about it then.

deitch (approved):

PR #3853, opened 2022-10-12
Faster image export

dgageot:
Creating the container without a network takes 50 to 80ms less.


deitch:
True, we only ever call this to create a tar reader, but it is possible we might want to actually run it. Any reason not to make it a boolean parameter and pass it as `false`?

dgageot:
@deitch No reason. I saw it was used only in that case. I can change that

dgageot:
@deitch Should be better now. Next time we have to add a flag to this function, it'll probably better to directly pass a `container.Config`.

deitch (approved):

PR #3855, opened 2022-10-17
Fix initrd image padding

dgageot:
The new code creates initrd files with trailing garbage.
When uncompressed, gzip will print a warning.

@deitch This issue has postponed a release of Docker Desktop.
We switched back to an old version of linuxkit.


dgageot:
@deitch yes, the "fix" looked awfully wrong to me. The linter agrees. In fact, my new theory is that this whole pad4 didn't do anything at all in the past. Now it does but it does it wrong.

What happens is that we produce an `initrd.dmg` file that's not a totally valid gzip file. It ends with garbage and gunzip will succeed in decompressing but will print this warning:
```
trailing garbage ignored
```

I wonder if we could remove pad4 altogether. Where does this requirement come from?

deitch:
It only is used for initrd [here](https://github.com/linuxkit/linuxkit/blob/93896eb201427e5cb56d74b30d59f994826501d5/src/cmd/linuxkit/initrd/initrd.go#L19). The comment [here](https://github.com/linuxkit/linuxkit/blob/f75b5cb18aa94ce3e0d8e5b954a3ce40170466f2/src/cmd/linuxkit/pad4/pad4.go#L8) says:

```go
// A Writer is an io.WriteCloser. Writes are padded with zeros to 4 byte boundary
```

It really wants them to have a 4 byte boundary padded. Is that a requirement somewhere? I don't know. @justincormack wrote the original element, perhaps he knows?

Is there an issue with various kernel versions?

dgageot:
@deitch I'm pushing the change of order between writers. It seems to be the "right" fix

deitch:
So the current model is:

```
write -> cpio -> gzip -> pad4 -> output
```

You are changing it to:

```
write -> cpio -> pad4 -> gzip -> output
```

The current one seems to assume that the gzip output should have a precise multiple of 4 bytes, which doesn't make much sense to me either, but maybe it is a kernel requirement? I never heard of it before, but that only shows the limits of my knowledge.

Your change assumes that the output of the cpio should be a precise multiple of 4 bytes before being passed to gzip. I find that strange as well.

To be honest, I really do not know enough about the requirements for either. I like your approach, but @justincormack hopefully knows more.

dgageot:
@deitch I'm trying to make sense of https://docs.kernel.org/driver-api/early-userspace/buffer-format.html

dgageot:
> @deitch I'm trying to make sense of https://docs.kernel.org/driver-api/early-userspace/buffer-format.html

From what I'm reading, the pad(4) operations mentioned in the spec seem to be taken care of by `github.com/surma/gocpio/writer.go`

deitch:
It sure looks like it. I vote for taking it out of the loop. If tests pass, we accept it until someone complains.

dgageot:
@deitch I've updated the PR to remove pad4 altogether. Let's see if the tests pass. I've build an image locally and it looks just fine.

deitch (approved):

rn (approved):

PR #3858, opened 2022-10-24
move rtf commands from GHA into make test; remove unused artifacts

deitch:

**- What I did**

* eliminated unused gcp artifact hack in `test/Makefile`
* added support for specifying test suites in `make test`
* replaced all direct calls to `rtf` in GHA workflows with `make test` calls

This is worth it on its own. Once it is done, I will use matrix to create parallel packages tests, which should speed up CI.

**- How I did it**

Changes to the two Makefiles in `ci.yaml`

**- How to verify it**

CI. If it works, then nothing has broken.

rtf call refactor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,9 +174,7 @@ jobs:
           ${{ runner.os }}-linuxkit-
 
     - name: Run Tests
-      run: |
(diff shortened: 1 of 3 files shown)


rn (approved):
don't think we use GCP anymore

PR #3860, opened 2022-10-27
add sharding option

deitch:

**- What I did**

Added sharding of packages tests, to try and get it from 90 mins serial to up to 10 parallel shards in ~9 mins each.

**- How I did it**

1. Updated rtf to support shards
2. Updated to latest rtf version
3. Add shard support to `Makefile` and `tests/Makefile`
4. Add matrix to GHA under packages tests so that it has 10 shards

**- How to verify it**

Run CI and then see if the results are as expected

Sharding support for much faster tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,9 @@ jobs:
     name: Packages Tests
     needs: [ build_packages, build ]
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


deitch:
Well that worked a little too well. Each shard ran in < 1 minutes, clearly it missed something, Back to the shard algorithm.

PR #3869, opened 2022-11-08
Build with bake

dgageot:
Build all the binaries with `docker buildx bake`. This allows building linuxkit binaries on any OS, even the macOS binary that depends on CGO.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,26 @@
 # syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.19.2
-ARG XX_VERSION=1.1.2
-ARG OSXCROSS_VERSION=12.3-r0-alpine
-
 # xx is a helper for cross-compilation
-FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.1.2 AS xx
 
 # osxcross contains the MacOSX cross toolchain for xx
-FROM crazymax/osxcross:${OSXCROSS_VERSION} AS osxcross
+FROM crazymax/osxcross:12.3-r0-alpine AS osxcross
 
-FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine3.16 AS base
+FROM --platform=$BUILDPLATFORM golang:1.19.2-alpine3.16 AS build
 COPY --link --from=xx / /
 RUN apk add --no-cache clang git lld llvm make
 ARG TARGETPLATFORM
 WORKDIR /src
-COPY --link src/cmd/linuxkit/ .
-
-FROM base AS build-linux
-RUN --mount=type=cache,target=/root/.cache \
-    --mount=type=cache,target=/go/pkg/mod <<EOT
-  set -ex
-  xx-go --wrap
-  LOCAL_TARGET=/out/linuxkit make local-build
-  xx-verify /out/linuxkit
-EOT
-
-FROM base AS build-darwin
-RUN --mount=type=cache,target=/root/.cache \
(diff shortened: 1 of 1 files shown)


deitch:
Darwin needs to be built on macOS only - or to mount all of the tools and libs in - because it compiles in direct access to Hypervisor Virtualization.Framework. Truth is, qemu from 6.something on has direct support, so I have been thinking that maybe we should drop back to using qemu on macOS, and let that interact with HVF. Not sure about pros and cons.

Would the `Makefile` change? How would we use this? It looks like the Dockerfile is complete on its own?

dgageot:
> Darwin needs to be built on macOS only - or to mount all of the tools and libs in - because it compiles in direct access to Hypervisor Virtualization.Framework. Truth is, qemu from 6.something on has direct support, so I have been thinking that maybe we should drop back to using qemu on macOS, and let that interact with HVF. Not sure about pros and cons.

The Dockerfile is supposed to mount the macOS SDK and use it to build linuxkit. But I couldn't make it fully work so far.
 
> Would the `Makefile` change? How would we use this? It looks like the Dockerfile is complete on its own?

Right now, I don't know what would be the impact on the Makefile. We would probably call the makefile from the Dockerfile.

deitch:
Opposite, I think. Look at the makefile for how it builds `linuxkit`, [here](https://github.com/linuxkit/linuxkit/blob/master/Makefile#L60-L61) which called [here](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/Makefile#L33-L40)

It basically tars up the current directory, passing it as stdin to `docker run linuxkit/go-compile` and saves the results to a tar file, which is extracted.

 I do not remotely recall why we do it this way.

What is better about this approach, and how would we use it?

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# osxcross contains the MacOSX cross toolchain for xx
+FROM crazymax/osxcross:${OSXCROSS_VERSION} AS osxcross
+
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine3.16 AS base
+COPY --link --from=xx / /
+RUN apk add --no-cache clang git lld llvm make
+ARG TARGETPLATFORM
+WORKDIR /src
+COPY --link src/cmd/linuxkit/ .
```
crazy-max, line 18: Remove this line, it's better to use `RUN` mount, see comment about single stage.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+FROM base AS build-windows
+RUN --mount=type=cache,target=/root/.cache \
+    --mount=type=cache,target=/go/pkg/mod <<EOT
+  set -ex
+  xx-go --wrap
+  LOCAL_TARGET=/out/linuxkit.exe make local-build
+  xx-verify /out/linuxkit.exe
+EOT
+
+FROM build-$TARGETOS AS build
```
crazy-max, line 49: Single `build` stage is enough, see https://github.com/docker/docker-credential-helpers/pull/249:

```suggestion
RUN --mount=type=bind,source=src/cmd/linuxkit,target=. \
    --mount=type=cache,target=/root/.cache \
    --mount=type=cache,target=/go/pkg/mod \
    --mount=type=bind,from=osxcross,src=/osxsdk,target=/xx-sdk <<EOT
  set -ex
  xx-go --wrap
  LOCAL_TARGET=/out/linuxkit make local-build
  xx-verify /out/linuxkit
  [ "$TARGETOS" = "windows" ] && mv /out/linuxkit /out/linuxkit.exe
EOT
```

Dockerfile
dgageot: :-) I was exactly working on that

PR #3870, opened 2022-11-08
when cannot get remote manifest in checking registry, that is just a …

deitch:
…sign that it does not exist there

**- What I did**

When checking if image is in registry for `pkg build`, it failed on errors, when it should treat errors as, "I could not find it there, so go ahead and build."

**- How I did it**

Change handling of `if err != nil`

**- How to verify it**

CI

Proper registry error handling


deitch:
I have no issues building `pkg/sshd` from source, including if I make minor changes to the package to force a "dirty" build.

I am going to merge this in as it urgently fixes a breaking issue. If there is a separate issue - with build in general or `pkg/sshd` in particular - open a new issue on it please.

PR #3871, opened 2022-11-09
pkg/sshd: Remove default bind of /root/.ssh

the-maldridge:

**- What I did**
Removed the default bind mount of `/root/.ssh` in the sshd package.  When using the metadata package to summon keys, the ssh directory will not be present in root's home folder on the host filesystem because the keys are summoned dynamically.  However, its not possible to just add the new bind for the `authorized_keys` file because the bind to the nonexistent directory prevents sshd from being able to start.  This change removes the default bind since it can be trivially added back using `binds.add` for anyone who wishes to statically provide an `authorized_keys` file at build time.

**- How I did it**
A single elegant emacs command.

**- How to verify it**
Observe the diff, see that the bind mount is no longer present in `build.yml`.

Remove `/root/.ssh` default bind to improve portability.  Bind may be re-added with `binds.add`.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -63,6 +63,21 @@ This hierarchy can then be used by individual containers, who can bind
 mount the config sub-directory into their namespace where it is
 needed.
 
+## A note on SSH
+
+Supported providers will extract public keys from metadata to a file
+located at `/run/config/ssh/authorized_keys`.  You must bind this path
+into the `sshd` namespace in order to make use of these keys.  Use a
+configuration similar to the one shown below to enable root login
+based on keys from the metadata service:
+
+```
+  - name: sshd
+    image: linuxkit/sshd:4696ba61c3ec091328e1c14857d77e675802342f
+    binds.add:
+     - /run/config/ssh/authorized_keys:/root/.ssh/authorized_keys
+```
+
 # Metadata image creation
 
 `linuxkit run` backends accept two options to pass metadata to the VM in a platform specific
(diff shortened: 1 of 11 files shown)


deitch:
I want to say that I am not sure I get this, but I am pretty sure I don't get this.

By default, it binds `/root/.ssh`. This is necessary because you need the system wide `authorized_keys`; don't want them in an ephemeral container.

I think you are saying that if `/root/.ssh` does not exist, then the container fails to start. Is that because the source dir for the bind is not there? Why not just make it a required directory?

the-maldridge:
I suppose you could make it a required directory, but that pushes the fix onto any user running ssh on linuxkit in the cloud, which feels like a kludge to me.  By default it will bind `/root/.ssh` which would be great if this existed, but assuming that you aren't baking keys into images, you're likely fetching them with the metadata package, which writes them to /run/config since that's writeable and /etc/ isn't.

deitch:
So you are saying current design of `linuxkit/sshd` requires `/root/.ssh`; if it doesn't exist, sshd fails to start. Rather than having them automatically go to something that might or might not exist, do not define the mount, but be explicit in the docs about, "it looks for keys in `/root/.ssh`, bind mount that from where you want."

Is that it? If so, the docs and examples prob should reflect that as well.

the-maldridge:
I have amended my PR to include documentation on the metadata package.  There didn't appear to be any documentation about sshd itself, but I can create some if you feel this is appropriate.

Some of the examples had conflicting behavior, so where there was question as to which key should be the active key I maintained whatever the previous behavior was.  I think in general though it would be better to defer to the key from metadata as that is the established pattern for cloud providers.

deitch (approved):

PR #3879, opened 2022-11-30
unify apk installed db for base layer

deitch:

**- What I did**

Fixed an issue wherein multiple layers in `init` trounce each other's apk installed db file.

Specifically, let's say I have some images:

```yml
init:
  - linuxkit/init:8f1e6a0747acbbb4d7e24dc98f97faa8d1c6cec7
  - linuxkit/runc:f01b88c7033180d50ae43562d72707c6881904e4
  - linuxkit/containerd:de1b18eed76a266baa3092e5c154c84f595e56da
  - linuxkit/ca-certificates:c1c73ef590dffb6a0138cf758fe4a4305c9864f4
```

Each of the above gets laid out, in order, on the root filesystem. 

If all of them have apk packages installed, then all of them have `/lib/apk/db/installed`. However, as this is just a tar stream, the last one (in the above example, from `linuxkit/ca-certificates`) _overwrites_ the previous. That means that the installed db is out of date.

This PR changes it so that we capture all of the `installed` files and concatenate them to have a single one that includes all of them.

Right now, it just concatenates, so it likely has duplicates, but it still works. If I can get deduplication before this goes in, I will comment here.

**- How I did it**

Added a tar wrapper for when the init sections are streamed, so it captures the `/lib/apk/db/installed` databases and then merges them when told to.

**- How to verify it**

CI, plus a manual test to see that it is merged.

init captures all apk installed databases


deitch:
I can confirm this works. The tar stream might be smaller if we had only the final version, but that can be a later optimization.

PR #3884, opened 2022-12-16
switch from flags to cobra

deitch:

**- What I did**

Replaced all of our custom flag handling with spf13/cobra

Provides much greater simplicity, reduces the need to handle things like help, number of arguments, etc.,
makes it easier to add features, aliases built in.

It also gives us things like persistent flags, so we can set cache-dir once, and verbose/quiet once/

**- How I did it**

Slowly and painfully

**- How to verify it**

CI

Use modern flag libraries


deitch:
Ah, lovely, CI finally green.

PR #3886, opened 2022-12-29
fix alpine base update docs

deitch:

**- What I did**

Fixed a few misleading lines on the alpine base update doc

**- How I did it**

markdown

**- How to verify it**

Read it

update docs

### DIFF
--- a/docs/alpine-base-update.md
+++ b/docs/alpine-base-update.md
@@ -130,7 +130,7 @@ following which is an explanation of each one.
 ```sh
 # Update tools packages
 cd $LK_ROOT/tools
-$LK_ROOT/scripts/update-component-sha.sh --pkg $LK_ROOT/tools/alpine
+$LK_ROOT/scripts/update-component-sha.sh --image $LK_ALPINE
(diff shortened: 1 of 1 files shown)


docs/alpine-base-update.md
```diff
@@ -130,7 +130,7 @@ following which is an explanation of each one.
 ```sh
 # Update tools packages
 cd $LK_ROOT/tools
-$LK_ROOT/scripts/update-component-sha.sh --pkg $LK_ROOT/tools/alpine
+$LK_ROOT/scripts/update-component-sha.sh --image $LK_ALPINE
```
rn, line 133: what's the reason for this change? Does `update-component-sha.sh` no longer take the `--pkg` option?

rn (approved):

docs/alpine-base-update.md
deitch: It does, but `tools/alpine` does not have a `build.yml`. To be fair, I have no idea how it worked before, but if this works, and we already are storing the `LK_ALPINE` var, might as well use it.

Issue #3885, opened 2022-12-28
Unable to build with 1.0.1


devisbal:
tar cf - -C . . | docker run --rm --net=none --log-driver=none -i -e GOARCH= linuxkit/go-compile:7b1f5a37d2a93cd4a9aa2a87db264d8145944006 --package github.com/linuxkit/linuxkit/src/cmd/linuxkit --ldflags "-X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.GitCommit= -X github.com/linuxkit/linuxkit/src/cmd/linuxkit/version.Version="v0.8+"" -o linuxkit > tmp_linuxkit_bin.tar
gofmt...
vendor/github.com/Code-Hex/vz/v3/internal/objc/finalizer_118.go:8:18: expected '(', found '['
vendor/github.com/moby/buildkit/frontend/attest/sbom.go:75:13: expected '(', found '['
vendor/github.com/moby/buildkit/frontend/frontend.go:15:28: expected ';', found '['
vendor/github.com/moby/buildkit/frontend/gateway/client/client.go:17:28: expected ';', found '['
vendor/github.com/moby/buildkit/solver/result/result.go:16:15: expected ']', found any
vendor/github.com/moby/buildkit/solver/result/result.go:26:2: expected declaration, found 'if'
vendor/github.com/moby/buildkit/solver/result/result.go:68:3: expected declaration, found 'return'
vendor/github.com/moby/buildkit/solver/result/result.go:91:2: expected declaration, found 'if'

deitch:
Local build works fine:

```
$ git checkout v1.0.1
Note: switching to 'v1.0.1'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 860163a9c Merge pull request #3863 from deitch/rationalize-build-pull
$ make -C ./src/cmd/linuxkit local-build
[...]

PR #3888, opened 2022-12-29
Alpine 317

deitch:
**- What I did**

Bumped `linuxkit/alpine` from basing on `alpine:3.14` to `alpine:3.17`. Too many packages were too far out of date, especially go.

Note that I did _not_ rebuild for s390x, as we do not have a build server for it.

**- How I did it**

Followed the doc. Slowly.

**- How to verify it**

CI

alpine:3.17 base


deitch:
fixes #3885

deitch:
Additionally had to do the following:

* update `go-compile.sh` to work, as well as reformat a few go files for the go1.19 rules.
* update some packages

tools/alpine/versions.aarch64
```diff
@@ -1,370 +1,387 @@
-# linuxkit/alpine:6f9efe87b3c394196949985f0ed6b5263432670a-arm64
+# linuxkit/alpine:1e9d030f156213bf857e32cb824e86a834e29d5f-
```
rn, line 1: looks like the arch is missing here. I thought this is handled automatically by the build/makefile

deitch: So did I. How odd. The really frustrating thing is that this is in the very first step. I loathe the idea of running through the whole thing again, having done it 3 times. First to update the first `FROM`, then when I realized we `FROM` again later, and then because of a bug in `go-compile.sh`. I will look into it, though.

This does not pass CI yet.

deitch: It did the right thing on `x86_64`. Huh

deitch: Ah shoot. The arm64 builder was not running that build in buildx. All other packages we build with `linuxkit pkg build`, which ensures bulldkit in a container, but `tools/alpine` is built directly with `docker build`, which doesn't have build if older. I need to upgrade docker there.

deitch: Done. Now it includes it.

deitch:
The `kmsg` test is failing, apparently nothing is being written to `/var/log` by `memlogd`. Since it takes all of the logs, I do not know how to debug it.

It looks like `examples/logging.yml` fails as well; nothing ends up in `/var/log` That should make it a bit easier to debug, as it has getty configured, so you can log in and see things, but there is nothing to see. I don't know if this is something broken by the changes here, or something from earlier that did not get picked up until now.

deitch:
Once #3890 is in, I can rebase and it should fix it.

deitch:
#3890 is in, that should fix the failing tests around logwrite and kmsg. 

I also had to fix `ci.yml` and `test/pkg/Makefile`, which I originally did as a commit on this PR, but it could be broken out separately, so I broke it into #3892. Once that one is in, this should be a clean "update to alpine 3.17" PR and should pass.

deitch:
OK #3890 and #3892 are in. Rebased and pushed, hopefully it all will be clean now.

deitch:
Yes! And now CI is clean!

rn (approved):
wow, great effort @deitch. LGTM

PR #3891, opened 2023-01-02
Add virtiofs command line option for qemu run

fredericdalleau:
**- What I did**

On linux, it requires root to start the virtiofsd and add some options to qemu processes for file sharing. I could test the C version of virtiofsd on Linux, `virtiofsd` is installed as part of `qemu-system-common` on an ubuntu system.
On Mac, it requires macOS 12+ and should do nothing on earlier OSes.

Let me know what you think.

**- How I did it**

Add an option to qemu command
Add an option to virtualization command

**- How to verify it**
```
 (sleep 5; echo "mount -t virtiofs virtiofs0 /mnt && ls /mnt && poweroff" ) | sudo ./src/cmd/linuxkit/linuxkit run qemu --mem 2048 --virtiofs $HOME linuxkit

 (sleep 5; echo "mount -t virtiofs virtiofs0 /mnt && ls /mnt && poweroff" ) | ./src/cmd/linuxkit/linuxkit run virtualization --mem 2048 --virtiofs $HOME linuxkit
```

Add a `--virtiofs` option to `linuxkit run qemu` to create virtiofs mounts.
Add a `--virtiofs` option to `linuxkit run virtualization` to create virtiofs mounts.

### DIFF
--- a/src/cmd/linuxkit/run_qemu.go
+++ b/src/cmd/linuxkit/run_qemu.go
@@ -23,28 +23,30 @@ const (
 
 // QemuConfig contains the config for Qemu
 type QemuConfig struct {
-	Path           string
-	ISOBoot        bool
-	UEFI           bool
-	SquashFS       bool
-	Kernel         bool
-	GUI            bool
-	Disks          Disks
-	ISOImages      []string
-	StatePath      string
-	FWPath         string
-	Arch           string
-	CPUs           string
-	Memory         string
-	Accel          string
-	Detached       bool
-	QemuBinPath    string
-	QemuImgPath    string
-	PublishedPorts []string
-	NetdevConfig   string
-	UUID           uuid.UUID
-	USB            bool
-	Devices        []string
+	Path             string
+	ISOBoot          bool
+	UEFI             bool
+	SquashFS         bool
+	Kernel           bool
+	GUI              bool
+	Disks            Disks
+	ISOImages        []string
+	StatePath        string
+	FWPath           string
+	Arch             string
+	CPUs             string
+	Memory           string
+	Accel            string
+	Detached         bool
+	QemuBinPath      string
+	QemuImgPath      string
+	PublishedPorts   []string
+	NetdevConfig     string
+	UUID             uuid.UUID
+	USB              bool
+	Devices          []string
+	VirtiofsdBinPath string
+	VirtiofsShares   []string
 }
 
 const (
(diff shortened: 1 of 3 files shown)


src/cmd/linuxkit/run_qemu.go
```diff
@@ -398,6 +404,27 @@ func runQemuLocal(config QemuConfig) error {
 		return fmt.Errorf("Detached mode is only supported when running in a container, not locally")
 	}
 
+	if len(config.VirtiofsShares) > 0 {
+		args = append(args, "-object", "memory-backend-memfd,id=mem,size="+config.Memory+"M,share=on", "-numa", "node,memdev=mem")
+	}
+	for index, source := range config.VirtiofsShares {
+		socket := fmt.Sprintf("%s%d", "/tmp/virtiofs", index)
```
rn, line 411: This doesn't seem to work if I want to run multiple instances of qemu with virtiofs enabled.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -398,6 +404,27 @@ func runQemuLocal(config QemuConfig) error {
...
+
+		cmd := exec.Command("/usr/lib/qemu/virtiofsd", "--socket-path="+socket,
```
rn, line 413: This hardcodes the path to `virtiofsd`. We try not to do this. either rely on `$PATH` or have a list of possible location with a command line option to override.

src/cmd/linuxkit/run_qemu.go
```diff
@@ -398,6 +404,27 @@ func runQemuLocal(config QemuConfig) error {
...
+			"--socket-group="+"docker",
```
rn, line 414: I'm not sure what this does and the documentation from https://gitlab.com/virtio-fs/virtiofsd does not shed any light

fredericdalleau: Will drop it, it's a remnant of trying to as non root

src/cmd/linuxkit/run_qemu.go
fredericdalleau: Seem it would make sense in `config.StatePath`, there is already the qemu pid in there. I'll do that, lmk if anything.

src/cmd/linuxkit/run_qemu.go
fredericdalleau: It's not in path at least here. Wrt the list of possible locations, I only see `/usr/local/lib/qemu` and `/usr/lib/qemu`
With the `list` of possible locations which needs to be documented, otherwise the user does not know which one is in use.
I propose to reverse the path from qemu path.

djs55 (approved):
LGTM

I tested on Mac by
```
./bin/linuxkit run virtualization --virtiofs `pwd`  examples/getty
```
and then logging in and
```
mkdir /tmp/mnt
mount virtiofs0 -t virtuous /tmp/mnt
ls /tmp/mnt
```

rn:
can you please squash the first three commits? the second and third seem to address review comments and should be squashed. the virt framework one should stay as a separate commit

rn (approved):

djs55 (approved):

djs55:
Just needs a rebase. I'll take a quick look

djs55:
Scratch that, the PR seems to be merged but still open. I'll close it. Github glitch?

PR #3902, opened 2023-01-23
Bump go-diskfs to latest

Itxaka:
**- What I did**

Bump go-diskfs in the metadata.
Fix the provider_cdrom to use the new Open method of diskfs

**- How I did it**

`go get -u github.com/diskfs/go-diskfs@master && go mod vendor`

Crazy I know, sometimes my genius… it's almost frightening :p

**- How to verify it**

Run metadata in the background with the full list of providers so the block devices are opened to get the size, then try to do something on those block devices, like create a new partition table. It will complain that the kernel cannot be informed of the changes to the block device due it being in use by something else. 
Problem was mainly that if you run the metadata will the full list of providers and no metadata, it will take a while for all those providers to time out. So while the metadata process is running, the GC wont clean up the open block devices. Once the metadata process ends, then GC seems to close the devices properly so its kind of difficult to trigger this in a normal way.
We only found about this becuase we were running the metadata in the brackground and our installer was creating new partition tables at the exact same time. If you run the installer sooner you wont hit it, or a bit later and also wont hit it.

Bump go-diskfs to latest and adapt provider_cdrom to the go-diskfs changes.

pisco


Itxaka:
cc @deitch ;)

deitch (approved):

Issue #3873, opened 2022-11-13
The go.mod file for the module providing named packages contains one or more replace directives.


lifestreamy:
Running 
`go install github.com/linuxkit/linuxkit/src/cmd/linuxkit@latest` 
in Windows powershell results in 
```
go: github.com/linuxkit/linuxkit/src/cmd/linuxkit@latest (in github.com/linuxkit/linuxkit/src/cmd/linuxkit@v0.0.0-20221110081417-aa2482123637):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```
![image](https://user-images.githubusercontent.com/44506462/201540618-e6369474-9563-431f-b471-eb05c36cfa2e.png)

Petrichia:
In fedora 36 I have the same issue.

deitch:
I don't understand why this is an issue? `github.com/linuxkit/linuxkit/src/cmd/linuxkit` **is** the main module.

Petrichia:
I don't really know. I test it with go version 1.17, just in case. And the issue remains.
I test it with different go version because I had issues with go1.19 when trying to install linuxkit with `make local`. It was beacuse go1.19 doesn't let you execute `go get` if you don't executed it where a go.mod is located.

Petrichia:
In [Go-issue-40276](https://github.com/golang/go/issues/40276), on section `Why can't module replace directives be used?` seems to explain why replace directives should report an error with go install.

Nosmoht:
@deitch 

it's an issue because it's mentioned in the docs to do so:
```
If you already have go installed you can use go install github.com/linuxkit/linuxkit/src/cmd/linuxkit@latest to install the linuxkit tool.
```

So after all `linuxkit` is not installed after executing the command. People just expect to have a working docu ;)

deitch:
Wrapping my head around this, so:

* I cannot use `replace` directives in my `go.mod` unless I am actually installing from inside the directory: `cd src/cmd/linuxkit && go install .`
* Outside of it does not work: `go install github.com/linuxkit/linuxkit/src/cmd/linuxkit` will not work if the `go.mod` has `replace` directives; I **must** either have no such directives, or explicitly clone the repo, cd into it, and then `go install .`

and, of course, `replace` directives are ignored in transient dependencies, only in the main module... but I cannot use them there if I want to do a straight `go install` from anywhere.

[...]

PR #3912, opened 2023-02-16
update to latest manifest-tool, oras-v2, remove all replace in go.mod

deitch:
**- What I did**

Now that manifest-tool includes only latest oras/v2, updated to both of those.

More importantly, this removes the last `replace` commands in `go.mod`, which enables linuxkit to be embedded without all of the build issues, and `go install github.com/linuxkit/linuxkit/src/cmd/linuxkit@latest` to work again.

**- How I did it**

`go get` then remove the `replace` statements, then `go mod tidy` then `go mod vendor`.

**- How to verify it**

CI.

go install works again


deitch:
Fixes #3873

rn (approved):

PR #3913, opened 2023-03-10
Retain /lib/apk/db for SBOM tools

eriknordmark:
**- What I did**
This allows SBOM tools to look at /lib/apk/db/installed to determine which package versions are included in the container. This should probably be applied across all of the linuxkit containers.

**- How I did it**

Preserve /etc/apk and /lib/apk by copying it to the container

**- How to verify it**

Inspect the produced containers to check that they have /lib/apk/db/installed

Retain /lib/apk/db for SBOM tools

### DIFF
--- a/pkg/containerd/Dockerfile
+++ b/pkg/containerd/Dockerfile
@@ -12,3 +12,5 @@ COPY --from=containerd-dev /usr/bin/containerd /usr/bin/ctr /usr/bin/containerd-
 COPY --from=alpine /usr/share/zoneinfo/UTC /etc/localtime
 COPY --from=alpine /etc/init.d/ /etc/init.d/
 COPY etc etc/
+COPY --from=alpine /etc/apk /etc/apk/
+COPY --from=alpine /lib/apk /lib/apk/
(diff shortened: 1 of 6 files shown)


pkg/containerd/Dockerfile
```diff
@@ -12,3 +12,4 @@ COPY --from=containerd-dev /usr/bin/containerd /usr/bin/ctr /usr/bin/containerd-
 COPY --from=alpine /usr/share/zoneinfo/UTC /etc/localtime
 COPY --from=alpine /etc/init.d/ /etc/init.d/
 COPY etc etc/
+COPY lib/apk lib/apk/
```
deitch, line 15: I think you meant this to be `COPY --from=build /lib/apk /lib/apk`

Same for the others

deitch (changes requested):

pkg/containerd/Dockerfile
eriknordmark: @deitch updated PR with --from

deitch:
Let CI run.

deitch (approved):

deitch:
Done and merged. Give it about 10-15 minutes and new images should be pushed out automatically.

PR #3924, opened 2023-05-04
kernel: enable CONFIG_PSI

tonistiigi:
**- What I did**

Enabled `CONFIG_PSI` for 5.15 kernels. PSI https://facebookmicrosites.github.io/psi/docs/overview enables `cpu/memory/io.pressure` files in cgroupv2 (already enabled) as well as `/proc/pressure` that is useful for analyzing the performance of applications and the system. 

**- How I did it**

**- How to verify it**

Build kernel and check `/proc/pressure`.

```
- Enable CONFIG_PSI in 5.15.x kernel configuration
```


djs55 (approved):
Makes sense to me

errordeveloper:
FYI I am also making an update to the latest kernel versions in https://github.com/linuxkit/linuxkit/pull/3919, with the intention to enable BTF after that also.

errordeveloper:
Yeah, I was just saying, cause I noticed that BTF PR too.

PR #3940, opened 2023-07-10
Update README.md

liamsegura:
Documentation update, fixed punctuation.

**- What I did**

Fixed documentation

**- How I did it**

with incredible difficulty 

**- How to verify it**

eyes

Huge documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,17 +4,17 @@
 
 LinuxKit, a toolkit for building custom minimal, immutable Linux distributions.
 
-- Secure defaults without compromising usability
-- Everything is replaceable and customisable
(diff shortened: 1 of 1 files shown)


deitch:
Why is this needed?

stappersg:
> Why is this needed?

Some documentation conventions say _each bullet point ends with a point (as a regular sentence)_

Some documentation conventions say _each bullet point is just a few words (and don't need a point at end)_

This comment is however a plea for _make a decission_  so we can move on. In stead of letting this open issue drain more human energy.

deitch:
Sure, this is fine, then. But it does need the signing and DCO.

PR #3942, opened 2023-07-24
move poweroff to full-fledged package

deitch:
**- What I did**

Moved `test/pkg/poweroff/` to `pkg/poweroff/`

It made sense to be in `test/` if and when it was consumed only by things in `test/`. But the aws builder inside linuxkit CLI uses it, as evidenced by its presence in `src/cmd/linuxkit/moby/mkimage.yaml`. Hence, it should be treated as a first-class (not test-only) package. This also makes it part of the usual build pipeline.

**- How I did it**

`mv test/pkg/poweroff pkg/`

After done and committed, I checked that the `show-tag` had not changed (it hadn't), so nothing else to update.

**- How to verify it**

CI

`pkg/poweroff` in top of tree instead of under `test/`


deitch:
Fixes #3943

MatheusAfinovicz (approved):

PR #3947, opened 2023-09-18
Add basic support for CVM scenario

yuguorui:
**- What I did**
Confidential virtual machines is a security technology currently being invested in by major CSPs. [[1]](https://www.alibabacloud.com/help/en/ecs/user-guide/build-a-tdx-confidential-computing-environment) [[2]](https://techcommunity.microsoft.com/t5/azure-confidential-computing/azure-confidential-vms-using-sev-snp-dcasv5-ecasv5-are-now/ba-p/3573747) [[3]](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html)

It aims to minimize the security impact of the CSP itself on tenant VM instances. By encrypting and isolating memory, registers, interrupt handling, etc., CSPs or unauthorized third parties will not be able to view customer data running on their platforms. The current mainstream confidential virtual machine technologies include: [Intel TDX](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-trust-domain-extensions.html) and [AMD SEV-SNP](https://www.amd.com/system/files/TechDocs/SEV-SNP-strengthening-vm-isolation-with-integrity-protection-and-more.pdf), etc.

We believe that immutable distributions like linuxkit are the key technology to support CVM scenarios, because traditional distributions contain too much software (which will lead to unacceptable TCB sizes) and include by default programs that may be considered backdoors in CVM scenarios (such as cloud-init), and cannot be easily rebuilt, etc.

So I added basic CVM support to linuxkit.

**- How I did it**
I mainly did the following work:
1. Adapt to the upstream kernels, TDX and SNP require at least v5.19 kernel version;
2. Allow the use of larger initramfs (larger than 4GiB).
3. Add Alibaba Cloud image upload support, which supports Intel TDX currently.

Please note step 2 is not necessary, but in practice we find that some applications often require larger initramfs, such as LLM (Large language model).

**- How to verify it**
You can refer to README-CVM.md to verify my work :).

Add basic support for CVM scenario.

A shiba dog play with League of Legends and get winning


deitch:
Hi @yuguorui ; thanks for this. I have been following the various forms of secure compute for a while, so this is an interest.

I think you might be based on some old commits. I see 9 commits here, and 1,140 files changed, when I can see many of those are changes that already exist. Can you check that out, see if you need to rebase, and then bring it down to just the important commit(s)? If it is a lot of changes, but they can be done in sequence, then we can have separate commits.

I am looking forward to having this in.

yuguorui:
> Hi @yuguorui ; thanks for this. I have been following the various forms of secure compute for a while, so this is an interest.
> 
> I think you might be based on some old commits. I see 9 commits here, and 1,140 files changed, when I can see many of those are changes that already exist. Can you check that out, see if you need to rebase, and then bring it down to just the important commit(s)? If it is a lot of changes, but they can be done in sequence, then we can have separate commits.
> 
> I am looking forward to having this in.

Thank you for your positive opinion, it means a lot to me. I will rearrange the commits as soon as possible to meet your requirements.

yuguorui:
I have removed the patch "Add image push support for Alibaba Cloud" section. 

It introduced too many changes due to the new dependencies, which leads to an update of the vendor directory. I think this PR should look more readable.

deitch:
Yes, this looks easier and cleaner. Probably should squash it to a single commit. I will comment inline with some questions.

README-CVM.md
```diff
@@ -0,0 +1,39 @@
+## How to build a CVM compatiable image
+
+```sh
+git clone https://github.com/linuxkit/linuxkit.git
+cd linuxkit
+git fetch origin pull/3947/head:cvm && git checkout -b cvm
```
deitch, line 6: This mostly should go away. This describes how to do it as long as the PR is open. The assumption of the PR is that this will merge into master soon.

deitch:
Can you explain the structure of the changes in here?

- what is different about CVM-compatible image? Is it a kernel config that is enabled, in which case, why would we not just enable it always? Or does it slow it down?
- what is in `contrib/foreign-kernels`, with the "`Dockerfile.rpm`? 
- what is different about the example file?
- why do we have this additional `grub/Dockerfile.rhel`?
- what is different about the new builder `mkimage-raw-efi-ext4`?  There always is a `mkimage-raw-efi`, which builds an EFI disk image, what does this change?

Thanks

PR #3960, opened 2023-11-25
FIX disk parameter usage

Fabsolute:
**- What I did**
The `-disk` parameter is not working with `linuxkit run virtualization`. Fixed

**- How I did it**
Changed conditions about disk check and created image if file not provided

**- How to verify it**
CI

`-disk` usage fixed for `linuxkit run virtualization`

Here is my cat Sanji
sanji

### DIFF
--- a/src/cmd/linuxkit/run_virtualizationframework_darwin_cgo_enabled.go
+++ b/src/cmd/linuxkit/run_virtualizationframework_darwin_cgo_enabled.go
@@ -182,18 +182,33 @@ func runVirtualizationFramework(cfg virtualizationFramwworkConfig, path string) 
 
 	var storageDevices []vz.StorageDeviceConfiguration
 	for i, d := range cfg.disks {
-		var id, diskPath string
+		var id string
 		if i != 0 {
 			id = strconv.Itoa(i)
 		}
+		if d.Size != 0 && d.Format == "" {
+			d.Format = "raw"
+		}
+		if d.Format != "raw" && d.Path == "" {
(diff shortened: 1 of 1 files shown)


deitch:
Cat is cute. 😄

deitch (commented):
Thanks for fixing this. I see how you did it, makes sense.

Only question I have is, why change `d` (which is the item coming from the `range`), as opposed to using `diskPath` the var? Changing an item in a loop, when your purpose is not to change the item, is a bit anti-pattern, isn't it?

Fabsolute:
Actually I made a little bit consistent with the other implementations like qemu, vmware etc.

deitch:
Oh, yeah? Then I won't complain. 😁

deitch (approved):

deitch:
CI can take a while.

nustiueudinastea:
Hey @deitch any chance we can merge this? PR looks pretty good

deitch:
Yeah. We have one failing test, we should rerun it, but I cannot find the button to do so.

Fabsolute:
> Yeah. We have one failing test, we should rerun it, but I cannot find the button to do so.

I can create a dummy commit to retrigger the tests 🤔

deitch:
Would you please? I am wrangling with the GitHub CLI to get it to do the same thing, and it is not making it easy.

deitch:
Running.

Fabsolute:
Oh, snap! It failed again. I rebased and pushed again.

PR #3968, opened 2023-12-18
logread: do not close socket too early

christoph-zededa:
only close socket once reading is finished

**- What I did**
Fix bug where `logread` does not return anything

**- How I did it**
Closing socket only after reading from socket is finished.

**- How to verify it**
Run `logread` in eve and see that it now prints out log messages.

Fix logread bug but not closing socket before reading.

https://www.youtube.com/watch?v=AV9z0c1hjnA&pp=ygUPYmVzdCByYXQgdHJpY2tz

### DIFF
--- a/pkg/memlogd/cmd/logread/main.go
+++ b/pkg/memlogd/cmd/logread/main.go
@@ -67,7 +67,6 @@ func StreamLogs(socketPath string, follow, dump bool) (<-chan LogEntry, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer conn.Close()
 
 	var n int
 	switch {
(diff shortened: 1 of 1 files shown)


deitch:
Was the problem that `conn.Close()` was being called when `StreamLogs()` is finished, but the goroutine lasts much longer, and therefore it should not be closed until after the goroutine is complete?

christoph-zededa:
> Was the problem that `conn.Close()` was being called when `StreamLogs()` is finished, but the goroutine lasts much longer, and therefore it should not be closed until after the goroutine is complete?

Yes.

deitch (approved):

christoph-zededa:
```
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Atonistiigi%2Fbinfmt%3Apull&service=registry.docker.io": read tcp 10.1.0.5:59566->54.196.99.49:443: read: connection reset by peer.
```
does not look like my change caused this ...

deitch:
> does not look like my change caused this

It looks like a network issue, but much more fun if I get to blame someone. 😆 

I will restart those tests.

PR #3969, opened 2024-01-08
Add support for loongarch64

yzewei:
**- What I did**
Add binfmt support for loongarch64 architecture.
**- How I did it**
 Update pkg/binfmt/Dockerfile and pkg/binfmt/etc/binfmt.d/00_linuxkit.conf
 Add QEMU masks and magics to loongarch
**- How to verify it**
docker buildx build --load -t linuxkit/binfmt:addloongarch64 -f Dockerfile .
docker run --rm --privileged -it linuxkit/binfmt:addloongarch64 
docker run --rm -it yangzewei2023/debian:sid uname -a
Linux 2a333d9eb0ea 6.1.0-11-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-4 (2023-08-08) loongarch64 GNU/Linux

Add support for cross-building loongarch64 images

image


deitch:
I am not sure I understand the purpose. I understand adding loongarch64 support to binfmt. But `linuxkit/binfmt` is used almost exclusively inside linuxkit images, i.e. to be able to run other commands inside a linuxkit image running somewhere (virtual, cloud, metal, etc.). Nothing stops you from using them elsewhere, but that is the primary purpose. Is that what you are after?

If so, you need to update all of the dependencies so that CI will pick it up. From within the root, run:

```bash
./scripts/update-component-sha.sh --pkg pkg/binfmt
```

That will update all of the tests and examples and everything else that use `linuxkit/binfmt`, and so ensure that CI catches it.

yzewei:
> I am not sure I understand the purpose. I understand adding loongarch64 support to binfmt. But `linuxkit/binfmt` is used almost exclusively inside linuxkit images, i.e. to be able to run other commands inside a linuxkit image running somewhere (virtual, cloud, metal, etc.). Nothing stops you from using them elsewhere, but that is the primary purpose. Is that what you are after?
> 
> If so, you need to update all of the dependencies so that CI will pick it up. From within the root, run:
> 
> ```shell
> ./scripts/update-component-sha.sh --pkg pkg/binfmt
> ```
> 
> That will update all of the tests and examples and everything else that use `linuxkit/binfmt`, and so ensure that CI catches it.


Thank you, adding loongarch64 support allows users to build loongarch64 in other architectures. Since the mainstream architecture is mainly amd64 at this stage, I think this is necessary for lovers who want to use the loongarch64 architecture.

deitch:
> Thank you, adding loongarch64 support allows users to build loongarch64 in other architectures. Since the mainstream architecture is mainly amd64 at this stage, I think this is necessary for lovers who want to use the loongarch64 architecture

That is what I am trying to understand. The mainstream architecture is equally amd64 and arm64 (we used to have s390x support, but it was not being used by anyone seriously, and getting a build server was difficult, so we dropped it).

It depends what you mean by "allows users to build loongarch64 in other architectures".

* If you mean, "I want to run a linuxkit-composed OS image, and then run loongarch64 process (or containers) on it", then, yes, what you did should do it.
* If you mean, "I want to build linuxkit-composed OS images for loongarch64", then this part is just a drop in the bucket. You would need to extend all images, update some parts of the linuxkit binary, update build pipelines, create kernels, and a whole bunch I am missing.

Which are you after?

yzewei:
> > Thank you, adding loongarch64 support allows users to build loongarch64 in other architectures. Since the mainstream architecture is mainly amd64 at this stage, I think this is necessary for lovers who want to use the loongarch64 architecture
> 
> That is what I am trying to understand. The mainstream architecture is equally amd64 and arm64 (we used to have s390x support, but it was not being used by anyone seriously, and getting a build server was difficult, so we dropped it).
> 
> It depends what you mean by "allows users to build loongarch64 in other architectures".
> 
> * If you mean, "I want to run a linuxkit-composed OS image, and then run loongarch64 process (or containers) on it", then, yes, what you did should do it.
> * If you mean, "I want to build linuxkit-composed OS images for loongarch64", then this part is just a drop in the bucket. You would need to extend all images, update some parts of the linuxkit binary, update build pipelines, create kernels, and a whole bunch I am missing.
> 
> Which are you after?

Thank you for your patience. As you said, linuxkit is huge scale. At the current stage, what I want to run a linuxkit-composed OS image, and then run loongarch64 process (or containers) on it.

deitch:
In that case, I think this works. Let's run CI and be sure it is clean.

deitch (approved):

yzewei:
> In that case, I think this works. Let's run CI and be sure it is clean.

There seems to be no problem.

PR #3970, opened 2024-01-15
Add USB/IP

### LINKED ISSUE #609: remote usb
See http://usbip.sourceforge.net/ and https://kernelnewbies.org/Linux_4.7#head-b998aa95aff18633d421b139478371d2759bdeb4 - should be able to remote USB devices from the Mac or Windows in theory...

beriberikix:
**- What I did**

Enable USB/IP support so USB devices from hosts can be used from within a container over the USB/IP protocol.

Only enables the Core & VHCI so that a container can act as a USB "client." The Host module is not enabled, nor are verbose logging.

Fixes #609

**- How I did it**

I started with a `defconfig` from a local kernel source.Enabled USB/IP and the VHCI module, and copied the changed values.

**- How to verify it**

Build a custom build of LinuxKit and test on an OS without USB support, like macOS. This Rust client works on macOS: https://github.com/jiegec/usbip.

Note:  I have not explicitly tested this with LinuxKit but have used USB/IP many times, including on Windows.

Add support for USB/IP so hosts like macOS can passthrough USB devices to a container.


deitch:
I am all in favour of this. It won't be enough, though. You need to generate new kernel images, then update the `.yaml` files that depend on them (test, example, etc.). I don't quite remember how, but I am pretty sure it is documented in the `kernel` folder. Give it a shot, if you get stuck with something that needs a maintainer, let me know.

beriberikix:
Thank you for the reply @deitch. I've been digging through the docs and I haven't found instructions on how to contribute new kernel images - can you share any pointers?

deitch:
I have done kernels only rarely. @rn always was the expert. I think @djs55 has also done some work on them.

I assume you read [this](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#modifying-the-kernel-config)?

beriberikix:
I did, thanks! AFAICT that simply describes how to customize the `.config` files, which is already done. I think the changes you described are in `[releasing.md](https://github.com/linuxkit/linuxkit/blob/master/docs/releasing.md)`. Unfortunately, "Releases can be done by any maintainer." :/ IUC, new kernels need to be pushed first to Dockerhub and then the yaml files can be updated, which can only be done by maintainers.

deitch:
Not quite. That releasing doc is about cutting semver linuxkit releases, not kernels.

Looks like it is just `make build` and then `make push` in `kernels/`. You can do `make build`, but not `make push`, since it requires a maintainer, so I won't trouble you. `make build` is running now in a codespaces; I imagine it will take a while.

beriberikix:
Does this need approval to actually [run the CI](https://github.com/linuxkit/linuxkit/actions/runs/7523119641)?

deitch:
Yeah, it does, but no point to it until we get the rebuilt kernel images pushed out. Will get it.

deitch:
They are building and will be pushed out.

deitch:
I ran into an issue, more related to the build process. Did you modify the `config-*` files by hand? Or via `make kconfig` and running the `linuxkit/kconfig` image?

beriberikix:
Ah, I did modify them by hand. Let me revert and do it correctly.

deitch:
I would hold on that. The builds were based on much older base linux, which caused issues. I have an open PR #3971 to update those. As soon as those are working and it is pushed out, I will merge that, and then you can rebase on that.

deitch:
@beriberikix what version of the kernel are you actually using?

beriberikix:
Pretty recent, 6.5.0-14. I'm happy to use `make kconfig` & use `linuxkit/kconfig` with the current base image to verify it can work and then update to the newer base image and squash commits once your other PR is merged!

deitch:
Ah, so you aren't using these older ones. That is good to know. I have been trying to update those, and running into some issues.

deitch:
It has been a long haul, down a deep rabbit hole of updating lots of other things. But I know you are waiting for this @beriberikix .

I managed to get 6.6.13 built, after updating the dependencies (alpine build base for kernels from 3.13 to 3.17 and then 3.19),  but that in turn required updating the base build image for _everything_ and then all of its dependencies, as well as some missing dependencies in the kernel builds to do so. The nice part is that the 6.x series required no patching, which makes it so much cleaner.

I expect the base builder image update to be ready tomorrow (Friday), barring any unusual CI issues. Then the kernel update should be ready after the weekend. At that point, you can rebase this all, and add the support.

We do not intend to keep rebuilding the 5.x kernels (except _perhaps_ 5.15.27, the most recent maintained 5.x version). Is there any reason not to put this support just in the 6.6.x series? It will have a config file as soon as that PR #3971 is in, as I said above, hopefully early next week.

beriberikix:
Thank you for slogging through that! From my perspective, it would be fine to focus on 6.x, my use case is focused on more recent kernels anyway! Just let me know what versions to target (ex. if I should include 5.15.27 or not.)

deitch:
Alpine update is complete, building 6.6.13 and toolset on top of it. Still need 6.6.13-dbg. If all goes well, you should be able to add this easily after that

deitch:
Finally! 

alpine is updated, kernel 6.6.13 is in! #3971 just closed.

Please rebase on the latest `master`, and go ahead and make your changes.

beriberikix:
@deitch I somehow messed up my tree to merge-hell oblivion and couldn't dig myself out. I decided it was easier to just start a new branch and make the minor changes from a clean slate. #3979 should be ready for review :crossed_fingers:

PR #3983, opened 2024-02-15
Rebased and re-implemented USBIP configs

### LINKED ISSUE #609: remote usb

beriberikix:
Old PR autoclosed so created a new one. `amd64` built but `arm64` is still having issues.

Fixes #609


deitch:
I can try it locally. But why are there so many changes in aarch64? I would have thought just the USBIP-related ones?

beriberikix:
I was wondering that too - I followed the exact same steps as I did for `x86_64` except I ran `make ARCH=arm64 defconfig` before `make ARCH=arm64 menuconfig` per the docs. Let me see if that's the source (wouldn't explain _why_.)

beriberikix:
So I switched back to `master` on my local machine and ran:

```
make kconfig
docker run --rm -ti -v $(pwd):/src linuxkit/kconfig
make ARCH=arm64 defconfig
cp .config /src/config-6.6.x-aarch64
```

I'd expect no changes but it introduced most of what we're seeing. Maybe something changed in `/linux/arch/arm64/configs/defconfig` before the last time the config was built?

deitch:
I can confirm that amd64 on your branch is fine, and arm64 is fine on `master` (so is amd64, but that isn't the subject of our discussion).

beriberikix:
Okay! I also wanted to make sure that USB/IP doesn't enable any transitive dependencies, so I verified that against main. Now with the manual changes I still can't build when I run `make build ARCH=arm64` on my branch or `master`. Any ideas?

deitch:
I suspect that has more to do with an error in cross-compilation. They work fine on `master` when I build directly on those architectures. I will try your branch on there.

Switch this from draft to ready?

deitch:
Successfully built and pushed out. Mark this as ready and we can run CI.

beriberikix:
Ready and squashed!

deitch:
CI running, let it rip!

deitch (approved):

deitch:
Thanks for the patience as we worked this out. I pushed out the images, so download it and check that it all works as you expect?

beriberikix:
Verified with QEMU! Planning on testing in Virtualization Framework next, but QEMU is enough to prove it's there.

Thank you for shepherding this through!

deitch:
Great! One of these lifetimes, we will get the kernel process part of the CI build. All packages used to be like that, so there s hope.

PR #3993, opened 2024-02-29
switch kernel builds to linuxkit pkg build for simplicity

deitch:
**- What I did**

Replace kernel complex Makefile build with simple wrappers around `linuxkit pkg build`.

With the latest changes to `linuxkit`, we are able to simplify long Makefile lines with simple commands. This is because `linuxkit pkg build` and `linuxkit pkg push` already have all of the logic to handle local and remote buildkit runners, push only if it exists, merging images in a remote index into what we push, etc.

**- How I did it**

Changes mainly to Makefile and documents. Added `build-args` files

**- How to verify it**

I run it and build and the kernel is correct.

Simplified kernel build process


deitch:
Still testing it, takes a while.

deitch:
And it works nicely. The internal makefile steps for building and pushing are simplified materially, and take advantage of linuxkit cache and runners and all. Next step will be to use these to add kernel builds and pushes as part of CI.

PR #3995, opened 2024-03-01
push arch-specific tags optional

deitch:
**- What I did**

Expanding on #3991 and #3992, make the pushing of arch-specific tags optional.

**- How I did it**

Changed the name of a parameter to `cache.Push()`, used it, and then made it available as a CLI flag

**- How to verify it**

Test. I did.

`linuxkit cache push` has option whether or not to add arch-specific tags


deitch:
Fixes #3996

PR #4020, opened 2024-03-19
README.md: fix format parameter

christoph-zededa:
linuxkit build expects '--format' or '-f' and not
'-format'

**- What I did**

**- How I did it**

**- How to verify it**


deitch (approved):

deitch:
Thanks. We switched from plan9-style to Unix-style a while back, so we could use the cobra package. I guess we missed a doc.

PR #4021, opened 2024-04-12
linuxkit virtualization run: fixed disk path argument

nustiueudinastea:
This PR fixes the virtualization run subcommand (on MacOS) which was not working previously.

Previously the diskPath variable was not correctly set when trying to attach a disk.

A bigger issue, which this PR does not tackle, is that generally linuxkit attempts to create a disk file if the file is not there, but the virtualization code does not do that. 

**- What I did**

Fixed the diskPath variable so that it get's correctly set and passed to vz.NewDiskImageStorageDeviceAttachment

**- How I did it**

**- How to verify it**

Run linuxkit as follows:

```
linuxkit run virtualization --disk file="disk.raw" ...rest of parameters
```


nustiueudinastea:
Silly me, there was already a better PR fixing this: https://github.com/linuxkit/linuxkit/pull/3960

PR #4025, opened 2024-04-18
include image reference as source in every tar file header

deitch:
**- What I did**

The initial output from `linuxkit build` is a tar of the rootfs. This usually (but not always) gets converted to another format, e.g. raw disk or squashfs or iso, etc.

This adds an extended header `LINUXKIT.packagesource` to every tar file in the initial tar output. This does not get carried over into other formats. This is good for record-keeping, and knowing the sources of things. More importantly, it can be used to create incremental updates. For example, if you run `linuxkit build --format tar foo.yaml`, and then change a single docker image and rerun it, it does not need to rebuild the whole thing, which can take some time. If the images are large, and each layer is gzipped, it can take a while. In the future, we could read `foo.yaml`, look at an input tar, and just copy over any files from the input tar that are in the `foo.yaml`. Only those that are not in the modified `foo.yaml` are left out, and then just those need to be added.

Also added a test.

This will not quite be ready yet, as we only report the source, e.g. `linuxkit/init:abcd1234`, and not why it is in a specific location, e.g. is it part of onboot or services or init?

**- How I did it**
Changed source, added tests

**- How to verify it**

CI, since I added tests

Track source of every file


deitch:
Now this includes 2 PAX headers, one each for the package that it is sourced from (or linuxkit itself) and one for the location in the file.

Issue #4032, opened 2024-04-26
[init] Make cgroups v2 the default


jacobweinstock:
**Description**

Hello. I would love to see cgroups v2 be the default in the init package. Currently, cgroups v1 is the default. See [ref](https://github.com/linuxkit/linuxkit/blob/master/pkg/init/cmd/rc.init/main.go#L228-L250). While there is a kernel parameter, `linuxkit.unified_cgroup_hierarchy=1`, to enable cgroups v2 this presents a challenge when some services (like Docker) in the Linuxkit build want/require cgroups v2. We must rely on the consumer that runns the Linuxkit build to provide the kernel parameter. While documenting this is fine, it leaves open the possibility of a user not setting this, having services fail, and making the experience challenging. I'm willing to make the PR. Thanks!

FYI, making cgroups v2 was referenced in #3719 back in 2021.

**Steps to reproduce the issue:**

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

PR #4033, opened 2024-04-26
Make cgroups v2 the default in the init pkg

jacobweinstock:
**- What I did**
cgroups v2 has been out since 2015. Not having to set a kernel parameter helps improve the user experience by not requiring it when it is required by services in a build. Making this the default was discussed back in 2021 here #3719 . This PR make cgroups v2 the default and cgroups v1 opt in via the kernel parameter: `linuxkit.unified_cgroup_hierarchy=0`.

Fixes: #4032 

**- How I did it**
Reverse the current logic to look for a kernel parameter to enable cgroups v1.

**- How to verify it**

make cgroups v2 the default in the init pkg.

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -2,7 +2,7 @@ kernel:
   image: linuxkit/kernel:6.6.13
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
-  - linuxkit/init:45a1ad5919f0b6acf0f0cf730e9434abfae11fe6
+  - linuxkit/init:8a7b6cdb89197dc94eb6db69ef9dc90b750db598
   - linuxkit/runc:6062483d748609d505f2bcde4e52ee64a3329f5f
   - linuxkit/containerd:e7a92d9f3282039eac5fb1b07cac2b8664cbf0ad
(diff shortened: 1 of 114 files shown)


deitch:
This looks great, and, yes, it is time to make it the default.

Can you update all of the downstream dependencies, mainly tests and some sample build files? There is simple tooling to do it. From the repository root directory, run:

```bash
./scripts/update-component-sha.sh --pkg ./pkg/init
```

That should update all of the files, and then you can commit them in (same commit).

jacobweinstock:
> This looks great, and, yes, it is time to make it the default.
> 
> Can you update all of the downstream dependencies, mainly tests and some sample build files? There is simple tooling to do it. From the repository root directory, run:
> 
> ```shell
> ./scripts/update-component-sha.sh --pkg ./pkg/init
> ```
> 
> That should update all of the files, and then you can commit them in (same commit).

Thanks, @deitch ! Updated. Hope I did it correctly?!

jacobweinstock:
Ok, I think maybe i got it working correctly now?!

deitch (approved):

deitch:
Yup, that looks right. Let the CI run, when it comes back green, we can merge it in.

PR #4038, opened 2024-05-02
linuxkit/build: implement --modified-only option

rouming:
This implements `--modified-only` option, which should complement recent `--input-tar` feature by generating resulting tarball with modified only files.

Sure thing. Sad Hamster himself:

### DIFF
--- a/src/cmd/linuxkit/build.go
+++ b/src/cmd/linuxkit/build.go
@@ -56,6 +56,7 @@ func buildCmd() *cobra.Command {
 		noSbom             bool
 		sbomOutputFilename string
 		inputTar           string
+		modifiedOnly       bool
 		sbomCurrentTime    bool
 		dryRun             bool
 	)
@@ -116,6 +117,9 @@ The generated image can be in one of multiple formats which can be run on variou
 
 			if inputTar != "" && pull {
 				return fmt.Errorf("cannot use --input-tar and --pull together")
+			}
+			if inputTar == "" && modifiedOnly {
+				return fmt.Errorf("--modified-only can be used together with the --input-tar")
 			}
 
 			var outfile *os.File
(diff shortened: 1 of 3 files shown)


rouming:
Does not copy files to the resulting tarball if they are not modified. The resulting tarball will be populated only with files, which we get through the regular OCI procedure. For example here the eve/kdump was modified:
```  
Create OCI config for docker.io/lfedge/eve-kdump:3f3131a76693faef830263d635d2b3beef78787c-dirty-ad49020-amd64
Image docker.io/lfedge/eve-kdump:3f3131a76693faef830263d635d2b3beef78787c-dirty-ad49020-amd64 arch amd64 found in local cache, not pulling
Image docker.io/lfedge/eve-kdump:3f3131a76693faef830263d635d2b3beef78787c-dirty-ad49020-amd64 arch amd64 found in local cache, not pulling
```
So only kdump will be included. The rest is skipped.

deitch:
Let's run through a simple use case:

1. I run `lkt build -o foo.tar`, it creates a tar file with the following 10 files: `a`,`b`,`c`,`d`,`e`,`f`,`g`,`h`,`i`,`j`.
2. I modify my input so that only files `g` and `h` are changed, replaced with `r` and `s`.
3. I rerun build with `--input-tar foo.tar`. It builds a new tar, call it `new.tar`. The contents are `a`,`b`,`c`,`d`,`e`,`f`,`i`,`j`,`r`,`s` sourced as follows:
    * `a`,`b`,`c`,`d`,`e`,`f`,`i`,`j` - copied from `foo.tar`
    * `r`, `s` - from whatever OCI image generated them

You want to create an option so that I have a tar file that contains _only_ `r` and `s`?

rouming:
> You want to create an option so that I have a tar file that contains only r and s?

That's correct. Motivation is the same: creation of full copy ~900Mb tar and then copy it to the rootfs can take a while. Changed files are usually not greater than ~150Mb, which saves ~10-15 seconds from overall image build time.

But! The issue is (about which I completely forgot) that eventually we need to merge both tars, but that can be done later, let's say if sizes of a tar with modified files grows to the half of the original (--input-tar foo.tar) tar size. Need to do some experiments.  Don't you mind if we keep this PR hanging here for a while?

deitch:
> Don't you mind if we keep this PR hanging here for a while?

Sure, we can keep it open. I need time to think this one through.

rouming:
Will close this for now.

PR #4039, opened 2024-05-07
move moby components that do not have runtime dependencies to own directory

deitch:
**- What I did**

The source directory `src/cmd/linuxkit/moby/` contains all of the following:

* types
* utility functions (mostly related to those types)
* complex functions, those that actually build images, or convert them to/from formats, etc.

This means that anything that accesses or imports the types or utility functions _also_ imports the dependencies of the complex functions, which means pulling in docker, etc., with all of the dependency management that implies.

This PR moves the first two into `src/cmd/linuxkit/moby/util/`, so it is possible to import those without all of the rest.

In retrospect, I easily could have done the reverse: leave those in `moby/` and move the complex ones into `moby/build/` or similar, but this worked fine.

**- How I did it**

Created a directory, moved files over, figured out vet issues, fixed them.

**- How to verify it**

CI. No change in functionality, so either it works or it doesn't.

Internal split of functionality between packages for moby.


deitch:
OK, I changed my mind and reversed it. Types are in `moby/` and build functions (that depend on them) are in `moby/build/`

PR #4046, opened 2024-06-19
Improve support for third-party registry images

### LINKED ISSUE #4045: ReferenceExpand causes image lookup failures with containerd image store
**Description**

We use the [aws-nitro-enclaves-cli](https://github.com/aws/aws-nitro-enclaves-cli) to build Nitro Enclaves images. The Nitro CLI uses LinuxKit to actually build the EIF files.

After upgrading systems to newer versions of Docker (26.1.3) and enabling the containerd image store, we started seeing build failures. We build a local docker image and pass the image reference to the Nitro CLI (which then itself passes it to `linuxkit` with the `--docker` flag).

The failures we are seeing are:
> Linuxkit reported an error while creating the customer ramfs: "Add init containers:\nProcess init image: docker.io/bastion.io/signer-nitro-enclave:local\nImage docker.io/bastion.io/foo:local not found in local cache, pulling\ntime=\"2024-06-19T00:46:13Z\" level=fatal msg=\"Failed to build init tarball from docker.io/bastion.io/foo:local: Could not pull image docker.io/bastion.io/foo:local: error getting manifest for trusted image docker.io/bastion.io/foo:local: GET https://index.docker.io/v2/bastion.io/foo/manifests/local: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:bastion.io/foo Type:repository]]\"\n"

While investigating the failures, I tried to reproduce the issue and came to the realization that [util.ReferenceExpand](https://github.com/linuxkit/linuxkit/blob/4f89f4f67e392ffa8c8bab63dfaf31746e3c11a0/src/cmd/linuxkit/util/reference.go#L23-L42) generated bad references in certain scenarios.

The root cause is that `ReferenceExpand` will always append `docker.io/`, even if the image is tagged with a registry domain. This worked for some reason with earlier Docker engine versions, but with newer versions the prepended `docker.io/` will cause the local image to not be found (and will cause Docker to try and pull the image from the wrong registery).

If we re-tag the local image as `bastion/foo:local` and pass that to LinuxKit, everything works fine:
```
INFO[0000] Add init containers:                         
INFO[0000] Process init image: docker.io/bastion/foo:local 
DEBU[0000] image tar: docker.io/bastion/foo:local rootfs/ 
DEBU[0000] docker inspect image: docker.io/bastion/foo:local 
DEBU[0000] docker inspect image: docker.io/bastion/foo:local 
DEBU[0000] docker inspect image: docker.io/bastion/foo:local...Done 
DEBU[0000] docker create: docker.io/bastion/foo:local 
DEBU[0000] docker create: docker.io/bastion/foo:local...Done 
DEBU[0000] docker export: 0ed87efeff945cdabb534e4f453d92038d18d711b4d89e0468ce20e200000000 

```

1. Build or tag a docker image with a non-docker registry URL: `docker tag alpine:latest bastion.io/alpine:latest`
2. Create a YAML config file that references an init image with a non-docker registry:
    ```
    prefix: rootfs/
    init:
      - bastion/alpine:latest
    files:
      - path: rootfs/dev
        directory: true
        mode: "0755"
    ```
3. Run LinuxKit with `--docker`: `linuxkit -v build --name customer-initrd.img --docker --format kernel+initrd --no-sbom conf.yaml`
4. Observe failures trying to local the image locally. 

```
Error: failed to build init tarball from docker.io/bastion.io/alpine:latest: Could not pull image docker.io/bastion.io/alpine:latest: error getting manifest for trusted image docker.io/bastion.io/alpine:latest: GET https://index.docker.io/v2/bastion.io/alpine/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:bastion.io/alpine Type:repository]]
2024/06/19 04:04:14 error during command execution: failed to build init tarball from docker.io/bastion.io/alpine:latest: Could not pull image docker.io/bastion.io/alpine:latest: error getting manifest for trusted image docker.io/bastion.io/alpine:latest: GET https://index.docker.io/v2/bastion.io/alpine/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:bastion.io/alpine Type:repository]]
```

I expected the image to be found locally from the Docker image store.

I believe the solution is to simply avoid prepending the `doker.io/` prefix if the image is already prepended with a domain.

jalaziz:
**- What I did**

Fix `util.ReferenceExpand` to support image references from non-DockerHub remote registries.

**- How I did it**

If the first part of an image reference looks like a domain, we simply return it as is.

**- How to verify it**

Improve support for image references from remote registries.

fixes #4045

PXL_20240420_190655037 PORTRAIT ORIGINAL

### DIFF
--- a/src/cmd/linuxkit/util/reference_test.go
+++ b/src/cmd/linuxkit/util/reference_test.go
@@ -1,0 +1,59 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReferenceExpand(t *testing.T) {
+	tests := []struct {
+		name    string
+		ref     string
+		options []ReferenceOption
+		want    string
+	}{
+		{
+			"basic image name should expand to docker.io/library image",
+			"redis",
+			nil,
+			"docker.io/library/redis",
+		},
+		{
+			"image name with user/org should expand to docker.io image",
+			"foo/bar",
+			nil,
+			"docker.io/foo/bar",
+		},
+		{
+			"custom registry image name should not expand",
+			"myregistry.io/foo",
+			nil,
+			"myregistry.io/foo",
+		},
+		{
+			"image name with more than three parts should not expand",
+			"foo/bar/baz",
+			nil,
+			"foo/bar/baz",
+		},
+		{
+			"with tag should add latest if image does not have tag",
+			"redis",
+			[]ReferenceOption{ReferenceWithTag()},
+			"docker.io/library/redis:latest",
+		},
+		{
+			"with tag should not add latest if image already has tag",
+			"redis:alpine",
+			[]ReferenceOption{ReferenceWithTag()},
+			"docker.io/library/redis:alpine",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := ReferenceExpand(tt.ref, tt.options...)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
--- a/src/cmd/linuxkit/util/reference.go
+++ b/src/cmd/linuxkit/util/reference.go
@@ -32,7 +32,12 @@ func ReferenceExpand(ref string, options ...ReferenceOption) string {
 	case 1:
 		ret = "docker.io/library/" + ref
 	case 2:
-		ret = "docker.io/" + ref
+		// If the first part is not a domain, assume it is a DockerHub user/org.
+		// This logic is copied from moby:
+		// https://github.com/moby/moby/blob/e7347f8a8c2fd3d2abd34b638d6fc8c18b0278d1/registry/search.go#L148C29-L149C71
+		if !strings.Contains(parts[0], ".") && !strings.Contains(parts[0], ":") && parts[0] != "localhost" {
+			ret = "docker.io/" + ref
+		}
 	}
 
 	if opts.withTag && !strings.Contains(ret, ":") {


deitch:
Thanks, this makes sense. I think all of the original images were on Docker hub, so it just fell that way naturally.

Do you feel like adding some simple unit tests in `reference_test.go`? I imagine that this is the easiest set to write, just some input strings and expected output strings.

jalaziz:
tests added

src/cmd/linuxkit/util/reference_test.go
```diff
@@ -0,0 +1,59 @@
...
+			"docker.io/foo/bar",
+		},
+		{
+			"custom registry image name should not expand",
+			"myregistry.io/foo",
+			nil,
+			"myregistry.io/foo",
+		},
+		{
+			"image name with more than three parts should not expand",
```
deitch, line 35: Why should it not? Isn't this the equivalent of `docker.io/foo/bar/baz`? Since it has no host, isn't the default docker.io?

deitch (commented):
Looking good. The tests made a question stand out.

src/cmd/linuxkit/util/reference_test.go
jalaziz: The previous logic didn't expand in this scenario and a three part image wouldn't be a valid Docker Hub image afaik.

jalaziz: > Your repository name must contain a combination of alphanumeric characters and may contain the special characters ., _, or -. Letters must be lowercase.

Docker Hub doesn't allow slashes in a name, so a three+ part repo is almost definitely not a Docker Hub repo.

deitch: > The previous logic didn't expand in this scenario and a three part image wouldn't be a valid Docker Hub image afaik.

That is a valid point. Let's leave this, then, until someone in product management at Docker Hub decides to expand that. We can worry about it then.

deitch:
Strange, the 2 CI failures. Hoping they are transient, kicking them off again.

deitch (approved):

PR #4055, opened 2024-07-11
properly save linuxkit version

deitch:
**- What I did**

The version for the `linuxkit` binary was consistently giving `v0.8+` from something hard-coded into the Makefile.

This changes the `src/cmd/Makefile` so it calculates the version. The version will be almost the equivalent of `git describe --tags`.

* If the most recent tag of semver pattern `v*` points to the HEAD commit, then version is injected as `vX.Y.Z` (the semver)
* If the most recent tag points to an earlier commit, then version is injected as `vX.Y.Z-<count>-<shorthash>`, i.e. the equivalent of `git describe --tags`
* In all cases, if there are uncommitted changes, append `-dirty`

**- How I did it**

Changes to `src/cmd/linuxkit/Makefile`

**- How to verify it**

CI obviously, although this should affect nothing in CI. Build it with various commits and run `linuxkit version` and see.

After the change:

```sh
$ linuxkit version
linuxkit version v1.2.0-92-ga7c8ed81f
commit: a7c8ed81f14cf5c11b6fbd50004ce1b1af6e2d26
```

That is _much_ better, and will be helpful in downloaded binaries from GitHub releases.

Proper versions included in linuxkit

### DIFF
--- a/src/cmd/linuxkit/Makefile
+++ b/src/cmd/linuxkit/Makefile
@@ -1,5 +1,24 @@
-VERSION?="v0.0-dev"
+# determine the version we save in the build binary
+# we always include the git commit.
+# the version is the current semver if it this commit matches the tag,
+# else it is the following: <tag>-<commits since tag>-<short commit hash>
+# if the git tree is dirty, append "-dirty"
+# most recent commit
 GIT_COMMIT=$(shell git rev-list -1 HEAD)
+# whether or not it is dirty, i.e. has uncommitted changes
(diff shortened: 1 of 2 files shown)


deitch:
Forgot to write. It also includes an option to `linuxkit version --short` to show just the version, which is convenient for machine processing. Same for `--commit`.

```sh
$ linuxkit version --short
v1.2.0-92-g6bf2010e6
$ linuxkit version --commit
6bf2010e6c8d89cf9838cb84a7f90fb190fd2db5
```

PR #4072, opened 2024-08-27
moby: check architecture for docker image

christoph-zededa:
Fixes that under certain cases the container image is already in the local docker registry, but with the wrong architecture; in this case just pretend it is not there and let the caller decide if they want to build it

**- What I did**

Check that linuxkit does accidentaly take a docker image with the wrong architecture.

**- How I did it**

Check architecture before taking it.

**- How to verify it**

https://github.com/lf-edge/eve/pull/4150 - enable DockerCache and run `make test`

https://media2.giphy.com/media/4Ho5aPCXJEzOqcwddj/giphy.webp?cid=ecf05e47j41oybtdpecws1b12gvms2d2wdg82uciccnty2bd&ep=v1_gifs_search&rid=giphy.webp&ct=g

### DIFF
--- a/src/cmd/linuxkit/docker/cmd.go
+++ b/src/cmd/linuxkit/docker/cmd.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


deitch (approved):
Nice catch, thank you.

PR #4097, opened 2025-01-20
Fix raw efi build image size calculation

### LINKED ISSUE #4095: linuxkit build fails for format raw-efi with "Disk full" error
The "Disk full" is printed by the command `mcopy -i $ESP_FILE $INITRD ::/` in the script `tools/mkimage-raw-efi/make-efi` when running `linuxkit build --format raw-efi`.
My guess is that the following calculation might be wrong under some circumstances (or edge cases):

```bash
ESP_FILE_SIZE_KB=$(( ( ( ($ESP_FILE_SIZE+1024-1) / 1024 ) + 1024-1) / 1024 * 1024 ))
```

Changing `ESP_HEADROOM=$(( 1024 * 1024 ))` to `ESP_HEADROOM=$(( 1024 * 1024 * 2 ))` solves the problem for me.

ChrisIgel:
fixes #4095

**- What I did**
Adjusted the calculations in the raw-efi build script.

**- How I did it**
Looked into the FAT design (https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system) and `mkfs.fat` source code (https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c).

**- How to verify it**

I think the problem for the too small image size is the growing File Allocation Table. The current assumption of 511 KiB of headroom is only true for images with a size of around 250 MB.
This is still only an approximation of the total size an no guarantee that it will work under all circumstances as the calculations in https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c are quite convoluted/complex.

### DIFF
--- a/src/cmd/linuxkit/moby/build/images.yaml
+++ b/src/cmd/linuxkit/moby/build/images.yaml
@@ -3,7 +3,7 @@
  iso-efi:        linuxkit/mkimage-iso-efi:4bff02040897e4177a2220038f4a5a08cd556afd
  iso-efi-initrd: linuxkit/mkimage-iso-efi-initrd:6d087ae7aa61fe6d840e314dcb2d6b7b2a5db1b8
  raw-bios:       linuxkit/mkimage-raw-bios:c64318cfc28b39010b05c7cd48673f397c45c8df
- raw-efi:        linuxkit/mkimage-raw-efi:8bc9d4afbbd5b01488a8c699b3db721484a9eda1
+ raw-efi:        linuxkit/mkimage-raw-efi:6d9a9ff0a575f8ba5fcd5bf25feac89d0aa343d0
  squashfs:       linuxkit/mkimage-squashfs:d9de62071bed575377e35e944004bcb23c50afab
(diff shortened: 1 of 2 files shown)


deitch (dismissed):

deitch:
This is mighty nice. Can you run the update script so it catches everything? You need to do the following from linuxkit root directory:

```bash
./scripts/update-component-sha.sh --pkg ./tools/mkimage-raw-efi
```

However, I am not someplace that I can build and push out the package. Hopefully CI will do it correctly.

ChrisIgel:
I updated the hash :)

deitch (approved):

deitch:
Thanks. Hopefully CI will push out the built packages. I never can remember which ones it does and which have to be done in advance. If I need to, I will do it, but it will have to wait a few days.

deitch:
Nope, it did not. `pkg/` stuff gets built automatically, and pushed out when merged-to-master; `tools/` stuff does not.

deitch:
Thanks!

PR #4099, opened 2025-01-22
Try resizing all found devices

### LINKED ISSUE #4098: pkg/extend should not fail early but try to resize all found devices
Currently pkg/extend fails early when it finds a device that cannot be resized. Instead it should just log the error and ignore/skip the device and continue with the remaining devices.

ChrisIgel:
fixes #4098 

**- What I did**
Change the pkg/extend so that extension is tried on all found devices.

**- How I did it**

**- How to verify it**

### DIFF
--- a/pkg/extend/extend.go
+++ b/pkg/extend/extend.go
@@ -22,8 +22,9 @@ import (
 const timeout = 60
 
 var (
-	fsTypeVar string
-	driveKeys []string
+	fsTypeVar   string
+	stopOnError bool
+	driveKeys   []string
 )
 
 // Fdisk is the JSON output from libfdisk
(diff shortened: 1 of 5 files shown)


deitch (changes requested):
Are we sure we always want it to keep going? Are there failure types that should be fatal?

Separately, you will need to update all of the dependencies. From the linuxkit root directory, run `./scripts/update-component-sha.sh --pkg ./pkg/extend` to do so.

deitch (approved):

deitch:
OK, I pushed out the updated image and kicked off CI again. Hopefully this time it will clear.

deitch:
The failures are not related to this change, so overriding.

PR #4100, opened 2025-01-29
containerd 20

deitch:
**- What I did**
Updated containerd to v2.x (specifically v2.0.2).

This then involved changing all downstream dependencies, as well as the binaries that use the containerd library surface. Those are:

- pkg/init
- src/cmd/linuxkit

I then updated the dependencies again.

**- How I did it**o

- `go get` for the modules
- update the library surface manually, then `go mod tidy` and `go mod vendor`
- update dependencies using `update-component-sha.sh`
- `lkt pkg push` for updated `tools/containerd-dev` and `pkg/init` and `pkg/containerd`

**- How to verify it**
CI

containerd v2.x


deitch:
This needs to be updated to containerd v2.0.3 once released. That will contain the fix to `make root-test` necessary to make it pass CI.

PR #4101, opened 2025-01-29
pkg/extend fix panic for empty partition tables

ChrisIgel:
**- What I did**

Fix a go panic when trying to resize a device with a partition table with 0 partitions.

**- How I did it**

Check the condition and return early.

**- How to verify it**

Create and mount a dummy device with an empty partition table and let `pkg/extend` try to resize the device.
The following error should no longer occur:
```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.extend({0xc000013880, 0xa}, {0x5341a3, 0x4})
        /<redacted>/linuxkit/pkg/extend/extend.go:93 +0x119c
main.autoextend({0x5341a3, 0x4})
        /<redacted>/linuxkit/pkg/extend/extend.go:60 +0x1b0
main.main()
        /<redacted>/linuxkit/pkg/extend/extend.go:333 +0x8e
```

### DIFF
--- a/pkg/extend/extend.go
+++ b/pkg/extend/extend.go
@@ -82,6 +82,11 @@ func extend(d, fsType string) error {
 		return fmt.Errorf("Unable to unmarshal partition table from sfdisk: %v", err)
 	}
 
+	if len(f.PartitionTable.Partitions) == 0 {
+		log.Printf("Disk %s has no partitions. Skipping", d)
+		return nil
+	}
+
 	if len(f.PartitionTable.Partitions) > 1 {
 		log.Printf("Disk %s has more than 1 partition. Skipping", d)
 		return nil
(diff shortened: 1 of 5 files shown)


deitch (changes requested):
This is fine, can you run `./scripts/update-component-sha.sh --pkg ./pkg/extend` from within the linuxkit repo root, so the examples and everything pick it up?

ChrisIgel:
> This is fine, can you run `./scripts/update-component-sha.sh --pkg ./pkg/extend` from within the linuxkit repo root, so the examples and everything pick it up?

I already did that :)

deitch:
Yeah, you are right. I was sure it was elsewhere, but it is not.

deitch (approved):

PR #4104, opened 2025-02-07
feat(format): allow custom partition layout specification in linuxkit format

SmartArray:
**- What I did**
Add an option to `linuxkit/format` that allows specifying multiple partitions.

**- How I did it**
I modified the `format.go` source code by adding a `partitionLayoutVar ` that allows specifying a different value than ";" (create one partition)

**- How to verify it**

```bash
/usr/bin/format --layout "size=20M,type=83\nsize=30M,type=83\n" /dev/vda;
```

should create two partition.

allow specification of partition layout in linuxkit/format

Screenshot 2025-02-07 at 07 42 25


SmartArray:
Yeah this was the goal! I would have edited exactly this documentation without you pointing it out! 🙂

 I need to work out the specifics and I am currently checking whether the tooling is working as expected. It looks good so far!

Issue #4131, opened 2025-06-26
Parallel pkg build can trounce each other


deitch:
Parallel runs of `linuxkit pkg build` can trounce each other. Specifically, they can corrupt the linuxkit cache.

The process of saving both blobs and the `index.json` assume that only one process is accessing it at once. If you run 2 parallel builds, they might corrupt `index.json` as they both try to write it.

They also might corrupt actual blob files, but that is less likely.

This issue also documents all of the places that we do writes to the cache, so that we can clean it up.

Everything is in [`src/cmd/linuxkit/cache`](https://github.com/linuxkit/linuxkit/tree/master/src/cmd/linuxkit/cache).

Very few files actually write to the cache, and we don't care quite as much about parallel reads, which are easy.

The 3 files that actually write to the cache are:

* [pull.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/cache/pull.go)
* [push.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/cache/push.go)
* [remove.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/cache/remove.go)
* [write.go](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/cache/write.go)

We will separate between those that write blobs to `blob/sha256/`, and those that write to `index.json`.

In all of the listed examples, `cache`, as in "calling `cache.Func`, refers to the external library that actually writes to the on-disk cache.

## Blobs

We are not concerned for now about writing blobs, for several reasons:

* To cause an issue, we would need two or more parallel processes writing the exact same blob, which is very unlikely.
* Fixing it requires an approach like how containerd does it, with an `ingest/` dir and giving unique filenames to downloads, e.g. `<digest>.<id>`, and then doing a single, quick, atomic `mv`.
* We rely entirely on the underlying [ggcr library](https://github.com/google/go-containerregistry), which likely does not support what we are trying to do.
 
If it becomes an issue in the future, we can deal with it then.

References to direct writing of blobs via GGCR:

* [pull.go calls cache.WriteIndex](https://github.com/linuxkit/linuxkit/blob/3f54a80824cca1dc00a2c500e13716df09740b95/src/cmd/linuxkit/cache/pull.go#L195)
[...]

PR #4132, opened 2025-06-26
Centralize safe cache writes

deitch:
**- What I did**
Made concurrent write access to the cache's index safe.

**- How I did it**
1. Centralized all write access to the cache's `index.json` into a single file
2. Added file locking utility using fcntl on Unix and Unix-like systems
3. Leveraged the file locking for cache writes

Ignored read locks for now, although the utility supports it. Too many places it could be added.

**- How to verify it**
CI to ensure no regressions, manual testing for lock timing.

Concurrent access to writing `index.json` in cache


deitch:
Fixes #4131

PR #4148, opened 2025-07-27
provide mirror support

### LINKED ISSUE #4142: Feature Request: Support overriding default registry during builds
Currently, [ReferenceExpand](https://github.com/linuxkit/linuxkit/blob/c0c5668116cd87141be341d7eebd9715bf638391/src/cmd/linuxkit/util/reference.go#L23) is hardcoded to `"docker.io"`, which makes it difficult to redirect image pulls to an internal or mirrored registry during builds.

I initially tried using Docker registry mirrors, but I’m guessing docker pull isn’t used under the hood, so that had no effect.

As a workaround, we’re currently using a "just-in-time" `sed` pass on the YAML file to rewrite image references (e.g. `image: "foo/bar"` -> `image: "some.repo.com/foo/bar"`). Another approach would be to pre-populate the linuxkit cache from our mirror, but that requires parsing the YAML and manually pulling images - doable, but even clunkier.

Assuming there isn’t already a flag or mechanism for this that I’ve missed, would the maintainers be open to a PR introducing a CLI flag or environment variable to override the default registry used in `ReferenceExpand`? I’m not proposing anything as broad as vendoring or full mirror resolution, which is likely out of scope (and certainly out of my current Go skills), just something minimal to support alternate registry defaults cleanly.

deitch:
Fixes #4142 

**- What I did**
Added support for using mirrors to override default registry for specific registries or all registries

You will be able to do:

```sh
$ lkt --mirror http://mymirror.local      # to mirror all
$ lkt --mirror ghcr.io=http://mymirror.local      # to mirror just ghcr.io
$ lkt --mirror ghcr.io=http://mymirror.local/ghcr --mirror http://mymirror.local/other      # to mirror ghcr and rest differently
```

**- How I did it**
Added the CLI flag, then rewrite the URL

**- How to verify it**
CI

We probably should add a specific test for this, but I struggled to figure out how. Maybe I will get it

mirror support

### DIFF
--- a/src/cmd/linuxkit/cache/pull.go
+++ b/src/cmd/linuxkit/cache/pull.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/partial"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/validate"
+	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/registry"
 	lktspec "github.com/linuxkit/linuxkit/src/cmd/linuxkit/spec"
 	"github.com/linuxkit/linuxkit/src/cmd/linuxkit/util"
 	imagespec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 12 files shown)


deitch:
Good, got a test in

PR #4152, opened 2025-07-31
Enable CONFIG_AMT in 6.6.x

kkroo:
Automatic Multicast Tunneling (RFC7450) in Linux is quite mature now, and of importance for IPTV workflows

**- What I did**

**- How I did it**

**- How to verify it**


deitch:
Thanks for this @kkroo. Can you follow the kernel update process [here](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#building-and-modifying)?

kkroo:
Thanks for the quick review, @deitch. Will do and get CI to pass. My intent is to get this into Docker Desktop. Am I in the right place?

deitch:
Yes and no. I believe that Docker Desktop uses a derivative of linuxkit for the desktop, but I don't think it is direct. We have some maintainers here from Docker, maybe @djs55 can comment?

kkroo:
@deitch I've made the edits to the yaml's by hand, the Makefile in the instructions seems to be broken on OSX. I don't have the permission to push the image to the repo, so Kernel Tests are still failing.

@djs55 keen on your input. Looking at my docker desktop now and its using 6.10.14-linuxkit where is the source for that?

kkroo:
bump

deitch:
Yeah, we would need to build and push it out. We don't yet have CI set up for that, and it is slow (building kernels).

We don't really like pushing out existing kernels again, as who knows what it can break. Is there a really good reason to do this on 6.6.71? Or can we cut a new version (I think 6.6.102 is the most recent 6.6.x), or even better yet a new series (6.12 has been out for a while, 6.16 just came out)? If you can follow the process there (same doc), I can push it out.

kkroo:
thanks for the response @deitch the goal is to get this upstreamed to docker desktop which is currently using 6.10.14 and introduce as little change as possible (AMT) which should not affect anything else. happy to help prep next kernel version target for docker desktop. should i stick with 6.10.x or is better to use a newer version?

deitch:
Just to reiterate, I have zero control over, or even insight into, Docker Desktop's VM image. Actually, I am kind of surprised they used 6.10, given its short life.

The list of kernel versions and their support lifetime is [on wikipedia](https://en.wikipedia.org/wiki/Linux_kernel_version_history). Given 6.12's long life, I would recommend that.

PR #4156, opened 2025-08-12
add support for dynamically calculated build arg sets

deitch:
**- What I did**
Following on from #4155 which added support to calculated build args using:

```
MYARG=@lkt:pkg:/path/to/pkg
```

This adds multiple dynamic packages with `@lkt:pkgs:/path/to/*` via:

```
MYARG_%=@lkt:pkgs:/path/to/*
```

It is explained in more detail in the docs, also included in this PR. In short, everything in the glob path will be treated as a package dir, have its `show-tag` resolved. It then will create multiple build-args, where each one has:

* value: the value of `lkt pkg show-tag <dir>`
* key: the key as provided (in the example above), with the single `%` replaced with the name of the image in the directory, upper-cased, with characters `/`, `-`, `:` converted to `_`

**- How I did it**
Changed the `TransformBuildArgValue()` func from what already existed, added docs, added tests.

**- How to verify it**
CI

Dynamic calculated build args


deitch:
Note that I have yet to add the actual tests for this change. They are coming shortly. I wanted CI to make sure I didn't break anything while I did it.

PR #4162, opened 2025-08-27
erofs output

deitch:
- **add erofs as output option**
- **unify nearly identical functions**

**- What I did**
Created a new tool for erofs output, and added it to linuxkit build. Also unified some identical functions.

And of course a test.

**- How I did it**
By changing the code.

**- How to verify it**
CI. I added a new test.

erofs output


djs55 (approved):
nice! I really like EROFS.

deitch:
Thanks @djs55 ; Justin has been showing me its ins and outs.

We might want to start looking at replacing the structure of containers - onboot, services, onshutdown - with erofs at some point, which makes it even more immutable. But that is in the future.

Do you know of any broadly used images that are storing their filesystems on Docker Hub in erofs?

PR #4164, opened 2025-08-28
 pkglib/git: remove trailing spaces of GOPKGVERSION

christoph-zededa:
this makes `--dry-run` inconvenient as it includes a newline

**- What I did**

trimming all whitespaces around the GOPKGVERSION

**- How I did it**

**- How to verify it**

Run
```
~/src/linuxkit/src/cmd/linuxkit/linuxkit pkg build --platforms linux/amd64 --force --dry-run pkg/pillar
```
 check that after `--build-arg=GOPKGVERSION=v0.0.0-20250828163257-a2e7a9467df0` there is no newline

Fix trailing newline after GOPKGVERSION

ChatGPT Image Aug 28, 2025, 04_47_16 PM

### DIFF
--- a/src/cmd/linuxkit/pkglib/git.go
+++ b/src/cmd/linuxkit/pkglib/git.go
@@ -242,5 +242,8 @@ func (g git) goPkgVersion() (string, error) {
 		}
 		version = fmt.Sprintf("%s-%s", lastSemver, dateCommit)
 	}
+
+	version = strings.TrimSpace(version)
+
 	return version, nil
 }


deitch:
I am not sure I agree with this. If I say "dry run", and it finds it in the cache, then it is telling what it would build: nothing. If you would force a build, then `pkg build --force`, whose equivalent is `pkg build --force --dry-run`.

christoph-zededa:
> I am not sure I agree with this. If I say "dry run", and it finds it in the cache, then it is telling what it would build: nothing. If you would force a build, then `pkg build --force`, whose equivalent is `pkg build --force --dry-run`.

`pkg build --force --dry-run` is a good workaround for me.
I will remove that commit so that only the fix for the newline is left.

deitch:
LGTM. Unfortunately, we need CI to run before merging, which can take a while.

Issue #4166, opened 2025-09-02
Enable `CONFIG_INET_DIAG_DESTROY` kernel config


PhilipSchmid:
### Description

Linuxkit does **not** enable the `CONFIG_INET_DIAG_DESTROY` [kernel config](https://www.kernelconfig.io/config_inet_diag_destroy) by default:
```bash
root@node:/# cat /proc/config.gz | gunzip | grep CONFIG_INET_DIAG_DESTROY | grep DIAG
# CONFIG_INET_DIAG_DESTROY is not set
```

However, other `CONFIG_INET_*` related flags are already set:
```bash
root@node:/# cat /proc/config.gz | gunzip | grep CONFIG_INET | grep DIAG
CONFIG_INET_DIAG=y
CONFIG_INET_TCP_DIAG=y
CONFIG_INET_UDP_DIAG=y
# CONFIG_INET_RAW_DIAG is not set
# CONFIG_INET_DIAG_DESTROY is not set
CONFIG_INET_SCTP_DIAG=y
```

Would it be an option to enable it for all 3 supported architectures? [x86_64](https://github.com/linuxkit/linuxkit/blob/179f74ac245b1d9011851b426f10cd98e2813f64/kernel/6.6.x/config-x86_64#L1131), [aarch64](https://github.com/linuxkit/linuxkit/blob/179f74ac245b1d9011851b426f10cd98e2813f64/kernel/6.6.x/config-aarch64#L1029), and [riscv64](https://github.com/linuxkit/linuxkit/blob/master/kernel/6.6.x/config-riscv64#L828)?

## Motivation

When trying to run [Cilium](https://github.com/cilium/cilium) on KIND clusters that run on the linuxkit kernel which doesn't have the `CONFIG_INET_DIAG_DESTROY=y` kernel config set, Cilium isn't able to terminate stale UDP sockets when running in it's optimized [Kube-Proxy Replacement](https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free) mode with [eBPF Host Routing](https://docs.cilium.io/en/stable/operations/performance/tuning/#ebpf-host-routing) enabled. See the [documented Limitations](https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/#:~:text=When%20socket%2DLB,and%20CONFIG_INET_DIAG_DESTROY.).

[...]

PR #4167, opened 2025-09-02
enabled CONFIG_INET_DIAG_DESTROY kernel config

PhilipSchmid:
Fixes https://github.com/linuxkit/linuxkit/issues/4166

**- What I did**
Enabled the `CONFIG_INET_DIAG_DESTROY` kernel config for all three supported CPU architectures. More details about the why in the issue.

Enabled the `CONFIG_INET_DIAG_DESTROY` kernel config

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:6.6.71
+  image: linuxkit/kernel:6.6.103
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
   - linuxkit/init:680da6e6f79bb8236a095147d532cd2160e23c9f
--- a/examples/cadvisor.yml
+++ b/examples/cadvisor.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:6.6.71
+  image: linuxkit/kernel:6.6.103
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0 console=ttysclp0"
 init:
   - linuxkit/init:680da6e6f79bb8236a095147d532cd2160e23c9f
--- a/examples/containerd-debug.yml
+++ b/examples/containerd-debug.yml
@@ -1,6 +1,6 @@
 # example with volumes, both blank and populated
 kernel:
-  image: linuxkit/kernel:6.6.71
+  image: linuxkit/kernel:6.6.103
   cmdline: "console=tty0 console=ttyS0 console=ttyAMA0"
 init:
   - linuxkit/init:680da6e6f79bb8236a095147d532cd2160e23c9f
--- a/examples/dm-crypt-loop.yml
+++ b/examples/dm-crypt-loop.yml
@@ -1,5 +1,5 @@
 kernel:
-  image: linuxkit/kernel:6.6.71
+  image: linuxkit/kernel:6.6.103
   cmdline: "console=tty0 console=ttyS0"
 init:
   - linuxkit/init:680da6e6f79bb8236a095147d532cd2160e23c9f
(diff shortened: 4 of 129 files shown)


deitch:
In principle, this looks good. It needs a little more, though, to actually get this in. The process is a bit burdensome (nowhere as much as it used to be, thankfully). It is documented in [kernels.md](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md), specifically the section [building and modifying](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#building-and-modifying).

You won't be able to push out new images, but if you get the right changes in here, I can build and push them.

The question is which kernel versions (and maybe series) should get this. We try really hard not to overwrite existing versions. Thankfully, you didn't try to backport anything into 5.x, which is not being updated.

Best would be to cut a new kernel version, also described in the doc under [adding a new version](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#adding-a-new-kernel-version). It only is a few more steps (except for the build itself, which is on me anyways; one of these days we will get CI to do it). The current version is 6.6.71 per [this file](https://github.com/linuxkit/linuxkit/blob/master/kernel/6.6.x/build-args#L1). As far as I know 6.6.103 is the most recent, so bump it to 6.6.103. If you feel like cutting a new series to 6.12.x, that would be even better, but not required.

PhilipSchmid:
Hi @deitch 

Thank you very much for all the super helpful input! I just followed the linked guide and force-pushed all the updated files.

deitch:
Thanks @PhilipSchmid and apologies for the week it took me to get to respond to this. I would build and push it out, but it does take a while to build, and I won't get to it for a few days.

I do have a question and a request. 

First, I see a lot of config changed other than the one you specifically requested. e.g. [This single one in arrch64](https://github.com/linuxkit/linuxkit/pull/4167/files#diff-81ae084eab14630e6a8af67d798014e6116253d69c3f702c88fb8b734a6ba91dL971), but multiple in riscv64 and quite a few in x86_64. What is the impact of those changes?

Second, something unfortunately only [partially documented in `kernels.md`](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#kernels-in-examples-and-tests), can you update everything that uses `6.6.71` to `6.6.103`, and those with hashes?  There are targets in [kernel/Makefile](https://github.com/linuxkit/linuxkit/blob/master/kernel/Makefile), specifically `make update-kernel-yamls`.

PhilipSchmid:
Hi @deitch

Sorry that it took so long. I was on PTO and only found now the time to work on that.

The config changes are not from any manual editing or external source. They're the automatic result of the Linux kernel's own `make defconfig` process when moving from kernel `6.6.71` to `6.6.103`. I also just ran `make update-kernel-yamls` but it looks like this doesn't also work correctly. For example, it also bumped a `5.4.x` version to `6.6.103` (see [here](https://github.com/linuxkit/linuxkit/pull/4167/files#diff-950246171baf9b4c1fc888542cc44d7c7f2a27c203d783e36f0dbb7a2c94e9ca)) and replaced `# linuxkit/kernel:<kernel>.<major>.<minor>-<hash> and once as` with `# linuxkit/kernel:6.6.103 and once as` (see [here](https://github.com/linuxkit/linuxkit/pull/4167/files#diff-7b84d9b96fc6b7160441c46db2ba1df3137072914af06a97f979ca38eda64f9e)).

deitch:
That is fine.

FYI, @ChrisIgel is releasing 6.12.x in #4185. Do you have any interest in seeing if he wants to include just these config changes in there, and it can be in 6.12.x going forward? Or do you want to proceed with this on 6.6.x?

PhilipSchmid:
Superseded by https://github.com/linuxkit/linuxkit/pull/4185. Thanks a lot!

PR #4170, opened 2025-09-08
remove rngd package.

justincormack:
The kernel now injects all the random seeds by default and this is no longer needed, and provides options to trust or not trust the different sources

CONFIG_RANDOM_TRUST_CPU=y and CONFIG_RANDOM_TRUST_BOOTLOADER=y are now defaults, and the kernel randomness has been reworked to remove the reasons to not do this before. Most hardware devices are now credited to the kernel entropy pool directly as well, on all sorts of devices.

### DIFF
--- a/examples/addbinds.yml
+++ b/examples/addbinds.yml
@@ -18,8 +18,6 @@ services:
     binds.add:
       # this will keep all of the existing ones as well
       - /var/tmp:/var/tmp
-  - name: rngd
-    image: linuxkit/rngd:80f22b0f60d23c29ce28d06674bc77fe3775a38b
 files:
   - path: etc/getty.shadow
     # sample sets password for root to "abcdefgh" (without quotes)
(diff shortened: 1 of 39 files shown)


deitch:
It also needs to be removed from the various build ymls, such as `examples/`, `linuxkit.yml`, and anywhere in `tests/`. Not sure if it would be anywhere else.

deitch:
A [quick check](https://github.com/search?q=repo%3Alinuxkit%2Flinuxkit%20%22linuxkit%2Frngd%22&type=code) shows it in `projects`, `examples`, `linuxkit.yml` and a few `test/cases/`

justincormack:
I think I removed it from all those. Its hard to see with the giant diff!

deitch:
Nope. See attached.
rngd

justincormack:
oops forgot to push those changes, added.

deitch:
CI failed on something

justincormack:
So I am no longer in the linuxkit docker hub org - can you add me back?

djs55:
@justincormack looking at the linuxkit hub org, it's currently at "6 of 3 seats filled". I'll ask about increasing it

justincormack:
@djs55 any luck with that?

PR #4181, opened 2025-10-01
Fix cache import from stdin

europaul:
**- What I did**

Fixed a bug in the `linuxkit cache import` command where it was failing when trying to import from stdin. The command was incorrectly treating the dash "-" (which conventionally represents stdin in Unix tools) as a regular filename instead of recognizing it as the stdin indicator.

**- How I did it**

Modified the stdin handling logic in `src/cmd/linuxkit/cache_import.go` to properly detect when "-" is passed as the input argument and route it to stdin processing instead of file processing. Also fixed Go formatting issues in `pkglib/build.go` that were preventing the `make local-check` target from passing.

**- How to verify it**

Try importing a docker image from stdin like
```bash
docker image save <image:tag> | linuxkit cache import -
```
it should not throw an error like
```
FATA[0000] unable to open -: open -: no such file or directory
```

Fix linuxkit cache import command to properly handle stdin input when using "-" argument

### DIFF
--- a/src/cmd/linuxkit/Makefile
+++ b/src/cmd/linuxkit/Makefile
@@ -117,7 +117,7 @@ sign:
 
 install:
 	cp -R ./bin/* $(PREFIX)/bin
-	
+
 .PHONY: clean
 clean:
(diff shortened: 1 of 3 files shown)


deitch:
Something really weird was going on with GHA runners. I wait a day and restarted again now.

Issue #4125, opened 2025-06-22
raw-efi image results in too big initrd


ChrisIgel:
When building a LinuxKit image with a big initrd in the format `raw-efi` the booting fails with `error: the initrd is too big.` being output by Grub. I can continue the boot process but that will then result in a kernel panic like this:

```
[    1.032085] clk: Disabling unused clocks
[    1.183678] ata4: SATA link down (SStatus 0 SControl 300)
[    1.185050] ata3: SATA link up 1.5 Gbps (SStatus 113 SControl 300)
[    1.186557] ata1: SATA link up 1.5 Gbps (SStatus 113 SControl 300)
[    1.188465] ata2: SATA link up 1.5 Gbps (SStatus 113 SControl 300)
[    1.190351] ata3.00: ATAPI: QEMU DVD-ROM, 2.5+, max UDMA/100
[    1.191738] ata3.00: applying bridge limits
[    1.193030] ata5: SATA link down (SStatus 0 SControl 300)
[    1.194562] ata2.00: ATA-7: QEMU HARDDISK, 2.5+, max UDMA/100
[    1.195969] ata2.00: 41943040 sectors, multi 16: LBA48 NCQ (depth 32)
[    1.197401] ata2.00: applying bridge limits
[    1.198600] ata6: SATA link down (SStatus 0 SControl 300)
[    1.200000] ata1.00: ATA-7: QEMU HARDDISK, 2.5+, max UDMA/100
[    1.201462] ata1.00: 1712128 sectors, multi 16: LBA48 NCQ (depth 32)
[    1.202820] ata1.00: applying bridge limits
[    1.203862] ata3.00: configured for UDMA/100
[    1.205110] ata1.00: configured for UDMA/100
[    1.206202] ata2.00: configured for UDMA/100
[    1.206318] scsi 0:0:0:0: Direct-Access     ATA      QEMU HARDDISK    2.5+ PQ: 0 ANSI: 5
[    1.209330] sd 0:0:0:0: [sda] 1712128 512-byte logical blocks: (877 MB/836 MiB)
[    1.209372] sd 0:0:0:0: Attached scsi generic sg0 type 0
[    1.210629] sd 0:0:0:0: [sda] Write Protect is off
[    1.212055] scsi 1:0:0:0: Direct-Access     ATA      QEMU HARDDISK    2.5+ PQ: 0 ANSI: 5
[...]

PR #4183, opened 2025-10-09
Switch to systemd-boot for raw-efi builds

ChrisIgel:
**- What I did**
Replaced grub with systemd-boot and make use of the systemd-efistub together with a UKI. Only for `raw-efi` and `amd64` and `aarch64` for now. Unfortunately, I could not keep the `riscv64` support because alpine does not have systemd-boot packages built for `riscv64`. I'm not sure if this would be a dealbreaker or if we maybe could just compile them quickly ourselves.

Also I initially thought enabling `CONFIG_EFI_STUB` in the kernel would not be necessary because I made use of the systemd-efistub but without the configuration I get the following error during boot:
```
../src/boot/linux.c:120@linux_exec: Bad kernel image: Load error
../src/boot/boot.c:2640@image_start: Failed to execute LinuxKit (\EFI\Linux\linuxkit.efi): Load error
```

The PR is related to these two issues, with the latter one being the reason for this PR: https://github.com/linuxkit/linuxkit/issues/4123 https://github.com/linuxkit/linuxkit/issues/4125

**- How I did it**
I added a new tools package `systemd-boot` that contains the EFI stubs for both `amd64` and `aarch64`. This is then used in the `mkimage-raw-efi` package instead of the grub image. Inside the `make-efi` script the grub config was replaced with a systemd-boot config and the kernel and initrd together with the cmdline were combined into a single UKI, that is then put into the ESP and is being auto-detected by systemd-boot.

**- How to verify it**
Build the new kernel (only necessary for amd64) and build a simple image with the kernel and with the `raw-efi` output. Let the resulting image run in UEFI mode, e.g. with:
```bash
linuxkit run qemu \
	--uefi \
	--fw /usr/share/ovmf/OVMF.fd \
	--cpus 2 \
	--mem 2048 \
	linuxkit-efi.img
```
If the resulting image is really small this PR would also be required for the file to be valid: https://github.com/linuxkit/linuxkit/pull/4139

Replaced grub with systemd-boot for `raw-efi` output

### DIFF
--- a/test/cases/010_platforms/000_qemu/060_run_raw_efi/test.sh
+++ b/test/cases/010_platforms/000_qemu/060_run_raw_efi/test.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+# SUMMARY: Check that EFI BIOS ISO boots in qemu
+# LABELS:
+
+set -e
+
+# Source libraries. Uncomment if needed/defined
+#. "${RT_LIB}"
+. "${RT_PROJECT_ROOT}/_lib/lib.sh"
+
+NAME=qemu-raw-efi
+
+clean_up() {
+	rm -rf ${NAME}-*
+}
+trap clean_up EXIT
+
+# see https://github.com/linuxkit/linuxkit/issues/1872 this is very flaky in qemu
+# disabling for now until we fix a config that works
+exit $RT_CANCEL
+
+if command -v qemu-system-x86_64; then
+	if [ ! -f /usr/share/ovmf/bios.bin ]; then
+		exit $RT_CANCEL
+	fi
+fi
+
+linuxkit build --format raw-efi --name "${NAME}" test.yml
+[ -f "${NAME}-efi.img" ] || exit 1
+linuxkit run qemu --uefi "${NAME}-efi.img" | grep -q "Welcome to LinuxKit"
+
+exit 0
--- a/test/cases/010_platforms/000_qemu/060_run_raw_efi/test.yml
+++ b/test/cases/010_platforms/000_qemu/060_run_raw_efi/test.yml
@@ -1,0 +1,10 @@
+kernel:
+  image: linuxkit/kernel:6.12.52
+  cmdline: "console=ttyS0"
+init:
+  - linuxkit/init:26e546e4ef2669bd44d8c46e3164eb5ba7ae1a5c
+  - linuxkit/runc:98ddef9b5b87cb44783566ed2a86cea95439a631
+onboot:
+  - name: poweroff
+    image: linuxkit/poweroff:4059858a555bed90c2280fa9b060b7a8f8de6d45
+    command: ["/bin/sh", "/poweroff.sh", "10"]
--- a/tools/mkimage-raw-efi/README.md
+++ b/tools/mkimage-raw-efi/README.md
@@ -1,0 +1,20 @@
+# ESP Structure Overview
+
+The script initializes an EFI System Partition using `mkfs.vfat` and populates it with directories and files for **systemd-boot** and a Linux Unified Kernel Image (UKI).
+
+## Partition Layout
+
+```bash
+ESP
+├── EFI
+│   ├── BOOT # contains exactly one of the bootloader binaries below for the respective architecture
+│   │   ├── BOOTX64.EFI # amd64
+│   │   ├── BOOTAA64.EFI # arm64
+│   │   └── BOOTRISCV64.EFI # riscv64
+│   └── Linux
+│       └── linuxkit.efi # LinuxKit Unified Kernel Image (UKI)
+└── loader
+    └── loader.conf # systemd-boot configuration file
+```
+
+UKIs in `EFI/Linux` do not need an explicit entry in `loader/entries` but are automatically picked up by `systemd-boot`.


deitch:
I really like this idea. I have been looking into systemd-boot and UKIs for other purposes, and would like to see them here.

A few points on this PR. Some will make it easier, some will make it more complex.

First, I don't think you need a `tools/systemd-boot/`. If all it does is extract the correct `*.EFI` for the given platform, then you can just take it `FROM alpine` in `mkimage-raw-efi` directly. That should save a bit of headache. You just would need to do the multi-platform thing so you have it for all target platforms, as `mkimage-raw-efi` could be built for (and run on) amd64 while building for arm64, or vice-versa, or for its own platform, etc.

The reason we have a separate grub package, is because we need to create the artifacts in grub-dev `grub-mkimage -O arm64-efi`, etc., and then someplace just to get them, and it gets used in multiple places downstream.

On the other hand, if you want to build it from source so you have it for riscv64 (assuming that is doable?), then a separate package is the way to go. 

Next, alpine. Updating alpine is a bit of a pain. It cannot just be updated in `tools/alpine`, there is a process for it, and everything downstream needs to be updated. It is [documented](https://github.com/linuxkit/linuxkit/blob/master/docs/alpine-base-update.md). 

Finally, kernel. It needs more than just updating the config. There is a process for updating kernel config, [documented here](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#modifying-the-kernel-config), but it is easier to do so in the context of a new kernel version, [documented here](https://github.com/linuxkit/linuxkit/blob/master/docs/kernels.md#adding-a-new-kernel-version).

Unfortunately, each of those steps is some work. I would suggest handling each one in a separate PR:

* update the kernel version (do the latest 6.6 version, or even 6.x) to include the changes you need
* update alpine to include the packages
* consume both of the above

ChrisIgel:
Yeah I know alpine and the kernel need more work for the full update. I just wanted to outline my ideas and the required changes without all the noise the other changes would introduce.
I could try the `riscv64` build but for me personally it would not be of importance/relevance. If it's like a 10 minute thing I'll do it though.

ChrisIgel:
I think the building for `riscv64` works. I just have no real way of verifying it works / is correct other than maybe using `qemu`.

I'll update the kernel and the alpine next, now that I know what packages I really need and where I need them.

deitch:
And of course, I commented on the wrong PR.

ChrisIgel:
I rebased the PR on the newest version with our alpine and kernel changes.

tools/systemd-boot/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+WORKDIR /
+COPY --from=alpine-amd64 /usr/lib/systemd/boot/efi/* /usr/lib/systemd/boot/efi/
+COPY --from=alpine-arm64 /usr/lib/systemd/boot/efi/* /usr/lib/systemd/boot/efi/
+
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.efi.stub /usr/lib/systemd/boot/efi/
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.elf.stub /usr/lib/systemd/boot/efi/
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.efi /usr/lib/systemd/boot/efi/
+
+# this is just a non-platform specific python file so we only need it once for all architectures
+COPY --from=alpine-riscv64 --chmod=755 /work/src/ukify/ukify.py /usr/sbin/ukify
```
deitch: No objection to this approach, but I am curious. Is `apk add ukify` not sufficient? Or not in alpine, so not worth the headache?

tools/systemd-boot/build.yml
```diff
@@ -0,0 +1,6 @@
+image: systemd-boot
+network: true
```
deitch: Why do we need `network: true`? All of the `apk add` are local using `linuxkit/alpine`, and `ADD https://github.com/systemd/systemd.git#v258.1` does not require it either. Or is it the `RUN meson ...` that needs it?

tools/systemd-boot/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM scratch
+ENTRYPOINT []
+WORKDIR /
+COPY --from=alpine-amd64 /usr/lib/systemd/boot/efi/* /usr/lib/systemd/boot/efi/
+COPY --from=alpine-arm64 /usr/lib/systemd/boot/efi/* /usr/lib/systemd/boot/efi/
+
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.efi.stub /usr/lib/systemd/boot/efi/
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.elf.stub /usr/lib/systemd/boot/efi/
+COPY --from=alpine-riscv64 /work/builddir/src/boot/*.efi /usr/lib/systemd/boot/efi/
```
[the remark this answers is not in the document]
ChrisIgel: The additional (selective) copies are because the build process (as we are building from source instead of using the pre-built alpine packages like for `amd64` and `arm64`) generates some more files we don't need.

tools/systemd-boot/Dockerfile
ChrisIgel: Yeah basically the latter. Would we include it in `amd64` **and** `arm64` despite being identical and from which one would be copy it? That's why I just left the `ukify` package out and did it like this.

ChrisIgel:
> I rather like this. Should existing tests cover it? I would think so.

I think I'd actually write an additional test for `raw-efi`. Maybe similar to the `020_run_efi` one.

> Can you describe what the filesystem layout looks like before and after? I can kind of reverse-engineer it from the `make-efi` code, but better to be explicit. Maybe a quick README in `mkimage-raw-efi` would be helpful.

Yeah sure, I'll add it.

> Also, what other build formats would benefit from this?

I think basically everything that targets `efi`. Also the `aws` one.

deitch:
And CI is clean, including your tests. Happy to merge it in, and thank you for your patience.

PR #4221, opened 2026-08-02
Fix LinuxKit BCC and perf builds after per-series hashing change

SmartArray:
**- What I did**

I am currently upgrading the dependencies of our cyber security product. The kernel tool build stopped working after upgrading LinuxKit.

I fixed the BCC and perf builds.

**- How I did it**

Commit `50025b884067d9913c18399ad074c6cec0c91468` (https://github.com/linuxkit/linuxkit/tree/50025b884067d9913c18399ad074c6cec0c91468) introduced per-series kernel hashes. BCC and perf still used the full kernel directory hash as far as I am concerned. They referenced a kernel image that did not exist.

I passed the series hash using `--hash`. The tool builds now reference the correct kernel image.

**- How to verify it**

Build the ARM64 kernels and tools:

```bash
make -C kernel ARCH=arm64 build
```

Verify that the kernel, BCC, and perf builds use the same hash.

Fixed BCC and perf builds after the introduction of per-series kernel hashes.

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -178,7 +178,7 @@ buildtool-%:
 	$(eval KERNEL_VERSION=$(call toolkernel,$*))
 	$(eval KERNEL_SERIES=$(call series,$(KERNEL_VERSION)))
 	$(eval SERIES_HASH=$(call series_hash,$(KERNEL_SERIES)))
-	linuxkit pkg build . $(FORCE) --platforms $(BUILD_PLATFORM) --build-yml ./$(COMMON_DIR)/build-$(TOOL).yml --tag "$(KERNEL_VERSION)-$(SERIES_HASH)" --build-arg-file $(KERNEL_SERIES)/build-args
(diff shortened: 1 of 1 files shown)


deitch (approved):
I had to review it for a bit to understand why this is necessary, including rereading how `lkt pkg build` utilizes it. I probably could have saved time by just reading how the Makefile builds the other kernels, where it already passes in the `--hash` explicitly. All good. Thank you for this.

deitch:
Need to sign the DCO.

SmartArray:
Rebased and signed the DCO

SmartArray:
> I had to review it for a bit to understand why this is necessary, including rereading how `lkt pkg build` utilizes it. I probably could have saved time by just reading how the Makefile builds the other kernels, where it already passes in the `--hash` explicitly. All good. Thank you for this.

Tbh, it took me over an hour to figure out what was going on using print out debugging, because it always worked prior fetching upstream changes

deitch:
> took me over an hour to figure out what was going on using print out debugging

Some complex make and other scripts here. But it is building kernels, so not too surprising.

PR #4214, opened 2026-04-03
Update buildkit to v1.29; fix authprovider API change in buildkit v1.29

eriknordmark:
This is to handle CVE-2026-33748 and CVE-2026-33747.
Fixes #4213 

**- What I did**

go get -u of buildkit

This brought in some need to update other packages, which make us hit the incompatible change in docker around AuthProvider, so this also fixes authprovider API change in buildkit v1.29

**- How I did it**

**- How to verify it**

Normal regression testing.

**- Description for the changelog**

Update the buildkit dependency to version v1.29

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4217, opened 2026-05-26
bump containerd to v2.2.4 (CVE-2026-46680)

eriknordmark:
**- What I did**

Bumped containerd from v2.2.0 to v2.2.4 in `tools/containerd-dev/Dockerfile`. This fixes [CVE-2026-46680](https://github.com/advisories/GHSA-fqw6-gf59-qr4w): an image whose `User` directive is a numeric string outside int32 range (e.g. `"9999999999"`) is treated as a username and resolved against the image's `/etc/passwd`. A crafted `/etc/passwd` mapping the same string to UID 0 then makes the container run as root, defeating Kubernetes `runAsNonRoot`. Fixed on the 2.2 line in v2.2.4.

**- How I did it**

1. `tools/containerd-dev/Dockerfile`: `CONTAINERD_COMMIT=v2.2.0` → `v2.2.4`.
2. Updated the `FROM linuxkit/containerd-dev:<hash>` line in `pkg/containerd/Dockerfile`, `pkg/init/Dockerfile`, and `test/pkg/containerd/Dockerfile` (the three consumers called out by the comment in `tools/containerd-dev/Dockerfile`).
3. Swept every example/test/project YAML that pins the resulting `linuxkit/containerd:<hash>`, `linuxkit/init:<hash>`, and `linuxkit/test-containerd:<hash>` tree hashes to the new values.

Mirrors the scope of the v2.2.0 bump (`2dd12173`, 2025-11-18).

**- How to verify it**

`make -C pkg build` will fail on this PR until the new `linuxkit/containerd-dev:405449f30ba384a0f67ad1b75f78a501604f2e44` image is published to Docker Hub. I don't have push access to the `linuxkit/` org, so a maintainer needs to build & push that image (and the dependent `linuxkit/containerd`, `linuxkit/init`, `linuxkit/test-containerd` tags) before CI can resolve the FROMs. Once the dev image is on Hub, CI should run cleanly.

**- Description for the changelog**

Update containerd to v2.2.4 (fixes CVE-2026-46680).

PR #4222, opened 2026-09-02
src/cmd/linuxkit: update buildkit to v0.31.1

eriknordmark:
buildkit removed the deprecated `SolveOpt.LocalDirs` field in v0.30.0, which
is what currently holds linuxkit at v0.29.0: `linuxkit pkg build` passes the
package directory as the Dockerfile and build-context local mounts through
that field.

The first commit switches those to `SolveOpt.LocalMounts`, wrapping each path
in an fsutil filesystem the same way buildkit itself did for `LocalDirs`
entries before dropping them. It builds unchanged against the currently
pinned v0.29.0, where both fields still exist.

The second commit moves the pin to v0.31.1, which carries two low-severity
security fixes (GHSA-7236-3392-c5c6, a custom frontend bypassing
Seccomp/AppArmor restrictions, and GHSA-72x6-4j93-7w86, a runtime DoS via
unbounded group parsing). `go mod tidy` brings a matching round of transitive
updates and raises the go directive to 1.25.9; it also retires the vendored
`golang.org/x/net/context` copy, unused since gcp moved to the standard
library context.

Verified at the branch tip: `go build ./...`, `go vet ./...`, `go test ./...`
and `golangci-lint run` (v2.11.4, matching CI) are all clean, with master
linted as a control. The first commit was separately built and vetted on its
own against v0.29.0, so both commits compile.

Motivation is downstream: lf-edge/eve vendors `pkglib` in two tool modules and
is pinned below buildkit 0.27 and 0.30 by the `ConfigFile` and `LocalDirs`
removals respectively. The first was fixed in #4214; this finishes the job so
EVE can re-pin linuxkit and take the buildkit security updates.


eriknordmark:
@deitch if/when #4217 is done, take a look at this one to enable the update of buildkit and they do the update in the second commit.

eriknordmark:
@deitch this is my last known update to the dependencies for linuxkit.

deitch:
Aw, but we will miss your updates!

deitch (approved):

eriknordmark:
> Aw, but we will miss your updates!

I said "known" and since the future is unknowable ... ;-)
Is this good to merge @deitch ?

